Issue #8, opened 2024-01-16
soroban-rpc: ingestion engine should set user-agent header on history archive requests


mollykarcher:
### What version are you using?

20.2.0

### What did you do?
run rpc server

### What did you expect to see?
* outbound http archive url requests initiated by rpc server's ingestion should have the http header `User-Agent: "horizon/<horizon_version> golang/<go_mod_version>`
* a prometheus metric is emitted for the history archive request

### What did you see instead?
the outbound archive url request has the default golang http header instead `User-Agent: "Go-http-client`, no metrics for history archive request emitted.

See parent/root issue https://github.com/stellar/go/issues/5163

PR #56, opened 2024-02-08
rpc: added user agent config on ha archive pool

sreuland:
### What

include user agent on archive pool config

### Why

correct http client identity for history archive requests
Closes: #8


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -144,12 +148,24 @@ func MustNew(cfg *config.Config) *Daemon {
...
+
+	historyArchive, err := historyarchive.NewArchivePool(
+		cfg.HistoryArchiveURLs,
 		historyarchive.ArchiveOptions{
+			NetworkPassphrase:   cfg.NetworkPassphrase,
 			CheckpointFrequency: cfg.CheckpointFrequency,
+			ConnectOptions: storage.ConnectOptions{
+				Context:   context.Background(),
+				UserAgent: fmt.Sprintf("soroban-rpc/%s", config.Version)},
+			CacheConfig: historyarchive.CacheOptions{
```
tamirms, line 160: I don't think it's necessary to cache history archive requests in soroban-rpc because soroban-rpc does not have a state verification routine which repeatedly downloads history archive files.

The only time soroban-rpc downloads the history archive files is if it's ingesting from a completely empty DB. In that case, soroban-rpc will ingest all ledger entries from the most recent checkpoint ledger. Otherwise, soroban-rpc will only ingest Tx Meta emitted by captive core.

If we add this history archive cache, I think we will not end up caching any requests and end up using extra disk for files we don't need.

sreuland: ok, that makes sense, will re-work this slightly to use the new ArchivePool interfaces still but disable cache usage.

sreuland: @tamirms , i see the [captive core backend also creates it's own internal archive pool](https://github.com/stellar/go/blob/master/ingest/ledgerbackend/captive_core_backend.go#L176) from urls, and then rpc also creates it's own archive pool locally from same urls, looking at the captive core backend, it primarily uses the pool up front to get root has and a ledger header, so, not heavy usage, but i'm wondering if it's worth consolidating the archive pool instances and letting the captive core backend receive optional pool instance rather than creating another one, such as this:
https://github.com/stellar/go/compare/master...sreuland:cc_archivepool

do you think it's worthwhile?

Shaptic: even more general could be providing an `ArchiveInterface` (since `ArchivePool` conforms to that) xor the list of URLs

sreuland: I re-worked this to not enable ha caching and not emit archive metrics since there is such a limited usage of them from rpc as called out, only about 2-3 requests at start time is about it between rpc and captive core backend. less code is best I think for now rather than trying to get fancy, focused on getting the user agent config'd.

tamirms: @sreuland sounds good!

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -144,12 +145,21 @@ func MustNew(cfg *config.Config) *Daemon {
...
+
+	historyArchive, err := historyarchive.NewArchivePool(
+		cfg.HistoryArchiveURLs,
 		historyarchive.ArchiveOptions{
+			NetworkPassphrase:   cfg.NetworkPassphrase,
 			CheckpointFrequency: cfg.CheckpointFrequency,
+			ConnectOptions: storage.ConnectOptions{
+				Context:   context.Background(),
+				UserAgent: cfg.HistoryArchiveUserAgent},
+			CacheConfig: historyarchive.CacheOptions{
```
tamirms, line 157: now that we're setting `Cache` to false, you can omit the `CacheConfig` property entirely

Shaptic: just for future reference, this will need a minor update once we back-merge 2.28.2 which will just be changing the cache path to be empty

cmd/soroban-rpc/internal/test/archive_test.go
```diff
@@ -0,0 +1,25 @@
...
+		historyArchiveProxyCallback: func(r *http.Request) {
+			userAgents.Store(r.Header["User-Agent"][0], "")
+		},
+	}
+	NewTest(t, cfg)
+
+	_, ok := userAgents.Load("testing")
+	assert.True(t, ok, "rpc service should set user agent for history archives")
+
+	_, ok = userAgents.Load("testing/captivecore")
```
sreuland, line 23: kinda cool to see the multiple user agents captured, there was also the captive core `curl` agent in here, but not relevant to rpc changes, so left that out.

cmd/soroban-rpc/internal/daemon/daemon.go
sreuland: yes, i had emitted it locally and then the visual of disabled here seemed worth it, but np removing as well - https://github.com/stellar/soroban-rpc/pull/56/commits/0600fe94542e395191065eb2ca51550f2c66ff65

cmd/soroban-rpc/internal/test/archive_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestArchiveUserAgent(t *testing.T) {
+	userAgents := sync.Map{}
+	cfg := &Test{
+		historyArchiveProxyCallback: func(r *http.Request) {
+			userAgents.Store(r.Header["User-Agent"][0], "")
+		},
+	}
+	NewTest(t, cfg)
```
tamirms, line 18: it seems unintuitive for NewTest() to both take in a *Test parameter and return a *Test value. I think the functional options pattern could be a good fit here:

https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

sreuland: yes, great point, i went with similar pattern as horizon internal integration tests, using a TestConfig to encapsulate, thanks! https://github.com/stellar/soroban-rpc/pull/56/commits/cdf9481bef898ce8b878400b2901b9fcabb02980

tamirms (approved):

Issue #22, opened 2023-08-31
soroban-rpc: create grafana dashboard


tsachiherman:
### What ?

The goal of this task is to create a grafana dashboard that would give us monitoring capabilities of the deployed servers.

aditya1702:
The dashboard has been updated to show the relevant metrics - https://grafana.stellar-ops.com/d/ut9rEquVz/soroban-rpc?orgId=1

Ops will figure out how to publish it to public Grafana this week. And the JSON has already been added to puppet - https://github.com/stellar/puppet-v4/pull/3694

PR #67, opened 2024-02-16
Add prometheus hook to count different log levels

aditya1702:
### What

Part of #22 

Adds a hook to the default RPC logger so that it counts the number of error, warning, debug etc... logs generated and shows it on grafana dashboard.

### Why

Our current dashboard does not seem to calculate the values.

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/test/metrics_test.go
+++ b/cmd/soroban-rpc/internal/test/metrics_test.go
@@ -47,19 +47,6 @@ func TestMetrics(t *testing.T) {
 	assert.Equal(t, val, 2.0)
 }
 
-//func TestLogMetrics(t *testing.T) {
-//	logMetrics := logmetrics.New("log_metrics_test")
-//	logger := supportlog.New()
-//	logger.AddHook(logMetrics)
-//
-//	err := errors.Errorf("test-error")
-//	logger.WithError(err).Error("test error 1")
-//	logger.WithError(err).Error("test error 2")
-//
-//	val := testutil.ToFloat64(logMetrics[logrus.ErrorLevel])
-//	assert.Equal(t, val, 2.0)
-//}
-
 func getMetrics(test *Test) string {
 	metricsURL, err := url.JoinPath(test.adminURL(), "/metrics")
 	require.NoError(test.t, err)
--- a/cmd/soroban-rpc/internal/test/metrics_test.go
+++ b/cmd/soroban-rpc/internal/test/metrics_test.go
@@ -2,6 +2,12 @@ package test
 
 import (
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/sirupsen/logrus"
+	"github.com/stellar/go/support/errors"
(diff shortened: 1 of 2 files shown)


cmd/soroban-rpc/internal/test/metrics_test.go
```diff
@@ -27,6 +33,19 @@ func TestMetrics(t *testing.T) {
 	require.Contains(t, metrics, buildMetric)
 }
 
+func TestLogMetrics(t *testing.T) {
```
tamirms, line 36: this test is not relevant to soroban-rpc because it is testing the correctness of the logmetrics package which is implemented in the go monorepo. Also, I think any functionality exercised in this test is already duplicated in the existing logmetrics test:

https://github.com/stellar/go/blob/master/support/logmetrics/main_test.go#L16

I think you should remove TestLogMetrics() entirely and, instead, you can add an extra check in `TestMetrics()` which verifies that the log metrics are getting included in the metrics response

cmd/soroban-rpc/internal/test/metrics_test.go
```diff
@@ -25,6 +28,23 @@ func TestMetrics(t *testing.T) {
...
+	assert.NoError(t, err)
+	var metric *io_prometheus_client.MetricFamily
+	for _, mf := range metricFamilies {
+		if *mf.Name == "soroban_rpc_log_error_total" {
+			metric = mf
+			break
+		}
+	}
+	val := metric.Metric[0].Counter.GetValue()
+	assert.Equal(t, val, 2.0)
```
tamirms, line 47: There might be some other parts of the code which log errors when the daemon is initialized in `NewTest(t, nil)`. and in that case the `soroban_rpc_log_error_total ` metric could be greater than 2. So, it would be more robust if this test asserted that value was positive rather than exactly 2

aditya1702: Ah yeah nice catch. I can add a GreaterOrEqual assertion here.

cmd/soroban-rpc/internal/test/metrics_test.go
```diff
@@ -25,6 +28,23 @@ func TestMetrics(t *testing.T) {
...
+	metricFamilies, err := test.daemon.MetricsRegistry().Gather()
+	assert.NoError(t, err)
+	var metric *io_prometheus_client.MetricFamily
+	for _, mf := range metricFamilies {
+		if *mf.Name == "soroban_rpc_log_error_total" {
+			metric = mf
+			break
+		}
+	}
+	val := metric.Metric[0].Counter.GetValue()
```
sreuland, line 46: may want to assert that `metric` is non-nil first to emit a valid assertion error instead of potentially just a nil pointer panic.

sreuland (approved):
interesting usage of logs as a source of metrics also, lgtm, but would probably confirm @tamirms approval also as he had prior comments.

tamirms (approved):

PR #68, opened 2024-02-16
Incorporate `soroban-spec-tools` dependency and remove all Soroban CLI references

psheth9:
### What

- Add soroban spec tools crate
- Remove any CLI related code
- Remove soroban-test crate

### Why

We have circular dependency RPC --> cli --> RPC. More context [here](https://stellarfoundation.slack.com/archives/C04B02ABF37/p1707936067776849) and [here](https://stellarfoundation.slack.com/archives/C02B04RMK/p1708125065281039)

### Known limitations

- We have not explored all the edge cases yet, still in progress. 
- It might be a good idea to pull out some of the common deps from tools and rpc in order to avoid cyclic mess in future

Soroban-tool PR to link spec-tools: https://github.com/stellar/soroban-tools/pull/1203


Makefile
```diff
@@ -41,8 +41,8 @@ Cargo.lock: Cargo.toml
 	cargo update --workspace
 
 install_rust: Cargo.lock
-	cargo install soroban-cli --version 20.2.0
-	cargo install --path ./cmd/crates/soroban-test/tests/fixtures/hello --root ./target --debug --quiet
+	#cargo install soroban-cli --version 20.2.0
+	#cargo install --path ./cmd/crates/soroban-test/tests/fixtures/hello --root ./target --debug --quiet
```
Shaptic, line 45: Can you check whether or not this actually affects things now that we've resolved the circular dependencies? If it's still possible to install the `soroban` CLI from Cargo, we can still use it to build the `test_hello_world.wasm` rather than hard-coding it.

Shaptic: I specifically mean the first line (L44) not the one involving fixtures.

Shaptic (approved):

Makefile
psheth9: Discussed offline !! [will add it back later if nothing breaks]

cmd/crates/soroban-spec-tools/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "soroban-spec-tools"
+description = "Tools for using a contract's XDR spec"
+homepage = "https://github.com/stellar/soroban-tools"
```
sreuland, line 4: should these refs to soroban-tools be changed soroban-rpc now?

psheth9: @sreuland  Yes already on it :)

PR #70, opened 2024-02-20
Perform fee padding in a larger bit width of integers

Shaptic:
### What
Modify fee calculation to use 64-bit integers + tests for overflow.

### Why
In situations where the resource fee padding causes the transaction fee to exceed its maximum value, it will overflow and create an invalid transaction, instead (i.e. one where the transaction fee is lower than the resource fee).

This code ensures that the padding is done in a higher bit width (i.e. with a `u64`) and then cast down to a `u32`.

### Known limitations
- [x] Releasing this is **blocked heavily** on #68.
- [x] I'm in the process of adding unit tests.


cmd/crates/soroban-rpc/src/txn.rs
```diff
@@ -315,13 +315,17 @@ pub fn assemble(
...
     let classic_transaction_fees = crate::DEFAULT_TRANSACTION_FEES;
-    // Pad the fees up by 15% for a bit of wiggle room.
-    tx.fee = (tx.fee.max(
-        classic_transaction_fees
-            + u32::try_from(simulation.min_resource_fee)
-                .map_err(|_| Error::LargeFee(simulation.min_resource_fee))?,
-    ) * 115)
+
+    // Pad the fees up by 15% for a bit of wiggle room
+    let padded_fee = u64::from(
```
dmkozh, line 320: I don't think this logic is correct. We should add 15% on top of the resource fee, not the combined fee (or even more specifically to the refundable portion of the resource fee, as non-refundable fee has almost no wiggle room).
It's also weird that this computation adds the inclusion fee at all; that should probably rather happen on the client side, or at least the inclusion fee should be passed as a part of simulation request. But that's a separate issue.

cmd/crates/soroban-rpc/src/txn.rs
```diff
@@ -315,13 +315,17 @@ pub fn assemble(
...
+        tx.fee.max(
+            classic_transaction_fees
+                + u32::try_from(simulation.min_resource_fee)
+                    .map_err(|_| Error::LargeFee(simulation.min_resource_fee))?,
+        ),
+    ) * 115
         / 100;
+    tx.fee = u32::try_from(padded_fee).map_err(|_| Error::LargeFee(padded_fee))?;
```
dmkozh, line 328: I don't think it's correct to return an error if the padded result overflows; if the simulated fee fits into `u32`, then I don't think we should return an error. I.e. I think we should clamp the max fee at `u32::MAX` (as long as simulated fee fits it, that is).

cmd/crates/soroban-rpc/src/txn.rs
leighmcculloch: Agreed, but this change is addressing the overflow issue, and we should keep it narrowly focused on that.

dmkozh: Yeah, that's true for the second part of my comment. But I still think it would be less confusing if we adjusted the resource fee, then added classic fee (at the very least it would make the code easier to read).

cmd/crates/soroban-rpc/src/txn.rs
Shaptic: yeah fair enough

cmd/crates/soroban-rpc/src/txn.rs
Shaptic: This is the client side - you can think of this code as the Rust RPC client.

> We should add 15% on top of the resource fee, not the combined fee (or even more specifically to the refundable portion of the resource fee

Sure, no argument here - I'm trying to maintain the original logic and only add the overflow checks to minimize the behavior check. Totally fine with tweaking the algorithm itself but we should do that separately.

dmkozh: > This is the client side - you can think of this code as the Rust RPC client.
FWIW I referred here specifically to the interface exposed to the user. I find it surprising that it's currently not possible to adjust the inclusion fee.
Anyway, I don't think this logic is set in stone, but I'm also fine with leaving it as is for now, given that this is going to be changed soon.

Shaptic: > it's currently not possible to adjust the inclusion fee

you can, since it takes the max of the existing vs. estimated fees. but agreed that it's moot anyway.

dmkozh: What I mean is to explicitly set an inclusion fee for transaction. IIUC CLI only allows overriding the whole fee. If we also do max of user-provided fee and resource fee, then we're confusing the user (they might say they want to spend 1000 stroops, while simulation ends up with 10 XLM).

Shaptic: yep good point, they should be separate overrides :+1:

leighmcculloch (approved):

mn13:
Hey guys, could you please explain some details why fee should fit 32 bits?

Why: with soroban 20.3.1  I tried to call `soroban contract install` with a 61 kb wasm in testnet and got `error: Fee was too large 4946582915`. After some investigation I found that error occurs in https://github.com/stellar/soroban-rpc/blob/main/cmd/crates/soroban-rpc/src/txn.rs#L321 because `classic_tx_fee + simulation.min_resource_fee = 4946582915` or 494.6582915 XLM.
Why values that exceed U32_MAX is not allowed for fee? Does it mean that there is no way to install wasm in mainnet with soroban-cli?

2opremio:
Hi @mn13 ! Thanks for reporting your problem. However, could you please create a separate ticket for it? Thanks!

leighmcculloch:
The 32-bit size is limited by the protocol, so any change to the size is a protocol change for the Stellar network itself, and not just a change to the rpc, or another tool.

@mn13 The best place to discuss possible protocol changes is currently over here, if you'd like to open an issue there: https://github.com/stellar/stellar-protocol/discussions

PR #87, opened 2024-02-27
feat: add configurable timeout instead of magic number

willemneal:
Currently running into issues with timeouts and this allows editing the timeout.

### DIFF
--- a/cmd/crates/soroban-rpc/src/lib.rs
+++ b/cmd/crates/soroban-rpc/src/lib.rs
@@ -548,6 +548,7 @@ pub struct FullLedgerEntries {
 
 pub struct Client {
     base_url: String,
+    timeout_in_secs: u64,
 }
 
 impl Client {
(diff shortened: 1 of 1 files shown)


cmd/crates/soroban-rpc/src/lib.rs
```diff
@@ -580,9 +581,17 @@ impl Client {
         tracing::trace!(?uri);
         Ok(Self {
             base_url: uri.to_string(),
+            timeout_in_secs: 30,
```
Shaptic, line 584: this changes the default to 30 instead of 10, below, right? we should keep things consistent

Shaptic (approved):
one nit on a behavior change but otherwise lgtm :+1:

cmd/crates/soroban-rpc/src/lib.rs
willemneal:
```suggestion
            timeout_in_secs: 10,
```

PR #92, opened 2024-03-01
Use the refactored version of soroban-simulation library

dmkozh:
The refactored library encapsulates the network configuration in the library, so there is no longer a need to track the bucket list size from soroban-rpc. It also accepts the adjustment config that allows customizing the resource/refundable fee leeway without modifying the simulation crate itself.

Besides the refactoring, soroban-simulation library is switched to use a more accurate recording mode implementation, which allows to decrease the necessary CPU instruction leeway significantly.

### DIFF
--- a/cmd/soroban-rpc/internal/preflight/preflight.go
+++ b/cmd/soroban-rpc/internal/preflight/preflight.go
@@ -2,7 +2,6 @@ package preflight
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"runtime/cgo"
 	"time"
@@ -36,7 +35,7 @@ type snapshotSourceHandle struct {
 }
 
 const (
-	defaultInstructionLeeway uint64 = 3000000
+	defaultInstructionLeeway uint64 = 0
 )
 
 // SnapshotSourceGet takes a LedgerKey XDR in base64 string and returns its matching LedgerEntry XDR in base64 string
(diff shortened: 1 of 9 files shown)


cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -36,7 +35,7 @@ type snapshotSourceHandle struct {
 }
 
 const (
-	defaultInstructionLeeway uint64 = 3000000
+	defaultInstructionLeeway uint64 = 0
```
2opremio, line 38: @dmkozh This will surely have an impact for the end user.

Are you sure this won't lead to the instruction/fee estimation falling short for the end user? I think it's better to overestimate.

dmkozh: Tightening the simulation estimates is the main functional goal of this whole project and this change reflects that (+3M leeway has always been a hack and we wanted to remove it ASAP). As you can see, the integration tests are still passing and there is also env side coverage for all the host fns and pretty complex setups with auth and cross-contract calls. These tests are passing with just 1% leeway.

I also think it's better to store defaults in the simulation library itself (https://github.com/stellar/rs-soroban-env/blob/8c9ab83c406bd86f56d52eae3e39dccf6b45b3da/soroban-simulation/src/simulation.rs#L315). This way they can reflect the state of simulation accuracy. That said, it is easy to modify `SimulationAdjustmentConfig` on the RPC side in case if we find that something is off without doing any changes in the simulation crate itself. The same goes for providing user-customizable leeways for more resources - it should be easy to do that without changing anything about the simulation library.

2opremio: Alright, makes sense!

2opremio (approved):
LGTM

The `internal_error` hack is pretty ugly (it would be way cleaner if an appropriate or custom `HostError`s could be used instead).

dmkozh:
> The internal_error hack is pretty ugly (it would be way cleaner if an appropriate or custom HostErrors could be used instead).

FWIW internal error is probably the most appropriate thing we can return currently as simulation will be incorrect in case if storage is somehow broken. It would be tricky to introduce new host error codes that are exclusive to simulation only (probably it's not impossible, but pretty cumbersome).

Issue #84, opened 2024-02-26
rename soroban-rpc crates to soroban-rpc-client


mollykarcher:
### What problem does your feature solve?

Confusion in spoken/written communication, including code reviews around distinction between RPC the service and RPC the crate/client.

### What would you like to see?

A more specific name to have a distinction between RPC the service and the client of the service.

### What alternatives are there?

Leave them the same

PR #93, opened 2024-03-02
Deprecate soroban-rpc and publish stellar-rpc-client

psheth9:
### What

Rename `soroban-rpc` to `soroban-rpc-crate` and also remove outdated `soroban-tools` refs 

Note crate.io is immutable repository so In order to rename we are following these steps: 

**New plan:

- Remove soroban-rpc crate entirely
- Publish new stellar-rpc-client crate (cloned from soroban-rpc)**

Old plan: This could have introduced breaking change so avoided !!

- Update readme in` soroban-rpc` to redirect users to new `stellar-rpc-client` crate, remove everything else
- publish soroban-rpc and verify crates.io
- publish stellar-rpc-client and verify crates.io

### Why

See #84 it had created lot of confusion when referring to `soroban-rpc` client crate as sometimes people percieve it as `soroban-rpc` service and not client crate. we are just better off with renaming to `soroban-rpc-client` !!

### Known limitations

We might need to update any dependent repositories or downstream components !!


Shaptic:
I really don't have a strong opinion on this whatsoever but maybe something like `rs-rpc-sdk` would deliver more information to match things like `js-stellar-sdk`/`java-stellar-sdk`/`rs-soroban-sdk`?

psheth9:
@Shaptic  Agreed but I think we would be doing that if we actually move out this crate to its own repo. Suggestion from Lei [here](https://github.com/stellar/soroban-rpc/issues/77#issuecomment-1965900721)

cmd/crates/soroban-rpc-client/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
-name = "soroban-rpc"
+name = "soroban-rpc-client"
```
leighmcculloch, line 2: I think we should update the `soroban-rpc` crate's readme on https://crates.io/crates/soroban-rpc somehow to point to the new crate name. I think that the only way to do that is to publish a new version of it though. Maybe we make a soroban-rpc crate in this repo that is empty other than a readme and publish it?

cmd/crates/soroban-rpc-client/README.md
```diff
@@ -0,0 +1,3 @@
+# soroban-rpc-client
+
+Tools and utilities for soroban rpc client.
```
leighmcculloch, line 3:
```suggestion
A rust client for communicating with the soroban-rpc.
```

cmd/crates/soroban-rpc-client/Cargo.toml
leighmcculloch, line 2: Should we be renaming it to stellar-rpc-client given the rpc will eventually be losing the soroban name?

psheth9: Valid question !! Asking in `soroban-dev` to get feedback from product folks

psheth9: Yes thats the only option seems like. publish empty soroban-rpc with updated readme and gonna yank it just to deprecate.

cmd/crates/soroban-rpc/Cargo.toml
```diff
@@ -1,50 +1,12 @@
...
-description = "Soroban RPC client for rust"
-homepage = "https://github.com/stellar/soroban-tools"
-repository = "https://github.com/stellar/soroban-tools"
+description = "Stellar RPC client for rust"
+homepage = "https://github.com/stellar/soroban-rpc"
+repository = "https://github.com/stellar/soroban-rpc"
 authors = ["Stellar Development Foundation <info@stellar.org>"]
 license = "Apache-2.0"
 readme = "README.md"
 version.workspace = true
```
leighmcculloch, line 9: I have a second though on publishing a new soroban-rpc with a deprecated message. The problem is if we post a minor or patch version, as will happen here, folks might upgrade to it by accident. Also, it'd be odd to publish a v21. I think we could just drop publishing it or the moment. And then once we've moved over to stellar-rpc-client, we can yank all the soroban-rpc versions and that will display in the crates.io UI.

psheth9: Yeah it will be a breaking change when they upgrade bc they actually get empty crate. 

so will there be any problem with crates.io, If I rollback to initial option of removing` soroban-rpc` crate from the code and publishing new ` stellar-rpc-client `  @leighmcculloch

leighmcculloch: No problem, I think that was the right approach and I sent us on an unnecessary detour, apologies for that.

psheth9: No worries, All good :)

leighmcculloch (approved):

PR #101, opened 2024-03-11
feat: abstract signer to allow for different methods of signing transactions and auth entries

willemneal:


janewang:
Making a trait / interface in its own crate and keeping it in the rpc crate for the time being. May need to move to a new crate. Would be a breaking changes

janewang:
A breaking change for RPC. This code will move to a new crate

willemneal:
closed in favor of #157

PR #120, opened 2024-04-02
Add ledger entry diff to simulateTransaction response

### LINKED ISSUE #11: soroban-rpc: add effects/changes a transaction makes over an account in the simulation response
### What problem does your feature solve?
The ability to show users what could be the state of their accounts after a transaction has been signed and sent to the network. With Stellar Classic operations was easy to show them all the operations happening on a transaction, but with Soroban transactions is harder since each contract is unique so being able to at least show them something like "XLM: -20,239" could improve the UX and in some way security a lot.

### What would you like to see?
A field in the simulation response similar to the result of the transaction (when transactions return value) but with the calculated simulation effects/changes. My idea is that I could show users something along the lines of "You are authorizing 200 USDC, the simulation result shows you could be charged 100 USDC". I understand the end result could be different, but is still valuable

### What alternatives are there?
Currently, I'm testing to show what is being authorized from the XDR transaction, that still helps with the UX but I think both cases are complementary

2opremio:
### What

It adds the following field to `SimulateTransactionResponse`

```go
StateDiff       []LedgerEntryDiff            `json:"stateDiff,omitempty"`       // If present, it indicates how the state (ledger entries) will change as a result of the transaction execution.
```

where:

```go
// LedgerEntryDiff designates a change in a ledger entry. Before and After cannot be be omitted at the same time.
// If Before is omitted, it constitutes a creation, if After is omitted, it constitutes a delation.
type LedgerEntryDiff struct {
	Before string `json:"before,omitempty"` // LedgerEntry XDR in base64
	After  string `json:"after,omitempty"`  // LedgerEntry XDR in base64
}
```

### Why

Closes #11 

### Known limitations

It can considerably increase the size of simulate transaction responses (particularly for contract uploads, which will contain the full wasm).

### DIFF
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -35,6 +35,13 @@ type RestorePreamble struct {
 	MinResourceFee  int64  `json:"minResourceFee,string"`
 }
 
+// LedgerEntryDiff designates a change in a ledger entry. Before and After cannot be be omitted at the same time.
+// If Before is omitted, it constitutes a creation, if After is omitted, it constitutes a delation.
+type LedgerEntryDiff struct {
+	Before string `json:"before,omitempty"` // LedgerEntry XDR in base64
+	After  string `json:"after,omitempty"`  // LedgerEntry XDR in base64
+}
+
 type SimulateTransactionResponse struct {
 	Error           string                       `json:"error,omitempty"`
 	TransactionData string                       `json:"transactionData,omitempty"` // SorobanTransactionData XDR in base64
@@ -43,6 +50,7 @@ type SimulateTransactionResponse struct {
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`         // an array of the individual host function call results
 	Cost            SimulateTransactionCost      `json:"cost,omitempty"`            // the effective cpu and memory cost of the invoked transaction execution.
 	RestorePreamble *RestorePreamble             `json:"restorePreamble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
+	StateDiff       []LedgerEntryDiff            `json:"stateDiff,omitempty"`       // If present, it indicates how the state (ledger entries) will change as a result of the transaction execution.
 	LatestLedger    uint32                       `json:"latestLedger"`
 }
 
@@ -149,6 +157,12 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 			}
 		}
 
+		stateDiff := make([]LedgerEntryDiff, len(result.LedgerEntryDiff))
+		for i := 0; i < len(stateDiff); i++ {
+			stateDiff[i].Before = base64.StdEncoding.EncodeToString(result.LedgerEntryDiff[i].Before)
+			stateDiff[i].After = base64.StdEncoding.EncodeToString(result.LedgerEntryDiff[i].After)
+		}
+
 		return SimulateTransactionResponse{
 			Error:           result.Error,
 			Results:         results,
@@ -161,6 +175,7 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 			},
 			LatestLedger:    latestLedger,
 			RestorePreamble: restorePreamble,
+			StateDiff:       stateDiff,
 		}
 	})
 }
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -3,7 +3,10 @@ package methods
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/creachadair/jrpc2/handler"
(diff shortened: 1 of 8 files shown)


cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -43,6 +50,7 @@ type SimulateTransactionResponse struct {
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`         // an array of the individual host function call results
 	Cost            SimulateTransactionCost      `json:"cost,omitempty"`            // the effective cpu and memory cost of the invoked transaction execution.
 	RestorePreamble *RestorePreamble             `json:"restorePreamble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
+	StateDiff       []LedgerEntryDiff            `json:"stateDiff,omitempty"`       // If present, it indicates how the state (ledger entries) will change as a result of the transaction execution.
```
leighmcculloch, line 53: The comment indicates that it may or may not be present. When would the state diff be omitted? If a transactions results in no affect to state then I think the state diff should be an empty array rather than missing from the response. It tells a clear story that state was unaffected rather than being silent on the matter.

leighmcculloch (commented):
Couple thoughts on the API.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
leighmcculloch, line 53: The term diff doesn't feel quite right, at the least what's included in the API doesn't really live up to the term. A diff is not actually returned, but rather the before and after snapshots are returned and it's up to the client to determine the diff if they need a diff.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -35,6 +35,13 @@ type RestorePreamble struct {
 	MinResourceFee  int64  `json:"minResourceFee,string"`
 }
 
+// LedgerEntryDiff designates a change in a ledger entry. Before and After cannot be be omitted at the same time.
+// If Before is omitted, it constitutes a creation, if After is omitted, it constitutes a delation.
+type LedgerEntryDiff struct {
```
leighmcculloch, line 40: Has there been any discussion about also including ledger keys XDR? While all ledger keys are constructable from the ledger entry, someone has to write code to do it and it might be helpful to also export the ledger key because it's common to surface IDs in APIs alongside their data for telling the overall story for the data across different APIs. For example, the RPC function for getting ledger entries accepts a ledger key as the ID.

leighmcculloch, line 40: I think there's value in including the type of change explicitly in a field. While it can be worked out with before / after values being null, the API is self documenting if the type is included.

For example, also including the key per the other comment:
```json
"stateDiff": [
    { "key": "ledgerkeyxdr", "type": "created", "before": null, "after": "new" },
    { "key": "ledgerkeyxdr", "type": "updated", "before": "old", "after": "new" },
    { "key": "ledgerkeyxdr", "type": "deleted", "before": "old", "after": null },
]
```

We can still use before and after so that the data is not polymorphic based on type, and for applications that just want to pull the after value, they can do that.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -35,6 +35,13 @@ type RestorePreamble struct {
...
+	Before string `json:"before,omitempty"` // LedgerEntry XDR in base64
+	After  string `json:"after,omitempty"`  // LedgerEntry XDR in base64
```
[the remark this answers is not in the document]
2opremio: Sure, I can do that.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: The diff will be omitted when there are no changes or ... for operations other than InvokeHostFunction (i.e. Restore/Extend footprint).

> I think the state diff should be an empty array rather than missing from the response

What you suggest makes sense but we already omit pretty much every other field when empty. So, I am reluctant to do that for consistency.

2opremio: > We could use the term `stateChanges` instead? Other ideas?

Makes sense, I will change it.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Uhm, it feels a bit redundant, but it matches what we do in [getLedgerEntries](https://developers.stellar.org/network/soroban-rpc/methods/getLedgerEntries), so I will do that.

2opremio: I feel this is pretty redundant, but I can add it.

2opremio:
@leighmcculloch PTAL

cmd/soroban-rpc/internal/methods/simulate_transaction.go
leighmcculloch: Thanks for the context, I didn't understand that the other fields were also being omitted when empty.

leighmcculloch (approved):
👏🏻

Shaptic (approved):

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -34,6 +37,108 @@ type RestorePreamble struct {
...
+	var (
+		entryXDR   []byte
+		changeType LedgerEntryChangeType
+	)
+	switch {
+	case beforePresent:
+		entryXDR = diff.Before
+		if afterPresent {
+			changeType = LedgerEntryChangeTypeUpdated
+		} else {
```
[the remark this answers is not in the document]
2opremio: If it gets here it can’t get to like 103, or am I missing something?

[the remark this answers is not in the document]
2opremio: No problem! It actually prompted me to add a unit test.

PR #127, opened 2024-04-09
Improve startup logging

2opremio:
### What

Log version and commit right after starting and add more detailed logging around in-memory storage initialization.

### Why

We found situations in which RPC took 30 mins to start

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -34,10 +34,11 @@ import (
 )
 
 const (
-	prometheusNamespace          = "soroban_rpc"
-	maxLedgerEntryWriteBatchSize = 150
-	defaultReadTimeout           = 5 * time.Second
-	defaultShutdownGracePeriod   = 10 * time.Second
+	prometheusNamespace                   = "soroban_rpc"
+	maxLedgerEntryWriteBatchSize          = 150
+	defaultReadTimeout                    = 5 * time.Second
+	defaultShutdownGracePeriod            = 10 * time.Second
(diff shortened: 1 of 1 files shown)


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -196,7 +202,20 @@ func MustNew(cfg *config.Config) *Daemon {
...
+	var initialSeq uint32
+	var currentSeq uint32
 	err = db.NewLedgerReader(dbConn).StreamAllLedgers(readTxMetaCtx, func(txmeta xdr.LedgerCloseMeta) error {
+		currentSeq = txmeta.LedgerSequence()
+		if initialSeq == 0 {
+			initialSeq = currentSeq
+			logger.WithFields(supportlog.F{
+				"seq": currentSeq,
+			}).Info("initializing in-memory store")
+		} else if (currentSeq-initialSeq)%inMemoryInitializationLedgerLogPeriod == 0 {
```
sreuland, line 214: this is great to print ongoing status, maybe the period of 1M ledgers does not get reached, is it worthwhile to use a counter that the function closure increments and uses as trigger instead?

sreuland (approved):
lgtm, thanks for adding this!

sreuland:
this was triggered during investigation of https://github.com/stellar/soroban-rpc/issues/128

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: There are around 40M ledgers in pubnet. I don’t want to flood the logs

PR #129, opened 2024-04-09
Pass bucket list size from meta to simulation `NetworkConfig`.

dmkozh:
### What

Pass bucket list size from meta to `NetworkConfig`.

### Why

Unfortunately, Core currently doesn't emit the bucket list size config entry changes in the meta, so the only way to fetch the proper value as of now is to use the meta field directly.

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -463,9 +463,9 @@ checksum = "9ea835d29036a4087793836fa931b08837ad5e957da9e23886b29586fb9b6650"
 
 [[package]]
 name = "ecdsa"
-version = "0.16.9"
+version = "0.16.7"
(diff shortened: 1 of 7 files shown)


leighmcculloch (approved):

2opremio:
Are the updated dependencies based off protocol 20 or 21? (e.g. XDR-wise)

dmkozh:
> Are the updated dependencies based off protocol 20 or 21? (e.g. XDR-wise)

The dependencies here were protocol 20 (the bump was necessary due to meta extension changes). But since we have decided to only include the fix with v21 env release, the deps in this PR don't make much sense. So I guess you could patch the code (but not dependency) changes from this PR into the PR that would do the 20->21 transition.

PR #132, opened 2024-04-10
Add get_version_info endpoint

psheth9:
Fixes https://github.com/stellar/soroban-rpc/issues/10

This PR does following things: 

- Call /info endpoint with backoff exp retry to fetch captive core version
- Reuse `getBucketListSizeAndProtocolVersion` to fetch protocol version
- Update MakeFile to populate version, commit_hash and build_time_stamp

get_version_info Response --> 

```
        Version            string `json:"version"`
	CommitHash         string `json:"commit_hash"`
	BuildTimestamp     string `json:"build_time_stamp"`
	CaptiveCoreVersion string `json:"captive_core_version"`
	ProtocolVersion    uint32 `json:"protocol_version"`
```


cmd/soroban-rpc/internal/methods/get_version_info.go
```diff
@@ -0,0 +1,42 @@
...
+	BuildTimestamp         string `json:"build_time_stamp"`
+	ProtocolVersion        int    `json:"protocol_version"`
+	XDRVersion             string `json:"xdr_version"`
+	CaptiveCoreVersionInfo string `json:"captive_core_version"`
+}
+
+func NewGetVersionInfoHandler(daemon interfaces.Daemon) jrpc2.Handler {
+	coreClient := daemon.CoreClient()
+	return handler.New(func(ctx context.Context, request GetVersionRequest) (GetVersionResponse, error) {
+		info, err := coreClient.Info(ctx)
```
Shaptic, line 26: maybe it's better that we still return as much as we can in this case? so then, for example, if the Captive Core is bogged down, we can still provide version info

Shaptic: actually, giving user the ability to talk to Core at all is dangerous - it's not built to handle requests like that. since we're just using it for the protocol version, we should cache that on startup, instead!

Shaptic (commented):
I know it's still a draft but I was just drivin' by

cmd/soroban-rpc/internal/methods/get_version_info.go
psheth9: Both great points !!

psheth9: Removed ProtocolVersion from response: From stellar-rpc document, it seems like we also want `passphrase` along with `protocol version`, so it might be just better to use `getNetwork` for fetching protocol version. (We could merge both the endpoints later in future but keeping it out of scope for this PR)

cmd/soroban-rpc/internal/methods/get_version_info.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func NewGetVersionInfoHandler() jrpc2.Handler {
+	return handler.New(func(ctx context.Context, request GetVersionInfoRequest) (GetVersionInfoResponse, error) {
+
+		return GetVersionInfoResponse{
+			Version:                config.Version,
+			CommitHash:             config.CommitHash,
+			BuildTimestamp:         config.BuildTimestamp,
+			CaptiveCoreVersionInfo: config.CaptiveCoreVersionInfo,
```
tamirms, line 27: this assumes that the captive core version is static but sometimes we rollout new versions of core while horizon / soroban-rpc is still running:

https://github.com/stellar/go/issues/3602

2opremio: You may want to reuse https://github.com/stellar/soroban-rpc/pull/142

psheth9: Thank you, Used `/info` endpoint to fetch captive core version and reused `getBucketListSizeAndProtocolVersion` to fetch protocol version

Makefile
```diff
@@ -2,6 +2,9 @@ all: check build test
 
 export RUSTFLAGS=-Dwarnings -Dclippy::all -Dclippy::pedantic
 
+
+CAPTIVE_CORE_VERSION_INFO := $(shell stellar-core version)
+
```
2opremio, line 7: We shouldn't need this anymore

Makefile
```diff
@@ -17,7 +20,8 @@ BUILD_TIMESTAMP ?= $(shell date '+%Y-%m-%dT%H:%M:%S')
 GOLDFLAGS :=	-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Version=${REPOSITORY_VERSION}' \
 				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.CommitHash=${REPOSITORY_COMMIT_HASH}' \
 				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.BuildTimestamp=${BUILD_TIMESTAMP}' \
-				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Branch=${REPOSITORY_BRANCH}'
+				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Branch=${REPOSITORY_BRANCH}' \
+				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.CaptiveCoreVersionInfo=${CAPTIVE_CORE_VERSION_INFO}'
```
2opremio, line 24: Nor this

cmd/soroban-rpc/internal/methods/get_version_info.go
```diff
@@ -0,0 +1,68 @@
...
+	CaptiveCoreVersion string `json:"captive_core_version"`
+	ProtocolVersion    uint32 `json:"protocol_version"`
+}
+
+func NewGetVersionInfoHandler(logger *log.Entry, ledgerEntryReader db.LedgerEntryReader, ledgerReader db.LedgerReader, daemon interfaces.Daemon) jrpc2.Handler {
+	coreClient := daemon.CoreClient()
+	return handler.New(func(ctx context.Context, request GetVersionInfoRequest) (GetVersionInfoResponse, error) {
+
+		var captiveCoreVersion string
+		info, err := coreClient.Info(ctx)
```
2opremio, line 29: I would cache this so that we don't invoke a process every time we call the endpoint. To be safe, you can cache it for certain time (e.g. 1 min)

Makefile
```diff
@@ -17,7 +18,7 @@ BUILD_TIMESTAMP ?= $(shell date '+%Y-%m-%dT%H:%M:%S')
 GOLDFLAGS :=	-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Version=${REPOSITORY_VERSION}' \
 				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.CommitHash=${REPOSITORY_COMMIT_HASH}' \
 				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.BuildTimestamp=${BUILD_TIMESTAMP}' \
-				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Branch=${REPOSITORY_BRANCH}'
+				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Branch=${REPOSITORY_BRANCH}' \
```
2opremio, line 21:
```suggestion
				-X 'github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config.Branch=${REPOSITORY_BRANCH}'
```

cmd/soroban-rpc/internal/methods/get_version_info.go
Shaptic: @2opremio I called this out [above](https://github.com/stellar/soroban-rpc/pull/132/files#r1560149665) but was even more strict: no RPC method should call out to Captive Core directly since it's not made to handle HTTP requests at scale. We should cache it once on startup/CC restart.

psheth9: Yes I agree with @Shaptic, we have to fetch core version only during bootup and restart. 

I added retry logic with ExponentialBackOff to fetch captive core version during daemon.run()

tamirms: as I mentioned in https://github.com/stellar/soroban-rpc/pull/132#discussion_r1566877185 the core version can change so it's better to cache it for a limited period of time as @2opremio suggested

psheth9: @tamirms oh I totally misunderstood it, I thought RPC process would still restart itself when we change the underlying core. 

I am more inclined towards doing cache part in follow up PR, because `getNetwork` also has the same problem where it talks to core info endpoint. Need to refactor that in same way. if everything else looks okay, I can revert this specific part in order to get the green signal. Thoughts?

tamirms: ok, I think the best solution would be to implement a function in the captive core implementation which exposes the most recent info response from core because we are already periodically querying the info endpoint to expose metrics:

https://github.com/stellar/go/blob/master/ingest/ledgerbackend/captive_core_backend.go#L218-L246

perhaps you could take that approach in a separate PR?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -185,6 +186,24 @@ func MustNew(cfg *config.Config) *Daemon {
...
+			return err
+		}
+		config.CaptiveCoreVersion = info.Info.Build
+		return nil
+	}
+
+	backoffStrategy := backoff.NewExponentialBackOff()
+	backoffStrategy.MaxElapsedTime = 4 * time.Second
+	err = backoff.Retry(fetchCaptiveCoreVersion, backoffStrategy)
+	if err != nil {
```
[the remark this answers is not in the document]
Shaptic: Does this mean the backoff strategy itself failed? imo yes I think it's safe to stop RPC here since it means something is very wrong with Captive Core, but maybe extend the `MaxElapsedTime` to 10s or even 30s.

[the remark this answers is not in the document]
tamirms: captive core is not running at this point in the code which is why the integration tests are probably failing. captive-core only starts running during the ingestion service

cmd/soroban-rpc/internal/methods/get_version_info.go
psheth9: Thanks @tamirms  Okay I will target that in separate PR !! For now, I am reverting code to initial approach of calling /info endpoint similar to what we have in `getNetwork`

cmd/soroban-rpc/internal/daemon/daemon.go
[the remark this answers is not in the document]
psheth9: As per this [discussion](https://github.com/stellar/soroban-rpc/pull/132#discussion_r1570670897) setting core version in daemon won't work, so removed this logic. Please note I am going back to initial approach of calling /info endpoint. Adding the support of caching along with new [function](https://github.com/stellar/soroban-rpc/pull/132#discussion_r1570930941) in core impl will be done in follow up PR. But for now I want to push this to prod to have something for clients to rely on :)

cmd/soroban-rpc/internal/test/get_version_info_test.go
```diff
@@ -0,0 +1,68 @@
...
+	client := jrpc2.NewClient(ch, nil)
+	request := methods.GetVersionInfoRequest{}
+
+	var result methods.GetVersionInfoResponse
+	err := client.CallResult(context.Background(), "getVersionInfo", request, &result)
+	assert.NoError(t, err)
+
+	assert.NotEmpty(t, result.Version)
+	assert.NotEmpty(t, result.BuildTimestamp)
+	assert.NotEmpty(t, result.CommitHash)
```
tamirms, line 33: you can assert that they match the values from config

cmd/soroban-rpc/internal/test/get_version_info_test.go
```diff
@@ -0,0 +1,68 @@
...
+	assert.NotEmpty(t, result.CaptiveCoreVersion)
+	assert.NotEmpty(t, result.ProtocolVersion)
```
tamirms, line 35: the protocol version is available in `test` . we should assert that `result.ProtocolVersion` matches the expected protocol version

cmd/soroban-rpc/internal/methods/get_version_info.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"github.com/creachadair/jrpc2"
+	"github.com/creachadair/jrpc2/handler"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/config"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/daemon/interfaces"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db"
+)
+
+type GetVersionInfoRequest struct {
```
tamirms, line 13: if a json rpc endpoint doesn't take any arguments you can just omit the request entirely. see cmd/soroban-rpc/internal/methods/get_latest_ledger.go for example

tamirms (approved):

PR #135, opened 2024-04-15
Bump dependencies for protocol 21

2opremio:


.github/workflows/soroban-rpc.yml
```diff
@@ -111,7 +111,7 @@ jobs:
     env:
       SOROBAN_RPC_INTEGRATION_TESTS_ENABLED: true
       SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN: /usr/bin/stellar-core
-      PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 20.2.0-1716.rc3.34d82fc00.focal
+      CORE_DEBIAN_PKG_VERSION: 20.4.1-1807.b152dc51d.focal
```
Shaptic, line 114: Do we need to have some sort of directive (maybe in the Captive Core config) to tell Core to run with Protocol 21 enabled? Or is this already implied by this build (which is weird given the `20` in the version..)

2opremio: It's implicit for captive core when it connects to the network but we also have the protocol number hardcoded in the simulation library (we should probably automate that)

2opremio: https://github.com/stellar/soroban-rpc/issues/137

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -174,10 +174,11 @@ func getLedgerInfo(params PreflightParameters) (C.ledger_info_t, error) {
 	li := C.ledger_info_t{
 		network_passphrase: C.CString(params.NetworkPassphrase),
 		sequence_number:    C.uint32_t(simulationLedgerSeq),
-		protocol_version:   20,
+		protocol_version:   21,
```
dmkozh, line 177: It would be nice to cover two protocol versions in the test suite: the 'current' one (20 in our case) and the 'next' one (21 in our case). The reason is that there is some period of time where a new version of RPC will still run for p20. That's not necessary for this PR (probably better to keep it concise), but just something to do as a followup for the additional test coverage.

2opremio: Yes, and we shouldn't hardcode the version in the simulation library https://github.com/stellar/soroban-rpc/issues/137

2opremio: https://github.com/stellar/soroban-rpc/issues/138

2opremio:
System tests are not going to work until protocol 21 support is included in quickstart.

cmd/soroban-rpc/internal/preflight/preflight.go
dmkozh: Yes, env itself has to support all the protocol versions. We also have the recording mode/simulation coverage for both protocols 20 and 21 and the plan is for simulation library to always cover the curr/next protocol in the way that I've described in my initial comment.

dmkozh: FWIW Core will run two versions of env during the transition period, but that's for reducing the risks of divergence; I don't think this is relevant/necessary for the simulation.

2opremio: Awesome, thanks!

2opremio:
@dmkozh I enabled protocol 20 integration testing and the simulation library fails with:

```
cannot create budget from network configuration
        	            	
        	            	Caused by:
        	            	    HostError: Error(Budget, InternalError)
        	            	    DebugInfo not available
```

I double checked that we are passing down the right protocol number (20).

All protocol 21 tests are passing without problems.

2opremio:
@sreuland I presume system tests are still using version 20 of the protocol, which would explain why they are failing.

.github/workflows/soroban-rpc.yml
```diff
@@ -105,13 +105,18 @@ jobs:
...
+        go: [ 1.22 ]
+        protocol-version: [ 20, 21 ]
     runs-on: ${{ matrix.os }}
     env:
       SOROBAN_RPC_INTEGRATION_TESTS_ENABLED: true
+      SOROBAN_RPC_INTEGRATION_TESTS_CORE_MAX_SUPPORTED_PROTOCOL: ${{ matrix.protocol-version }}
       SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN: /usr/bin/stellar-core
-      PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 20.2.0-1716.rc3.34d82fc00.focal
+      PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 21.0.0-1812.rc1.a10329cca.focal
+      PROTOCOL_20_CORE_DOCKER_IMG: stellar/unsafe-stellar-core:21.0.0-1812.rc1.a10329cca.focal
```
tamirms, line 117: these values match the protocol 21 environment variables below, is that intentional?

cmd/soroban-rpc/internal/test/docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     # Note: Please keep the image pinned to an immutable tag matching the Captive Core version.
     #       This avoids implicit updates which break compatibility between
     #       the Core container and captive core.
-    image: ${CORE_IMAGE:-stellar/unsafe-stellar-core:20.2.0-1716.rc3.34d82fc00.focal}
+    image: ${CORE_IMAGE:-stellar/unsafe-stellar-core:20.4.1-1807.b152dc51d.focal}
```
tamirms, line 18: should we be using the protocol 21 version of core here?

.github/workflows/soroban-rpc.yml
2opremio: Yep, the protocol 21 should be able to run protocol 20 too

cmd/soroban-rpc/internal/test/docker-compose.yml
2opremio: They are only really set through the `CORE_IMAGE` variable

tamirms (approved):

Issue #110, opened 2024-03-21
add getTransactions endpoint


mollykarcher:
### What problem does your feature solve?

Horizon will cease to serve txmeta sometime in Q2 (see [discord discussion](https://discord.com/channels/897514728459468821/900374272751591424/1199438109796999298)) due to data storage size concerns of Soroban transactions. The ecosystem needs an alternative to this, and we want that alternative to be the RPC. 

### What would you like to see?

Users need to be able to page through all transactions available in RPC's retention window. See this [issue](https://github.com/stellar/soroban-rpc/issues/61) for some prior discussion/context on why we've chosen not to implement an actual stream/websocket/sse implementation here. 

The `getTransactions` endpoint should be able to support pagination and should provide all metadata about the transaction that is currently available via `getTransaction`.

Scope for this will be limited to simple pagination with no filters, and we will work/propose additional filters on this as separate issues. Pagination should follow the format/user experience of the `getEvents` endpoint, _except_ see (see https://github.com/stellar/soroban-rpc/issues/118) for changes that we want to make to pagination and follow the decision there.

#### Request format:
```
{
  "jsonrpc": "2.0",
  "id": <id>,
  "method": "getTransactions",
  "params": {
    "startLedger": <startLedger>,
    "endLedger": <endLedger
    "pagination": {
      "limit": <limit>,
      "cursor": <cursor>
    }
  }
}
```

#### Response format
```
{
  "jsonrpc": "2.0",
  "id": <id>,
  "result": {
    "transactions": [
        {
            "status": <status>,
            "applicationOrder": <order>,
            "envelopeXdr": <envelopeXdr>,
            "resultXdr": <resultXdr>,
            "resultMetaXdr": <resultMetaXdr>,
            "ledger": <ledger>,
            "createdAt": <createdAt>
        }
    ],
    "latestLedger": <latestLedger>,
    "latestLedgerCloseTime": <closeTime>,
    "oldestLedger": <oldestLedger>,
    "oldestLedgerCloseTime": <oldestCloseTime>,
    "cursor": <cursor / pagingToken>
  }
}
```
```[tasklist]
### Tasks
```

mollykarcher:
@psheth9 @aditya1702 I added an example request/response to the description here, per our conversation yesterday. Let me know what you think

Shaptic:
```
    "latestLedger": <latestLedger>,
    "latestLedgerCloseTime": <closeTime>,
    "oldestLedger": <oldestLedger>,
[...]

Issue #111, opened 2024-03-21
add getLedgers endpoint


mollykarcher:
Users need to be able to page through all ledgers available in RPC's retention window. The `getLedgers` endpoint should be able to support pagination and should provide all [ledger header fields](https://developers.stellar.org/docs/learn/encyclopedia/network-configuration/ledger-headers), similar to Horizon's [`ledgers` endpoint](https://developers.stellar.org/network/horizon/resources/ledgers/object)

Scope for this will be limited to simple pagination with no filters, and we will work/propose additional filters on this as separate issues.

**Request format**
```
{
  "jsonrpc": "2.0",
  "id": <id>,
  "method": "getLedgers",
  "params": {
    "startLedger": <startLedger>,
    "pagination": {
      "limit": <limit>,
      "cursor": <cursor>
    }
  }
}
```

**Response format**
```
{
  "jsonrpc": "2.0",
  "id":<id>,
  "result": {
    "ledgers": [
      {
        "id": <id>,
        "hash": <hash>,
        "sequence": <sequence>,
        "transaction_count": <count>,
        "successful_transaction_count": <count>,
        "failed_transaction_count": <count>,
        "operation_count": <count>,
        "tx_set_operation_count": <count>,
        "closed_at": <time>,
        "total_coins": <count>,
        "fee_pool": <fee_pool>,
        "base_fee_in_stroops":<base_fee>,
        "base_reserve_in_stroops": <base_reserve>,
        "max_tx_set_size": <set_size>
      }
    ],
    "latestLedger": <latestLedger>,
    "latestLedgerCloseTime": <closeTime>,
    "oldestLedger": <oldestLedger>,
    "oldestLedgerCloseTime": <oldestCloseTime>,
    "cursor": <cursor / pagingToken>
  }
}
```

The above example represents all fields that are currently in Horizon's response. However, we should assess whether all of these are necessary for RPC.

mollykarcher:
[...]

PR #136, opened 2024-04-15
Add `getTransactions` endpoint

aditya1702:
### What

Implement the `getTransactions` endpoint. It reads the ledger close meta stored in sqlite and decodes the necessary transaction information. 

A quick summary of what is added in the PR:
- RPC handler for `getTransactions`
- Pagination support on the new endpoint
- Unit tests
- Integration tests

### Why

See #110 

### Known limitations

N/A


Shaptic (commented):
just a quick first pass

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,267 @@
...
+	err = ledgerRange.isValid(ctx, h.ledgerReader, h.ledgerEntryReader)
+	if err != nil {
+		return GetTransactionsResponse{}, &jrpc2.Error{
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+
+	// Get all ledgers within the range from db
+	ledgers, err := h.ledgerReader.GetLedgers(ctx, ledgerRange.Start.LedgerSequence, ledgerRange.End.LedgerSequence)
```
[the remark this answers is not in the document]
aditya1702: @Shaptic Yes I just quickly added this so that I can do a quick implementation. This is something we should discuss though - doing it one ledger at a time vs in batches.

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,267 @@
...
+
+	start := transactions.NewCursor(request.StartLedger, 0, 0)
+	endLedger, found, err := h.ledgerReader.GetLedger(ctx, request.EndLedger)
+	if err != nil || !found {
+		return GetTransactionsResponse{}, &jrpc2.Error{
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+	end := transactions.NewCursor(request.EndLedger, uint32(endLedger.CountTransactions()), 0)
```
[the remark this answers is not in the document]
aditya1702: @Shaptic Yeah I still need to make sure the indexing is proper everywhere.

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -69,6 +70,21 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
 	}
 }
 
+func (r ledgerReader) GetLedgers(ctx context.Context, startSequence uint32, endSequence uint32) ([]xdr.LedgerCloseMeta, error) {
```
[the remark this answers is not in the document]
aditya1702: @Shaptic Yeah good idea. Will add that

cmd/soroban-rpc/internal/methods/get_transactions.go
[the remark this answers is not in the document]
aditya1702: Actually I will start with just reading one ledger at a time since the pagination limit will be on transactions which would be hit sooner than we finish off all the ledgers in the range.

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -111,26 +111,30 @@ LedgerLoop:
...
+			// Seek returns EOF so we can move onto next ledger
+			continue
 		}
 
 		// Decode transaction info from ledger meta
 		txCount := ledger.CountTransactions()
-		for i := start.TxIdx; i < uint32(txCount); i++ {
-			cursor = transactions.NewCursor(ledger.LedgerSequence(), i, 0)
+		for i := startTxIdx; i < txCount; i++ {
+			cursor = transactions.NewCursor(ledger.LedgerSequence(), uint32(i), 0)
```
Shaptic, line 140: `< txCount` is wrong, right? It should be `txCount - startTxIdx` (maybe +/-1? check me). Or, better might just be to iterate from `startTxIdx` until you get `io.EOF` from the reader

aditya1702: @Shaptic No the `txCount` would work here because only the startIdx is shifting towards the end. And we get the startIdx from the cursor which is the last txIdx we read from it. The indexing is not changing here.

cmd/soroban-rpc/internal/events/cursor_test.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func TestParseCursor(t *testing.T) {
-	for _, cursor := range []Cursor{
+	for _, cursor := range []*Cursor{
```
Shaptic, line 12: this doesn't seem necessary?

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,242 @@
...
+	// Iterate through each ledger and its transactions until limit or end range is reached
+	var txns []TransactionInfo
+	var cursor transactions.Cursor
+LedgerLoop:
+	for ledgerSeq := start.LedgerSequence; ledgerSeq <= request.EndLedger; ledgerSeq++ {
+		// Get ledger close meta from db
+		ledger, found, err := h.ledgerReader.GetLedger(ctx, ledgerSeq)
+		if (err != nil) || (!found) {
+			if err == nil {
+				err = errors.New("ledger close meta not found")
```
Shaptic, line 109: suggestion: add `ledgerSeq` into the error message

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,242 @@
...
+		// Move the reader to specific tx idx
+		startTxIdx := 0
+		if ledgerSeq == start.LedgerSequence {
+			startTxIdx = int(start.TxIdx)
+		}
+		err = reader.Seek(startTxIdx)
+		if err != nil {
+			// Seek returns EOF so we can move onto next ledger
+			continue
+		}
```
Shaptic, line 135: I know Seek only ever returns `io.EOF || nil` but an explicit check would be defensive here

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,242 @@
...
+		LatestLedgerCloseTimestamp: latestLedgerCloseTime,
+		Pagination: &TransactionsPaginationOptions{
+			Cursor: &cursor,
+			Limit:  limit,
+		},
+	}, nil
+
+}
+
+func (h *transactionsRPCHandler) getTransactionDetails(tx ingest.LedgerTransaction) ([]byte, []byte, []byte, error) {
```
Shaptic, line 195: TODO: Add docs to describe each return value, otherwise it's not obvious without reading the code

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+type TransactionInfo struct {
+	Result           []byte `json:"result"`
+	Meta             []byte `json:"meta"`
+	Envelope         []byte `json:"envelope"`
+	FeeBump          bool   `json:"feeBump"`
+	ApplicationOrder int32  `json:"applicationOrder"`
+	Successful       bool   `json:"successful"`
+	LedgerSequence   uint   `json:"ledgerSequence"`
```
Shaptic, line 57: `uint32` is fine here, no?

cmd/soroban-rpc/internal/methods/get_transactions.go
Shaptic: suppose startTxIdx = 5 and txCount = 10, what I'm saying is that this will definitely encounter EOF after just 5 iterations and return an error (EOF check should be a `break` I think)

Shaptic: nvm I can't read

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -247,6 +247,28 @@ func (cfg *Config) options() ConfigOptions {
...
+			Name:         "max-transactions-limit",
+			Usage:        "Maximum amount of transactions allowed in a single getTransactions response",
+			ConfigKey:    &cfg.MaxTransactionsLimit,
+			DefaultValue: uint(10000),
+		},
+		{
+			Name:         "default-transactions-limit",
+			Usage:        "Default cap on the amount of transactions included in a single getTransactions response",
+			ConfigKey:    &cfg.DefaultTransactionsLimit,
+			DefaultValue: uint(100),
```
Shaptic, line 260: let's do 10 by default with a max of 200

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,242 @@
...
+
+	latestLedger, found, err := h.ledgerReader.GetLedger(ctx, latestSequence)
+	if (err != nil) || (!found) {
+		if err == nil {
+			err = errors.New("ledger close meta not found")
+		}
+		return 0, 0, err
+	}
+
+	return int64(latestSequence), int64(latestLedger.LedgerHeaderHistoryEntry().Header.ScpValue.CloseTime), nil
```
Shaptic, line 226: there's a `latestLedger.LedgerCloseTime()`

cmd/soroban-rpc/internal/methods/get_transactions.go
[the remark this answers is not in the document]
aditya1702: We dont seem to use that convention in other rpc methods - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/get_events.go#L71.

I found that Google's JSON guide uses and recommends camelCase - https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format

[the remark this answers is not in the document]
Shaptic: sgtm! thanks for checking

cmd/soroban-rpc/internal/events/cursor_test.go
aditya1702: @Shaptic There were some receivers in https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/events/cursor.go#L45 that use pointer and some dont. For sake of uniformity, I converted all receivers to pointers. Do not know if that is a bad idea. 

Maybe@tamirms @2opremio have any suggestions?

tamirms: a Cursor is an immutable type so it should not be using pointer receivers except for UnmarshalJSON(). I don't think there should be any changes to cmd/soroban-rpc/internal/events/cursor.go

cmd/soroban-rpc/internal/transactions/cursor.go
```diff
@@ -0,0 +1,99 @@
+package transactions
```
tamirms, line 1: this code is not necessary. instead use the `stellar/go/toid` package

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,259 @@
...
+// TransactionsPaginationOptions defines the available options for paginating through transactions.
+type TransactionsPaginationOptions struct {
+	Cursor *transactions.Cursor `json:"cursor,omitempty"`
+	Limit  uint                 `json:"limit,omitempty"`
+}
+
+// GetTransactionsRequest represents the request parameters for fetching transactions within a range of ledgers.
+type GetTransactionsRequest struct {
+	StartLedger uint32                         `json:"startLedger"`
+	EndLedger   uint32                         `json:"endLedger"`
```
tamirms, line 27: should EndLedger be optional? the limit will implicitly define the end ledger

Shaptic: the limit is on the number of transactions, not ledgers, right? so not necessarily

aditya1702: @tamirms Yes, as George pointed out, the limit is number of transactions, not ledgers. And from a user perspective, it would be much easier to just specify the end ledger rather than the API figuring it out for them.

tamirms: @aditya1702 

The `endLedger` filter is only useful if there exists some ledger range where the number of transactions in `[startLedger, endLedger]` is less than `limit`. However, if you take a look at the number of successful + failed transactions included in recent ledgers, https://horizon.stellar.org/ledgers?order=desc , you will find the sum is greater than the default `max-transactions-limit` value (200). So, for pretty much all requests the `limit` parameter will end up restricting how many transactions are included in the response and not the `endLedger` parameter.

Even if there were ledger ranges which had fewer transactions than the requested limit, the client could simply just ignore the transactions in the response which are outside the desired ledger range.

BTW the horizon `/transactions` endpoint does not have an `endLedger` parameter and I'm not aware of any complaints about that being a missing feature.

Some other differences between this endpoint and the horizon `/transactions` endpoint:

- `startLedger` is not necessary because the `startLedger` can be encoded is a toid and included as a cursor
- horizon has an `order` parameter which allows you to return transactions in ascending or descending order which is useful because it provides an easy way to query the newest or oldest transactions

Shaptic: We should keep in mind, though, that the purpose of this new endpoint is to facilitate, as @tomerweller put it, "poor man's ingestion." With that in mind, we don't need `order` and can optimize for the forward-moving case, and I think a `startLedger` still makes sense given the fact that TOIDs are too opaque for users to be able to pass along without an SDK to help them. Users won't see a cursor until they make an initial request. Maybe we can do `startLedger xor cursor`?

aditya1702: @Shaptic I agree and that is what I have added in this implementation  (similar to `getEvents`) -  on the first request, user will specify a `startLedger` value and then on the subsequent requests they can just use the cursor returned from last response.

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,259 @@
...
+	ApplicationOrder int32  `json:"applicationOrder"`
+	Successful       bool   `json:"successful"`
+	LedgerSequence   uint32 `json:"ledgerSequence"`
+}
+
+// GetTransactionsResponse encapsulates the response structure for getTransactions queries.
+type GetTransactionsResponse struct {
+	Transactions               []TransactionInfo `json:"transactions"`
+	LatestLedger               int64             `json:"latestLedger"`
+	LatestLedgerCloseTimestamp int64             `json:"latestLedgerCloseTimestamp"`
```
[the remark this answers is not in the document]
aditya1702: @Shaptic I understand the use-case for having `LatestLedger` but what purpose would `OldestLedger` serve here apart from keeping it consistent with the getTransaction endpoint?

cmd/soroban-rpc/internal/methods/get_transactions.go
tamirms: Without the order parameter it is not possible retrieve the last n ledgers within a single request.

startLedger xor cursor is fine by me. But I’m not seeing the utility of the end ledger parameter

cmd/soroban-rpc/internal/methods/get_transactions.go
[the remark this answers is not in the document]
Shaptic: the intent of `OldestLedger`, afaict, is to indicate to the user what the RPC instance's history range is

cmd/soroban-rpc/internal/methods/get_transactions.go
Shaptic: I agree, I don't think we need it since it's covered implicitly by `limit`. As for 

> the last n ledgers 

we can add this once it's requested instead of preemptively.

aditya1702: Resolving this since we all agreed to remove endLedger in the platform discussion

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -206,6 +206,13 @@ func NewJSONRPCHandler(cfg *config.Config, params HandlerParams) Handler {
 			queueLimit:           cfg.RequestBacklogGetTransactionQueueLimit,
 			requestDurationLimit: cfg.MaxGetTransactionExecutionDuration,
 		},
+		{
+			methodName:           "getTransactions",
+			underlyingHandler:    methods.NewGetTransactionsHandler(params.Logger, params.LedgerReader, params.LedgerEntryReader, cfg.MaxTransactionsLimit, cfg.DefaultTransactionsLimit, cfg.NetworkPassphrase),
+			longName:             "get_transactions",
+			queueLimit:           cfg.RequestBacklogSendTransactionQueueLimit,
```
sreuland, line 213: just driving by, noticed `getTransactions` is applying existing tx sub queue and duration limit cfg values, should it be using a new pair of queue/duration limit config keys, e.g. `cfg.RequestBacklogGetTransactionsQueueLimit` , it seems like the request/load profiles would be different.

aditya1702: Ah good catch. Yes it should be using a different config separate to this endpoint

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,195 @@
...
+	if req.Pagination != nil && req.Pagination.Limit > maxLimit {
+		return fmt.Errorf("limit must not exceed %d", maxLimit)
+	}
+
+	return nil
+}
+
+// GetTransactionsResponse encapsulates the response structure for getTransactions queries.
+type GetTransactionsResponse struct {
+	Transactions          []db.Transaction `json:"transactions"`
```
tamirms, line 50: I don't think `db.Transaction` is the right representation to include in the response. The `db.Transaction` struct doesn't define any json tags on any of its fields

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,195 @@
...
+		reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(h.networkPassphrase, ledger)
+		if err != nil {
+			return GetTransactionsResponse{}, &jrpc2.Error{
+				Code:    jrpc2.InvalidParams,
+				Message: err.Error(),
+			}
+		}
+
+		// Move the reader to specific tx idx
+		startTxIdx := 0
```
tamirms, line 129: the transaction order in the toid starts from 1

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,195 @@
...
+				return GetTransactionsResponse{}, &jrpc2.Error{
+					Code:    jrpc2.InvalidParams,
+					Message: err.Error(),
+				}
+			}
+		}
+
+		// Decode transaction info from ledger meta
+		txCount := ledger.CountTransactions()
+		for i := startTxIdx; i < txCount; i++ {
```
tamirms, line 142: the last transaction order can equal txCount since the index begins at 1

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @tamirms I was trying to avoid defining a duplicate struct and just reuse this. Since db.Transaction has exactly all the fields we want to return in the response.

tamirms: I understand but does the get transactions json response actually match the specification in the issue?

aditya1702: I agree. It is better to add a separate struct

cmd/soroban-rpc/internal/db/mocks.go
```diff
@@ -72,12 +76,31 @@ func (txn *mockTransactionHandler) GetTransaction(ctx context.Context, hash xdr.
...
-		itx, err := ParseTransaction(*txn.ledgers[hash.HexString()], tx)
+		itx, err := ParseTransaction(*txn.txHashToMeta[hash.HexString()], tx)
 		return itx, txn.ledgerRange, err
 	}
 }
 
 func (txn *mockTransactionHandler) RegisterMetrics(_, _ prometheus.Observer) {}
 
-var _ TransactionReader = &mockTransactionHandler{}
-var _ TransactionWriter = &mockTransactionHandler{}
```
Shaptic, line 83: why remove? :confused:

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+type transactionInfo struct {
+	// Successful indicates whether the transaction was successful or not
+	Successful bool `json:"status"`
+	// ApplicationOrder is the index of the transaction among all the transactions
+	// for that ledger.
+	ApplicationOrder int32 `json:"applicationOrder,omitempty"`
+	// FeeBump indicates whether the transaction is a feebump transaction
+	FeeBump bool `json:"feeBump,omitempty"`
```
Shaptic, line 56: what would `omitempty` mean in this context?

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,230 @@
...
+	err = request.isValid(h.maxLimit, ledgerRange)
+	if err != nil {
+		return GetTransactionsResponse{}, &jrpc2.Error{
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+
+	// Move start to pagination cursor
+	start := toid.New(int32(request.StartLedger), 1, 0)
```
Shaptic, line 111: operation index should likewise start at 1, this is defined in the [TOID spec](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0035.md): 

> **Operation Index:** the index of the operation within its parent transaction. This takes up 12 bits of the id, from bits 52 to 63, inclusive. The operation index value starts at 1. The maximum supported number of operations per transaction is 4095.

cmd/soroban-rpc/internal/db/mocks.go
Shaptic: in fact you should instead add a line ensuring `mockLedgerReader` conforms to the `db.LedgerReader` interface

Shaptic (commented):
Super close :clap: couple more minor comments and this should be ready to merge

cmd/soroban-rpc/internal/db/mocks.go
aditya1702: @Shaptic Ah I think I did not realise that these lines are actually needed so that it gives an error if the structs stop implementing the interfaces. I will add them again and also the mockLedgerReader

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: Copied it from getTransaction response but did not realise the omitempty tags were present. Except the diagnosticEventsXDR, it is not required here

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	// Iterate through each ledger and its transactions until limit or end range is reached.
+	// The latest ledger acts as the end ledger range for the request.
+	var txns []transactionInfo
+	var cursor *toid.ID
+LedgerLoop:
+	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
+		// Get ledger close meta from db
+		ledger, found, err := h.ledgerReader.GetLedger(ctx, uint32(ledgerSeq))
```
Shaptic, line 132: This is a thought for the future: we should probably do something more like `getLedgers()` and then process ledgers in batches so that we don't have to have one SQL query for each iteration (while still respecting `limit`, of course, but a little extra memory is worth saving queries).

We can include this as part of (or after, at the very least) the `getLedgers` work in #111.

Shaptic (approved):
LGTM, nice work :clap: but wait for another look from @tamirms before merging

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if req.Pagination != nil && req.Pagination.Limit > maxLimit {
+		return fmt.Errorf("limit must not exceed %d", maxLimit)
+	}
+
+	return nil
+}
+
+type transactionInfo struct {
```
tamirms, line 49: this struct should be public because it is referenced in other packages (namely test/get_transactions_test.go)

cmd/soroban-rpc/internal/test/get_transactions_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/toid"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/methods"
+)
+
+const (
+	Cursor int64 = 55834578945
```
tamirms, line 19: it's not clear what's the significance of 55834578945 and how it's used in TestGetTransactions(). How does the test guarantee there will be 2 transactions after this specific cursor?

cmd/soroban-rpc/internal/test/get_transactions_test.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// buildTxParams constructs the parameters necessary for creating a transaction from the given account.
+//
+// account - the source account from which the transaction will originate. This account provides the starting sequence number.
+//
+// Returns a fully populated TransactionParams structure.
+func buildTxParams(account txnbuild.SimpleAccount) txnbuild.TransactionParams {
```
tamirms, line 27: this seems to build a set options transaction. perhaps this function should be renamed to `buildSetOptionsTxParams()`>

cmd/soroban-rpc/internal/test/get_transactions_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var hashes []string
+	for i := 0; i <= 2; i++ {
+		account := txnbuild.NewSimpleAccount(address, int64(i))
+		tx, err := txnbuild.NewTransaction(buildTxParams(account))
+		assert.NoError(t, err)
+
+		hash, err := sendTransaction(t, client, kp, tx)
+		assert.NoError(t, err)
+		hashes = append(hashes, hash)
+		time.Sleep(1 * time.Second)
```
tamirms, line 85: why is the sleep necessary?

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @tamirms This struct is not referenced anywhere except the get_transactions.go file.

cmd/soroban-rpc/internal/test/get_transactions_test.go
aditya1702: @tamirms I took the cursor string value coming from the previous request's result - https://github.com/stellar/soroban-rpc/pull/136/files/2d660841df55a3e5d5b569c07948ea66aad1481d#diff-dd051ad5301911cdd3964f2129a8d3e93066ca9bd9a9490166fdab6cf28d3952R123.

cmd/soroban-rpc/internal/test/get_transactions_test.go
aditya1702: Because if I send the txns one after the other, Core gives a `TRY_AGAIN_LATER` error response. This means its currently unavailable to perform requests.

cmd/soroban-rpc/internal/methods/get_transactions.go
tamirms: https://github.com/stellar/soroban-rpc/blob/2d660841df55a3e5d5b569c07948ea66aad1481d/cmd/soroban-rpc/internal/test/get_transactions_test.go#L112-L114

here there is code outside the methods package which refers to fields defined in this struct

cmd/soroban-rpc/internal/test/get_transactions_test.go
tamirms: I looked at the code you linked to and I don't see where you extract the cursor out from the response

cmd/soroban-rpc/internal/test/get_transactions_test.go
tamirms: there is a `sendSuccessfulTransaction()` helper function which is used in many of the other integration tests. You can try using that instead of creating your helper function which does roughly the same thing

cmd/soroban-rpc/internal/test/get_transactions_test.go
aditya1702: @tamirms I had manually recorded the value of the previous result's cursor and added it to the constant variable. However, I realise that is confusing so I have refactored the code to directly convert the string cursor and use it instead - https://github.com/stellar/soroban-rpc/pull/136/files#diff-dd051ad5301911cdd3964f2129a8d3e93066ca9bd9a9490166fdab6cf28d3952R92

Lmk if this works now. Thanks

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/toid"
+
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/ledgerbucketwindow"
+)
+
+// TransactionsPaginationOptions defines the available options for paginating through transactions.
+type TransactionsPaginationOptions struct {
+	Cursor *toid.ID `json:"cursor,omitempty"`
```
tamirms, line 22: the cursor type in the request does not match the cursor type in GetTransactionsResponse. The cursor you provide in the request should probably be a string and not a json object

aditya1702: @tamirms I used the json object because that is how we use it for getEvents pagination options. If we change this cursor to string, we should also do it for the events cursor to keep it same. What do you think?

tamirms: @aditya1702 the curser for getEvents is a string.  https://developers.stellar.org/network/soroban-rpc/api-reference/methods/getEvents

the events cursor struct has a custom json implementation which serializes the struct into a string

https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/events/cursor.go#L39-L57

aditya1702: @tamirms Gotcha. So in this case I would have to explicitly parse the cursor string using toid.Parse right? As I have done in this test - https://github.com/stellar/soroban-rpc/pull/136/files#diff-dd051ad5301911cdd3964f2129a8d3e93066ca9bd9a9490166fdab6cf28d3952R92

tamirms: yeah, you can make it a string and parse the value into a toid.ID or you could add a wrapper around toid.ID which defines json marshaling / unmarshaling functions

aditya1702: I will just parse it as string and convert to toid.ID

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @Shaptic Yeah I agree. We should do the batch processing as its more efficient.

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -135,7 +144,7 @@ LedgerLoop:
 				err = errors.Errorf("ledger close meta not found: %d", ledgerSeq)
 			}
 			return GetTransactionsResponse{}, &jrpc2.Error{
-				Code:    jrpc2.InvalidParams,
+				Code:    jrpc2.InternalError,
```
Shaptic, line 147: why internal now? if it's not in the DB then their request is out of bounds

Shaptic: same comment for all of the below error changes imo

aditya1702: @Shaptic Yeah you are correct. I refactored this code again to use InvalidParams for "ledger close meta not found" or for any other db related error InternalError code.

However, for these 3 errors:
- https://github.com/stellar/soroban-rpc/pull/136/files#diff-5276db8001730f32b267cab637b316958b69aef6f8184343338855fdc118b700R158
- https://github.com/stellar/soroban-rpc/pull/136/files#diff-5276db8001730f32b267cab637b316958b69aef6f8184343338855fdc118b700R169
- https://github.com/stellar/soroban-rpc/pull/136/files#diff-5276db8001730f32b267cab637b316958b69aef6f8184343338855fdc118b700R169

InternalError would make more sense dont you think? Since they are generated due to some internal parsing issues?

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -0,0 +1,241 @@
...
+		if request.Pagination.Cursor != "" {
+			cursorInt, err := strconv.ParseInt(request.Pagination.Cursor, 10, 64)
+			if err != nil {
+				return GetTransactionsResponse{}, &jrpc2.Error{
+					Code:    jrpc2.InvalidParams,
+					Message: err.Error(),
+				}
+			}
+
+			*start = toid.Parse(cursorInt)
```
tamirms, line 124: I think you should validate the cursor here

tamirms (approved):

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -139,14 +139,14 @@ LedgerLoop:
 	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
 		// Get ledger close meta from db
 		ledger, found, err := h.ledgerReader.GetLedger(ctx, uint32(ledgerSeq))
-		if (err != nil) || (!found) {
-			code := jrpc2.InternalError
-			if err == nil {
-				err = errors.Errorf("ledger close meta not found: %d", ledgerSeq)
-				code = jrpc2.InvalidParams
+		if !found {
```
tamirms, line 142: you need to check `err` first because it is possible that `err` is not nil `found` is false

aditya1702: Ah the default case. Gotcha

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @tamirms Is this what you had in mind for validation?

`if start.String() != request.Pagination.Cursor {
		return GetTransactionsResponse{}, &jrpc2.Error{
					Code:    jrpc2.InternalError,
					Message: fmt.Sprintf("could not parse cursor: expected %s, got %s", request.Pagination.Cursor, start.String()),
				}
}`

tamirms: @aditya1702 sorry, I think you can ignore this comment. I think the check above `cursorInt, err := strconv.ParseInt(request.Pagination.Cursor, 10, 64)` should cover it

Issue #115, opened 2024-03-27
[epic] move in-memory datastructures to the database


2opremio:
### What problem does your feature solve?

Reduces memory consumption so that we can extend the default history retention to one week (currently at 2 hours for transactions and 24 hours for events). See [the analysis document](https://docs.google.com/document/d/1NSrOuogVTHKTk1QLuMS-d4u-aT20ZeAzIJqiqPn0RHg/edit#heading=h.1my3tarepa0c) for details.

### What would you like to see?

I would like to see the in-memory data-structures (Currently based around `LedgerBucketWindow`) stored in the database instead of in-memory.

Regarding the DB schema, we can either roughly mirror what's kept in memory or use some sort of indexing on the txmeta which is already kept in the DB.

We should also have [the bigger picture](https://docs.google.com/document/d/1tTuzL7HBHedp9TE-LonD1QV67SaBSReQ8YFj9xQIgtA/edit#heading=h.iacejigpgv59) in mind when modifying the schema. The chosen schema should consider extensibility to implement the new Soroban RPC endoints.

I suggest presenting a schema design which should be peer reviewed before jumping to the implementation.

As an output of this ticket we should also get a performance comparison (e.g. through a purposely built Golang benchmark)

### What alternatives are there?

The performance shouldn't considerably degrade. If it does, we should consider caching recent data (particularly for transactions). However, a hybrid in-memory/db approach increases complexity.

## Subtasks

```[tasklist]
### Design
- [x] Design doc for Transactions: https://github.com/stellar/soroban-rpc/issues/130
- [x] Design doc for Events: https://github.com/stellar/soroban-rpc/issues/131
```

```[tasklist]
### Implementation (Transactions)
- [x] Add support for reading transactions with an offset (https://github.com/stellar/go/pull/5274)
- [ ] https://github.com/stellar/soroban-rpc/issues/153
- [ ] https://github.com/stellar/soroban-rpc/issues/140
- [ ] https://github.com/stellar/soroban-rpc/issues/154
- [ ] https://github.com/stellar/soroban-rpc/issues/155
- [ ] https://github.com/stellar/soroban-rpc/issues/110
```

```[tasklist]
### Implementation \[Events]
- [ ] https://github.com/stellar/soroban-rpc/issues/189
- [ ] https://github.com/stellar/soroban-rpc/issues/190
- [ ] https://github.com/stellar/soroban-rpc/issues/191
[...]

PR #141, opened 2024-04-16
Increase `getTransaction` window via ingestion into a local database

Shaptic:
### What
This changes the handler for `getTransactions` to be a new, DB-connected `TransactionHandler` structure which supports two methods: fetching a specific transaction hash and fetching the ingested ledger range (oldest & latest).

To support fetching this, there is a new database migration with a table `tx hash | ledger seq | apply order` that gets filled on each ledger close. Note that when actually fetching the ledger, it is extracted on-the-fly from the `LedgerCloseMeta` via a `JOIN` rather than from a denormalized database row, per the design doc.

### Why
See #140 as part of #115.

### Known limitations
n/a


cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	rows := sq.
+		Select("t.application_order", "lcm.meta").
+		From(fmt.Sprintf("%s t", transactionTableName)).
+		Join(fmt.Sprintf("%s lcm ON (t.ledger_sequence = lcm.sequence)", ledgerCloseMetaTableName)).
+		Where(sq.Eq{"hash": rawHash}).
+		Limit(1).
+		RunWith(txn.stmtCache).
+		QueryRow()
```
Shaptic, line 132: @sydneynotthecity @chowbao is this a dumb way to do this? I think the `JOIN` is inevitable. Also, would `USING (sequence)` be superior in any way?

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,203 @@
...
+		return xdr.LedgerCloseMeta{}, ingest.LedgerTransaction{}, err
+	}
+
+	var lcm xdr.LedgerCloseMeta
+	if err := lcm.UnmarshalBinary(row.meta); err != nil {
+		return lcm, ingest.LedgerTransaction{}, err
+	}
+
+	reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(txn.passphrase, lcm)
+	reader.Seek(row.txIndex - 1)
```
Shaptic, line 148: this is expected to fail until https://github.com/stellar/go/pull/5274 gets in

cmd/soroban-rpc/internal/db/transaction.go
sydneynotthecity: `USING (sequence)` is not superior in any way. Some argue it's easier to read, but the SQL interpreter will resolve the two the exact same way.

The biggest performance gain you'll get for this query is to confirm that the first listed table is indexed by your join key. ie, does the `transactions` table have a index on `sequence`? or does the `lcm`? Whichever table contains that index needs to come first.

sydneynotthecity: There are implicit indexes on primary keys, but not foreign keys.

I'm not super familiar with the sqlite query planner--you may be right that the `WHERE` clause kicks in first....but it could also be the `JOIN` 🤷‍♀️ I think it might be worthwhile to just run an `EXPLAIN ANALYZE` against a join version of the query vs subquery to see if there is any performance differences. My guess is the differences will be negligible. If you do change the query, I would opt for a subquery instead of two queries. The query planner will resolve as two queries anyway.

Shaptic: Good call, thanks! It looks like it does the smart thing for this version, anyway:

```
EXPLAIN QUERY PLAN 
  SELECT t.application_order, lcm.meta 
  FROM transactions t JOIN ledger_close_meta lcm ON (t.ledger_sequence = lcm.sequence) 
  WHERE t.hash = '29994b6f84a0fcd19b5bc5f35e692f418feb4ea3ef9a2edf71c61c4f217e079d'
  LIMIT 1;

QUERY PLAN
|--SEARCH t USING INDEX sqlite_autoindex_transactions_1 (hash=?)
`--SEARCH lcm USING INTEGER PRIMARY KEY (rowid=?)
```

which tells me, if I'm reading it right, that it does the `WHERE` first using the pkey index, then uses the pkey index on the `ledger_sequence` from that result row :ok_hand: which is exactly what I expanded it to anyway.

Yay for smort planners. Thanks again!

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
```diff
@@ -0,0 +1,16 @@
...
+
+-- index to find transactions in ledgers by hash
+CREATE TABLE transactions (
+    hash NCHAR(64) PRIMARY KEY,
+    ledger_sequence INTEGER NOT NULL,
+    application_order INTEGER NOT NULL,
+    -- possibly a useful search index for Soroban-oriented queries, e.g.
+    -- transactions that involve the SAC or have just events in general
+    is_soroban BOOLEAN NOT NULL,
+    FOREIGN KEY (ledger_sequence)
```
tamirms, line 11: what's the motivation for using a foreign key constraint here? support for foreign key constraints are disabled in sqlite by default

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+-- index to find transactions in ledgers by hash
+CREATE TABLE transactions (
+    hash NCHAR(64) PRIMARY KEY,
+    ledger_sequence INTEGER NOT NULL,
+    application_order INTEGER NOT NULL,
+    -- possibly a useful search index for Soroban-oriented queries, e.g.
+    -- transactions that involve the SAC or have just events in general
+    is_soroban BOOLEAN NOT NULL,
```
tamirms, line 10: I think it's better to remove this column unless there is an immediate need. BTW, there is no index on this column so it won't be efficient to filter by `is_soroban`. If in the future, we need to implement an indexing feature for soroban transactions we can always add this column and an index in a subsequent migration. Given that soroban-rpc has a limited retention window, it isn't as expensive to reingest compared to horizon

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+-- index to find transactions in ledgers by hash
+CREATE TABLE transactions (
+    hash NCHAR(64) PRIMARY KEY,
+    ledger_sequence INTEGER NOT NULL,
```
tamirms, line 6: should there be an index on this column so you can trim the transactions table as transactions fall outside the retention window?

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+-- index to find transactions in ledgers by hash
+CREATE TABLE transactions (
+    hash NCHAR(64) PRIMARY KEY,
```
tamirms, line 5: if you store the hash in binary instead of the hexadecimal form you only require 32 bytes

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -133,6 +133,7 @@ func newCaptiveCore(cfg *config.Config, logger *supportlog.Entry) (*ledgerbacken
 func MustNew(cfg *config.Config) *Daemon {
 	logger := supportlog.New()
 	logger.SetLevel(cfg.LogLevel)
+	supportlog.SetLevel(cfg.LogLevel)
```
tamirms, line 136: why is this change included?

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+	return ledgerRange
+}
+
+// GetTransaction conforms to the interface in
+// methods/get_transaction.go#NewGetTransactionHandler so that it can be used
+// directly against the RPC handler.
+//
+// Errors (i.e. the bool being false) should only occur if the XDR is out of
+// date or is otherwise incorrect/corrupted or the tx really isn't in the DB.
+func (txn *transactionReaderTx) GetTransaction(log *log.Entry, hash xdr.Hash) (
```
tamirms, line 266: instead of taking a logger parameter, this function should return an error and then the caller can decide how to handle the error (e.g. if it's worth logging)

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -236,17 +229,48 @@ func MustNew(cfg *config.Config) *Daemon {
...
+		[]string{"operation"},
+	)
+	txCountMetric := prometheus.NewSummary(prometheus.SummaryOpts{
+		Namespace: daemon.MetricsNamespace(), Subsystem: "transactions",
+		Name:       "count",
+		Help:       "count of transactions ingested, sliding window = 10m",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+	})
+
+	daemon.metricsRegistry.MustRegister(txDurationMetric, txCountMetric)
```
tamirms, line 249: these metrics should be defined an encapsulated in the component which uses the metrics (e.g. cmd/soroban-rpc/internal/db/transaction.go )

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -265,3 +311,22 @@ func runMigrations(db *sql.DB, dialect string) error {
 	_, err := migrate.ExecMax(db, dialect, m, migrate.Up, 0)
 	return err
 }
+
+// NewTestDB is here so that other packages in sRPC can use it
```
tamirms, line 315: The db functionality should be unit tested within this package. For example, the code in cmd/soroban-rpc/internal/db/transaction.go should be unit tested here instead of in  cmd/soroban-rpc/internal/methods/get_transaction_test.go . In cmd/soroban-rpc/internal/methods/get_transaction_test.go , you can use mocks for the transaction store. And in cmd/soroban-rpc/internal/test/transaction_test.go we have the integration tests which exercises everything end to end with real implementations

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+		hexHash := hex.EncodeToString(hash[:])
+		is_soroban := (tx.UnsafeMeta.V == 3 && tx.UnsafeMeta.V3.SorobanMeta != nil)
+		query = query.Values(hexHash, lcm.LedgerSequence(), tx.Index, is_soroban)
+	}
+	_, err = query.RunWith(txn.stmtCache).Exec()
+
+	L.WithError(err).
+		WithField("duration", time.Since(start)).
+		WithField("sql_duration", time.Since(mid)).
+		Infof("Ingested %d transaction lookups", len(transactions))
```
tamirms, line 135: I don't think this is worth logging because we have metrics reporting this info

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+		// and inner hashes.
+		if tx.Envelope.IsFeeBump() {
+			transactions[tx.Result.InnerHash()] = tx
+		}
+		transactions[tx.Result.TransactionHash] = tx
+	}
+
+	mid := time.Now()
+	L.WithField("passphrase", txn.passphrase).
+		Debugf("Ingesting %d transaction lookups from ledger", len(transactions))
```
tamirms, line 121: I don't think this is worth logging

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+	L := log.
+		WithField("ledger_seq", lcm.LedgerSequence()).
+		WithField("tx_count", txCount)
+
+	if txn.stmtCache == nil {
+		return errors.New("TransactionWriter incorrectly initialized without stmtCache")
+	}
+
+	if txCount == 0 {
+		L.Warnf("No transactions present in ledger: %+v", lcm)
```
tamirms, line 93: I don't think this is worth logging. If you're running against  stand alone network there will be lots of these cases logged

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+}
+
+func NewTransactionWriter(db db.SessionInterface, stmtCache *sq.StmtCache, passphrase string) TransactionWriter {
+	return &transactionHandler{db: db, stmtCache: stmtCache, passphrase: passphrase}
+}
+
+func (txn *transactionHandler) InsertTransactions(lcm xdr.LedgerCloseMeta) error {
+	start := time.Now()
+	txCount := lcm.CountTransactions()
+	L := log.
```
tamirms, line 84: the pattern soroban rpc has adapted is having a logger be passed in to each component. you can checkout how it's done in cmd/soroban-rpc/internal/ingest/service.go . Also, I noticed that we don't log anything in any of the other db components so I think it makes sense to avoid logging here so that we can have a consistent style.

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	return &transactionReaderTx{ctx: ctx, db: sesh, passphrase: txn.passphrase}, nil
+}
+
+// GetLedgerRange will fetch from the database in an isolated tx.
+func (txn *transactionHandler) GetLedgerRange(ctx context.Context) ledgerbucketwindow.LedgerRange {
+	reader, err := txn.NewTx(ctx)
+	defer reader.Done()
+	if err != nil {
```
tamirms, line 182: the error should be returned so the caller can be able to interpret it

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+}
+
+func (txn *transactionReaderTx) Done() error {
+	return txn.db.Rollback()
+}
+
+// GetLedgerRange pulls the min/max ledger sequence numbers from the database.
+func (txn *transactionReaderTx) GetLedgerRange() ledgerbucketwindow.LedgerRange {
+	var ledgerRange ledgerbucketwindow.LedgerRange
+	log.Debugf("Retrieving ledger range from database: %+v", txn)
```
tamirms, line 197: I don't think we should be logging anything here. Most of the log invocations are for error cases. Instead of doing that we should return the error and let the caller handle it as they wish (e.g. it can be logged on the caller side)

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+type TransactionWriter interface {
+	InsertTransactions(lcm xdr.LedgerCloseMeta) error
+	RegisterMetrics(ingest, insert, count prometheus.Observer)
+}
+
+// TransactionReader is used to serve requests and just returns a cloned,
+// read-only DB session to perform actual DB reads on.
+type TransactionReader interface {
+	NewTx(ctx context.Context) (TransactionReaderTx, error)
+	GetLedgerRange(ctx context.Context) ledgerbucketwindow.LedgerRange
```
tamirms, line 48: is it necessary to define a GetLedgerRange() on both this interface and in TransactionReaderTx?

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,387 @@
...
+
+		// We use EOF to mark "this tx doesn't exist," since that isn't a real error.
+		return xdr.LedgerCloseMeta{}, ingest.LedgerTransaction{}, io.EOF
+	}
+
+	var (
+		txIndex int
+		meta    []byte
+	)
+	if err := rows.Scan(&txIndex, &meta); err != nil {
```
tamirms, line 366: xdr.LedgerCloseMeta implements a db scan function so you can scan an instance of xdr.LedgerCloseMeta here instead of manually unmarshalling the binary a few lines below

tamirms:
We should have benchmarks on the db implementation of getTransaction so we have an idea about the performance. You can ingest 24 hours worth ledgers and then try querying a random sample of transaction hashes. Of course I expect the db performance to be much slower than the in-memory one but it's still good to know what the expected duration will be

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: nice tip!

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: this was a weird workaround for the `methods.LedgerRangeGetter` interface but I got rid of that entirely in 82ccd3af2dd2bda439d728d41cc71613c85c98cf

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: changed to return errors :+1:

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: good call, done :+1:

cmd/soroban-rpc/internal/daemon/daemon.go
Shaptic: I'm open to this, but then this means passing around the daemon/metrics registry to every single component in the system which seems silly. The way I have it here creates a separation of concerns of the component saying "here's what I can observe" and the daemon saying "here's how I want it observed". The daemon handles all creation (i.e. the type-specific `prometheus.Observer`), configuration (buckets, etc.), registration, etc. while each component just reports observations.

cmd/soroban-rpc/internal/daemon/daemon.go
Shaptic: Because otherwise when you do `supportlog.Debugf(...)` you get no logs :/ I get that we should probably have component-specific loggers, but since we don't, why do we not just use the global logger everywhere, anyway?

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
Shaptic: Good call! I didn't think about this as the column is sorted so it should be easy but obviously sqlite doesn't know better.

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
Shaptic: just seemed strictly more correct, but it doesn't really add any value if the code is correct

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
Shaptic: > it isn't as expensive to reingest compared to horizon

Yeahh the idea was that it was still quite expensive to do so (vs. just a migration), being easier than Horizon is kind of a low bar heh, and I felt like it could be valuable to add in advance (I think we'll want it for `getEvents`), but you're right that maybe it's premature - dropped the column :+1:

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: True but metrics are aggregated - there have been many instances where it's been very useful to see individual timelines for certain cases in Horizon

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: > you can checkout how it's done in cmd/soroban-rpc/internal/ingest/service.go

Can you elaborate? I see the initial [logger creation](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/ingest/service.go#L82) and its usage in the `service` itself (e.g. [here](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/ingest/service.go#L267)), but I don't see it passed along to any subservice.

> Also, I noticed that we don't log anything in any of the other db components so I think it makes sense to avoid logging here so that we can have a consistent style.

I'd argue it makes more sense to start logging in the other DB components, especially since we're moving towards a DB-only architecture and we want the same logging/metrics/visibility that we used to have in the in-memory components.

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: yeah, valid - done!

Shaptic:
@tamirms this is a great idea, but the ingestion lifecycle in sRPC is basically non-existent. I'll start a thread on Slack to discuss options as this is a longer topic!

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
Shaptic: I wanted to do this at first but I couldn't figure out how to make `BLOB` play nicely with `[]byte` when calling `.Values(...)`. I think it kept wanting to expand the slice into individual components... but I tried again in f265e19f0c8a64808f5c0b85caa31fbbe983b2a5 and it worked this time for some reason lol.

A downside is that it makes manual inspection of the database harder (since `select hash from transactions limit 1;` gives you nonsense that you can't forward to a `curl getTransaction` RPC call)

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,347 @@
...
+// read-only DB session to perform actual DB reads on.
+type TransactionReader interface {
+	NewTx(ctx context.Context) (TransactionReaderTx, error)
+}
+
+// TransactionReaderTx provides all of the public ways to read from the DB.
+// Note that `Done()` *MUST* be called to clean things up.
+type TransactionReaderTx interface {
+	GetTransaction(hash xdr.Hash) (Transaction, ledgerbucketwindow.LedgerRange, error)
+	GetLedgerRange() (ledgerbucketwindow.LedgerRange, error)
```
aditya1702, line 54: Would it be better to have GetLedgerRange in db/ledger.go instead?

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -211,18 +211,22 @@ func (cfg *Config) options() ConfigOptions {
 		},
 		{
 			Name: "event-retention-window",
-			Usage: fmt.Sprintf("configures the event retention window expressed in number of ledgers,"+
-				" the default value is %d which corresponds to about 24 hours of history", ledgerbucketwindow.DefaultEventLedgerRetentionWindow),
+			Usage: fmt.Sprintf(
```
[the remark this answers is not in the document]
Shaptic: yeah I just didn't want to duplicate work in case all of us were going to make the same changes but all slightly differently - in fact we should probably target a separate branch for this work so that we can isolate changes like this one

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: maybe as an interface, yeah, but then TransactionReader depends on LedgerReader which would be a bunch of premature refactoring - I think it'll be better to change this once we flesh out the interfaces more with the other pending db work

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,347 @@
...
+
+	return &transactionReaderTx{ctx: ctx, db: sesh, passphrase: txn.passphrase}, nil
+}
+
+func (txn *transactionReaderTx) Done() error {
+	return txn.db.Rollback()
+}
+
+// GetLedgerRange pulls the min/max ledger sequence numbers from the database.
+func (txn *transactionReaderTx) GetLedgerRange() (ledgerbucketwindow.LedgerRange, error) {
```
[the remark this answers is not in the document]
Shaptic: good idea - I renamed in ed23473, wdyt?

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
tamirms: you can use the `hex()` function in the sql query to render the BLOB as a hex string:

https://evanhahn.com/how-to-format-sqlite-blob-columns-as-hex/

tamirms:
> this is a great idea, but the ingestion lifecycle in sRPC is basically non-existent.

@Shaptic I was actually suggesting that we benchmark the  get transaction by hash db query https://github.com/stellar/soroban-rpc/blob/940f5d0145eb7f5ac8a8ca43764c864b7d3125de/cmd/soroban-rpc/internal/db/transaction.go#L53

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,355 @@
...
+	var ledgerRange ledgerbucketwindow.LedgerRange
+
+	sqlTx := txn.db.GetTx()
```
tamirms, line 179: you can use `UNION ALL` to query the newest and oldest ledger within a single SQL query:

https://stackoverflow.com/questions/1485391/how-to-get-first-and-last-record-from-a-sql-query

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,355 @@
...
+		Where(sq.Lt{"ledger_sequence": cutoff}).
+		Exec()
+	return err
+}
+
+// NewTx creates a read-only SQL transaction on a cloned database session. You
+// MUST call `.Done()` on the resulting reader if there are no errors.
+func (txn *transactionHandler) NewTx(ctx context.Context) (TransactionReader, error) {
+	sesh := txn.db.Clone()
+	if err := sesh.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
```
tamirms, line 164: I don't think it's necessary to start a transaction for `GetTransaction(hash xdr.Hash)` and `GetLedgerRange()`. The reason we start a repeatable read transaction for the `LedgerEntryReader` is because we may need to query multiple ledger entries for a single simulate transaction or get ledger entries request and, in that scenario, we want to ensure all the ledger entries are consistent with the same ledger.

Transactions, unlike ledger entries, are immutable so we don't need to have a snapshot isolation level for those queries. Furthermore, we are only querying a single transaction in the `getTransaction` endpoint

cmd/soroban-rpc/internal/db/transaction.go
tamirms: you can pass in a logger when creating the DB instance instead of using the global logger. this thread is also relevant:

https://github.com/stellar/soroban-rpc/pull/141/files/67d5ed9e6f2c7378d78b4a89223a3e415d6a6aa5#r1578459290

cmd/soroban-rpc/internal/db/migrations/02_transactions.sql
Shaptic: protip, thanks!

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Agreed! I opted for the tx-based approach to conform to the other stuff (I asked about it [on Slack](https://stellarfoundation.slack.com/archives/G01FUHTUQ8Z/p1713217110553329), too), but it felt excessive. I've refactored in d74e30d with this simplification!

cmd/soroban-rpc/internal/db/transaction.go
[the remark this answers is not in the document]
Shaptic: refactored to avoid needing sql transactions entirely!

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Yep you're right, I think I was just plumbing it incorrectly when I tried it before. Fixed up in f187405!

cmd/soroban-rpc/internal/db/transaction.go
tamirms: yes you're right. but as a work around you can use sub queries: https://stackoverflow.com/a/22609948

Shaptic: Nice find! Done in 7bfc43a

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,336 @@
...
+			errors.Wrapf(err, "failed row read for txhash %s", hash)
+	}
+	if !rows.Next() {
+		if ierr := rows.Err(); ierr != nil {
+			return xdr.LedgerCloseMeta{}, ingest.LedgerTransaction{},
+				errors.Wrapf(ierr, "failed to fetch txhash %s", hash)
+		}
+
+		// We use EOF to mark "this tx doesn't exist," since that isn't a real error.
+		return xdr.LedgerCloseMeta{}, ingest.LedgerTransaction{}, io.EOF
```
tamirms, line 274: I think it would be cleaner to return a boolean indicating that that transaction is present / not present rather than returning `io.EOF` as an indicator that the transaction is not present

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,336 @@
...
+		return lcm, ingest.LedgerTransaction{},
+			errors.Wrapf(err, "failed to index to tx %d in ledger %d (txhash=%s)",
+				txIndex, lcm.LedgerSequence(), hash)
+	}
+
+	ledgerTx, err := reader.Read()
+	return lcm, ledgerTx, err
+}
+
+func ParseTransaction(lcm xdr.LedgerCloseMeta, ingestTx ingest.LedgerTransaction) (Transaction, error) {
```
tamirms, line 298: is this function called outside the db package? if not, then I don't think we need to export it

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,336 @@
...
+				Limit(1),
+			"m2",
+		)
+
+	sql, args, err := oldestQ.ToSql()
+	if err != nil {
+		return ledgerRange, errors.Wrap(err, "couldn't build ledger range query")
+	}
+
+	rows, err := txn.db.Query(ctx, newestQ.Suffix(" UNION ALL "+sql, args...))
```
tamirms, line 177: you can use the `txn.db.Select()` function to get all the rows and that would allow you to simplify the error checking  because you wouldn't need to call `rows.Next() ` and `rows.Scan()` twice

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,336 @@
...
+	_, err := sq.StatementBuilder.
+		RunWith(txn.stmtCache).
+		Delete(transactionTableName).
+		Where(sq.Lt{"ledger_sequence": cutoff}).
+		Exec()
+	return err
+}
+
+// GetLedgerRange pulls the min/max ledger sequence numbers from the database.
+func (txn *transactionHandler) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
```
tamirms, line 143: I just realized that we could maintain the ledger range as in-memory fields that are updated by the ingestion thread:

Upon start up you query the oldest and newest ledger. Then whenever we successfully ingest a new ledger we increment the newest ledger field. And, whenever we trim an old ledger we increment the oldest ledger field.

That would allow us to reduce the amount of db queries we make.

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Four return values? :sweat_smile: how about `db.ErrNoTx`, instead? It overloads the term but actually kinda works lol

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: The intent was for @aditya1702 to use it in a later refactor of [this code diff](https://github.com/stellar/soroban-rpc/pull/136/files#diff-5276db8001730f32b267cab637b316958b69aef6f8184343338855fdc118b700R211), i.e. replacing `transactionsRPCHandler.getTransactionDetails(tx ingest.LedgerTransaction) ([]byte, []byte, []byte, error)` which is doing the same stuff.

cmd/soroban-rpc/internal/db/transaction.go
tamirms: sure, that works!

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Thanks for the tip! I figured it might've been possible but since there weren't any docs I had trouble finding the right syntax combination - gave it another go and fixed it up in 57f6f78.

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Nice!! I had a similar idea but not to this level of depth because it was before I knew about trimming. I propose that we do this in a later PR (note that this PR doesn't target `main`) to isolate the changes as well as enhance testing!

tamirms: sure, sounds good!

cmd/soroban-rpc/internal/db/transaction.go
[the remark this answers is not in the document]
aditya1702: That is great!

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: opted for a custom error, instead :+1:

Shaptic:
@tamirms I've added benchmarks! Here's the log on my machine:

```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkLedgerUpdate-12                      	       86	   15459833 ns/op
BenchmarkTransactionFetch-12                  	    10000	     114646 ns/op
```

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -130,21 +137,78 @@ func getLatestLedgerSequence(ctx context.Context, q db.SessionInterface, cache *
...
+	passphrase            string
+
+	metrics *ReadWriterMetrics
+}
+
+// NewReadWriter constructs a new readWriter instance and configures the size of
+// ledger entry batches when writing ledger entries and the retention window for
+// how many historical ledgers are recorded in the database, hooking up
+// metrics for various DB ops.
+func NewReadWriterWithMetrics(
```
tamirms, line 158: I think we should only have one public constructor for ReadWriter because we probably never want to construct an instance without metrics. I would remove the other constructor (`NewReadWriter()`) and rename this one to `NewReadWriter()`. For tests, you can call `interfaces.MakeNoOpDeamon()` to make a testing daemon stub.

cmd/soroban-rpc/internal/db/db.go
tamirms: @Shaptic I think this comment still applies. I think we should move `NewTestDB()` back to a testing file in this package

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -0,0 +1,308 @@
...
+	start, mid := time.Now(), time.Now()
+	txCount := lcm.CountTransactions()
+	L := txn.log.
+		WithField("ledger_seq", lcm.LedgerSequence()).
+		WithField("tx_count", txCount)
+
+	defer func() {
+		if txn.ingestMetric != nil {
+			txn.ingestMetric.Observe(time.Since(start).Seconds())
+			txn.insertMetric.Observe(time.Since(mid).Seconds())
```
tamirms, line 71: it seems like these two metrics are measuring the same values?

tamirms: ok, I see that `mid` gets updated later in the function. In practice I think the time spent collecting the transactions from the ledger close meta will be extremely small so having two different metrics is probably overkill

cmd/soroban-rpc/internal/methods/get_transaction.go
```diff
@@ -88,16 +89,22 @@ func GetTransaction(getter transactionGetter, request GetTransactionRequest) (Ge
...
+	tx, storeRange, err := reader.GetTransaction(ctx, txHash)
+
 	response := GetTransactionResponse{
 		LatestLedger:          storeRange.LastLedger.Sequence,
 		LatestLedgerCloseTime: storeRange.LastLedger.CloseTime,
 		OldestLedger:          storeRange.FirstLedger.Sequence,
 		OldestLedgerCloseTime: storeRange.FirstLedger.CloseTime,
 	}
-	if !found {
+	if err == db.ErrNoTransaction {
```
tamirms, line 100: better to use https://pkg.go.dev/errors#Is in case the error is wrapped

cmd/soroban-rpc/internal/methods/health.go
```diff
@@ -17,18 +16,18 @@ type HealthCheckResult struct {
...
+) jrpc2.Handler {
 	return NewHandler(func(ctx context.Context) (HealthCheckResult, error) {
-		ledgerRange := ledgerRangeGetter.GetLedgerRange()
-		if ledgerRange.LastLedger.Sequence < 1 {
+		ledgerRange, err := reader.GetLedgerRange(ctx)
+		if err != nil || ledgerRange.LastLedger.Sequence < 1 {
 			return HealthCheckResult{}, jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: "data stores are not initialized",
+				Message: "data stores are not initialized " + err.Error(),
```
tamirms, line 30: is there a possibility that `err` is nil?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -341,3 +343,7 @@ func (d *Daemon) Run() {
 		return
 	}
 }
+
+func (d *Daemon) GetGlobals() (*db.DB, *log.Entry, *prometheus.Registry) {
```
tamirms, line 347: I don't think this function is used

cmd/soroban-rpc/internal/methods/get_transaction.go
Shaptic: great catch! done in 9fc64a9

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: `mid` was specifically to measure the sql/db portion of the ingestion, but I do agree with you that it's probably going to be nearly identical to `start` to which the distinction probably isn't worth it :+1: done in 9fc64a9

cmd/soroban-rpc/internal/db/db.go
Shaptic: sure, fair enough! done in 9fc64a9

cmd/soroban-rpc/internal/db/db.go
Shaptic: Ah yes I forgot about this.

>  In cmd/soroban-rpc/internal/methods/get_transaction_test.go , you can use mocks for the transaction store.

In the [current tests](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/get_transaction_test.go#L154), you can actually see that this does not happen: we use a real store. That's why when I transitioned the tests, I did the same thing. I saw that we have this weird imitation of a fake backend for [get_latest_ledger_test.go](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/get_latest_ledger_test.go#L76) but I couldn't understand what that was testing tbh.

Anyway, I added a fake in-memory store for use in testing in 5e49d7d and refactored the methods to use it. I think that's sufficient!

tamirms (approved):

PR #142, opened 2024-04-16
Obtain simulation protocol version dynamically

2opremio:
### What

Obtain protocol version for simulation dynamically

### Why

It used to be hardcoded to version 20 but. Now that we are releasing a new protocol version we cannot do that anymore.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -7,11 +7,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync/atomic"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
 
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/daemon/interfaces"
 	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db"
 	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/preflight"
 )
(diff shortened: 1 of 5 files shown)


cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -298,3 +307,19 @@ func getBucketListSize(ctx context.Context, ledgerReader db.LedgerReader, latest
...
+	if version != 0 {
+		return int(version), nil
+	}
+	info, err := client.Info(ctx)
+	if err != nil {
+		return 0, err
+	}
+	cachedCoreProtocolVersion.CompareAndSwap(0, int32(info.Info.ProtocolVersion))
+	return info.Info.ProtocolVersion, nil
+}
```
2opremio, line 325: Is requesting and caching forever too much? I suppose Core could be restarted due to an error, but I doubt it would magically change the protocol version?

2opremio: I think something like that could potentially happen in Horizon, so I am not sure. @tamirms thoughts?

dmkozh: Where exactly does the protocol version come from in the `CoreClient`? During the protocol upgrade the version will change, without any restarts.

tamirms: @2opremio I see that you already get the ledger close meta and use it to configure the `BucketListSize`. The the ledger close meta also contains the protocol version (it is located in the ledger header `closeMeta.LedgerHeaderHistoryEntry().Header.LedgerVersion`). I think it would be better to obtain the protocol version from the ledger close meta instead of querying core's http endpoint

2opremio: Ah ... really good point. Thanks!

2opremio:
@tamirms PTAL

tamirms (approved):

PR #146, opened 2024-04-19
getHealth: add retention window to status

2opremio:
I am open to design modifications

### What

Status will be `"healthy"` when the retention window isn't full and "healthy, retention window full", when the window is full.

### Why

Make it simpler to write health probes

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/methods/health.go
+++ b/cmd/soroban-rpc/internal/methods/health.go
@@ -42,10 +42,17 @@ func NewHealthCheck(retentionWindow uint32, ledgerRangeGetter LedgerRangeGetter,
 				Message: msg,
 			}
 		}
+		status := "healthy"
+		latestLedger := ledgerRange.LastLedger.Sequence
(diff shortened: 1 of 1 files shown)


cmd/soroban-rpc/internal/methods/health.go
```diff
@@ -42,10 +42,17 @@ func NewHealthCheck(retentionWindow uint32, ledgerRangeGetter LedgerRangeGetter,
 				Message: msg,
 			}
 		}
+		status := "healthy"
+		latestLedger := ledgerRange.LastLedger.Sequence
+		oldestLedger := ledgerRange.FirstLedger.Sequence
+		rententionWindowFull := (latestLedger - oldestLedger) >= retentionWindow
+		if rententionWindowFull {
+			status += ", retention window full"
```
Shaptic, line 50: Are there a lot of places that do an `== "healthy"` check? This would break that. Maybe it'd be better to have a separate `RetentionWindowStatus: "full" | "incomplete"` kinda thing, instead?

PR #149, opened 2024-04-21
Remove unnecessary dependencies

2opremio:


Shaptic (approved):
:fire:

Cargo.toml
```diff
@@ -29,40 +29,11 @@ version = "=21.0.0"
 # rev = "c30bc769e379bef9b94a3ceb464aa78c1185eeb3"
 # path = "../rs-soroban-sdk/soroban-spec"
```
leighmcculloch, line 30: I think we can also remove `soroban-spec` as a dependency and `soroban-sdk` further down, which would then mean that soroban-rpc then has zero dependency on the `stellar/rs-soroban-sdk` repository.

The `stellar-rpc-client` uses those two crates to lookup the wasm specs from a .wasm file it downloads from the soroban-rpc. Imo it's outside the responsibility of the stellar-rpc-client to do the decoding of the wasm. The client should help an application download the .wasm, and then leave it up to the application to extract the spec. Applications that want to extract the spec, like the soroban-cli, can import `soroban-spec` and `soroban-sdk`.

leighmcculloch (commented):
👏🏻 Nice. One comment inline about a couple other deps I think we can remove, and now would be a good time because of the nature of the change being breaking.

Cargo.toml
2opremio: @willemneal could you push out the wasm spec lookup to soroban-cli?

PR #150, opened 2024-04-22
Add test upgrading Protocol 20 to Protocol 21

2opremio:
### What

Test upgrading protocol 20 to protocol 21

### Why

Make sure everything works

### Known limitations

N/A


cmd/soroban-rpc/internal/test/upgrade_test.go
```diff
@@ -0,0 +1,128 @@
...
+	newLedgerSequence := uint32(0)
+	for i := 0; i < 60; i++ {
+		time.Sleep(time.Second)
+		newLedgerSequence, err = db.NewLedgerEntryReader(rpcDB).GetLatestLedgerSequence(context.Background())
+		require.NoError(t, err)
+		if newLedgerSequence > initialLedgerSequence {
+			break
+		}
+	}
+	require.Greater(t, newLedgerSequence, initialLedgerSequence, "rpc didn't start ingesting protocol 21")
```
sreuland, line 68: could remove the `for loop` in favor of `require.Eventually(t, func() {...}, time.Minute, time.Second)`

cmd/soroban-rpc/internal/test/upgrade_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		BaseFee: txnbuild.MinBaseFee,
+		Preconditions: txnbuild.Preconditions{
+			TimeBounds: txnbuild.NewInfiniteTimeout(),
+		},
+	})
+	tx, err = txnbuild.NewTransaction(params)
+	assert.NoError(t, err)
+
+	assert.NoError(t, err)
```
sreuland, line 126: duplicate assertion on ln 124 and 126. 
is it worthwhile to check return value/SorobanData on the tx result to confirm invoke host fn response?

sreuland (approved):

cmd/soroban-rpc/internal/test/upgrade_test.go
2opremio: Nice, I will use this!

cmd/soroban-rpc/internal/test/upgrade_test.go
2opremio: Good catch

cmd/soroban-rpc/internal/test/upgrade_test.go
2opremio: We should probably use this in the state expiration tests as well

PR #167, opened 2024-05-09
Fix deadlock on events ingestion error

2opremio:
### What

Ensure mutex is unlocked in case of error

### Why

It causes a deadlock in case of error

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/events/events.go
+++ b/cmd/soroban-rpc/internal/events/events.go
@@ -207,6 +207,7 @@ func (m *MemoryStore) IngestEvents(ledgerCloseMeta xdr.LedgerCloseMeta) error {
 	}
 	m.lock.Lock()
 	if _, err = m.eventsByLedger.Append(bucket); err != nil {
+		m.lock.Unlock()
 		return err
 	}
 	m.lock.Unlock()


2opremio:
and this is why it would be great to get a block-level defer in Go. https://github.com/golang/go/issues/54249

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -207,6 +207,7 @@ func (m *MemoryStore) IngestEvents(ledgerCloseMeta xdr.LedgerCloseMeta) error {
 	}
 	m.lock.Lock()
```
psheth9, line 208: instead we can go with `defer m.lock.Unlock() `

Shaptic (approved):
nice catch :+1:

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -207,6 +207,7 @@ func (m *MemoryStore) IngestEvents(ledgerCloseMeta xdr.LedgerCloseMeta) error {
...
 	if _, err = m.eventsByLedger.Append(bucket); err != nil {
+		m.lock.Unlock()
```
sreuland, line 210: is there an idiom to defer Unlock() once immediately after the Lock() instead?

psheth9 (approved):
just one comment !!

cmd/soroban-rpc/internal/events/events.go
Shaptic: I think his idea is to unlock asap but I can't imagine deferring would make a significant performance difference for two `Observe` calls (cc @sreuland who had same comment)

psheth9: Yeah no performance diff but I prefer it that way so that further error branches dont have to worry about unlocking it. I think we have the similar pattern in [here](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/transactions/transactions.go#L124) in transaction.go [Anything works as long as we unlock it :) ]

Shaptic: agreed!

2opremio: not with the same granularity

PR #172, opened 2024-05-10
Add getFeeStats method

2opremio:
### What

Add `getFeeStats` method, with the following response data structure:

```go

type GetFeeStatsResult struct {
	SorobanInclusionFee FeeDistribution `json:"sorobanInclusionFee"`
	InclusionFee        FeeDistribution `json:"inclusionFee"`
	LatestLedger        uint32          `json:"latestLedger"`
}

type FeeDistribution struct {
	Max              uint64 `json:"max,string"`
	Min              uint64 `json:"min,string"`
	Mode             uint64 `json:"mode,string"`
	P10              uint64 `json:"p10,string"`
	P20              uint64 `json:"p20,string"`
	P30              uint64 `json:"p30,string"`
	P40              uint64 `json:"p40,string"`
	P50              uint64 `json:"p50,string"`
	P60              uint64 `json:"p60,string"`
	P70              uint64 `json:"p70,string"`
	P80              uint64 `json:"p80,string"`
	P90              uint64 `json:"p90,string"`
	P95              uint64 `json:"p95,string"`
	P99              uint64 `json:"p99,string"`
	TransactionCount uint32 `json:"transactionCount,string"`
	LedgerCount      uint32 `json:"ledgerCount"`
}
```

It includes extra fields with respect to https://github.com/stellar/soroban-rpc/issues/114 (a global `latestLedger` field`and a `count` field in each distribution, it indicates how many fees were included to compute it).

We can, optionally also include the initial ledger used for each distribution if that's interesting (which, together with the global `latestLedger` field, will tell the user what ledger window was used for each distribution).

### Why

Closes https://github.com/stellar/soroban-rpc/issues/114

### Known limitations

N/A


2opremio:
This is still untested but created early to gather feedback.

@tamirms I would like to know your opinion on the implementation. Unlike Horizon (which uses Postgres to compute the stats :S), we do everything in Go and book-keep an in-memory data structure.

@janewang Thoughts about the API?

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,178 @@
...
+		lastVal = fees[i]
+		localRepetitions = 0
+	}
+	count := uint64(len(fees))
+	// nearest-rank percentile
+	percentile := func(p uint64) uint64 {
+		// ceiling(p*count/100)
+		kth := ((p * count) + 100 - 1) / 100
+		return fees[kth-1]
+	}
```
tamirms, line 94: I find [this implementation](https://github.com/montanaflynn/stats/blob/v0.7.1/percentile.go#L53-L86) easier to understand

tamirms:
@2opremio in-memory approach makes sense to me especially considering we consider at most 50 ledgers

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,178 @@
...
+		ops := tx.Envelope.Operations()
+		if len(ops) == 1 {
+			switch ops[0].Body.Type {
+			case xdr.OperationTypeInvokeHostFunction, xdr.OperationTypeExtendFootprintTtl, xdr.OperationTypeRestoreFootprint:
+				if tx.Envelope.V1 == nil || tx.Envelope.V1.Tx.Ext.SorobanData != nil {
+					// this shouldn't happen
+					continue
+				}
+				inclusionFee := uint64(tx.Envelope.V1.Tx.Fee) - uint64(tx.Envelope.V1.Tx.Ext.SorobanData.ResourceFee)
+				classicFees = append(sorobanInclusionFees, inclusionFee)
```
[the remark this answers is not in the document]
janewang:
```
SorobanInclusionFee FeeDistribution `json:"sorobanInclusionFee"` => Soroban Inclusion Fee Distribution
InclusionFee        FeeDistribution `json:"inclusionFee"` => Classic Inclusion Fee Distribution
```
These two inclusion fees are separate and are unrelated distributions.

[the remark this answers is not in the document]
tamirms: currently `classicFees` only contains fees for soroban transactions because the if statement and the switch statement filters away transactions which have more than 1 operation and transactions which consist of non soroban operations

2opremio:
@janewang how about `feeCount`?

cmd/soroban-rpc/internal/feewindow/feewindow.go
2opremio: I considered using that but it's less efficient (it uses floats and we would need to sort the list multiple times for the other computations).

Maybe it's not worth optimizing it though ...

cmd/soroban-rpc/internal/feewindow/feewindow.go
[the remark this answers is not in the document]
2opremio: @tamirms that's a bug, it should be:

```suggestion
				sorobanInclusionFees = append(sorobanInclusionFees, inclusionFee)
```

Fixed

2opremio:
Actually, we could have `feeCount` and `ledgerCount`

cmd/soroban-rpc/internal/feewindow/feewindow.go
2opremio: We can use that implementation for testing though

2opremio: @tamirms that's a bug, it should be:

```suggestion
				sorobanInclusionFees = append(sorobanInclusionFees, inclusionFee)
```

Fixed

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		ops := tx.Envelope.Operations()
+		if len(ops) == 1 {
+			switch ops[0].Body.Type {
+			case xdr.OperationTypeInvokeHostFunction, xdr.OperationTypeExtendFootprintTtl, xdr.OperationTypeRestoreFootprint:
+				if tx.Envelope.V1 == nil || tx.Envelope.V1.Tx.Ext.SorobanData != nil {
+					// this shouldn't happen
+					continue
+				}
+				inclusionFee := uint64(tx.Envelope.V1.Tx.Fee) - uint64(tx.Envelope.V1.Tx.Ext.SorobanData.ResourceFee)
```
sydneynotthecity, line 157: You may want to check out how we implemented `inclusionFee` in stellar-etl: https://github.com/stellar/stellar-etl/blob/master/internal/transform/transaction.go#L155 We used a different formula: 
```
inclusion_fee = fee_charged - (resource_fee - resource_fee_refund)
```

sydneynotthecity:
> Actually, we could have `feeCount` and `ledgerCount`

I still think `feeCount` is a little confusing for end users. I would prefer either `sorobanTransactionCount` or `observationCount`. So that it's more clear _what_ fees are counted.

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,178 @@
...
+				sorobanInclusionFees = append(sorobanInclusionFees, inclusionFee)
+				continue
+			}
+		}
+		classicFees = append(classicFees, uint64(tx.Envelope.Fee()))
```
tamirms, line 162: I think horizon divides the fee by operation count:

https://github.com/stellar/go/blame/master/services/horizon/internal/db2/history/operation.go#L62-L79

2opremio: Ah, good point, the documentation isn't clear about it

cmd/soroban-rpc/internal/feewindow/feewindow.go
2opremio: I used github.com/montanaflynn/stats for testing. So, while I was at it, I created a benchmark:

```
goos: darwin
goarch: arm64
pkg: github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/feewindow
BenchmarkComputeFeeDistribution
BenchmarkComputeFeeDistribution/computeFeeDistribution
BenchmarkComputeFeeDistribution/computeFeeDistribution-20         	 8273469	       137.5 ns/op
BenchmarkComputeFeeDistribution/alternativeComputeFeeDistribution
BenchmarkComputeFeeDistribution/alternativeComputeFeeDistribution-20         	  362680	      3122 ns/op
PASS
```

The implementation is 20x faster. Not that, spending 6 microseconds (two computations) every ledger was going to cause any havoc ...

2opremio:
> > Actually, we could have `feeCount` and `ledgerCount`
> 
> I still think `feeCount` is a little confusing for end users. I would prefer either `sorobanTransactionCount` or `observationCount`. So that it's more clear _what_ fees are counted.

I will go with `transactionCount`

2opremio:
@tamirms @sydneynotthecity @janewang PTAL

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,191 @@
...
+	P90         uint64
+	P95         uint64
+	P99         uint64
+	FeeCount    uint32
+	LedgerCount uint32
+}
+
+type FeeWindow struct {
+	lock          sync.RWMutex
+	feesPerLedger *ledgerbucketwindow.LedgerBucketWindow[[]uint64]
```
Shaptic, line 35: Should we decide now whether or not we ever want this to be DB-backed or if the in-memory approach is more than sufficient?

2opremio: In memory should be more than enough considering it’s 50 ledgers by default.

cmd/soroban-rpc/internal/feewindow/feewindow.go
```diff
@@ -0,0 +1,191 @@
...
+	var classicFees []uint64
+	for {
+		tx, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		feeCharged := uint64(tx.Result.Result.FeeCharged)
```
tamirms, line 155: it looks like the horizon endpoint includes 2 distributions one for FeeCharged and another for MaxFee (which is the maximum fee the account is willing to pay to submit their transaction):

https://horizon.stellar.org/fee_stats

Here we only include the distribution on FeeCharged. Is that intentional or should we also include both distributions in the soroban-rpc endpoint?

2opremio: CC @janewang

janewang: The distribution makes sense to show is the actual feeCharged.

tamirms (approved):

PR #174, opened 2024-05-13
Add new `getTransactions` endpoint and database backend for transactions

aditya1702:
### What

This PR merges all the new changes related to `transactions`. This includes:
- Move transaction data from in-memory store to DB. Also, refactor `getTransaction` endpoint to use the db: #141 
- Add `getTransactions` endpoint: #110

### Why

Part of #115 

### Known limitations

NA


cmd/soroban-rpc/internal/config/options.go
```diff
@@ -211,18 +211,22 @@ func (cfg *Config) options() ConfigOptions {
...
 			Name: "transaction-retention-window",
-			Usage: "configures the transaction retention window expressed in number of ledgers," +
-				" the default value is 1440 which corresponds to about 2 hours of history",
+			Usage: fmt.Sprintf(
+				"configures the transaction retention window expressed in number of ledgers,"+
+					" the default value is %d which corresponds to about 24 hours of history",
+				ledgerbucketwindow.OneDayOfLedgers),
 			ConfigKey:    &cfg.TransactionLedgerRetentionWindow,
-			DefaultValue: uint32(1440),
+			DefaultValue: uint32(ledgerbucketwindow.OneDayOfLedgers),
```
Shaptic, line 229: oh we should undo this?

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -15,7 +15,9 @@ import (
 
 	"github.com/stellar/go/support/db"
 	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/daemon/interfaces"
```
Shaptic, line 20: does this cause a circular import? :/

aditya1702: @Shaptic Nope, I am able to compile this branch and run RPC successfully.

Shaptic: me too, but according to https://github.com/golang/go/issues/51180 it may cause issues in tests?? even though mine pass just fine locally

cmd/soroban-rpc/internal/methods/get_fee_stats.go
```diff
@@ -56,12 +58,18 @@ type GetFeeStatsResult struct {
 }
 
 // NewGetFeeStatsHandler returns a handler obtaining fee statistics
-func NewGetFeeStatsHandler(windows *feewindow.FeeWindows, ledgerRangeGetter LedgerRangeGetter) jrpc2.Handler {
+func NewGetFeeStatsHandler(windows *feewindow.FeeWindows, reader db.TransactionReader, logger *log.Entry) jrpc2.Handler {
```
Shaptic, line 61: We should still use PRs for any add'l changes (besides upstream sync) that we want in this branch to have a proper merge log + review process.

cmd/soroban-rpc/internal/config/options.go
psheth9: Yeah I think so, users can set it higher if they want to. Do we want to add some upper bound here? like dont go beyond 7 days or something? I know we have not added those restrictions for in-memory stores but might be a good idea.

cmd/soroban-rpc/internal/db/db.go
psheth9: Even if it works locally, I think we should not import `daemon/interface` here because db.go is at more lower level which should not have access to daemon package. Currently what we need is only `MetricsNamespace` and `MetricsRegistry` in db.go. 

It is circular bc daemon is already using db.go, it might be better to refactor now imo.

Shaptic: I think it's technically not circular because `daemon` imports `db` but `db` imports `daemon/interface`? So they're different. I totally agree that daemon has no business here but this came from review comments (https://github.com/stellar/soroban-rpc/pull/141#discussion_r1579522946) on my PR.

cmd/soroban-rpc/internal/config/options.go
Shaptic: nah I think it's fine, if you have disk why not :shrug:

2opremio:
I would be interested to get a performance comparison between the in-memory implementation and the db implementation.

Shaptic:
@2opremio me too! We can do a basic evaluation by seeing the change in metrics after updating (is RPC in Grafana?), but I can also try to see the results of the `BenchmarkTransactionFetch` after reverting #141 :+1:

Shaptic (approved):

PR #184, opened 2024-05-16
Recompile the test contract from the CLI repo

Shaptic:
### What
Checked out soroban-cli, went to 

```
cd cmd/crates/soroban-test/tests/fixtures/test-wasms/hello_world
cargo build --target wasm32-unknown-unknown --release
```

and copied the resulting wasm file from `/target/` here.

### Why
Gotta keep it in sync so integration tests are accurate.

### Known limitations
Doing this manually is annoying.


Shaptic:
We'll need to actually update tests since the optimized WASM stuff makes it perform differently, I think, so it's not just a binary update.

willemneal:
@leighmcculloch Perhaps we could move contract build into it's own crate? This way it could be used here as well to build contracts consistently and perhaps also include the test contracts.

leighmcculloch:
@willemneal It doesn't really matter if the special contract build is used or if cargo build is used in this context. The contract build process is only important for production deployments for reducing wasm size, and even then it's easily reproducible using a cargo command. So I'm not sure the value is significant and worth the cost of breaking the build command out.

PR #192, opened 2024-05-31
Ingest events into DB

psheth9:
Fixes: https://github.com/stellar/soroban-rpc/issues/189

This diff mainly covers the write path (ingestion) of events, follow up PR refactors getEvents in order to fetch it from DB

Changes:

- Introduce new event table
- Parse events from LCM and ingest in DB
- Tests 

Design Doc [here](https://docs.google.com/document/d/1Q5AJvszwFmwl3NzHh7PJIMj4mhN3HgHRn_7IOQGeMWg/edit)
Parent Epic [here](https://github.com/stellar/soroban-rpc/issues/115)

### DIFF
--- a/cmd/soroban-rpc/internal/db/event.go
+++ b/cmd/soroban-rpc/internal/db/event.go
@@ -5,6 +5,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/support/db"
+	"github.com/stellar/go/support/errors"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
 	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/events"
@@ -29,12 +30,21 @@ type eventHandler struct {
 	ingestMetric, countMetric prometheus.Observer
 }
 
-func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) (err error) {
+func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
+	txCount := lcm.CountTransactions()
+
+	if eventHandler.stmtCache == nil {
+		return errors.New("EventWriter incorrectly initialized without stmtCache")
+	} else if txCount == 0 {
+		return nil
+	}
 
 	var txReader *ingest.LedgerTransactionReader
-	txReader, err = ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
+	txReader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
 	if err != nil {
-		return
+		return errors.Wrapf(err,
+			"failed to open transaction reader for ledger %d",
+			lcm.LedgerSequence())
 	}
 	defer func() {
 		closeErr := txReader.Close()
@@ -51,7 +61,7 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) (err err
 			break
 		}
 		if err != nil {
-			return
+			return err
 		}
 
 		if !tx.Result.Successful() {
--- a/cmd/soroban-rpc/internal/db/event.go
+++ b/cmd/soroban-rpc/internal/db/event.go
@@ -73,6 +73,10 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 			return err
 		}
 
+		if len(txEvents) == 0 {
+			continue
+		}
+
 		query := sq.Insert(eventTableName).
 			Columns("id", "ledger_sequence", "application_order", "contract_id", "event_type")
 
--- a/cmd/soroban-rpc/internal/db/event.go
+++ b/cmd/soroban-rpc/internal/db/event.go
@@ -77,8 +77,12 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 			Columns("id", "ledger_sequence", "application_order", "contract_id", "event_type")
 
 		for index, e := range txEvents {
+			var contractId []byte
+			if e.Event.ContractId != nil {
+				contractId = e.Event.ContractId[:]
+			}
 			id := events.Cursor{Ledger: lcm.LedgerSequence(), Tx: tx.Index, Op: 0, Event: uint32(index)}.String()
-			query = query.Values(id, lcm.LedgerSequence(), tx.Index, e.Event.ContractId[:], int(e.Event.Type))
+			query = query.Values(id, lcm.LedgerSequence(), tx.Index, contractId, int(e.Event.Type))
 		}
 
 		_, err = query.RunWith(eventHandler.stmtCache).Exec()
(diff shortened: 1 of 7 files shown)


2opremio:
Please do not merge until there is a solution for data migration

psheth9:
> Please do not merge until there is a solution for data migration

Not going to merge to `main` but to `events-db-backend` !! @2opremio

2opremio:
Did you write migration code for the new schema? Running instances with the in-memory datasctructure should work when running this code.

2opremio:
You also need to remove the in-memory prefill from daemon.go

psheth9:
@2opremio Yes I will be adding the migration code and going to remove the entire in-memory storage of events in follow up PR. Not planning to do that in this PR bc need to refactor `getEvents` [to fetch it from DB instead of in-memory] 

Could you approve this PR if DB schema and ingestion looks good :) ? Updated the details of follow up refactoring task [here](https://github.com/stellar/soroban-rpc/issues/190) 

Note: This won't merge changes to `main` but only to` events-db-backend`

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/events"
+	"io"
+)
+
+const eventTableName = "events"
+
+type EventWriter interface {
+	InsertEvents(lcm xdr.LedgerCloseMeta) error
```
aditya1702, line 18: I think the interface should not be defined in the same file as its implementation: https://go.dev/wiki/CodeReviewComments#interfaces

Would like to get a recommendation from @tamirms about this.

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -0,0 +1,91 @@
...
+		tx, err = txReader.Read()
+		if err == io.EOF {
+			err = nil
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		if !tx.Result.Successful() {
```
aditya1702, line 67: Why are we skipping if the transaction is not successful? There can be diagnostic events for both successful and failed txns right?

@2opremio @tamirms

cmd/soroban-rpc/internal/db/event_test.go
```diff
@@ -0,0 +1,167 @@
...
+			),
+		))
+	}
+	ledgerCloseMeta := ledgerCloseMetaWithEvents(1, now.Unix(), txMeta...)
+
+	eventW := write.EventWriter()
+	err = eventW.InsertEvents(ledgerCloseMeta)
+	assert.NoError(t, err)
+
+	//TODO: Call getEvents and validate events data.
```
2opremio, line 166: Can you please resolve this before merging?

cmd/soroban-rpc/internal/db/migrations/03_events.sql
```diff
@@ -0,0 +1,16 @@
+-- +migrate Up
+
+-- indexing table to find events in ledgers by contract_id
+CREATE TABLE events (
+                              id TEXT PRIMARY KEY,
+                              ledger_sequence INTEGER NOT NULL,
+                              application_order INTEGER NOT NULL,
+                              contract_id BLOB NOT NULL,
+                              event_type INTEGER NOT NULL
```
2opremio, line 9: Nit: formatting

2opremio:
Can you please fix the tests and the TODO?

cmd/soroban-rpc/internal/db/event.go
psheth9: Yeah this is the existing behavior. Can update it if it is not true anymore.

aditya1702: Yeah I saw it is in current getEvents too. I wanted to confirm this is true since I saw diagnostic events in both types of txns.

cmd/soroban-rpc/internal/db/event_test.go
psheth9: yes, actually ignore this TODO as `getEvents` is backed by memory in this diff, this will be resolved in this follow up PR https://github.com/stellar/soroban-rpc/pull/215 (WIP) @2opremio

psheth9:
@2opremio  Fixed the tests PTAL :) Next is coming up soon

2opremio:
Ah, the in-memory store still exists? I would much rather review finished, self-contained code with proper tests.

Can you remove the in-memory store and add end-to-end tests here?

2opremio:
I have finished what I was going , so I will get to review this on Monday. Sorry for the delay.

cmd/soroban-rpc/internal/db/event.go
2opremio: You are right, but that's how it's done everywhere else in the db package

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -0,0 +1,99 @@
...
+	txReader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
+	if err != nil {
+		return errors.Wrapf(err,
+			"failed to open transaction reader for ledger %d",
+			lcm.LedgerSequence())
+	}
+	defer func() {
+		closeErr := txReader.Close()
+		if err == nil {
+			err = closeErr
```
2opremio, line 52: Don't overwrite the error, just use `errors.Join()`

2opremio: Also, note that err is shadowed down the line, so this may not be the `err` you want

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		if err != nil {
+			return err
+		}
+
+		if !tx.Result.Successful() {
+			continue
+		}
+
+		txEvents, err := tx.GetDiagnosticEvents()
```
2opremio, line 71: This shadows the `err`above, so the error handling in defer above won't do what you expect. You may want to use a named return parameter to make it cleaner.

2opremio:
The approach seems fine, but ... it's hard to tell without seeing the complexity of the querying side.

I would like to see a benchmark of querying and see how it compares with the in-memory store.

2opremio (approved):

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -0,0 +1,99 @@
...
+
+		query := sq.Insert(eventTableName).
+			Columns("id", "ledger_sequence", "application_order", "contract_id", "event_type")
+
+		for index, e := range txEvents {
+			var contractId []byte
+			if e.Event.ContractId != nil {
+				contractId = e.Event.ContractId[:]
+			}
+			id := events.Cursor{Ledger: lcm.LedgerSequence(), Tx: tx.Index, Op: 0, Event: uint32(index)}.String()
```
2opremio, line 88: @psheth9 I have just realized that the table already contains a ledger row, so the ledger info here is redundant.

Also:

1. if the op is always 0 (and we don't plan that to change) it shouldn't go into the DB.
2. the tx index isn't the same as the application order field?

Thus, unless there is a good reason for the opposite, I would suggest storing the tx index and event index instead of the id field.

PR #198, opened 2024-06-05
Refactor getVersionInfo and getNetwork RPC to not use HTTP endpoint /info on core

psheth9:
Fixes: https://github.com/stellar/soroban-rpc/issues/161

Predecessor task: Update captive_core_backend.go to expose ProtocolVersion instead of fetching from core client https://github.com/stellar/go/pull/5332

This PR adds following:

- Refactor of `getVersionInfo`
- Refactor of `getNetwork`
- Adds a new `getProtocolVersion` util method
- Benchmark and tests

Benchmark details:

```
goos: darwin
goarch: arm64
BenchmarkGetProtocolVersion
BenchmarkGetProtocolVersion-12    	   58597	     19727 ns/op
```


cmd/soroban-rpc/internal/methods/get_version_info.go
```diff
@@ -22,33 +22,17 @@ type GetVersionInfoResponse struct {
...
-		defer func() {
-			_ = readTx.Done()
-		}()
+func NewGetVersionInfoHandler(
+	logger *log.Entry,
+	ledgerEntryReader db.LedgerEntryReader,
+	ledgerReader db.LedgerReader,
+	daemon interfaces.Daemon,
+) jrpc2.Handler {
+	core := daemon.GetCore()
```
Shaptic, line 31: `core` can be `nil` here, right?

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,53 @@
...
+	ledgerReader db.LedgerReader,
+) (uint32, error) {
+	var protocolVersion uint32
+	readTx, err := ledgerEntryReader.NewCachedTx(ctx)
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		_ = readTx.Done()
+	}()
```
Shaptic, line 22: `defer readTx.Done()` has the same effect, right?

cmd/soroban-rpc/internal/methods/get_network.go
```diff
@@ -17,20 +19,23 @@ type GetNetworkResponse struct {
...
+) jrpc2.Handler {
 	return NewHandler(func(ctx context.Context, request GetNetworkRequest) (GetNetworkResponse, error) {
-		info, err := coreClient.Info(ctx)
+		protocolVersion, err := getProtocolVersion(ctx, ledgerEntryReader, ledgerReader)
 		if err != nil {
-			return GetNetworkResponse{}, (&jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: err.Error(),
-			})
+			logger.WithError(err).Info("error occurred while fetching protocol version")
```
Shaptic, line 32: Shouldn't it be an `Error` log?
```suggestion
			logger.WithError(err).Error("failed to fetch protocol version")
```

cmd/soroban-rpc/internal/methods/get_version_info.go
psheth9: hmm I dont think so we error out early if that happens in `daemon.go`

```
core, err := newCaptiveCore(cfg, logger)
	if err != nil {
		logger.WithError(err).Fatal("could not create captive core")
	}
```

cmd/soroban-rpc/internal/methods/util.go
psheth9: yeah actually it is same

psheth9: (I think golint will complain about ignoring error from readTx.Done() with one liner tho)

cmd/soroban-rpc/internal/methods/get_network.go
```diff
@@ -17,20 +19,23 @@ type GetNetworkResponse struct {
...
+) jrpc2.Handler {
 	return NewHandler(func(ctx context.Context, request GetNetworkRequest) (GetNetworkResponse, error) {
-		info, err := coreClient.Info(ctx)
+		protocolVersion, err := getProtocolVersion(ctx, ledgerEntryReader, ledgerReader)
 		if err != nil {
-			return GetNetworkResponse{}, (&jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: err.Error(),
-			})
+			logger.WithError(err).Error("failed to fetch protocol version")
```
2opremio, line 32: Why the behavior change here? If there is an error, we shouldn't be using `protocolVersion` later on

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,53 @@
...
+
+	latestLedger, err := readTx.GetLatestLedgerSequence()
+	if err != nil {
+		return 0, err
+	}
```
2opremio, line 27:
```suggestion
	readTx, err := ledgerEntryReader..GetLatestLedgerSequence(ctx)
	if err != nil {
		return 0, err
	}
```

cmd/soroban-rpc/internal/methods/util.go
2opremio: It's better to use `nolint` than adding a defer simply to ignore the error

2opremio: And even better than ignoring the error, add it to the returned error with `errors.Join()1

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,53 @@
...
+
+	_, protocolVersion, err = getBucketListSizeAndProtocolVersion(ctx, ledgerReader, latestLedger)
+	if err != nil {
+		return 0, err
+	}
+	return protocolVersion, nil
+}
+
+func getBucketListSizeAndProtocolVersion(
```
2opremio, line 36: This is too specific to have as a generic util function (that's why it was in `simulate_transaction.go`). The function is tiny, so it's fine to replicate the needed parts of `getBucketListSizeAndProtocolVersion()` into `getBucketListSizeAndProtocolVersion()` and leave it where it was.

This will also allow you to get the ledger sequence and ledger in the same transaction.

cmd/soroban-rpc/internal/methods/get_network.go
psheth9: oh yes, good catch !! We dont want to change this.

cmd/soroban-rpc/internal/methods/util.go
psheth9: Got it. replicated some part of it, but keeping `getProtocolVersion` in util

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return 0, err
+	}
+	if !ok {
+		return 0, fmt.Errorf("missing meta for latest ledger (%d)", latestLedger)
+	}
+	if closeMeta.V != 1 {
+		return 0, fmt.Errorf("latest ledger (%d) meta has unexpected verion (%d)", latestLedger, closeMeta.V)
+	}
+	return uint32(closeMeta.V1.LedgerHeader.Header.LedgerVersion), nil
```
Shaptic, line 41: This routine seems really heavy just to get the latest protocol version :/ Can we cache it? Also it shares a lot of code with `getBucketListSizeAndProtocolVersion`; at the very least this can use that after the `closeMeta` acquisition, yeah?

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,42 @@
...
+	ledgerEntryReader db.LedgerEntryReader,
+	ledgerReader db.LedgerReader,
+) (uint32, error) {
+	readTx, err := ledgerEntryReader.NewCachedTx(ctx)
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		doneErr := readTx.Done()
+		err = errors.Join(err, doneErr)
```
Shaptic, line 22: TIL :+1:

cmd/soroban-rpc/internal/methods/util.go
Shaptic: Oh I just saw that you're using a cached tx. That seems fine then? If you could throw in a benchmark that'd be sick

2opremio: It uses a cached transaction which will cache the entries when requested multiple times, but not the first time.

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,42 @@
...
+
+func getProtocolVersion(
+	ctx context.Context,
+	ledgerEntryReader db.LedgerEntryReader,
+	ledgerReader db.LedgerReader,
+) (uint32, error) {
+	readTx, err := ledgerEntryReader.NewCachedTx(ctx)
+	if err != nil {
+		return 0, err
+	}
```
2opremio, line 19: As I mentioned in a prior review, there is no point int creating a transaction just to obtain the latest ledger sequence. Transactions make sense when querying multiple things from the DB at disjoint times.

2opremio: you can obtain the latest ledger sequence without a transaction, but it would be better if you get the both latest ledger and and the latest ledger sequence within the same transaction

2opremio: Either option is fine but don't start a transaction to query a single thing.

cmd/soroban-rpc/internal/methods/util.go
psheth9: @Shaptic  Did some quick benchmarking for `getProtocolVersion`. Not bad at all as it is fetching data from cached tx.

```

goos: darwin
goarch: arm64
BenchmarkGetProtocolVersion
BenchmarkGetProtocolVersion-12    	   58597	     19727 ns/op
```

psheth9: Added benchmark in `util_test.go`

cmd/soroban-rpc/internal/methods/util.go
```diff
@@ -0,0 +1,32 @@
...
+	if !ok {
+		return 0, fmt.Errorf("missing meta for latest ledger (%d)", latestLedger)
+	}
+	if closeMeta.V != 1 {
+		return 0, fmt.Errorf("latest ledger (%d) meta has unexpected verion (%d)", latestLedger, closeMeta.V)
+	}
+	return uint32(closeMeta.V1.LedgerHeader.Header.LedgerVersion), nil
```
Shaptic, line 31: You can replace this with `getBucketListSizeAndProtocolVersion(ctx, ledgerReader, latestLedger)`, right?

psheth9: I intentionally kept it like that because I dont want to add dependency of `simulate_transaction.go` in here

Initially what I had in mind was to have just one common method which can be utilized by `simulate_txn` and `version_info` file but refactored that based on this [comment](https://github.com/stellar/soroban-rpc/pull/198#discussion_r1664979754).

2opremio (approved):
Approved! (modulo the benchmark providing a reasonable time for the endpoint)

PR #207, opened 2024-06-10
Migrate data to new transactions table

2opremio:
Also, create data migration infrastructure so that future migrations can be easily added.


cmd/soroban-rpc/internal/db/db.go
```diff
@@ -52,7 +53,14 @@ type dbCache struct {
...
 type DB struct {
 	db.SessionInterface
-	cache dbCache
+	cache *dbCache
+}
+
+func (db *DB) Clone() *DB {
+	return &DB{
+		SessionInterface: db.SessionInterface.Clone(),
+		cache:            db.cache,
```
tamirms, line 62: I'm not sure if the cache should be propagated to the cloned `DB` instance since the cache contains data which may be bound to the current running transaction. If you look at the implementation of `Clone()` on the db session you will find that it does not clone the transaction state:

https://github.com/stellar/go/blob/master/support/db/session.go#L120-L127

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -93,28 +101,50 @@ func OpenSQLiteDB(dbFilePath string) (*DB, error) {
...
-	sql := sq.Select("value").From(metaTableName).Where(sq.Eq{"key": latestLedgerSequenceMetaKey})
+func getMetaBool(ctx context.Context, q db.SessionInterface, key string) (bool, error) {
+	valueStr, err := getMetaValue(ctx, q, key)
+	if err != nil {
+		return false, err
+	}
+	return strconv.ParseBool(valueStr)
+}
+
+func setMetaBool(ctx context.Context, q db.SessionInterface, key string) error {
```
tamirms, line 119: nit: imo the name `setMetaBool()` implies you can set it to true or false

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -0,0 +1,174 @@
...
+	return g.migration.ApplicableRange()
+}
+
+func (g *guardedMigration) Commit(ctx context.Context) error {
+	if g.alreadyMigrated {
+		return nil
+	}
+	err := setMetaBool(ctx, g.db.SessionInterface, g.guardMetaKey)
+	if err != nil {
+		return errors.Join(err, g.Rollback(ctx))
```
tamirms, line 156: I noticed we don't rollback when encountering `Apply()` errors. But the lack of rollbacks don't cause any issues because we exit the process as soon as encountering an error in `mustInitializeStorage()`. So perhaps we don't need to rollback here either? Or maybe the `Rollback()` function can be removed entirely since any error in the migration is treated as a fatal error.

2opremio: Good catch. We don't currently roll back, but we might in the future (depending on how we end up applying the migrations). We should clean-up the code (and maybe not apply the migrations in the daemon).

tamirms:
@2opremio overall looks good! Could you add some tests

cmd/soroban-rpc/internal/db/db.go
2opremio: Uhm, I need to think about this. I think it's probably fine to clone it at selected places. I am tempted to remove the method and just copy it manually in the only `Clone()` invokation

2opremio: I just removed the Clone() method

2opremio:
@tamirms Sure thing!

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -0,0 +1,183 @@
...
+		if localErr := m.Rollback(ctx); localErr != nil {
+			err = errors.Join(err, localErr)
+		}
+	}
+	return err
+}
+
+// guardedMigration is a db data migration whose application is guarded by a boolean in the meta table
+// (after the migration is applied the boolean is set to true, so that the migration is not applied again)
+type guardedMigration struct {
```
Shaptic, line 104: Could we guard against duplicate migrations by comparing the `ledger_sequence` range in the transactions table against ledger meta rather than this method? Because wouldn't this prevent the migration from happening on subsequent releases? So if someone upgrades to 21.3.0 then 21.3.1 they wouldn't update their transactions table?

2opremio: > Could we guard against duplicate migrations by comparing the ledger_sequence range in the transactions table against ledger meta rather than this method?

The length of meta doesn't need to match the length of the transactions due to the configurable window size. What's more, it can have changed several times.

> Because wouldn't this prevent the migration from happening on subsequent releases? So if someone upgrades to 21.3.0 then 21.3.1 they wouldn't update their transactions table?

I am not sure I understand. Upgrades should only happen once.

Shaptic: Lemme redirect my question to something simpler so that I can understand the flow here: when does the "the migration happen" meta key get set back to `false`? Or do we need new meta keys every time we need a migration?

2opremio: There is one meta key per migration

2opremio:
@tamirms PTAL 

I added a pretty sophisticated integration test which tests the migration of all prior RPC versions (within the same protocol) to the current code.

It does so by pulling the previously released RPC containers, submitting a transaction to them and then running the current RPC version with the same database and network.

cmd/soroban-rpc/internal/db/migration.go
2opremio: We could maybe use a dedicated table for the data migration keys (instead of the meta table). But I doubt that's a considerable improvement (famous last words)

tamirms:
@2opremio the integration tests seem to be failing in github CI. If it's too difficult to get the integration tests to work I think a unit test in the db package should be sufficient

2opremio:
@tamirms integration tests work locally, let me try to debug what's going on in CI

2opremio:
I think I may have encountered a race condition in the web server. One of the integration test jobs is passing while others get stuck at :

```
2024-06-13T13:05:55.8549758Z time="2024-06-13T13:05:55.854Z" level=info msg="starting Soroban RPC" commit= pid=9794 version=0.0.0
2024-06-13T13:05:55.8696456Z     integration.go:236: Waiting for RPC to be up...
2024-06-13T13:05:55.8698645Z time="2024-06-13T13:05:55.869Z" level=info msg="starting HTTP server" addr="localhost:8000" pid=9794
2024-06-13T13:05:55.8701329Z     integration.go:252: RPC still unhealthy [-32603] Post "http://localhost:8000": dial tcp [::1]:8000: connect: connection refused 
[...]
2024-06-13T13:06:26.8937223Z     integration.go:256: RPC failed to get healthy in 30 seconds
```

Which means that even if the webserver is listening at localhost:8000, the client cannot reach it

2opremio:
Uhm, I am starting to think it's a problem with the JSONRPC client.

If I print the listening ports right after the error, I see that soroban RPC is indeed listening on port 8000

```
    integration.go:265: RPC still unhealthy Post "http://localhost:8000": dial tcp [::1]:8000: connect: connection refused 
[...]
Who is listening on RPC port?
COMMAND    PID   USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
test.test 9885 runner   17u  IPv4 115645      0t0  TCP localhost:8000 (LISTEN)
```

I think that the client may not retry after the first attempt :S

2opremio:
@tamirms This is finally ready to go.

The new migration test required rewriting most of the integration-test infra, so the PR ended up getting quite big (sorry about that).

As a result the integration test infra is way cleaner.

BTW, the spurious test failures I was fighting were due to https://github.com/creachadair/jrpc2/issues/118

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -107,8 +107,7 @@ func (txn *transactionHandler) InsertTransactions(lcm xdr.LedgerCloseMeta) error
 	}
 	_, err = query.RunWith(txn.stmtCache).Exec()
 
-	L.WithError(err).
```
tamirms, line 110: why is the error removed from the log?

tamirms (approved):
great job debugging https://github.com/creachadair/jrpc2/issues/118 , this looks great!

cmd/soroban-rpc/internal/test/docker-compose.rpc.yml
```diff
@@ -0,0 +1,17 @@
...
+    ports:
+      - "8000:8000"
+      - "8080:8080"
+    command: --config-path /soroban-rpc.config
+    volumes:
+      - ${RPC_CONFIG_MOUNT_DIR}/stellar-core-integration-tests.cfg:/stellar-core.cfg
+      - ${RPC_CONFIG_MOUNT_DIR}/soroban-rpc.config:/soroban-rpc.config
+      - ${RPC_SQLITE_MOUNT_DIR}:/db/
+    # Needed so that the sql database files created in the container
+    # have the same uid and gid as in the host
```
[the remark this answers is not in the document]
2opremio: Because the database needs to be accessiable by both the host and the container for the migrations test. Otherwise, after it's created by the container it isn't accessible by the host.

cmd/soroban-rpc/internal/db/transaction.go
2opremio: Because it can be nil (it normally is nil) and the consumer already gets the error.

PR #213, opened 2024-06-12
fix: return GetTransactionRaw with getTransaction

willemneal:
### What

`getTransaction` method should return a `GetTransactionResponseRaw` which then can be converted to `GetTransaction`. This was causing the struct to not be deserialized correctly.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/crates/stellar-rpc-client/src/lib.rs
+++ b/cmd/crates/stellar-rpc-client/src/lib.rs
@@ -799,7 +799,7 @@ impl Client {
 
     ///
     /// # Errors
-    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponse, Error> {
+    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponseRaw, Error> {
         let mut oparams = ObjectParams::new();
         oparams.insert("hash", tx_id)?;
         Ok(self.client().request("getTransaction", oparams).await?)
@@ -827,7 +827,7 @@ impl Client {
         let exponential_backoff: f64 = 1.0 / (1.0 - E.powf(-1.0));
         let mut sleep_time = Duration::from_secs(1);
         loop {
-            let response = self.get_transaction(tx_id).await?;
+            let response: GetTransactionResponse = self.get_transaction(tx_id).await?.try_into()?;
             match response.status.as_str() {
                 "SUCCESS" => {
                     // TODO: the caller should probably be printing this
--- a/cmd/crates/stellar-rpc-client/src/lib.rs
+++ b/cmd/crates/stellar-rpc-client/src/lib.rs
@@ -802,7 +802,8 @@ impl Client {
     pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponse, Error> {
         let mut oparams = ObjectParams::new();
         oparams.insert("hash", tx_id)?;
-        let resp: GetTransactionResponseRaw = self.client().request("getTransaction", oparams).await?;
+        let resp: GetTransactionResponseRaw =
+            self.client().request("getTransaction", oparams).await?;
         Ok(resp.try_into()?)
     }
 
--- a/cmd/crates/stellar-rpc-client/src/lib.rs
+++ b/cmd/crates/stellar-rpc-client/src/lib.rs
@@ -799,10 +799,11 @@ impl Client {
 
     ///
     /// # Errors
-    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponseRaw, Error> {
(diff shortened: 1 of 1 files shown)


Shaptic (approved):
seems good :+1: 

but why didn't this come up before?

mollykarcher:
> but why didn't this come up before?

Second this. This seems like something that the system-tests should have been able to catch. Did we fail to update the e2e workflow or something?

cmd/crates/stellar-rpc-client/src/lib.rs
```diff
@@ -799,7 +799,7 @@ impl Client {
 
     ///
     /// # Errors
-    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponse, Error> {
+    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponseRaw, Error> {
```
leighmcculloch, line 802: I think the issue here is that this should still be returning the GetTransactionResponse, but it should be converting from the raw to the non raw here. The raw isn't intended to be used externally I thought.

leighmcculloch (changes requested):
See comment above.

cmd/crates/stellar-rpc-client/src/lib.rs
leighmcculloch: Looking at the code prior to its last change it does appear that the function used to return the raw version, which really begs the question why the two versions exist if we expose the raw version to people anyway: https://github.com/stellar/soroban-rpc/commit/587a229541043ca3b2e3e4514296d6c87625b25c#diff-ef42a2e4c0c5ce57fe0118fafc5db591609074ffcb19fc3a7a8576b32026d423L860-L863

leighmcculloch: I think we should fix this by having all the public fns return the same type, `GetTransactionResponse`, and make GetTransactioResponseRaw unexported. cc @2opremio wdyt?

cmd/crates/stellar-rpc-client/src/lib.rs
```diff
@@ -799,7 +799,7 @@ impl Client {
...
         let mut oparams = ObjectParams::new();
         oparams.insert("hash", tx_id)?;
         Ok(self.client().request("getTransaction", oparams).await?)
```
leighmcculloch, line 805:
```suggestion
    pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponse, Error> {
        let mut oparams = ObjectParams::new();
        oparams.insert("hash", tx_id)?;
        let resp: GetTransactionResponseRaw = self.client().request("getTransaction", oparams).await?;
        Ok(resp.try_into()?)
```

cmd/crates/stellar-rpc-client/src/lib.rs
```diff
@@ -827,7 +827,7 @@ impl Client {
         let exponential_backoff: f64 = 1.0 / (1.0 - E.powf(-1.0));
         let mut sleep_time = Duration::from_secs(1);
         loop {
-            let response = self.get_transaction(tx_id).await?;
+            let response: GetTransactionResponse = self.get_transaction(tx_id).await?.try_into()?;
```
leighmcculloch, line 830:
```suggestion
            let response = self.get_transaction(tx_id).await?;
```

cmd/crates/stellar-rpc-client/src/lib.rs
Shaptic: yeah that's what the JS SDK does, e.g. for `(Raw)?LedgerEntryChange` [here](https://github.com/stellar/js-stellar-sdk/blob/master/src/rpc/api.ts#L171-L186).

cmd/crates/stellar-rpc-client/src/lib.rs
```diff
@@ -802,7 +802,8 @@ impl Client {
     pub async fn get_transaction(&self, tx_id: &Hash) -> Result<GetTransactionResponse, Error> {
         let mut oparams = ObjectParams::new();
         oparams.insert("hash", tx_id)?;
-        Ok(self.client().request("getTransaction", oparams).await?)
+        let resp: GetTransactionResponseRaw = self.client().request("getTransaction", oparams).await?;
```
leighmcculloch, line 805:
```suggestion
        let resp: GetTransactionResponseRaw =
            self.client().request("getTransaction", oparams).await?;
```

leighmcculloch (approved):

leighmcculloch:
I've approved and am merging this PR. @2opremio @mollykarcher Is there a way to ship this in a bug fix release, such as `21.3.1` instead of waiting for `21.4.0`?

PR #215, opened 2024-06-13
Add getEvents backed by DB 

psheth9:
Overall design:

- Ingest events into DB: store entire `eventXDR` along with metadata like contract_ids, topics etc..
- Combine `contract_ids` and `topics` from all the filters for a given request and pass it along to DB (to save multiple DB calls)
- Fetch events and apply filter again based on the request (on less data comparatively)

Other Refactor:
- Remove in-memory events store
- Introduce `endLedgerSeq`
- Add migration related logic for events

Other related PR:

- Migration failures and refactor daemon.go [PR](https://github.com/stellar/soroban-rpc/pull/262)

Data Flow:

image

Design Docs:
 
- [Eng doc](https://docs.google.com/document/d/1Q5AJvszwFmwl3NzHh7PJIMj4mhN3HgHRn_7IOQGeMWg/edit)

- Benchmarking doc [here](https://docs.google.com/document/d/14PTIC1dTNZBbuAtI91baFSPv9se4rO9sqcphV-81Yo8/edit#heading=h.sacw04vfsw6u)

Parent Epic https://github.com/stellar/soroban-rpc/issues/115
Fixes: https://github.com/stellar/soroban-rpc/issues/190


psheth9:
`ledgerCloseMetaWithEvents` in `get_events_test` does not seem to work during DB ingestion.

Error logs:
`unknown tx hash in LedgerCloseMeta: 4705598f90aa1c2affc48ffb10f42ab1a4d9545fcb8ed0adfd6b2ae141aa1c21`

Shaptic:
@psheth9 you're changing the hash of the transaction when you inject events into it after creation which makes it invalid

psheth9:
> you're changing the hash of the transaction when you inject events into it after creation which makes it invalid

This is not the case as I am not touching any fields related to hash. Also `ledgerCloseMetaWithEvents` is being used everywhere (in old code) to validate `get_events` RPC so this should work automatically with new code. Unfortunately it does not :(

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -6,13 +6,16 @@ import (
...
 	sq "github.com/Masterminds/squirrel"
 	_ "github.com/mattn/go-sqlite3"
 	"github.com/prometheus/client_golang/prometheus"
 	migrate "github.com/rubenv/sql-migrate"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"path"
+	"strconv"
+	"sync"
+	"testing"
```
2opremio, line 18: If this is accepted by the linter ... it's buggy

psheth9: hmm yeah I dont see that in [error logs](https://github.com/stellar/soroban-rpc/actions/runs/9606783845/job/26496854173?pr=215)

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,169 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 
 	return nil
 }
+
+type ScanFunction func(xdr.DiagnosticEvent, events.Cursor, int64, *xdr.Hash) bool
+
+// trimEvents removes all Events which fall outside the ledger retention window.
+func (eventHandler *eventHandler) trimEvents(latestLedgerSeq uint32, retentionWindow uint32) error {
```
aditya1702, line 116: I see this function is the exact copy of `trimTransactions` function. Would it be a good idea to just combine and use a common function and pass the event/txn table related info as params?

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,169 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+			return fmt.Errorf("failed to index to tx %d in ledger %d: %w", txIndex, lcm.LedgerSequence(), err)
+		}
+
+		err = reader.Seek(txIndex - 1)
+		if err != nil {
+			return fmt.Errorf("failed to index to tx %d in ledger %d: %w", txIndex, lcm.LedgerSequence(), err)
+		}
+
+		ledgerCloseTime := lcm.LedgerCloseTime()
+		ledgerTx, err := reader.Read()
```
aditya1702, line 183: Handle error by Read()

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,169 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		WithField("duration", time.Since(start)).
+		Debugf("Fetched and decoded all the events with filters - contractIDs: %v ", contractIDs)
+
+	return nil
+}
+
+// GetLedgerRange returns the min/max ledger sequence numbers from the events table
+// TODO: Once we unify all the retention window we would keep only one GetLedgerRange logic
+
+func (eventHandler *eventHandler) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
```
aditya1702, line 212: This can be simplified using min/max queries - https://github.com/stellar/soroban-rpc/pull/217/files#diff-0de90191c7a664162c6d9b9ddee3a00bf98a069c661fdaa3eb0bda6b728c22f7R143.

But yes, once these changes go in, we can just use 1 implementation for all 3 tables

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -302,25 +306,48 @@ type GetEventsResponse struct {
...
+	for contractID := range contractIDSet {
+		contractIDs = append(contractIDs, contractID)
+	}
+	return contractIDs
 }
 
-func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse, error) {
+func (h eventsRPCHandler) getEvents(ctx context.Context, request GetEventsRequest) (GetEventsResponse, error) {
+
 	if err := request.Valid(h.maxLimit); err != nil {
```
aditya1702, line 335: We can also include the same check for latestLedger <= start <= oldestLedger that we do here in `getTransactions` - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/get_transactions.go#L40

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -375,13 +402,19 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
 		}
 		results = append(results, info)
 	}
+
 	return GetEventsResponse{
-		LatestLedger: latestLedger,
+		LatestLedger: ledgerRange.LastLedger.Sequence,
```
aditya1702, line 407: Would be a good idea to include oldestLedger like other `getTransactions` endpoint

aditya1702 (commented):
Looking good! Did a first pass

cmd/soroban-rpc/internal/methods/get_events_test.go
```diff
@@ -1,8 +1,15 @@
 package methods
```
aditya1702, line 1: We should also add an integration test for `getEvents`

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: Yes It is a good idea, it is out of scope for this PR, created issue here: https://github.com/stellar/soroban-rpc/issues/231

2opremio:
Please refactor the loop at https://github.com/stellar/soroban-rpc/pull/215/files#diff-4dea5171d622e1239f0fb5dc9a2941a2e7a1cf15a4a953cfd71f8235e65e3be3R317-R344 . Now that you removed the in-memory datastructure, it should only loop the strictly necessary ledgers.

You will need to:

1. Calculate the strictly necessary range. I made sure migrations provide a range precisely for this. You will need to take into account the range needed by fee stats too.
2. Create a variant of `StreamAllLedgers()` so that it only streams over the necessary ledgers and invoke it

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE events(
     id                TEXT PRIMARY KEY,
     ledger_sequence   INTEGER NOT NULL,
     application_order INTEGER NOT NULL,
-    contract_id       BLOB,
+    contract_id       TEXT,
```
2opremio, line 8: Why is this? The DB shouldn't be storing text ids (i.e. hex) it should be binary, so that it occupies less space.

cmd/soroban-rpc/internal/db/db.go
2opremio: golang-ci is still failing, so I guess there is a problem here. Regardless, stdlib imports should be present in their own block at the very beginning.

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -184,10 +184,23 @@ func (g *guardedMigration) Rollback(ctx context.Context) error {
 func BuildMigrations(ctx context.Context, logger *log.Entry, db *DB, cfg *config.Config) (Migration, error) {
 	migrationName := "TransactionsTable"
 	factory := newTransactionTableMigration(ctx, logger.WithField("migration", migrationName), cfg.TransactionLedgerRetentionWindow, cfg.NetworkPassphrase)
-	m, err := newGuardedDataMigration(ctx, migrationName, factory, db)
+	m1, err := newGuardedDataMigration(ctx, migrationName, factory, db)
 	if err != nil {
 		return nil, fmt.Errorf("creating guarded transaction migration: %w", err)
 	}
 	// Add other migrations here
```
2opremio, line 191:
```suggestion
```

2opremio:
I don't think it makes sense to have a dedicated events package now that it only contains the cursor implementation

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +112,225 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		}
+		if latestLedger > retentionWindow {
+			firstLedgerToMigrate = latestLedger - retentionWindow
+		}
+
+		// Truncate the table, since it may contain data, causing insert conflicts later on.
+		_, err := db.Exec(ctx, sq.Delete(transactionTableName).Where(sq.Lt{"ledger_sequence": firstLedgerToMigrate}))
+		if err != nil {
+			return nil, fmt.Errorf("couldn't truncate the table %q: %w", transactionTableName, err)
+		}
```
2opremio, line 328: You don't need to do this since the migration is being applied on an empty DB. In the case of transactions we realized there was no migration after the code was in the wild.

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -366,3 +373,14 @@ func runSQLMigrations(db *sql.DB, dialect string) error {
...
+func NewTestDB(tb testing.TB) *DB {
+	tmp := tb.TempDir()
+	dbPath := path.Join(tmp, "db.sqlite")
+	db, err := OpenSQLiteDB(dbPath)
+	require.NoError(tb, err)
+	tb.Cleanup(func() {
+		assert.NoError(tb, db.Close())
+	})
+	return db
+}
```
2opremio, line 386: I am not sure why this is needed, but it shouldn't live with production code (it causes bloat). Consider placing it in a `_test.go` file or a package only used by tests.

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -81,12 +96,12 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 			Columns("id", "ledger_sequence", "application_order", "contract_id", "event_type")
 
 		for index, e := range txEvents {
-			var contractId []byte
+			var contractID string
 			if e.Event.ContractId != nil {
-				contractId = e.Event.ContractId[:]
+				contractID = strkey.MustEncode(strkey.VersionByteContract, (*e.Event.ContractId)[:])
```
2opremio, line 101: Please don't store strkeys in the DB (they occupy a lot more space). Just store the 32 byte hash.

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
2opremio: In fact, I think this could be a `BLOB(32)`

2opremio: Now I am noticing that the id field is also text.  Why is that? (sorry I didn't notice before)

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +112,225 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 
 	return nil
 }
+
+type ScanFunction func(xdr.DiagnosticEvent, events.Cursor, int64, *xdr.Hash) bool
```
2opremio, line 116: Please add parameter names (or use more specific types) so that the reader can understand what are the `int64` and `*xdr.Hash` parameters.

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +112,225 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+	for _, row := range rows {
+		eventCursorID, txIndex, lcm := row.EventCursorID, row.TxIndex, row.Lcm
+		reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
+		if err != nil {
+			return fmt.Errorf("failed to index to tx %d in ledger %d: %w", txIndex, lcm.LedgerSequence(), err)
+		}
+
+		err = reader.Seek(txIndex - 1)
+		if err != nil {
+			return fmt.Errorf("failed to index to tx %d in ledger %d: %w", txIndex, lcm.LedgerSequence(), err)
```
2opremio, line 186: The error message text is identical and not descriptive

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +112,225 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		return fmt.Errorf(
+			"db read failed for start ledger cursor= %v contractIDs= %v: %w",
+			cursorRange.Start.String(),
+			contractIDs,
+			err)
+	} else if len(rows) < 1 {
+		eventHandler.log.Infof("No events found for start ledger cursor= %v contractIDs= %v",
+			cursorRange.Start.String(),
+			contractIDs,
+		)
```
2opremio, line 173: This is too noisy

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +112,225 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		WithField("endLedgerSequence", cursorRange.End.Ledger).
+		WithField("duration", time.Since(start)).
+		Debugf("Fetched and decoded all the events with filters - contractIDs: %v ", contractIDs)
+
+	return nil
+}
+
+// GetLedgerRange returns the min/max ledger sequence numbers from the events table
+// TODO: Once we unify all the retention window we would keep only one GetLedgerRange logic
+
```
2opremio, line 221:
```suggestion
// GetLedgerRange returns the min/max ledger sequence numbers from the events table
// TODO: Once we unify all the retention windows we should  keep only one GetLedgerRange function
```

cmd/soroban-rpc/internal/db/event.go
2opremio: Yep, please simplify the query following the approach from https://github.com/stellar/soroban-rpc/pull/217

cmd/soroban-rpc/internal/methods/get_events.go
2opremio: Why is it necessary? It adds more pressure to the DB. I agree that we should be consistent across both endpoints but `GetLedgerRange`  does not seem to be particularly cheap.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -333,6 +359,16 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
 			limit = request.Pagination.Limit
 		}
 	}
+	end := events.Cursor{Ledger: request.StartLedger + ledgerbucketwindow.OneDayOfLedgers}
```
2opremio, line 362: Is this for limiting the maximum number of ledgers to query? If so it should probably be parameterized in the command-line like we do for transactions or at the very very least it should be commented and use a constant.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -333,6 +359,16 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
...
+	cursorRange := events.CursorRange{Start: start, End: end}
+
+	// Check if requested start ledger is within stored ledger range
+	if start.Ledger < ledgerRange.FirstLedger.Sequence || start.Ledger > ledgerRange.LastLedger.Sequence {
+		return GetEventsResponse{}, &jrpc2.Error{
+			Code:    jrpc2.InvalidRequest,
+			Message: fmt.Sprintf("startLedger must be within the ledger range: %d - %d", ledgerRange.FirstLedger.Sequence, ledgerRange.LastLedger.Sequence),
+		}
+	}
```
2opremio, line 371: As I mentioned above. Why is this  needed?

cmd/soroban-rpc/internal/methods/get_events.go
2opremio: Also, a day worth of ledgers can be quite a lot. We may need to rethink this based on how expensive it is to query events.

2opremio: We should probably limit by number of events and not ledgers

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -341,20 +377,25 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
...
+		}
+
+		if request.Matches(event) && cursor.Cmp(start) >= 0 {
+			cursorSet[cursorID] = struct{}{}
+			found = append(found, entry{cursor, ledgerCloseTimestamp, event, txHash})
+		}
+		return uint(len(found)) < limit
+	}
+
+	err = h.dbReader.GetEvents(ctx, cursorRange, contractIDs, f)
```
2opremio, line 398: I am concerned about the efficiency of this. When events were stored in-memory it was probably fine to scan through a potentially large range of events and filter them by applying a function to all of them.

I suspect that the same approach can be prohibitive when moving the events to a database and extending the default history to 7 days.

It's good that we filter by contract ID on the database but to ensure this performs decently we may either need to either:
1. Be more restrictive (e.g. enforce providing contract IDs, with a cap on how many contracts to provide)
2. Move more of the filtering to the DB

2opremio: Regardless I would like to see a benchmark for event querying

2opremio (changes requested):
This gets us much closer! I did leave quite a few comments. The most important ones concern the data format stored in the DB and the scanning approach.

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: Ideally this should never scan 7 days of history as we are restricting fetch query by [this](https://github.com/stellar/soroban-rpc/pull/215/files#diff-1f027a1543c912ae4bc1533d9fece665a2b555548e8f74a8e591a70fdc317b2fR362)

Some context around Ledger restriction for given request here in design [doc](https://docs.google.com/document/d/1Q5AJvszwFmwl3NzHh7PJIMj4mhN3HgHRn_7IOQGeMWg/edit#heading=h.ah30ftbvhnx7)

Ledger restriction parameter (N) should be more explicit and ideally should be part of the config. so for any given get request we will only look at start: start + N ledger. Also it might be also useful to include lastProcessedLedger as a part of response so client understands what to use as startLedger in follow up query. I will add this change once we agree.

I also liked the idea of making filters more restrictive because current filters are more loose ( contract_ids, event_type, topics all are optional) but this can be a additive change along with ledger restriction part.

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: limiting by events still won't solve the issue with current filters. We need to restrict scanning of ledgers for given request. 

> a day worth of ledgers can be quite a lot
hmmm we do look at day of ledgers in memory for events so this might not be true?

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: I agree that GetLedgerRange is expensive call but

1. This is the current behavior of getEvents to validate range and also to be consistent with other endpoints
2. We anyway going to call `GetLedgerRange` to include latest ledger in response

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
psheth9: id field here is basically a cursor id (TOID) of event. That is why it is stored as TEXT

cmd/soroban-rpc/internal/db/event.go
psheth9: Now that we are unifying all retention window to 1 day, It makes more sense to use the common `GetLedgerRange` function here. I will wait for Aditya's [change](https://github.com/stellar/soroban-rpc/pull/217/files#diff-0de90191c7a664162c6d9b9ddee3a00bf98a069c661fdaa3eb0bda6b728c22f7R143) to merge to main. otherwise it is gonna be a throwaway work on event side.

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
Shaptic: [TOID is a 64-bit number](https://github.com/stellar/go/blob/master/toid/main.go#L17-L18); can we have a 64-bit integer column in sqlite?

psheth9: Sorry my bad, I was storing the string representation of it. INTEGER makes sense.

cmd/soroban-rpc/internal/db/db.go
psheth9: All lint errors resolved now !! phew

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
tamirms: @psheth9 sorry for the late reply! that comment is basically saying if you have a collection of ledgers, operations, and transactions, the toid cannot uniquely identify an item in that collection because the ledger `n` has the same toid as the first operation in ledger `n` and the first transaction in ledger `n`. In that scenario, you don't know if it refers to the ledger, the transaction, or the operation.

However, the reason that comment does not apply in your situation is that you have a table consisting of only 1 type: events. If all objects have the same type in a collection then the toid will be unique.

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -13,11 +13,14 @@ import (
...
 	"github.com/stellar/go/network"
 	"github.com/stellar/go/support/strutils"
-
-	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/ledgerbucketwindow"
 )
 
-const defaultHTTPEndpoint = "localhost:8000"
+const (
+	// OneDayOfLedgers is (roughly) a 24 hour window of ledgers.
+	OneDayOfLedgers = 17280
```
Shaptic, line 20: This constant already exists somewhere, I think?

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -212,24 +215,36 @@ func (cfg *Config) options() Options {
...
+			ConfigKey:    &cfg.HistoryRetentionWindow,
+			DefaultValue: uint32(OneDayOfLedgers),
+			Validate:     positive,
+		},
+		// TODO: remove
 		{
 			Name: "event-retention-window",
 			Usage: fmt.Sprintf(
-				"configures the event retention window expressed in number of ledgers,"+
+				"(Deprecated, overidden by history-retention-window) configures the event retention window expressed in number of ledgers,"+
```
Shaptic, line 232: typo:
```suggestion
				"(Deprecated, overridden by history-retention-window) configures the event retention window expressed in number of ledgers,"+
```

Shaptic: Also it's not overridden, right? It's the max of the three. Maybe this is okay as help text, though, so that we can move users away from setting that flag.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -301,13 +288,12 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
 }
 
 // mustInitializeStorage initializes the storage using what was on the DB
-func (d *Daemon) mustInitializeStorage(cfg *config.Config) (*feewindow.FeeWindows, *events.MemoryStore) {
-	eventStore := events.NewMemoryStore(
-		d,
+func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows {
+	feewindows := feewindow.NewFeeWindows(
+		cfg.ClassicFeeStatsLedgerRetentionWindow,
+		cfg.SorobanFeeStatsLedgerRetentionWindow,
```
Shaptic, line 294: Separate question, cc @2opremio: should we bother with letting these be different? Or even configurable?

cmd/soroban-rpc/internal/db/cursor.go
```diff
@@ -27,6 +27,13 @@ type Cursor struct {
 	Event uint32
 }
 
+type CursorRange struct {
+	// Start defines the (inclusive) start of the range.
+	Start Cursor
+	// End defines the (exclusive) end of the range.
```
Shaptic, line 33: Can you double-check that this is the same behavior as for the other paginated endpoints? I thought we started the next page of reads at +1 the given start cursor (cc @aditya1702 for `getTransactions` context)

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -42,9 +55,9 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 	var txReader *ingest.LedgerTransactionReader
 	txReader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
 	if err != nil {
-		return errors.Wrapf(err,
-			"failed to open transaction reader for ledger %d",
-			lcm.LedgerSequence())
+		return fmt.Errorf(
+			"failed to open transaction reader for ledger %d: %w ",
+			lcm.LedgerSequence(), err)
```
Shaptic, line 60: why not wrap?

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,156 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+			cursorRange.Start.String(),
+			contractIDs,
+			err)
+	} else if len(rows) < 1 {
+		eventHandler.log.Debugf(
+			"No events found for ledger range: start ledger cursor= %v - end ledger cursor= %v contractIDs= %v",
+			cursorRange.Start.String(),
+			cursorRange.End.String(),
+			contractIDs,
+		)
```
Shaptic, line 177: recommendation: use `WithField` for some of these parameters, e.g.,
```suggestion
		eventHandler.log.
		  WithField("start", cursorRange.Start.String()).
		  WithField("end", cursorRange.End.String()).
		  WithField("contracts", contractIDs).
		  Debugf("...")
```

Also it'd be good to convert the IDs to hex for the output

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,156 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		return nil
+	}
+
+	for _, row := range rows {
+		txIndex, lcm := row.TxIndex, row.Lcm
+		reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(eventHandler.passphrase, lcm)
+		if err != nil {
+			return fmt.Errorf("failed to create ledger reader from LCM: %w", err)
```
Shaptic, line 185: `errors.Wrap` lets callers use `errors.Is` so it's preferable when you have an `err != nil`, whereas `fmt.Errorf` or `errors.New` is better when you are passing up a new error

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -65,10 +98,42 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
 		return results[0], true, nil
 	default:
-		return xdr.LedgerCloseMeta{}, false, fmt.Errorf("multiple lcm entries (%d) for sequence %d in table %q", len(results), sequence, ledgerCloseMetaTableName)
+		return xdr.LedgerCloseMeta{}, false, fmt.Errorf("multiple lcm entries (%d) for sequence %d in table %q",
+			len(results), sequence, ledgerCloseMetaTableName)
 	}
 }
 
+// GetLedgerRange pulls the min/max ledger sequence numbers from the meta table.
+func (r ledgerReader) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
```
Shaptic, line 107: Is this from an upstream merge? You should merge it into the base branch and then this PR diff will be smaller & easier to review

Shaptic (commented):
Did a quick pass

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -293,34 +298,61 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+	dbReader     db.EventReader
 	maxLimit     uint
 	defaultLimit uint
+	logger       *log.Entry
+	ledgerReader db.LedgerReader
 }
 
-func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse, error) {
+func combineContractIDs(filters []EventFilter) ([][]byte, error) {
+	contractIDSet := make(map[string]struct{})
```
Shaptic, line 319: We have https://github.com/stellar/go/tree/master/support/collections/set for this :+1:

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -293,34 +298,61 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+		for _, contractID := range filter.ContractIDs {
+			contractIDSet[contractID] = struct{}{}
+		}
+	}
+
+	contractIDs := make([][]byte, 0, len(contractIDSet))
+	for contractID := range contractIDSet {
+		id, err := strkey.Decode(strkey.VersionByteContract, contractID)
+		if err != nil {
+			return nil, fmt.Errorf("contract ID %v invalid", contractID)
```
Shaptic, line 331:
```suggestion
			return nil, errors.Wrapf(err, "invalid contract ID: %v", contractID)
```

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -333,28 +365,59 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
...
+	}
 
 	type entry struct {
-		cursor               events.Cursor
+		cursor               db.Cursor
 		ledgerCloseTimestamp int64
 		event                xdr.DiagnosticEvent
 		txHash               *xdr.Hash
 	}
 	var found []entry
```
Shaptic, line 395: Maybe it's worth giving this some large capacity to preallocate in case there are a lot of results that will lead to array expansion/copies?

Shaptic: e.g. using `limit`

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -333,28 +365,59 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
...
-			ClampEnd:   true,
-		},
-		func(event xdr.DiagnosticEvent, cursor events.Cursor, ledgerCloseTimestamp int64, txHash *xdr.Hash) bool {
-			if request.Matches(event) {
-				found = append(found, entry{cursor, ledgerCloseTimestamp, event, txHash})
-			}
-			return uint(len(found)) < limit
-		},
-	)
+	cursorSet := make(map[string]struct{})
```
Shaptic, line 396: same idea here, maybe give it `limit` capacity in advance? might not be able to use stellar/go's `collections.Set[T]` though? feel free to adapt it

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +109,156 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		}
+
+		// Find events based on filter passed in function f
+		for eventIndex, event := range diagEvents {
+			cur := Cursor{Ledger: lcm.LedgerSequence(), Tx: uint32(txIndex), Event: uint32(eventIndex)}
+			if !f(event, cur, ledgerCloseTime, &transactionHash) {
+				return nil
+			}
+		}
+	}
```
Shaptic, line 212: Could we parallelize this, maybe?? Unless the speed is acceptable - goroutines would probably increase allocs

cmd/soroban-rpc/internal/config/options.go
psheth9: hmmm this was always here in options.go. I did not create the new one

cmd/soroban-rpc/internal/config/options.go
psheth9: Yeah correct it is still max of 3, but we want users to just set the `history-retention-window `

cmd/soroban-rpc/internal/db/cursor.go
psheth9: Just to confirm the concern, you meant increment cursor id by 1 for txn/events if cursor is present in request?

Pagination for events is like this:
```

if request.Pagination.Cursor != nil {
			start = *request.Pagination.Cursor
			// increment event index because, when paginating,
			// we start with the item right after the cursor
			start.Event++
		}
```

psheth9: Equal to what we have for Transaction:
```

if request.Pagination != nil {
		if request.Pagination.Cursor != "" {
			cursorInt, err := strconv.ParseInt(request.Pagination.Cursor, 10, 64)
			if err != nil {
				return toid.ID{}, 0, &jrpc2.Error{
					Code:    jrpc2.InvalidParams,
					Message: err.Error(),
				}
			}
			*start = toid.Parse(cursorInt)
			// increment tx index because, when paginating,
			// we start with the item right after the cursor
			start.TransactionOrder++
		}
		if request.Pagination.Limit > 0 {
			limit = request.Pagination.Limit
		}
	}
```

cmd/soroban-rpc/internal/db/ledger.go
psheth9: Yeah this should not be here. Sorry I might have messed it up while resolving conflicts from upstream. Should be fixed now.

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: Nice !! Much cleaner

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -91,14 +100,48 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
 			}
-			query = query.Values(lcm.LedgerSequence(), tx.Index, contractID, int(e.Event.Type))
+
+			id := Cursor{Ledger: lcm.LedgerSequence(), Tx: tx.Index, Op: 0, Event: uint32(index)}.String()
+			eventBlob, err := xdr.MarshalBase64(e)
+			if err != nil {
+				return err
+			}
+
+			v0, ok := e.Event.Body.GetV0()
```
Shaptic, line 118: We don't need to store the body and the topics separately, you should just store the `data` field w/in the body

psheth9:
Benchmark Results for new getEvents design:


**For 3 million events:**

```
DB size stays at 1.8 GB
Latency ranges from 0.08 ms/op to 0.527 ms/op


```

**For 30 million events (upper bound):**

```
DB size stays at ~16 GB:
Latency ranges from 0.27 ms to 157 ms 

```


Allocs and memory usage capped at: `10K allocs and 0.8 MB`, on an avg it is very low. 

Note this is a huge improvement from the benchmarking results of the last design !!

cmd/soroban-rpc/internal/db/event.go
psheth9: What do you mean? We have to store topics separately in order to add index on db right?

cmd/soroban-rpc/internal/db/event.go
psheth9: I am not planning to do this refactor in this PR to reduce the chaos :)  This is a good idea and added the issue [here](https://github.com/stellar/soroban-rpc/issues/259)

cmd/soroban-rpc/internal/db/event.go
psheth9: Good point !! I was using standard `errors` package from go which has no `Wrap` method I think, okay will update all/most errors using `"github.com/stellar/go/support/errors"`

cmd/soroban-rpc/internal/db/event.go
psheth9: Latency has improved significantly with the new design so I have not experimented with it but we will refactor as we go forward. but honestly performance should not degrade overall for quite some time as current traffic is very low.

cmd/soroban-rpc/internal/db/event.go
psheth9: oh well now I remember linter complains when I use `errors.Wrapf ` :|

`use of `errors.Wrapf` forbidden because "Do not use stellar/go/support/errors, use the standard 'errors' package and fmt.Errorf()." (forbidigo)`

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: Bad advice... modern Go says you should use `errors.Join(err, fmt.Errorf("contract ID invalid: %v", contractID))` here :roll_eyes:

cmd/soroban-rpc/internal/db/event.go
Shaptic: We do have to store the topics separately to build indices, yes (though I honestly wonder how valuable those are given that most topics are really distinct? And thus not easily sort/searchable?), but we don't need to store the entire body in a separate column since topics are part of the body, so you can just store the `data` value rather than its `body = topics + data`.

cmd/soroban-rpc/internal/db/cursor.go
psheth9: cc @aditya1702  for another look and then I will resolve this if all looks good.

cmd/soroban-rpc/internal/db/event.go
psheth9: so you meant not to duplicate topics as they are already stored separately.

overall philosophy that I have in mind for this design is to store encoded `eventXDR` first and save other metadata to reduce the latency if needed e.g. topics, contract_ids etc..

so lets say if eventXDR evolves in future and topics are now increased from 5 to 7, this design would just work fine. We won't have to do any migration or addition of any new col in DB for added topics. 

Also currently topic1-topic5 are stored on DB but we can drop some of those too later on if we see first 2 topics are good enough to reduce the fetched data. 

>(though I honestly wonder how valuable those are given that most topics are really distinct? And thus not easily sort/searchable?)

hmm I mean it does help with requested invalid topic tho. but yeah without indexing it is at `~200-250 ms `when benchmarked against `30 millions events.` @Shaptic

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: Yeah I think we just need to stick to one pattern either `errors.Wrap` (from our internal package) vs the one you mentioned. Also our linter does not like `errors.Wrap` :(

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -78,15 +100,59 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+
+			var contractID []byte
 			if e.Event.ContractId != nil {
-				contractId = e.Event.ContractId[:]
+				contractID = e.Event.ContractId[:]
 			}
-			id := events.Cursor{Ledger: lcm.LedgerSequence(), Tx: tx.Index, Op: 0, Event: uint32(index)}.String()
-			query = query.Values(id, lcm.LedgerSequence(), tx.Index, contractId, int(e.Event.Type))
+
+			id := Cursor{Ledger: lcm.LedgerSequence(), Tx: tx.Index, Op: 0, Event: uint32(index)}.String()
```
tamirms, line 123: why not store this value as an integer? I think it will fit within 64 bits

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +163,182 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+func (eventHandler *eventHandler) trimEvents(latestLedgerSeq uint32, retentionWindow uint32) error {
+	if latestLedgerSeq+1 <= retentionWindow {
+		return nil
+	}
+
+	cutoff := latestLedgerSeq + 1 - retentionWindow
+	_, err := sq.StatementBuilder.
+		RunWith(eventHandler.stmtCache).
+		Delete(eventTableName).
+		Where(sq.Lt{"ledger_sequence": cutoff}).
```
tamirms, line 184: there is no `ledger_sequence` column. should this where clause be comparing the cutoff ledger with the id column?

cmd/soroban-rpc/internal/db/event.go
psheth9: You meant to store just TOID right? 

but existing [TOID](https://github.com/stellar/go/blob/master/toid/main.go#L17-L18) does not consider event index so in RPC we have Cursor which is basically TOID + event_index [this](https://github.com/stellar/soroban-rpc/blob/eb7cb8aa8ec4c4129bcb5dc0dfc8c6f963f967a5/cmd/soroban-rpc/internal/events/cursor.go#L31)

As a workaround I can replace `op_index` with `event index` as `op_index` is always 0 for events but might not be future proof?

cmd/soroban-rpc/internal/db/event.go
psheth9: oh yes totally forgot about updating this logic 👌

cmd/soroban-rpc/internal/db/event.go
tamirms: ok, I didn't realize that the toid + event index would not fit in 64 bits. I think it's fine to keep it as a string

cmd/soroban-rpc/internal/db/event.go
Shaptic: Good points all around, duplication of the structure seems worth it to avoid later complications :+1:

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -310,36 +303,46 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) (*feewindow.FeeWindow
...
-		// TODO: clean up once we remove the in-memory storage.
-		//       (we should only stream over the required range)
-		if r := dataMigrations.ApplicableRange(); r.IsLedgerIncluded(currentSeq) {
-			if err := dataMigrations.Apply(readTxMetaCtx, txmeta); err != nil {
-				d.logger.WithError(err).Fatal("could not run migrations")
+
+	// Merge migrations range and fee stats range to get the applicable range
+	latestLedger, err := db.NewLedgerEntryReader(d.db).GetLatestLedgerSequence(readTxMetaCtx)
+	if err != nil {
+		d.logger.WithError(err).Fatal("failed to get latest ledger sequence: %w", err)
```
Shaptic, line 310: nitpick: no need to log error twice
```suggestion
		d.logger.WithError(err).Fatal("failed to get latest ledger sequence")
```

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -310,36 +303,46 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) (*feewindow.FeeWindow
...
+	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
+	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: latestLedger - maxFeeRetentionWindow, LastLedgerSeq: latestLedger}
+	applicableRange := dataMigrations.ApplicableRange()
+	ledgerSeqRange = ledgerSeqRange.Merge(applicableRange)
+
+	err = db.NewLedgerReader(d.db).StreamLedgerRange(
+		readTxMetaCtx,
+		ledgerSeqRange.FirstLedgerSeq,
+		ledgerSeqRange.LastLedgerSeq,
+		func(txmeta xdr.LedgerCloseMeta) error {
```
Shaptic, line 322: Would it be better to pass this by reference to avoid copies of the large ledger meta structure on every invocation?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -310,36 +303,46 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) (*feewindow.FeeWindow
...
-				d.logger.WithError(err).Fatal("could not run migrations")
+
+	// Merge migrations range and fee stats range to get the applicable range
+	latestLedger, err := db.NewLedgerEntryReader(d.db).GetLatestLedgerSequence(readTxMetaCtx)
+	if err != nil {
+		d.logger.WithError(err).Fatal("failed to get latest ledger sequence: %w", err)
+	}
+
+	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
+	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: latestLedger - maxFeeRetentionWindow, LastLedgerSeq: latestLedger}
```
Shaptic, line 314: nit / suggestion: `LedgerSeq` is implied by the structure name (`LedgerSeqRange`), so you could just have `LedgerSeqRange { First, Last }`

Shaptic: You could / we should probably also use `ingest.Range` and add a `.Merge()` to it to DRY

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -310,36 +303,46 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) (*feewindow.FeeWindow
...
+		readTxMetaCtx,
+		ledgerSeqRange.FirstLedgerSeq,
+		ledgerSeqRange.LastLedgerSeq,
+		func(txmeta xdr.LedgerCloseMeta) error {
+			currentSeq = txmeta.LedgerSequence()
+			if initialSeq == 0 {
+				initialSeq = currentSeq
+				d.logger.WithFields(supportlog.F{
+					"seq": currentSeq,
+				}).Info("initializing in-memory store")
```
Shaptic, line 328: iirc you can use `WithField` if there's just one and you won't need the `supportlog.F` thing

cmd/soroban-rpc/internal/db/cursor.go
Shaptic: Right, so doesn't that mean `Start` is not inclusive? So the comment is incorrect.

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -78,15 +100,60 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+				return err
+			}
+
+			v0, ok := e.Event.Body.GetV0()
+			if !ok {
+				return errors.New("unknown event version")
+			}
+
+			// Encode the topics
+			maxTopicCount := 4
```
Shaptic, line 135: This is also defined in get_events.go, we should unify

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -78,15 +100,60 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+			topicList := make([]string, maxTopicCount)
+			for index, segment := range v0.Topics {
+				seg, err := xdr.MarshalBase64(segment)
```
Shaptic, line 138: Why marshal topics? We can compare bytes in the SQL query instead to save space

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +164,185 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+// trimEvents removes all Events which fall outside the ledger retention window.
+func (eventHandler *eventHandler) trimEvents(latestLedgerSeq uint32, retentionWindow uint32) error {
+	if latestLedgerSeq+1 <= retentionWindow {
+		return nil
+	}
+	cutoff := latestLedgerSeq + 1 - retentionWindow
+	id := Cursor{Ledger: cutoff}.String()
+
+	_, err := sq.StatementBuilder.
+		RunWith(eventHandler.stmtCache).
```
Shaptic, line 184: Maybe dumb question: why would we run deletion with a cache?

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +164,185 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		rowQ = rowQ.Where(sq.Eq{"event_type": eventTypes})
+	}
+
+	if len(topics) > 0 {
+		var orConditions sq.Or
+		for i, topic := range topics {
+			if topic == nil {
+				continue
+			}
+			orConditions = append(orConditions, sq.Eq{fmt.Sprintf("topic%d", i+1): topic})
```
Shaptic, line 227: How does this work for wildcard topic filters?

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +164,185 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+	}
+	if !foundRows {
+		eventHandler.log.
+			WithField("duration", time.Since(start)).
+			WithField("start", cursorRange.Start.String()).
+			WithField("end", cursorRange.End.String()).
+			WithField("contracts", contractIDs).
+			WithField("eventTypes", eventTypes).
+			WithField("Topics", topics).
+			Debugf(
```
Shaptic, line 289: Is this worth logging? Might be very noisy

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +164,185 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+	return e.writer.InsertEvents(meta)
+}
+
+func newEventTableMigration(
+	logger *log.Entry,
+	retentionWindow uint32,
+	passphrase string,
+) migrationApplierFactory {
+	return migrationApplierFactoryF(func(db *DB, latestLedger uint32) (MigrationApplier, error) {
+		firstLedgerToMigrate := firstLedger
```
Shaptic, line 330: Am I blind or where is `firstLedger` defined?

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
```diff
@@ -1,16 +1,23 @@
...
+    topic1            BLOB,
+    topic2            BLOB,
+    topic3            BLOB,
+    topic4            BLOB
 );
 
-CREATE INDEX idx_ledger_sequence ON events(ledger_sequence);
-CREATE INDEX idx_contract_id ON events(contract_id);
+CREATE INDEX idx_contract_id ON events (contract_id);
+CREATE INDEX idx_topic1 ON events (topic1);
```
Shaptic, line 19: Wait why is this a delta instead of a new file? Shouldn't it be its own migration file?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -10,10 +10,19 @@ import (
...
 	"github.com/stellar/go/xdr"
 
-	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/events"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db"
+)
+
+const (
+	LedgerScanLimit     = 8000
+	maxContractIDsLimit = 5
+	maxTopicsLimit      = 5
```
Shaptic, line 24: 5?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -88,23 +106,24 @@ func (g *GetEventsRequest) Valid(maxLimit uint) error {
 	// Validate start
 	// Validate the paging limit (if it exists)
 	if g.Pagination != nil && g.Pagination.Cursor != nil {
-		if g.StartLedger != 0 {
-			return errors.New("startLedger and cursor cannot both be set")
+		if g.StartLedger != 0 || g.EndLedger != 0 {
+			return errors.New("startLedger/endLedger and cursor cannot both be set") //nolint:forbidigo
```
Shaptic, line 110: just a suggestion: a little neater but same meaning
```suggestion
			return errors.New("ledger ranges and cursor cannot both be set") //nolint:forbidigo
```

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -88,23 +106,24 @@ func (g *GetEventsRequest) Valid(maxLimit uint) error {
...
 
 	// Validate filters
-	if len(g.Filters) > 5 {
+	if len(g.Filters) > maxFiltersLimit {
 		return errors.New("maximum 5 filters per request")
 	}
 	for i, filter := range g.Filters {
 		if err := filter.Valid(); err != nil {
-			return errors.Wrapf(err, "filter %d invalid", i+1)
+			return errors.Wrapf(err, "filter %d invalid", i+1) //nolint:forbidigo
```
Shaptic, line 126: use https://github.com/pkg/errors instead of our support/errors and the linter will :shushing_face:

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: oh for the list of topic filters, not topics in a single topic filter lol

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -159,7 +186,7 @@ func (e *EventFilter) Valid() error {
 	}
 	for i, topic := range e.Topics {
 		if err := topic.Valid(); err != nil {
-			return errors.Wrapf(err, "topic %d invalid", i+1)
+			return fmt.Errorf("topic %d invalid: %w", i+1, err)
```
Shaptic, line 189: use [pkg/errors](https://github.com/pkg/errors) like I mentioned above to keep `errors.Is` compatibility and general Go-ness

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -293,76 +320,172 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+func combineEventTypes(filters []EventFilter) []int {
+	eventTypes := make(map[int]bool)
+	for _, filter := range filters {
+		for _, eventType := range filter.EventType.Keys() {
+			eventTypeXDR := getEventTypeXDRFromEventType()[eventType]
+			eventTypes[int(eventTypeXDR)] = true
+		}
+	}
+	uniqueEventTypes := make([]int, 0, len(eventTypes))
+	for eventType := range eventTypes {
```
Shaptic, line 370: You aren't checking `eventType == true/false` anywhere. Should this be a Set?

Shaptic (commented):
Left a buncha comments, mostly small stuff and a few questions

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: Yeah it is bit confusing but yes it is maximum 5 topics per filter: max 5 topics list --> each list has 4 topic strings.

cmd/soroban-rpc/internal/db/event.go
psheth9: Migration code has been merged so this should be all good now.

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
psheth9: It is delta because schema is already been define in base branch` 'events-db-backend'.` It is going to be a new file when we merge to `main`

cmd/soroban-rpc/internal/db/event.go
psheth9: I actually wanted to log this in order to understand the usage of `getEvents` so that we can tune it later. Even tho getting a log dump from partners going be interesting and something I am planning to look into :)

cmd/soroban-rpc/internal/db/sqlmigrations/03_events.sql
Shaptic: Ah duh!

cmd/soroban-rpc/internal/db/event.go
psheth9: image

Attaching small design diagram to explain it


1. Combine Filters: We ignore the wildcard topic entry when we combine all the filters together. This might be the only way because what we need to tell DB is to fetch all the entries when one of the topic is wildcard

2. Apply filter after fetching events from DB: In this step we will check the actual intersection of topics e.g where topic1 = "some bytes" and topic2 = " wildcard(*) " and topic3 = "some bytes"
note that events that we fetch from DB always does not end up in result set as we need to check for the final intersection of all the filters for given EventXDR.

I went with this approach in order to reduce latency and multiple DB calls. @Shaptic

cmd/soroban-rpc/internal/daemon/daemon.go
psheth9: I think this is done and refactored by migration code PR

cmd/soroban-rpc/internal/db/event.go
Shaptic: Got it! And we're doing the second-stage filtering client side to ensure the topics are OR'd not AND'd? Per the [docs](https://developers.stellar.org/docs/data/rpc/api-reference/methods/getEvents) (had to double-check myself):

> Events matching **any** of the filters are included. To match a filter, an event must match both a contractId and a topic.

psheth9: Yes second-stage filtering on service side. Well client side if you think in the context of DB hahaa 

Code:

```
if request.Matches(event) { --> This is 2nd stage filtering to exactly match the filter
			found = append(found, entry{cursor, ledgerCloseTimestamp, event, txHash})
		}

```

https://github.com/stellar/soroban-rpc/pull/215/files#diff-1f027a1543c912ae4bc1533d9fece665a2b555548e8f74a8e591a70fdc317b2fR475

psheth9: And this is the interesting part so wanted to make it as clear as possible.

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: wait but why do we need that check? if keys are populated that means it is part of our result set right?

cmd/soroban-rpc/internal/daemon/daemon.go
psheth9: Idk I am seeing this pattern everywhere in daemon.go. Also it was part of migration code, it might be showing up here as I have reformatted some of the code. I will keep it as is for now.

cmd/soroban-rpc/internal/db/event.go
psheth9: Good call out !! refactored :)

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: Yeah, I'm saying it doesn't seem like you need to be using a map to bool and can use `collections.Set[int]` instead for readability

cmd/soroban-rpc/internal/daemon/daemon.go
psheth9: This migration code will not run as frequent afaik.  Also this would require changing feeWindow impl and migration framework so I am not planning to do as part of this PR.

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +165,180 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+		cur, err := ParseCursor(id)
+		if err != nil {
+			return fmt.Errorf("failed to parse cursor: %w", err)
+		}
+
+		var eventXDR xdr.DiagnosticEvent
+		err = xdr.SafeUnmarshal(eventData, &eventXDR)
+		if err != nil {
+			return fmt.Errorf("failed to decode event: %w", err)
+		}
```
Shaptic, line 276: You might be able to Scan directly into the xdr.DiagnosticEvent structure? Or is this the thing we talked about where it was being weird for some reason

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -293,76 +315,172 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+		uniqueEventTypes = append(uniqueEventTypes, eventType)
+	}
+	return uniqueEventTypes
+}
+
+func combineTopics(filters []EventFilter) ([][][]byte, error) {
+	encodedTopicsList := make([][][]byte, db.MaxTopicCount)
```
Shaptic, line 372: I think you want `0, db.MaxTopicCount`, right? Otherwise you have empty bytes and then the `append` adds to the end of that? :thinking:

cmd/soroban-rpc/internal/db/event.go
Shaptic: Nice, I didn't see that line! Huge diff :laughing:

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -293,76 +315,172 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+		}
+
+		for _, topicFilter := range filter.Topics {
+			for i, segmentFilter := range topicFilter {
+				if segmentFilter.wildcard == nil && segmentFilter.scval != nil {
+					encodedTopic, err := segmentFilter.scval.MarshalBinary()
+					if err != nil {
+						return [][][]byte{}, fmt.Errorf("failed to marshal segment: %w", err)
+					}
+					encodedTopicsList[i] = append(encodedTopicsList[i], encodedTopic)
```
Shaptic, line 386: Wait how is this an array of an array of an array of bytes? Since you are going to pass all of them to the SQL flattened out, shouldn't it only be a 2D array? As in, a 4-element array of topics `{ {topic1, topic1, ...}, {topic2, ... } }`?

cmd/soroban-rpc/internal/db/event.go
```diff
@@ -97,3 +164,185 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
+			contractIDs,
+			eventTypes,
+			topics,
+			err)
+	}
+
+	defer rows.Close()
+
+	foundRows := false
+	for rows.Next() {
```
Shaptic, line 251: You also need to check `rows.Err()` somewhere, I think? I forget the exact semantics

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: ` { {topic1, topic1, ...}, {topic2, ... } }` --> Array of Array of String --> Array of Array of Array of byte

psheth9: image

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: No no if I do that it will fail at `encodedTopicsList[i] = append(encodedTopicsList[i], encodedTopic)` because it is 0 length but on the other hand I have created 4 nil slices which I can easily access using `encodedTopicsList[i]`

cmd/soroban-rpc/internal/db/event.go
psheth9: hmmm I dont understand the scan part what do you mean? We dont have to have to do anything as we are now directly storing entire eventXDR blob.

cmd/soroban-rpc/internal/db/event.go
psheth9: hahaa yes sorry for the big diff 😆

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: Oh duh :facepalm: thanks lol

cmd/soroban-rpc/internal/db/event.go
Shaptic: Like if you keep the type of eventXDR as an xdr.DiagnosticEvent it will automatically decode into the structure when you call `Scan`

psheth9: oh yes right right, sqlite does not understand it directly. I think I need to implement custom decoder or pass on the existing decoder somewhere so skipped that part for now

cmd/soroban-rpc/internal/db/cursor.go
psheth9: I think this comment was there in the context of DB operation afaik see below query. Also just to add this is existing cursor logic for events in memory too.

our SQL query is like this:

```
Where(sq.GtOrEq{"id": cursorRange.Start.String()}).
Where(sq.Lt{"id": cursorRange.End.String()}).
```

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -135,6 +155,14 @@ var eventTypeFromXDR = map[xdr.ContractEventType]string{
...
 }
 
+func getEventTypeXDRFromEventType() map[string]xdr.ContractEventType {
+	return map[string]xdr.ContractEventType{
+		EventTypeSystem:     xdr.ContractEventTypeSystem,
+		EventTypeContract:   xdr.ContractEventTypeContract,
+		EventTypeDiagnostic: xdr.ContractEventTypeDiagnostic,
+	}
+}
+
```
Shaptic, line 165: Can/Should this just be a `var` at the top of the file? You could then use `len(EventTypeMap)` instead of a hardcoded `3` constant elsewhere.

Shaptic (approved):
LGTM! I'd like @2opremio to have a last look, though, since he was driving most of the reviews. We should definitely add Prometheus metrics to these in a separate PR.

psheth9:
@Shaptic Thanks for review comments and stamp !! 

Yes anyway I have to create final PR for `events-db-backend` to `main`. @2opremio you can take a final look there. Thank you :)

cmd/soroban-rpc/internal/methods/get_events.go
psheth9: well initially it was `var` same as `eventTypeFromXDR` but linter does not like global var :( so it was more of temp fix. it will all be gone when we target all of this lint error together to have consistency in code.

PR #216, opened 2024-06-14
Refactor and parallelize integration tests

2opremio:
### What

Parallelize and refactor integration tests.

### Why

Tests look way cleaner and run 10x faster (at least in my local machine)

Followup to #207 

### Known limitations

The dynamic port allocation makes the code a little less readable


2opremio:
@tamirms the PR is pretty big, but I have created self-container commits which should be reviewable incrementally.

2opremio:
@tamirms It's still a draft until we check how it performs and I can address a couple of TODOs.

2opremio:
It turns out that the tests are around 2 times faster in CI (I will try to play with the parallelism level to see if it improves)

tamirms:
@2opremio you can use this approach to let docker-compose select open ports on the host for the exposed ports on the containers:

https://stackoverflow.com/questions/60109556/expose-random-port-to-docker-compose-yml

there is a docker-compose command you can use to find the port mapping between host and container

2opremio:
Ahh, interesting. I will look into that. That should reduce the clashes

2opremio:
@tamirms I applied your suggestion. The only preallocated port is Captive Core's peer port, because there doesn't seem to be a way to make Core allocate it dynamically.

We can even get rpc to use dynamic ports! :)

2opremio:
Now the problem is I cannot get an ordered output of the failing tests (since tests are running in parallel, golang outputs a mixed output) which makes it hard to see what went wrong.

2opremio:
The logging should be fixed. This blogpost was pretty useful: https://brandur.org/t-parallel

The trick was to use `testing.T.Log()` for everything

2opremio:
Tests are now failing some times because of https://github.com/stellar/go/issues/5350

2opremio:
I fixed the race at https://github.com/stellar/go/pull/5352

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
```diff
@@ -0,0 +1,196 @@
...
+			continue
+		}
+
+		return result
+	}
+	t.Fatal("GetTransaction timed out")
+	return result
+}
+
+func SendSuccessfulTransaction(t *testing.T, client *Client, kp *keypair.Full, transaction *txnbuild.Transaction) methods.GetTransactionResponse {
```
Shaptic, line 73: can/should these be methods of `Client`?

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
```diff
@@ -0,0 +1,196 @@
...
+	assert.GreaterOrEqual(t, response.LatestLedgerCloseTime, response.LedgerCloseTime)
+	return response
+}
+
+func SimulateTransactionFromTxParams(t *testing.T, client *Client, params txnbuild.TransactionParams) methods.SimulateTransactionResponse {
+	savedAutoIncrement := params.IncrementSequenceNum
+	params.IncrementSequenceNum = false
+	tx, err := txnbuild.NewTransaction(params)
+	require.NoError(t, err)
+	params.IncrementSequenceNum = savedAutoIncrement
```
Shaptic, line 135: Isn't this passed by value and so it doesn't need this preservation behavior?

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
```diff
@@ -0,0 +1,196 @@
...
+	request := methods.SimulateTransactionRequest{Transaction: txB64}
+	var response methods.SimulateTransactionResponse
+	err = client.CallResult(context.Background(), "simulateTransaction", request, &response)
+	require.NoError(t, err)
+	return response
+}
+
+func PreflightTransactionParamsLocally(t *testing.T, params txnbuild.TransactionParams, response methods.SimulateTransactionResponse) txnbuild.TransactionParams {
+	if !assert.Empty(t, response.Error) {
+		fmt.Println(response.Error)
```
Shaptic, line 147:
```suggestion
		t.Log(response.Error)
```

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
```diff
@@ -0,0 +1,196 @@
...
+	op := params.Operations[0]
+	switch v := op.(type) {
+	case *txnbuild.InvokeHostFunction:
+		require.Len(t, response.Results, 1)
+		v.Ext = xdr.TransactionExt{
+			V:           1,
+			SorobanData: &transactionData,
+		}
+		var auth []xdr.SorobanAuthorizationEntry
+		for _, b64 := range response.Results[0].Auth {
```
Shaptic, line 162: totally separate but we really ought to just write a Go client for Soroban RPC at this point

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -0,0 +1,717 @@
...
+		captiveCoreStoragePath: "/tmp/captive-core",
+		archiveURL:             fmt.Sprintf("http://%s:%d", inContainerCoreHostname, inContainerCoreArchivePort),
+		sqlitePath:             "/db/" + filepath.Base(i.sqlitePath),
+	}
+}
+
+func (i *Test) getRPConfigForDaemon() rpcConfig {
+	coreBinaryPath := os.Getenv("SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN")
+	if coreBinaryPath == "" {
+		i.t.Fatal("missing SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN")
```
Shaptic, line 282: Can we check the PATH by default? (e.g. `which stellar-core`) It's always mildly annoying to provide this

Shaptic (commented):
This PR is a lot to take in so my review isn't super comprehensive but I do have a main question: can you control "how parallel" the tests are or does it run them all at once? Gonna try to run them locally and I'm worried it won't work on older machines.

Edit: ah, your [linked post](https://brandur.org/t-parallel) explains you can use `-count=N` for it.

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
2opremio: It is an interface which is passed by reference

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
2opremio: Good catch!

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
2opremio: This code predates the refactoring though (although it's understandably hard to see with so many changes)

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
2opremio: We can do it separately from this change (that check predates the PR)

(I never have it in my path though :)

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
2opremio: Yes, I totally agree, I suggested it to @mollykarcher a while ago and I recall she mentioned it wasn't a priority (although refactoring the whole integration test infrastructure probably isn't either).

Now seriously, I think it makes sense to create an internal client at least. Exposing it externally carries a higher maintenance burden.

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
2opremio: Uhm .. not really since it's pretty opinionated. Sending a transaction (fullstop) probably should.

2opremio:
@Shaptic PTAL

Shaptic (approved):

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
Shaptic: Yeah it's hard to tell what moved and what changed but thanks for clarifying!

cmd/soroban-rpc/internal/integrationtest/infrastructure/client.go
Shaptic: Yeah internal makes a lot of sense so that we can make whatever changes we want.

Issue #210, opened 2024-06-11
getHealth's retention window is broken


2opremio:
It returns the transaction retention window, instead of the maximum retention window.

It broke in #174

In particular, this is not true:

https://github.com/stellar/soroban-rpc/blob/c818bb87eaaf705ca4b4c1f4b5ded8cd0cf859ff/cmd/soroban-rpc/internal/jsonrpc.go#L139-L142

The transaction window is still different than the events window, which can be larger.

PR #217, opened 2024-06-14
Fix ledger range calculation

### LINKED ISSUE #208: transactionHandler's GetLedgerRange is not accurate
It uses the meta table and not the transaction table. The meta table can have a longer history than the transaction table since it stores the maximum range of the events storage window and the transaction storage window.

Also, `GetLedgerRange` is untested. A test should be added since the function is pretty complex.

The problem was introduced in https://github.com/stellar/soroban-rpc/pull/174

aditya1702:
### What

Fixes #208 , #210 

- Change `GetLedgerRange` function to query the meta table in an optimised way. This will be used by the rest of the endpoints to get the oldest and latest ledger details.
- Add migration to create index on the `sequence` column in the `ledger_close_meta` table
- Add tests for `GetLedgerRange`

### Why

Resolving bugs that were introduced with the new transactions db changes.

### Known limitations

N/A


cmd/soroban-rpc/internal/test/transaction_test.go
```diff
@@ -136,8 +136,6 @@ func TestSendTransactionBadSequence(t *testing.T) {
 	err = client.CallResult(context.Background(), "sendTransaction", request, &result)
 	assert.NoError(t, err)
 
-	assert.NotZero(t, result.LatestLedger)
```
aditya1702, line 139: I have removed these assertions because in the integration test, the RPC inserts ledgers without any transactions so the txns db is empty and this assertion fails. I removed it for now since we have other checks that test `GetLedgerRange`'s functionality.

Thoughts @2opremio @Shaptic

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -31,6 +32,10 @@ const (
 	latestLedgerSequenceMetaKey = "LatestLedgerSequence"
 )
 
+type LedgerRangeReader interface {
```
2opremio, line 35:
```suggestion
type LedgerRangeGetter interface {
```

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -142,32 +142,45 @@ func (txn *transactionHandler) GetLedgerRange(ctx context.Context) (ledgerbucket
...
+
+	// Final query to join ledger_close_meta table and the sequence numbers we got from txnMinMaxLedgersQ
+	finalSql := sq.
+		Select("lcm.meta").
+		From(fmt.Sprintf("%s as lcm", ledgerCloseMetaTableName)).
+		JoinClause(fmt.Sprintf("JOIN (%s) as seqs ON lcm.sequence == seqs.ledger_sequence", txnMinMaxLedgersQ))
+
 	var lcms []xdr.LedgerCloseMeta
-	if err = txn.db.Select(ctx, &lcms, newestQ.Suffix("UNION ALL "+sql, args...)); err != nil {
+	if err = txn.db.Select(ctx, &lcms, finalSql); err != nil {
```
2opremio, line 183: This seems overly complicated. I think you can use the [min/max functions](https://www.sqlite.org/lang_aggfunc.html) to drastically simplify this.

Also, please benchmark this and make sure it takes a reasonable time.

cmd/soroban-rpc/internal/test/transaction_test.go
2opremio: I don't understand why it used to work and now it doesn't. The code you added shouldn't impact transaction submission (and certainly ingestion).

Also, you are submitting a transaction so the ledger shouldn't be empty, and, even if it were empty the latest ledger shouldn't be zero.

2opremio (changes requested):
Left a few comments with what should be changed.

Also, please update the transaction migration to make use of the correct `GetLedgerRange()` (I left a TODO there).

cmd/soroban-rpc/internal/db/transaction.go
Shaptic: Unfortunately you can't. Well, you can, but you cannot combine `min` and `max` it the same query because we also need the respective `meta` from the ledgers table to get the ledger close time alongside the sequence number. So you have to have separate `ORDER BY LIMIT 1`s. You can also use a subquery. @tamirms and I discussed this here: https://github.com/stellar/soroban-rpc/pull/141#discussion_r1578433440.

Also there are benchmarks for getTransactionByHash which tests this by extension (see https://github.com/stellar/soroban-rpc/pull/141#issuecomment-2088676124).

cmd/soroban-rpc/internal/test/transaction_test.go
Shaptic: I think it's because now we're checking the transaction table for its oldest/latest ledger rather than the meta table, so it'll be empty if you have empty ledgers. Which does make sense but it's really weird behavior.

Shaptic: So `getEvents`, `getLedgers` (when it comes), and `getTransactions` will all return different `oldestLedger` / `latestLedger` ranges.

cmd/soroban-rpc/internal/db/transaction.go
2opremio: I am pretty sure you can use max/min as a subquery and then a join

2opremio: I would like an explicit yes for this, and some numbers in the PR

cmd/soroban-rpc/internal/test/transaction_test.go
aditya1702: @Shaptic Yep thats what is happening here. Earlier, we used to check a different table so it would always have the ledgers in it. As I mentioned, the integration test RPC client ingests ledgers with no txns so the `LedgerRangeGetter` would always return empty.

I will try to resolve this today.

aditya1702:
> Left a few comments with what should be changed.
> 
> Also, please update the transaction migration to make use of the correct `GetLedgerRange()` (I left a TODO there).

@2opremio Is this the TODO you are referring to? - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/db/migration.go#L33

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -135,48 +139,36 @@ func (txn *transactionHandler) trimTransactions(latestLedgerSeq uint32, retentio
...
+	if ledgerSeqs.MinLedgerSequence == nil || ledgerSeqs.MaxLedgerSequence == nil {
+		return ledgerRange, nil
 	}
 
+	// Use the min and max ledger sequences to query the ledger_close_meta table
+	ledgerMetaSQL := sq.
+		Select("lcm.meta").
+		From(ledgerCloseMetaTableName + " as lcm").
+		Where(sq.Eq{"lcm.sequence": []uint32{*ledgerSeqs.MinLedgerSequence, *ledgerSeqs.MaxLedgerSequence}})
+
```
2opremio, line 168: You should be able to do it in a single query to the DB

cmd/soroban-rpc/internal/db/util.go
```diff
@@ -0,0 +1,107 @@
+package db
```
2opremio, line 1: This code only seems to be used in tests, it shouldn't be in the `db` package (since it gets incorporated into production code, causing bloat). Please place it in a `_test.go` file or in a package only used by tests.

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -1,6 +1,8 @@
+//nolint:mnd
```
2opremio, line 1: Why was this removed? We shouldn't be using magic numbers in production code.

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -1,8 +1,10 @@
+//nolint:intrange
```
2opremio, line 1: Why? Shouldn't we fix the code instead?

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -170,6 +174,10 @@ func NewTest(t *testing.T, cfg *TestConfig) *Test {
 		i.waitForRPC()
 	}
 
+	// We populate the transactions table with some rows so that GetLedgerRange called in transaction related
+	// integration tests returns a valid latest ledger.
+	i.insertTransactions()
```
2opremio, line 179: Please don't do this for every test, this is related to transaction-tests only. Also, why can't you submit transactions to populate the table instead of doing it artificially?

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -619,7 +661,11 @@ func (i *Test) UpgradeProtocol(version uint32) {
 
 func (i *Test) StopRPC() {
 	if i.daemon != nil {
-		i.daemon.Close()
+		err := i.daemon.Close()
+		if err != nil {
+			i.t.Logf("could not close RPC daemon: %v", err)
+			return
+		}
```
2opremio, line 668: Just use `require.NoError()` instead

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -29,7 +30,10 @@ import (
 // maxHTTPRequestSize defines the largest request size that the http handler
 // would be willing to accept before dropping the request. The implementation
 // uses the default MaxBytesHandler to limit the request size.
-const maxHTTPRequestSize = 512 * 1024 // half a megabyte
+const (
+	maxHTTPRequestSize          = 512 * 1024 // half a megabyte
```
2opremio, line 34: Please move the comment down to the definition

cmd/soroban-rpc/internal/db/transaction_test.go
```diff
@@ -19,12 +18,21 @@ import (
...
-	require.Error(t, err, ErrNoTransaction)
+	reader := NewTransactionReader(log.DefaultLogger, db, passphrase)
+
+	// Assert the ledger range
+	ledgerRange, err := reader.GetLedgerRange(ctx)
+	require.NoError(t, err)
+	assert.Equal(t, uint32(0), ledgerRange.FirstLedger.Sequence)
+	assert.Equal(t, int64(0), ledgerRange.FirstLedger.CloseTime)
+	assert.Equal(t, uint32(0), ledgerRange.LastLedger.Sequence)
+	assert.Equal(t, int64(0), ledgerRange.LastLedger.CloseTime)
```
2opremio, line 32: Please use dedicated tests for GetLedgerRange instead

cmd/soroban-rpc/internal/db/transaction_test.go
```diff
@@ -51,18 +59,26 @@ func TestTransactionFound(t *testing.T) {
...
 
-	// check 404 case
+	// Assert the ledger range
 	reader := NewTransactionReader(log, db, passphrase)
+	ledgerRange, err := reader.GetLedgerRange(ctx)
+	require.NoError(t, err)
+	assert.Equal(t, uint32(1234), ledgerRange.FirstLedger.Sequence)
+	assert.Equal(t, LedgerCloseTime(1234), ledgerRange.FirstLedger.CloseTime)
+	assert.Equal(t, uint32(1237), ledgerRange.LastLedger.Sequence)
+	assert.Equal(t, LedgerCloseTime(1237), ledgerRange.LastLedger.CloseTime)
```
2opremio, line 69: Please use dedicated tests for GetLedgerRange instead

2opremio:
e2e tests are failing

2opremio:
> Benchmarking for `GetLedgerRange`
> 
> ```
> goos: darwin
> goarch: arm64
> pkg: github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db
> BenchmarkGetLedgerRange
> BenchmarkGetLedgerRange-12    	     373	   3225746 ns/op
> ```

So, 3ms per operation. Not terrible, but it should improve when you merge everything into a single query

2opremio (changes requested):
It's moving in the right direction but I left a few comments on things to change.

cmd/soroban-rpc/internal/db/transaction.go
aditya1702: Yeah I did not realise combining everything in single would be so efficient.

cmd/soroban-rpc/internal/events/events.go
aditya1702: I removed this because of this line:

```
Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001}
```

Do I then need to define constants for each key-value pair here?

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
aditya1702: Well, because I added this in the first place was because the integration tests for sending transaction were failing here - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/integrationtest/transaction_test.go#L179. I thought it would defeat the purpose of testing the send functionality if we use it to populate the table artificially. What do you think?

However, I agree that we should not run these on each integration test and I will add a config to do it only for txn related tests.

cmd/soroban-rpc/internal/events/events.go
aditya1702: Anyways, I have fixed it now. I agree though - in general its good to not have any magic numbers defined anywhere.

aditya1702:
New Benchmark for `GetLedgerRange`:

```
goos: darwin
goarch: arm64
BenchmarkGetLedgerRange
BenchmarkGetLedgerRange-12    	   54640	     21908 ns/op
```

cmd/soroban-rpc/internal/db/util.go
aditya1702: I wanted to remove duplicated code since txMeta function is used in multiple places. However, I cant move it to infrastructure file since it creates an import cycle with the db package. I have reverted the util.go file creation and added the txMeta implementation to the infrastructure util.go file for now.

2opremio: Sounds good

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
2opremio: If that code is only used in getTransaction tests, then we should try to only incorporate it there.

I am not sure why existing tests were failing. It worries me that we need to artificially need to populate the db to fix them.

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -1,3 +1,4 @@
+//nolint:intrange
```
2opremio, line 1: Why is there an exception here?

cmd/soroban-rpc/internal/integrationtest/infrastructure/test.go
aditya1702: @2opremio No its not just used in getTransactions tests. There were around 11 integration tests that were failing due to not populating the the transactions db. So this piece of code is tied to multiple tests.

2opremio: > There were around 11 integration tests that were failing due to not populating the the transactions db

This is the part I don't understand. The normal operation of tests should populate the transactions table.

2opremio: i.e. if you want the transactions table to be populated, then submit transactions. If you need to artificially fill the table for some tests, then those tests should do it and the general `test.go` file shouldn't be poluted by it.

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -135,43 +139,23 @@ func (txn *transactionHandler) trimTransactions(latestLedgerSeq uint32, retentio
 	return err
 }
 
-// GetLedgerRange pulls the min/max ledger sequence numbers from the database.
+// GetLedgerRange pulls the min/max ledger sequence numbers from the transactions table.
 func (txn *transactionHandler) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
```
psheth9, line 143: @aditya1702 if we are refactoring `GetLedgerRange` to fetch entries from LCM table, we def want to move this to LedgerReader.  and then we want to pass `LedgerReader` in related RPCs instead of `TxnReader`

aditya1702: @psheth9 I am still refactoring things 😅 . Doing it in small small commits

aditya1702:
Benchmarking for the `GetLedgerRange` from meta `table`

```
goos: darwin
goarch: arm64
BenchmarkGetLedgerRange
BenchmarkGetLedgerRange-12    	   57433	     20941 ns/op
```

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -65,10 +68,39 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
 }
 
+// GetLedgerRange pulls the min/max ledger sequence numbers from the meta table.
+func (r ledgerReader) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
+	var ledgerRange ledgerbucketwindow.LedgerRange
+
+	query := sq.Select("lcm.meta").
+		From(ledgerCloseMetaTableName + " as lcm").
+		Where(sq.Expr("lcm.sequence = (SELECT MIN(sequence) FROM " + ledgerCloseMetaTableName + ") " +
+			"OR lcm.sequence = (SELECT MAX(sequence) FROM " + ledgerCloseMetaTableName + ")"))
```
2opremio, line 83: Is this faster than using `IN()`?

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -65,10 +68,39 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
+
+	var lcms []xdr.LedgerCloseMeta
+	if err := r.db.Select(ctx, &lcms, query); err != nil {
+		return ledgerRange, fmt.Errorf("couldn't query ledger range: %w", err)
+	} else if len(lcms) < getLedgerRangeMetaArrayCheckValue {
```
tamirms, line 88: what if there's only one ledger in the db? In that case the min and the max will be the same. I don't think this code handles that scenario. Could you please add a test for this scenario?

cmd/soroban-rpc/internal/db/ledger.go
2opremio: This seems to work:


```sql
select ledger_close_meta.* from ledger_close_meta, (select min(sequence) as min, max(sequence) as max from ledger_close_meta) minmax where ledger_close_meta.sequence in (minmax.min, minmax.max) ;
 ```

Comment (2opremio) [reply cmd/soroban-rpc/internal/db/ledger.go:83]: BTW, this is not a blocked, but I would encourage you to try it out.

Comment (2opremio) [reply cmd/soroban-rpc/internal/db/ledger.go:83]: Regardless, I think you can replace the Selects in `sq.Expr()` by actual `sq.Select()`

Comment (2opremio) [reply cmd/soroban-rpc/internal/db/ledger.go:83]: Also, you can use `sq.Or` instead of a string.

Comment (2opremio) [reply cmd/soroban-rpc/internal/db/ledger.go:83]: i.e. try to use `sq` as much as possible instead of text.

File: cmd/soroban-rpc/internal/methods/send_transaction.go [hunk 17]
```diff
@@ -73,7 +73,7 @@ func NewSendTransactionHandler(
 		}
 		txHash := hex.EncodeToString(hash[:])
 
-		ledgerInfo, err := reader.GetLedgerRange(ctx)
+		ledgerInfo, err := ledgerReader.GetLedgerRange(ctx)
```
2opremio, line 76: I know this is not directly related to this PR, but here we can get the latestledger instead (which is currently cheaper). Since we only need the latest ledger.

Optionally, you can fix it later, as part of https://github.com/stellar/soroban-rpc/issues/233

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -18,7 +18,11 @@ import (
 	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/ledgerbucketwindow"
 )
 
-const transactionTableName = "transactions"
+const (
+	transactionTableName              = "transactions"
+	firstLedgerToMigrate              = 2
+	getLedgerRangeMetaArrayCheckValue = 2
```
2opremio, line 24: I have seen the use of this variable and it makes the code less readable, please use a nolint instead.

Also, try to define variables in the same file as you use them.

2opremio (changes requested):
Almost there! Just a few minor things to go!

2opremio:
BTW, what's the benchmark result?

aditya1702:
> BTW, what's the benchmark result?

@2opremio I pasted the benchmarking here - https://github.com/stellar/soroban-rpc/pull/217#issuecomment-2195169517

cmd/soroban-rpc/internal/methods/send_transaction.go
aditya1702: @2opremio Yeah I agree we dont need the range here. I will fix it as part of #233

cmd/soroban-rpc/internal/db/transaction.go
aditya1702: Ah the `getLedgerRangeMetaArrayCheckValue ` was originally being used in the transaction.go file and I forgot to remove it.

aditya1702:
Benchmarking after changing the query to use sq.Select instead of raw expression:

```
goos: darwin
goarch: arm64
BenchmarkGetLedgerRange
BenchmarkGetLedgerRange-12    	   50593	     23656 ns/op
```

cmd/soroban-rpc/internal/db/ledger.go
2opremio: Did you get to try using `IN` instead?

2opremio: I suspect it may be more performant since it uses a single subquery.

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -65,10 +67,42 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
+
+	return ledgerbucketwindow.LedgerRange{
+		FirstLedger: ledgerbucketwindow.LedgerInfo{
+			Sequence:  lcms[0].LedgerSequence(),
+			CloseTime: lcms[0].LedgerCloseTime(),
+		},
+		LastLedger: ledgerbucketwindow.LedgerInfo{
+			Sequence:  lcms[len(lcms)-1].LedgerSequence(),
+			CloseTime: lcms[len(lcms)-1].LedgerCloseTime(),
+		},
```
tamirms, line 102: how do we know that the rows will be returned in sorted ledger sequence order?

cmd/soroban-rpc/internal/db/transaction.go
```diff
@@ -339,9 +282,6 @@ func newTransactionTableMigration(ctx context.Context, logger *log.Entry, retent
 		}
 		// Truncate the table, since it may contain data, causing insert conflicts later on.
 		// (the migration was shipped after the actual transactions table change)
-		// FIXME: this can be simply replaced by an upper limit in the ledgers to migrate
```
tamirms, line 342: has this FIXME been addressed?

cmd/soroban-rpc/internal/db/sqlmigrations/03_ledgers_index.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+
+-- creating an index in the ledgers table on the sequence number.
+CREATE INDEX index_lcm_sequence ON ledger_close_meta(sequence);
```
tamirms, line 4: isn't the `sequence` column a primary key for the `ledger_close_meta` table? If so, I'm don't think there's any benefit in having this index, see https://stackoverflow.com/questions/3379292/is-an-index-needed-for-a-primary-key-in-sqlite

2opremio: Really good point! It should be easy to figure out by running the benchmark with and without the index

cmd/soroban-rpc/internal/db/transaction.go
aditya1702: @2opremio and I decided to remove this FIXME. I will remove it in this PR.

cmd/soroban-rpc/internal/db/sqlmigrations/03_ledgers_index.sql
aditya1702: Yeah will run it with and without to check

aditya1702: Its the same without the index too.

Shaptic (commented):
I know this has gone through many many iterations but it seems we've finally landed on one that leads to quite a simple and clean implementation! LGTM from my perspective besides some tiny comments, but I'll wait for the other two to put a formal :heavy_check_mark: on this one :clap:

aditya1702:
@Shaptic Yeah I was also thinking of the same idea and added the ORDER BY ASC clause at the end. This just increases the query time by 1 microsecond (which is not a lot tbh).

cc @2opremio

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: @tamirms I added the order by ASC clause at the end to make sure its order is fixed.

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: @2opremio I think `IN` will complicate the query. squirrel does not have a built in `IN` function, and then we will have to add `UNION ALL` to combine both min and max queries for that.

2opremio: You can use IN in squirrel:

```go
stooges := users.Where(sq.Eq{"username": []string{"moe", "larry", "curly", "shemp"}})
three_stooges := stooges.Limit(3)
rows, err := three_stooges.RunWith(db).Query()

// Behaves like:
rows, err := db.Query("SELECT * FROM users WHERE username IN (?,?,?,?) LIMIT 3",
                      "moe", "larry", "curly", "shemp")
```

2opremio: Also, the query I suggested doesn't use `UNION ALL`

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -65,10 +67,42 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
 	}
 }
 
+// GetLedgerRange pulls the min/max ledger sequence numbers from the meta table.
+func (r ledgerReader) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
+	query := sq.Select("lcm.meta").
+		From(ledgerCloseMetaTableName + " as lcm").
+		Where(sq.Or{
+			sq.Expr("lcm.sequence = (?)", sq.Select("MIN(sequence)").From(ledgerCloseMetaTableName)),
+			sq.Expr("lcm.sequence = (?)", sq.Select("MAX(sequence)").From(ledgerCloseMetaTableName)),
```
2opremio, line 81: I am not totally sure, but I think you can use `sq.Eq{}` instead of  `"lcm.sequence = (?)"`

2opremio (approved):
Good job! I left a few extra comments, I am approving since I don't want to block the PR anymore, but please try the suggestions I added before merging.

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: Ahhh did not consider `sq.Eq`. That should work. I will try it out.

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: @2opremio It wont work because sq.Eq expects actual values and cannot infer SQL functions like MIN and MAX.

PR #218, opened 2024-06-17
Remove rpc-client crate

2opremio:
The crate has moved to another repo: https://github.com/stellar/rs-stellar-rpc-client


leighmcculloch (approved):
Does the rust preflight lib need any of the rust.yml CI?

If no, looks good to me! 🎉 

Thanks for following up the separation of the crate into the new repo. ❤️

2opremio:
Uhm, we already build it with RPC in all archs. So maybe I will just add a clippy test.

2opremio:
The PR is stuck due to the lack of `complete` job (marked as a required check in terraform).

@leighmcculloch Did you manage to find a workaround for https://github.com/orgs/community/discussions/12377 accross multiple workflows?

I think that you used `complete` for that but, as I understand it, it only works within a single workflow? I would like something similar across multiple workflows.

2opremio:
OK, I fixed it at d91b510514e369a0746eeb9c3fe70a185d9e2e2e

leighmcculloch:
If you add a complete job to multiple workflows, as long as the workflows all start before one finishes, then the PR should be blocked from merging until all compelte jobs pass. At least htat's how it used to work It hink.

2opremio:
I added a `complete` job in a new workflow (`required_status_check.yml`) which waits for the required workflows. 

It works fine, but ... the problem is that if something fails and is re-run manually then you also need to re-run `required_status_check`, which is a pain.

I may end up adding a `complete` job to every required workflow, but it's a pain ...

PR #220, opened 2024-06-17
Enforce Go recompilation when libpreflight changes

2opremio:
Before I had to resort to tricks like adding white lines to Go source.

Related: https://github.com/golang/go/issues/24355


2opremio:
This doesn't work, because you don't seem to be able to invalidate the cache of a single package

PR #222, opened 2024-06-18
use CORE_IMAGE_BIN_PATH on e2e for core image, skip compile

sreuland:
### What

use CORE_IMAGE_BIN_PATH with CORE_IMAGE on e2e

### Why

 for core image, skip compile, to build e2e faster


2opremio:
Nice!! Thanks!

It still takes 22 mins thought!

2opremio (approved):

sreuland:
> Nice!! Thanks!
> 
> It still takes 22 mins thought!

yes cli `make install_rust` from source and the rpc `make build-soroban-rpc` targets take up the majority of that remaining build time, system-test can optionally do `cargo install soroban-cli <ver>` , but I've noticed that takes almost as much time as build from repo/source since the the crate install builds rust locally anyways.

PR #234, opened 2024-06-27
Unify transaction and event retention windows

### LINKED ISSUE #199: Unify events and transactions retention window
Also, set a default retention of 7 days.

We should not do this before removing the in-memory data-structures ( #115 , otherwise the memory consumption of soroban-rpc will explode).

2opremio:
### What

Use a new, `unified history-retention-window` (with a default value of 1 day until  #115 is fixed) flag and mark the event/transaction window configuration parameters as deprecated.

Until we remove the event/transaction flag, `history-retention-window` will be initialized to the maximum of the three.

Also, initialize `cfg.HistoryArchiveUserAgent` in a more suitable place.

### Why

Simplifies the configuration. Closes #199

### Known limitations
N/A

### DIFF
--- a/cmd/soroban-rpc/internal/config/main.go
+++ b/cmd/soroban-rpc/internal/config/main.go
@@ -41,6 +41,7 @@ type Config struct {
 	PreflightWorkerQueueSize                       uint
 	PreflightEnableDebug                           bool
 	SQLiteDBPath                                   string
+	HistoryRetentionWindow                         uint32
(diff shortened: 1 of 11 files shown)


tamirms (approved):

aditya1702 (approved):

2opremio:
BTW, this closes https://github.com/stellar/soroban-rpc/issues/208 and https://github.com/stellar/soroban-rpc/issues/210

PR #237, opened 2024-07-04
Refine logging of migrations and tx ingestion

2opremio:
### What

Downgrade Tx ingestion logging to Debug (otherwise we flood the logs during the migration)

Make sure we log whenever a migration is applied

### Why

Make logs more intuitive and not excessively verbose

### Known limitations
N/A


mollykarcher (approved):

2opremio:
Integration tests work locally but fail immediately in CI without any logs. My bet is that they are getting killed due to running in parallel.

I will address that in a separate PR.

Issue #211, opened 2024-06-11
refactor getTransactionsByLedgerSequence


2opremio:
It's huge (140 lines), it should be broken down into smaller pieces to make it readable.

PR #244, opened 2024-07-09
Break `getTransactionsByLedgerSequence` into smaller sub-functions

aditya1702:
### What

#211 

### Why

Function was very long. Shortened it for readability.

### Known limitations

NA

### DIFF
--- a/cmd/soroban-rpc/internal/methods/get_transactions.go
+++ b/cmd/soroban-rpc/internal/methods/get_transactions.go
@@ -130,7 +130,7 @@ func (h transactionsRPCHandler) fetchLedgerData(ctx context.Context, ledgerSeq u
 	} else if !found {
 		return ledger, &jrpc2.Error{
 			Code:    jrpc2.InvalidParams,
-			Message: fmt.Sprintf("ledger close meta not found: %d", ledgerSeq),
+			Message: fmt.Sprintf("database does not contain metadata for ledger %d", ledgerSeq),
 		}
 	}
 	return ledger, nil
--- a/cmd/soroban-rpc/internal/methods/get_transactions.go
+++ b/cmd/soroban-rpc/internal/methods/get_transactions.go
@@ -238,7 +238,7 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 	// Iterate through each ledger and its transactions until limit or end range is reached.
 	// The latest ledger acts as the end ledger range for the request.
 	var txns []TransactionInfo
-	done := false
+	var done bool
 	cursor := toid.New(0, 0, 0)
 	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
 		ledger, err := h.fetchLedgerData(ctx, uint32(ledgerSeq))
--- a/cmd/soroban-rpc/internal/methods/get_transactions.go
+++ b/cmd/soroban-rpc/internal/methods/get_transactions.go
@@ -237,7 +237,6 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 	// The latest ledger acts as the end ledger range for the request.
 	var txns []TransactionInfo
 	cursor := toid.New(0, 0, 0)
-LedgerLoop:
 	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
 		ledger, err := h.fetchLedgerData(ctx, uint32(ledgerSeq))
 		if err != nil {
@@ -250,7 +249,7 @@ LedgerLoop:
 		}
 
 		if len(txns) >= int(limit) {
-			break LedgerLoop
+			break
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,141 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
...
+			return GetTransactionsResponse{}, err
+		}
+
+		cursor, err = h.processTransactionsInLedger(ledger, start, &txns, limit)
+		if err != nil {
+			return GetTransactionsResponse{}, err
+		}
+
+		if len(txns) >= int(limit) {
+			break LedgerLoop
```
2opremio, line 253: You can remove the tag, doing break by itself will lead to the same result.

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,141 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
...
 	cursor := toid.New(0, 0, 0)
-LedgerLoop:
-	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
-		// Get ledger close meta from db
-		ledger, found, err := h.ledgerReader.GetLedger(ctx, uint32(ledgerSeq))
-		if err != nil {
-			return GetTransactionsResponse{}, &jrpc2.Error{
+	reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(h.networkPassphrase, ledger)
+	if err != nil {
+		return cursor, &jrpc2.Error{
```
Shaptic, line 146: Do we not want to return the passed in cursor here, i.e. `*toid`? (after a `nil` check of course)

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,141 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
...
 				Code:    jrpc2.InternalError,
-				Message: err.Error(),
+				Message: ierr.Error(),
+			}
+		}
+	}
+
+	txCount := ledger.CountTransactions()
+	for i := startTxIdx; i <= txCount; i++ {
+		cursor = toid.New(int32(ledger.LedgerSequence()), int32(i), 1)
```
Shaptic, line 166: Shouldn't we just set the field that's changing, i.e. transaction index?

Shaptic (commented):
Just a coupla comments but otherwise LGTM!

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: Ah nice catch. Forgot to remove this.

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @Shaptic The ledger sequence will also change once all tranactions in a ledger are read.

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: @Shaptic Actually we can use return nil, err instead of the start cursor. What do you think?

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: Good idea. Instead of creating the new object each time, just reassign the tx number

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,141 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
...
+			Code:    jrpc2.InternalError,
+			Message: err.Error(),
+		}
+	}
+
+	startTxIdx := 1
+	ledgerSeq := ledger.LedgerSequence()
+	if int32(ledgerSeq) == start.LedgerSequence {
+		startTxIdx = int(start.TransactionOrder)
+		if ierr := reader.Seek(startTxIdx - 1); ierr != nil && !errors.Is(ierr, io.EOF) {
```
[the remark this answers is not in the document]
aditya1702: @Shaptic That is not necessary since this line will handle that:

```
for i := startTxIdx; i <= txCount; i++
```
The startTxIdx will be greater than txCount and it will exit automatically

2opremio (approved):

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,142 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 			limit = request.Pagination.Limit
 		}
 	}
+	return start, limit, nil
```
Shaptic, line 119: Maybe a nitpick but it doesn't seem like there's a reason for it to return a pointer since it's always non-nil

cmd/soroban-rpc/internal/methods/get_transactions.go
```diff
@@ -134,92 +116,142 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
...
-				break LedgerLoop
-			}
+	start, limit, err := h.initializePagination(request)
+	if err != nil {
+		return GetTransactionsResponse{}, err
+	}
+
+	// Iterate through each ledger and its transactions until limit or end range is reached.
+	// The latest ledger acts as the end ledger range for the request.
+	var txns []TransactionInfo
```
Shaptic, line 240: thoughts on a capacity pre-alloc here?

Shaptic (approved):
LGTM! Just a couple minor things to optionally consider :+1:

cmd/soroban-rpc/internal/methods/get_transactions.go
aditya1702: Good idea. Can just use the max limit to allocate here.

Issue #233, opened 2024-06-27
db: Remove latestledger key from meta table


2opremio:
The meta table contains the `LatestLedgerSequence` key, which can simply be obtained from the ledger table (`GetLedgerRange`)

This will also require updating the db API and, while we are at it, cache the full range, not just the latest ledger
```[tasklist]
### Tasks
```

2opremio:
We actually cannot remove the table itself, because now we are using it to store the migration state. But ... we should remove the latestledger key.

PR #246, opened 2024-07-10
[WIP] Replace `GetLatestLedgerSequence` calls with `GetLedgerRange`

aditya1702:
### What

#233 

### Why

### Known limitations

[TODO or N/A]


2opremio:
I don't think there is a need to replace the calls. I think the `GetLatestLedger` functions can (and should) exist. It's just the internal implementation which should change.

PR #249, opened 2024-07-15
Add framework for converting XDR structures to JSON

### LINKED ISSUE #124: JSON Support
Developers have the option to request data in JSON format. By default, responses are in the current XDR+JSON format unless specified otherwise in the request. If JSON format is requested, the system will dynamically convert the XDR data to JSON for easier readability and use by developers.

In order to:
- To request JSON format, include `'xdrFormat’: 'json'` in the request.
- To request the current XDR+JSON mixed response (how the API currently returns the response), simply don’t specify the `xdrFormat` in the request, or include `'xdrFormat': 'base64'` in the request. The response will return the current JSON and XDR in base64 string. 

The initial implementation is then backward compatible.

Example Usage for JSON response:

```
curl -X POST \
-H 'Content-Type: application/json' \
-H 'Accept': application/json \
-d '{
 "jsonrpc": "2.0",
 "id": 8675309,
 "method": "simulateTransaction",
 "params": {
   "xdrFormat": "json",
   "transaction": <in JSON>,
   "resourceConfig": {
     "instructionLeeway": 3000000
   }
 }
}' \
https://soroban-testnet.stellar.org:443 | jq
```

We have developed a canonical XDR ⇔ JSON schema for conversion between XDR and JSON: [Stellar JSON](https://docs.google.com/document/d/1fHQxS5ZEt7nwBVZg4qLHpKzKsUXxoF5Et6cFFYqrNMM/edit)

See https://docs.google.com/document/d/1tTuzL7HBHedp9TE-LonD1QV67SaBSReQ8YFj9xQIgtA/edit#heading=h.uedyb8nrk3a9

Shaptic:
### What
This PR adds support for bsae64-encoded XDR fields to instead be returned as unpacked JSON objects.

#### New Request Parameter
We introduce a new, optional parameter to all endpoints with XDR outputs:
```typescript
xdrFormat?: "base64" | "json"
```
When omitted, the behavior does not change and we encode fields as base64.

#### New Response Fields
There are new field names for the JSONified versions of XDR structures. Any field with an `Xdr` suffix (e.g., `resultXdr` in [`getTransaction()`](https://developers.stellar.org/docs/data/rpc/api-reference/methods/getTransaction)) will be replaced with one that has a `Json` suffix (e.g., `resultJson`) that is a JSON object verbosely and completely describing the XDR structure.

Certain XDR-encoded fields do not have an `Xdr` suffix, but those also have a `*Json` equivalent and are listed below:

* _getEvents_: `topic` -> `topicJson`, `value` -> `valueJson`
* _getLedgerEntries_: `key` -> `keyJson`, `xdr` -> `dataJson`
* _getLedgerEntry_: `xdr` -> `entryJson`
* _simulateTransaction_: `transactionData`, `events`, `results.auth`, `restorePreamble.transactionData`, `stateChanges.key|before|after` all have a `Json` suffix, and `results.xdr` is now `results.returnValueJson`

### Notes for reviewers
* It is much easier to see the changes when you [turn whitespace off](https://github.com/stellar/soroban-rpc/pull/249/files?diff=unified&w=1).
* Start with the bindings layer: 
  - `lib.rs` has the Rust code to perform the XDR -> JSON conversion via `stellar_xdr`.
  - `xdr_json.h` just has the header definitions to define what is accessible from Go -> Rust.
  - `conversion.go` has the actual binding invocation:
    * `convertAnyBytes` is the one that does the low-level Go -> C/FFI -> Rust work
    * the exported members let other packages use it and one is for validation
    * `TransactionToJSON` is an isolated way to see how the conversion is done for a transaction structure
    * the others are just helper abstractions
* Then, review one of the endpoints, e.g., `get_transaction.go`. Each one of these has three sets of changes:
  - Validation for the new request parameter
  - New fields with a `Json` (serialized) and `JSON` (Golang) suffix with a 1:1 mapping to their corresponding base64-encoded fields that should always be a `json.RawMessage`.
  - A `switch` statement in the handler that uses the new `xdr2json` package (i.e. `conversion.go`) to convert XDR structures to JSON.
* The same changes should be in every endpoint.
* Most test changes have a lot of variable renames (**not** serialized fields), so feel free to ignore those entirely.
* There are also new tests for the JSON behavior.

### Why
Closes #124.

#### Remaining Items
- [x] Complete the initial implementation
- [x] Add support for all endpoints:
  * [x] `getTransaction`
  * [x] `getTransactions`
  * [x] `getLedgerEntry`
  * [x] `getLedgerEntries`
  * [x] `getEvents`
  * [x] `sendTransaction`
  * [x] `simulateTransaction`
- [x] Testing for all of the above
- [x] Polish

### Alternative Implementations
Would it be possible to extend `xdrgen` to actually have `.JSON()`-like methods attached? This would be a significant lift there, since it'd need to do the Go <-> Rust FFI, but the calling code would be far cleaner **and** we'd get JSON output "for free" across any Golang project that used `stellar/go/xdr`...


cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -664,3 +668,32 @@ fn extract_error_string<T>(simulation_result: &Result<T>, go_storage: &GoLedgerS
...
+    let type_str = from_c_string(typename);
+    let the_type = match xdr::TypeVariant::from_str(&type_str) {
+        Ok(t) => t,
+        Err(e) => {
+            return string_to_c(format!(r#"{{ "error": "{e}", "type": "{type_str}" }}"#));
+        }
+    };
+
+    let xdr_bytearray = from_c_xdr(xdr);
+    let mut buffer = xdr::Limited::new(xdr_bytearray.as_slice(), DEFAULT_XDR_RW_LIMITS.clone());
```
Shaptic, line 690: Self-review: given that we know the size of the structure here (passed as `xdr.len`), would it make sense to make that the maximum limit read length for this and just use `DEFAULT_XDR_RW_LIMITS` for depth? cc @leighmcculloch or maybe @sisuresh? I'm not sure of the security/resource implications here.

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -664,3 +668,32 @@ fn extract_error_string<T>(simulation_result: &Result<T>, go_storage: &GoLedgerS
...
+// xdr_to_json takes in a string name of an XDR type in the Stellar Protocol
+// (i.e. from the stellar_xdr crate) as well as a raw byte structure and returns a
+// JSONified string of said structure.
+#[no_mangle]
+pub extern "C" fn xdr_to_json(typename: *mut libc::c_char, xdr: CXDR) -> *mut libc::c_char {
+    let type_str = from_c_string(typename);
+    let the_type = match xdr::TypeVariant::from_str(&type_str) {
+        Ok(t) => t,
+        Err(e) => {
+            return string_to_c(format!(r#"{{ "error": "{e}", "type": "{type_str}" }}"#));
```
Shaptic, line 685: Is this preferable or should I do a similar thing to preflight where we catch and unwind panics? I feel like this is pretty clean but cc @2opremio for opinions

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,166 @@
...
+	map[string]interface{},
+	map[string]interface{},
+	[]map[string]interface{},
+	error,
+) {
+	var err error
+	var result, envelope, resultMeta map[string]interface{}
+	var diagEvents []map[string]interface{}
+
+	result, err = Convert(xdr.TransactionResult{}, tx.Result)
```
Shaptic, line 131: Reviewer note: this provides type safety over passing e.g., `"TransactionResult"` which is prone to typos

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -66,22 +67,29 @@ func (e eventTypeSet) matches(event xdr.ContractEvent) bool {
...
+	InSuccessfulContractCall bool   `json:"inSuccessfulContractCall"`
+	TransactionHash          string `json:"txHash"`
+
+	// TopicXDR is a base64-encoded list of ScVals
+	TopicXDR  []string                 `json:"topic,omitempty"`
+	TopicJSON []map[string]interface{} `json:"topicJson,omitempty"`
+
+	// ValueXDR is a base64-encoded ScVal
+	ValueXDR  string                 `json:"value,omitempty"`
+	ValueJSON map[string]interface{} `json:"valueJson,omitempty"`
```
leighmcculloch, line 85: 💡 Given that the JSON is encoded by the Rust, I assume the presence of a map here means that the Go is first decoding the Rust JSON then re-encoding that here.

I suggest you replace this map with json.RawMessage ([godoc](https://pkg.go.dev/encoding/json#RawMessage)) which allows you to inject the precomputed JSON from the Rust directly in without having to do the encode-decode-encode dance.

Shaptic: Yes, you're right that the Rust code was returning a JSON string and that the Go code was giving that structure via `json.Unmarshal`. While it's a brilliant idea to just inject it directly, I'm worried the JSON serialization of the response will just treat it like a string? I'll have to try it out. Also, it unfortunately prevents any inspection of the structure (I am checking for an `error` field as a way to transmit errors across the FFI).

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -664,3 +668,36 @@ fn extract_error_string<T>(simulation_result: &Result<T>, go_storage: &GoLedgerS
...
+
+// xdr_to_json takes in a string name of an XDR type in the Stellar Protocol
+// (i.e. from the stellar_xdr crate) as well as a raw byte structure and returns a
+// JSONified string of said structure.
+//
+// On error, it will return a string with a JSON structure and an "error" key containing
+// the error message from Rust as well as a "type" key for the XDR type that it
+// attempted to decode from the given buffer (this should match `typename`).
+#[no_mangle]
+pub extern "C" fn xdr_to_json(typename: *mut libc::c_char, xdr: CXDR) -> *mut libc::c_char {
```
sreuland, line 680: question on organizational aspect, the xdr-json conversion isn't specific to preflight lib correct? is it worthwhile for conversion to be in a separate `lib/conversion`?

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,166 @@
...
+	goStr := convertStr(xdrTypeName, field)
+	return jsonify(goStr)
+}
+
+// ConvertAny takes a valid XDR object (`xdr`) and returns a
+// JSON-formatted serialization of that object.
+//
+// Unlike `Convert`, the value here needs to be valid and
+// serializable.
+func ConvertAny(xdr interface{}) (map[string]interface{}, error) {
```
sreuland, line 60: `ConvertAny` is ambiguous, it kinda overlaps semantics of `Convert` also, for consideration, how about rename:
`Convert` -> `ConvertBinary`
`ConvertAny` -> `ConvertInstance`

cmd/soroban-rpc/lib/preflight/src/lib.rs
Shaptic: It's not specific to preflight, but it does depend on some shared functionality like the `CXDR` structure and misc. memory management. In an ideal world we'd probably want to split that out into a shared lib but that seems a little preemptive for just a single function, since there'd be a bunch of busy work to make sure compiling/linking all works together across three separate Rust libs.

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: Nice, yeah! Didn't give the naming too much thought so thanks for the "fresh eyes" perspective :+1:

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -664,3 +668,36 @@ fn extract_error_string<T>(simulation_result: &Result<T>, go_storage: &GoLedgerS
...
+        }
+    };
+
+    let xdr_bytearray = from_c_xdr(xdr);
+    let mut buffer = xdr::Limited::new(xdr_bytearray.as_slice(), DEFAULT_XDR_RW_LIMITS.clone());
+
+    let t = match xdr::Type::read_xdr_to_end(the_type, &mut buffer) {
+        Ok(t) => t,
+        Err(e) => {
+            return string_to_c(format!(r#"{{ "error": "{e}", "type": "{type_str}" }}"#));
```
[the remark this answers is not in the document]
Shaptic: Nice idea! Done in 0ef8b4b742850a050f6bc4c993ef31a02ef25372

cmd/soroban-rpc/lib/preflight/src/lib.rs
Shaptic: I did the catch and unwind to incorporate @leighmcculloch's suggestion to avoid the JSON round-trip :+1:

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: Done, brilliant :pray:

leighmcculloch (commented):
The JSON has a version but that version isn't exposed in the http responses. How can a client identify which version of the JSON is in use?

We could expose the JSON version in http responses as a field where the version can be the major version of the Rust XDR lib in use.

This is something we could explore independently or later.

Maybe that version is identifiable through the RPCs version if they are consistent.

Shaptic:
@leighmcculloch it’d be same as the protocol XDR version, no?

leighmcculloch:
> @leighmcculloch it’d be same as the protocol XDR version, no?

Yup, the XDR's major version is same as the protocol version, and same as the major version of the rpc too.

The protocol version isn't really helpful because the rpc can be v22 while the protocol is still executing v21.

But for a client interacting with the RPC API how does it know when a response contains JSON from v21 vs v22?

Shaptic:
I'm still confused. Wouldn't the JSON for protocol 21 always be the same, and once it's protocol 22 it'd change?

2opremio:
All the `lib/preflight` changes belong to their own library e.g. `lib/xdr2json`

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,144 @@
+//nolint:lll
```
2opremio, line 1: What are the offending lines? Are they unavoidably long? Should we, otherwise tolerate longer lines in general (which is totally file, but I would rather do that than sprinkle files with `nolint:lll`)?

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,144 @@
...
+	return convertAnyBytes(xdrTypeName, field)
+}
+
+// ConvertInterface takes a valid XDR object (`xdr`) and returns
+// the raw JSON-formatted serialization of that object. If there is an
+// error, it returns an empty JSON object.
+//
+// Unlike `ConvertBytes`, the value here needs to be valid and
+// serializable.
+func ConvertInterface(xdr interface{}) (json.RawMessage, error) {
```
2opremio, line 63: If you expect a `BinaryMarshaler` why not enforce so directly?

Although, wouldn't it make more sense to enforce `xdr.EncodeTo`? 

Also, consider using an `xdr.Encoder` directly where possible to make encoding more efficient (less garbage(

2opremio: If the problem is you need the type name and you still need reflection, you can still do it internally, without exposing an `interface{}`

2opremio: But note that reflection is slow and has a high performance impact.

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,144 @@
...
+
+func convertAnyBytes(xdrTypeName string, field []byte) (json.RawMessage, error) {
+	var jsonStr, errStr string
+	// scope just added to show matching alloc/frees
+	{
+		goRawXdr := CXDR(field)
+		b := C.CString(xdrTypeName)
+
+		result := C.xdr_to_json(b, goRawXdr)
+		C.free(unsafe.Pointer(b))
```
2opremio, line 85: Is `unsafe.Pointer()` needed? I thought you could call `C.free` directly.

cmd/soroban-rpc/internal/xdr2json/conversion.go
2opremio: If the typename is strictly neccessary we can modify `xdrgen` to provide it thourgh a method.

leighmcculloch:
> I'm still confused. Wouldn't the JSON for protocol 21 always be the same, and once it's protocol 22 it'd change?

The structure of the JSON is defined by the structure of the XDR definitions.

When a new version of stellar-core is prepared such as v22 will be, it might contain changes to the structure of the XDR definitions. The XDR binary will always be backwards compatible, but the XDR definitions can change structure.

That new XDR definitions will be released with the software stellar-core v22 and all the downstream components. But the new elements of the v22 XDR won't show up until the network upgrades.

So the structure of the JSON can change the moment XDR v22 is imported, not activated.

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: It's the C compilation lines; you had the same directive in preflight.go

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: Nice, not sure why I didn't realize that I could just ask for the interface I wanted. I think `BinaryMarshaler` is fine since the only thing I need is `MarshalBinary`, so it feels prescriptive to require a higher-level interface.

EncodeTo is a good idea, but it will make caller code more verbose. I'll explore this to see if it's worthwhile! It may not be since each structure is separate in memory.

> But note that reflection is slow and has a high performance impact.
> If the typename is strictly neccessary we can modify xdrgen to provide it thourgh a method.

It is necessary, yeah. Totally agreed that it's better to have a `Name()` method on each XDR structure that gets generated; using reflection was the only way to get the name otherwise. Since we'd need to modify xdrgen, we should do it after this PR and patch it later; it'll be a small fix here.

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: Nope, `free` requires an unsafe.Pointer and `b` isn't one. You have the same code in [preflight.go here](https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/preflight/preflight.go#L76-L78) for `free`s.

Shaptic:
@2opremio for this bit

> All the lib/preflight changes belong to their own library though e.g. lib/xdr2json

While I agree with you in principal, you can see that there is a lot of code re-use for memory management for the FFI layer, e.g. `from_c_string`, `from_c_xdr`, `string_to_c`, etc. While we could extract that into its own library and share it across preflight and this new layer, this seems like an unnecessary pain in the ass (new compilation flow, refactoring preflight itself) for a relatively small gain in library structure. If we do it, we should do it separately from this PR to make it have its own set of CI checks and separate feedback evaluation imo.

Shaptic:
@leighmcculloch is bd2a7a0 akin to what you had in mind? The returned `jsonSchemaVersion` field will always match Core's major version, which means e.g. `22.0.0` will return `22` even if it isn't activated yet.

2opremio:
> @2opremio for this bit
> 
> > All the lib/preflight changes belong to their own library though e.g. lib/xdr2json
> 
> While I agree with you in principal, you can see that there is a lot of code re-use for memory management for the FFI layer, e.g. `from_c_string`, `from_c_xdr`, `string_to_c`, etc. While we could extract that into its own library and share it across preflight and this new layer, this seems like an unnecessary pain in the ass (new compilation flow, refactoring preflight itself) for a relatively small gain in library structure. If we do it, we should do it separately from this PR to make it have its own set of CI checks and separate feedback evaluation imo.

Yes, I think we should create a separate library for this and I don’t think we should to master until that is done. Sorry, but I don’t think that the preflight library should be polluted. I am ok with not having CI checks for the library in this PR (there were no CI checks before anyways).

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: I added a benchmark in 1ef0940, here are the results:

```
$ go test -bench=. -run='BenchmarkJSONTransactions' ./cmd/soroban-rpc/internal/methods/...
goos: linux
goarch: amd64
cpu: 13th Gen Intel(R) Core(TM) i7-1360P
BenchmarkJSONTransactions/JSON_format-16         	  199840	      6885 ns/op
BenchmarkJSONTransactions/XDR_format-16          	  709465	      2500 ns/op
```

You can see that JSON mode is ~4x slower than XDR mode, but this is not due to reflection. In fact, reflection is not even in the top 100 for the CPU trace. Profiling shows that the **additional time is spent almost exclusively in the cgocall** across the Go -> C (Rust) boundary:

```
(pprof) top5  
Showing nodes accounting for 1930ms, 42.05% of 4590ms total
Dropped 138 nodes (cum <= 22.95ms)
Showing top 5 nodes out of 156
      flat  flat%   sum%        cum   cum%
    1190ms 25.93% 25.93%     1510ms 32.90%  runtime.cgocall
     230ms  5.01% 30.94%      230ms  5.01%  runtime.duffcopy
     220ms  4.79% 35.73%      660ms 14.38%  runtime.mallocgc
     160ms  3.49% 39.22%      160ms  3.49%  encoding/base64.(*Encoding).Encode
     130ms  2.83% 42.05%      320ms  6.97%  runtime.scanobject
```

<hr>

Side note for @leighmcculloch: I wanted to see how much your suggestion about adding the `json.RawMessage` support helped. It led to a ~2.5x speedup since there's much less time doing marshaling! So you rock :guitar: Details:

<details>

```
$ go test -bench=. -run='BenchmarkJSONTransactions' ./cmd/soroban-rpc/internal/methods/...
goos: linux
goarch: amd64
cpu: 13th Gen Intel(R) Core(TM) i7-1360P
BenchmarkJSONTransactions/JSON_format-16         	   69666	     16455 ns/op
BenchmarkJSONTransactions/XDR_format-16          	  515062	      2454 ns/op
```

</details>

Shaptic:
@2opremio sure, that's fine with me! In that case, do you mind approving this to merge into a feature branch so that we can still isolate the preflight <-> xdr2json separation changes?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -392,35 +409,60 @@ func eventInfoForEvent(event xdr.DiagnosticEvent, cursor events.Cursor, ledgerCl
...
+		for _, topic := range v0.Topics {
+			topic, err := xdr2json.ConvertInterface(topic)
+			if err != nil {
+				return EventInfo{}, err
+			}
+			topics = append(topics, topic)
+		}
+		info.TopicJSON = topics
+
+		dataJs, err := xdr2json.ConvertInterface(v0.Data)
```
tamirms, line 435: nit: since the topics variable defined above does not have the Js suffix, for the sake of consistency, I think we should rename this variable to `data`

cmd/soroban-rpc/internal/xdr2json/conversion.go
```diff
@@ -0,0 +1,139 @@
...
+
+	resultMeta, err = ConvertBytes(xdr.TransactionMeta{}, tx.Meta)
+	if err != nil {
+		return result, envelope, resultMeta, err
+	}
+
+	return result, envelope, resultMeta, nil
+}
+
+func IsValidConversion(format string) error {
```
tamirms, line 130: nit: I think `IsValidFormat()` would be a better name

tamirms: also, I think this function, along with the FormatJSON and FormatBase64 constants, should be moved to the methods package. the xdr2json package seems to be solely focused on converting binary xdr to json and it doesn't need to be aware of other formats

cmd/soroban-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -101,32 +113,58 @@ func NewGetLedgerEntriesHandler(logger *log.Entry, ledgerEntryReader db.LedgerEn
...
+					KeyJSON:            keyJs,
+					DataJSON:           entryJs,
+					LastModifiedLedger: uint32(ledgerKeyAndEntry.Entry.LastModifiedLedgerSeq),
+					LiveUntilLedgerSeq: ledgerKeyAndEntry.LiveUntilLedgerSeq,
+				})
+
+			default:
+				keyXDR, err := xdr.MarshalBase64(ledgerKeyAndEntry.Key)
+				if err != nil {
+					logger.WithError(err).WithField("request", request).
```
tamirms, line 143: nit: we seem to be inconsistent about whether we log serialization errors. In this endpoint we log marshling errors only for bas64 but not for json. in other endpoints we never log  serialization errors

cmd/soroban-rpc/internal/methods/get_events.go
Shaptic: sure, fair enough! refactored a bit to not use temporary variables and have consistent naming :+1:

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: Yep, agreed :+1: moved to methods and renamed

cmd/soroban-rpc/internal/methods/get_ledger_entries.go
Shaptic: Yeah I agree with you but didn't want to modify this code (it's unchanged). Removed in abeffe9 :+1:

tamirms (approved):
LGTM!

cmd/soroban-rpc/internal/xdr2json/conversion.go
Shaptic: Following up from Slack. I ran a Python benchmark that queried for random transactions 5000 times. Here are the latency results:

```bash
$ python3 benchmark.py 
5000 queries with xdrFormat=base64 format took 7.89s
5000 queries with xdrFormat=json format took 11.33s
$ python3 benchmark.py 
5000 queries with xdrFormat=base64 format took 7.89s
5000 queries with xdrFormat=json format took 11.66s
$ python3 benchmark.py 
5000 queries with xdrFormat=base64 format took 8.19s
5000 queries with xdrFormat=json format took 12.00s
```

As you can see, JSON mode definitely does take noticeably longer (about ~40% overhead), but it's much less overhead when amortized over the entire request.

PR #251, opened 2024-07-17
Use `TransactionInfo` within `GetTransactionResponse`

Shaptic:
**WARNING**: This is a breaking API change; see _Known limitations_, below.

### What
Use `TransactionInfo` (from the `getTransactions()` handler) as part of the `getTransaction()` handler.

### Why
One is a subset of the other, and this makes it easier for both the code and documentation to reflect that.

It also guarantees that the endpoints don't diverge in the future.

This is partially related to #124, where I realized I was duplicating work by parsing out the transaction fields separately for each endpoint: this would make the work more streamlined.

### Known limitations
According to the [CI integration test runs](https://github.com/stellar/soroban-rpc/actions/runs/9967582043/job/27541488475?pr=251), this breaks `getTransaction`'s integration tests because `createdAt` is encoded as an `int64` rather than a `string` in the schema for `getTransactions`. We need to change this for JavaScript which is a breaking API change.

### DIFF
--- a/cmd/soroban-rpc/internal/methods/get_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/get_transaction.go
@@ -29,8 +29,6 @@ const (
 
 // GetTransactionResponse is the response for the Soroban-RPC getTransaction() endpoint
 type GetTransactionResponse struct {
-	// Status is one of: TransactionSuccess, TransactionNotFound, or TransactionFailed.
-	Status string `json:"status"`
 	// LatestLedger is the latest ledger stored in Soroban-RPC.
 	LatestLedger uint32 `json:"latestLedger"`
 	// LatestLedgerCloseTime is the unix timestamp of when the latest ledger was closed.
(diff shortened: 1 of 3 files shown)


cmd/soroban-rpc/internal/methods/get_transaction.go
```diff
@@ -29,8 +29,6 @@ const (
 
 // GetTransactionResponse is the response for the Soroban-RPC getTransaction() endpoint
 type GetTransactionResponse struct {
-	// Status is one of: TransactionSuccess, TransactionNotFound, or TransactionFailed.
-	Status string `json:"status"`
```
psheth9, line 33: We could still keep the `Status` field to show what actually happened with Txn. Any strong motive to remove this field? or is it rarely used by users?

psheth9 (approved):
Much cleaner !! just one comment on removing the `status` field

cmd/soroban-rpc/internal/methods/get_transaction.go
Shaptic: Well it's still part of `TransactionInfo`. It can take on one additional value (`TransactionNotFound`) for this endpoint, but since it's a `string` regardless this doesn't impact the code.

PR #255, opened 2024-07-24
Add CI concurrency group for some workflows

mollykarcher:
### What

Add CI concurrency group limiting to one build per branch, canceling any older runs.

### Why

See discussion [here](https://stellarfoundation.slack.com/archives/C02B04RMK/p1721777179370309). Goal is to reduce cost and limit CI resource usage. This change will make it so that CI runs are canceled when new commits are pushed, except on protected branches (ex. `main`) 

Note that the only workflow that is actually costing us money is `e2e`. Since this repo is public, all workflows that run on standard github runners (which is everything else) are free. So if for some reason we still want/like the runs for the other workflows to keep going when new commits are pushed, we could eliminate the changes there and just keep the `e2e` workflow changes.

### DIFF
--- a/.github/workflows/dependency-check.yml
+++ b/.github/workflows/dependency-check.yml
@@ -5,6 +5,10 @@ on:
     branches: [ main, release/** ]
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_protected == 'true' && github.sha || github.ref }}
+  cancel-in-progress: true
+
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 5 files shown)


Shaptic (approved):
You should've checked this by pushing some fixup commits :stuck_out_tongue_closed_eyes: LGTM though!

mollykarcher:
> You should've checked this by pushing some fixup commits 😝 LGTM though!

Yeah good call, we gucci

image
https://github.com/stellar/soroban-rpc/actions/runs/10081693493

PR #256, opened 2024-07-29
Optimise db performance of `GetLedgerRange`

### LINKED ISSUE #233: db: Remove latestledger key from meta table

aditya1702:
### What

Fixes #233 

- Removes the `LatestLedger` key from `meta` table. Use `GetLedgerRange` function since it already contains information about the latest ledger.
- Make the `GetLedgerRange` faster by caching `latestLedgerSeq` and `latestLedgerCloseTime`. This leads to a single query to get the oldest ledger and its close time.

### Why

The `latestLedgerSeq` key can be removed since we already get that info from ledger range function. The `GetLedgerRange` is used a lot now so it is important to make it faster by caching the values.

### Known limitations

NA


aditya1702:
@2opremio I will include caching the ledger range in another PR since I am testing things with it and this PR can be merged without it.

2opremio:
Caching the latestledger is pretty fundamental for performance, so I would rather wait to have caching in main.

aditya1702:
> Caching the latest ledger is pretty fundamental for performance, so I would rather wait to have caching in main.

@2opremio The `latestLedger` is still being cached - https://github.com/stellar/soroban-rpc/pull/256/files#diff-2b645f26f8eb502d1988698186482a5910b37509e606f968a3a60c927c397cfeL150. I was referencing the `oldestLedger` and the close times that comprise the range.

2opremio:
Ah ok, then the PR description and title are outdated. Can you update them?

2opremio:
Or do you mean that everything is cached and working except to the oldest ledger and the oldest ledger close time?

2opremio:
OK, I've taken a look at the code

> @2opremio The `latestLedger` is still being cached - https://github.com/stellar/soroban-rpc/pull/256/files#diff-2b645f26f8eb502d1988698186482a5910b37509e606f968a3a60c927c397cfeL150. I was referencing the `oldestLedger` and the close times that comprise the range.


As I recall, this places it into the downstream cache, but doesn't really use the cache (as you can see in the code the code doesn't check the cache for the value it just writes to it). We need to be careful about using `GetLedgerRange()` as opposed to `GetLatestLedger()` here because it's more expensive.

One option is to actually read from the cache there but I think there were some subtleties preventing from doing so. Can you take a look?

aditya1702:
Benchmarking `GetLedgerRange` after the new caching changes:

```
goos: darwin
goarch: arm64
BenchmarkGetLedgerRange
BenchmarkGetLedgerRange-12    	   81654	     14526 ns/op
```

This is a reduction of 40% from the previous benchmarking time - https://github.com/stellar/soroban-rpc/pull/217#issuecomment-2197552507

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -31,10 +32,92 @@ type ledgerReader struct {
...
+	}, nil
+}
+
 func NewLedgerReader(db *DB) LedgerReader {
 	return ledgerReader{db: db}
 }
 
+func (r ledgerReader) NewTx(ctx context.Context) (LedgerReaderTx, error) {
+	txSession := r.db.Clone()
+	if err := txSession.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
```
tamirms, line 109: this transaction is never closed and it's actually not passed into `ledgerReaderTx{}` .

I think a better approach would be to remove the `NewTx()` function and keep the `LedgerReader ` interface as it was before with `GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error)`.

`GetLedgerRange()` can encapsulate the transaction management

aditya1702: I am going to remove the transaction creation as discussed with @tamirms on a call.

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -75,6 +75,38 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
 
 // GetLedgerRange pulls the min/max ledger sequence numbers from the meta table.
 func (r ledgerReader) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error) {
+	var latestLedgerSeqCache uint32
+	var latestLedgerCloseTimeCache int64
+
+	r.db.cache.RLock()
+	latestLedgerSeqCache = r.db.cache.latestLedgerSeq
+	latestLedgerCloseTimeCache = r.db.cache.latestLedgerCloseTime
+	r.db.cache.RUnlock()
```
tamirms, line 84:
```suggestion
	r.db.cache.RLock()
	latestLedgerSeqCache := r.db.cache.latestLedgerSeq
	latestLedgerCloseTimeCache := r.db.cache.latestLedgerCloseTime
	r.db.cache.RUnlock()
```

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -75,6 +75,38 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
+				fmt.Sprintf("sequence = (SELECT MIN(sequence) FROM %s)", ledgerCloseMetaTableName),
+			)
+		var lcm []xdr.LedgerCloseMeta
+		if err := r.db.Select(ctx, &lcm, query); err != nil {
+			return ledgerbucketwindow.LedgerRange{}, fmt.Errorf("couldn't query ledger range: %w", err)
+		}
+
+		return ledgerbucketwindow.LedgerRange{
+			FirstLedger: ledgerbucketwindow.LedgerInfo{
+				Sequence:  lcm[0].LedgerSequence(),
```
tamirms, line 100: you should add a check to assert lcm is not empty. I know that we don't expect it to be empty because of the cache indicates that ledger exists but it is still worth including the check imo

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -128,33 +128,29 @@ func getMetaValue(ctx context.Context, q db.SessionInterface, key string) (strin
...
-		return "", fmt.Errorf("multiple entries (%d) for key %q in table %q", len(results), latestLedgerSequenceMetaKey, metaTableName)
+		return "", fmt.Errorf("multiple entries (%d) for key %q in table %q", len(results), key, metaTableName)
 	}
 	return results[0], nil
 }
 
-func getLatestLedgerSequence(ctx context.Context, q db.SessionInterface, cache *dbCache) (uint32, error) {
-	latestLedgerStr, err := getMetaValue(ctx, q, latestLedgerSequenceMetaKey)
+func getLatestLedgerSequence(ctx context.Context, ledgerReader LedgerReader, cache *dbCache) (uint32, error) {
+	ledgerRange, err := ledgerReader.GetLedgerRange(ctx)
```
tamirms, line 137: this call will end up querying the DB even if the cache is populated. I think we should first check the cache and return that value if it exists. otherwise, we can query the db and populate the cache

aditya1702: @tamirms This function only gets called during filling of entries via the checkpoint - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/ingest/service.go#L175. At all other instances, we have a check on the cache to see if its populated and return that value instead:
- https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/db/ledgerentry.go#L148
- https://github.com/stellar/soroban-rpc/pull/256/files#diff-c57d2d5677168ee9d46b7a74fb7a34a5ba6e11448ada5854e6c85d4246140fa2R345

I tried adding a check directly where you pointed, but it leads to race condition issues during concurrent reads and writes. @2opremio was also mentioning encountering this when he added this change first.

tamirms: > At all other instances, we have a check on the cache to see if its populated and return that value instead

The fact that this check is repeated more than once is a sign that the check should be encapsulated in `getLatestLedgerSequence()` so that all the callers of `getLatestLedgerSequence()` don't have to check the cache.

> I tried adding a check directly where you pointed, but it leads to race condition issues during concurrent reads and writes

if there's no way to resolve this issue we can keep the code as is. Could you describe the race condition in more detail?

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -140,15 +140,13 @@ type ledgerEntryReadTx struct {
...
 
 func (l *ledgerEntryReadTx) GetLatestLedgerSequence() (uint32, error) {
-	if l.latestLedgerSeqCache != 0 {
-		return l.latestLedgerSeqCache, nil
-	}
-	latestLedgerSeq, err := getLatestLedgerSequence(context.Background(), l.tx, l.globalCache)
+	latestLedgerSeq, err := getLatestLedgerSequence(context.Background(), l.ledgerReader, l.globalCache)
 	if err == nil {
 		l.latestLedgerSeqCache = latestLedgerSeq
 	}
```
tamirms, line 152: Sorry, I think there's a subtle bug here that I overlooked. Check out the comment here https://github.com/stellar/soroban-rpc/blob/f7ebe3ebccb47d02edc85946e63544282af6bb10/cmd/soroban-rpc/internal/db/ledgerentry.go#L348-L353

if we use the global cache we to get the latest ledger seq we might be getting a newer ledger than the one bound to the ledgerEntryReadTx.

I think there's also a pre-existing bug with `NewTx()`:

https://github.com/stellar/soroban-rpc/blob/f7ebe3ebccb47d02edc85946e63544282af6bb10/cmd/soroban-rpc/internal/db/ledgerentry.go#L372-L377

we should be locking before starting the transaction just like how it's done in `NewCachedTx()`. I wonder if we can get rid of `NewTx()` and instead always use  `NewCachedTx()`

@2opremio will be to confirm

aditya1702: @tamirms Yes you are right here. But I think the change to add the cache read in `getLatestLedgerSequence` will still work for the reader function here - https://github.com/stellar/soroban-rpc/blob/f7ebe3ebccb47d02edc85946e63544282af6bb10/cmd/soroban-rpc/internal/db/ledgerentry.go#L341. But yes, the one bound to tx should still use the local cache value. @2opremio Any thoughts?

cmd/soroban-rpc/internal/db/db.go
aditya1702: @tamirms It was a mistake on my side. I was not locking the caching during read and that was leading to the race condition. It now works fine and the tests are passing

cmd/soroban-rpc/internal/db/ledgerentry.go
2opremio: > I wonder if we can get rid of `NewTx()` and instead always use `NewCachedTx()`
> 
> @2opremio will be to confirm

I took a look at `NewTx()` vs `NewCachedTx()`. There is an advantage to not using the cache when it's not needed (e.g. in `getLedgerEntries`, where the the entries will never be queried multiple times) the cache would just cause GC pressure but ... we should measure whether it's worth the extra complexity.



> we should be locking before starting the transaction just like how it's done in `NewCachedTx()`

The cache isn't being used by the transaction for reads (AFAICT it's only passed to update the latestLedgerEntry when it's not initialized) so I don't think we need to lock earlier. Do you have a specific race condition in mind?

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -75,6 +75,39 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
+				CloseTime: lcm[0].LedgerCloseTime(),
+			},
+			LastLedger: ledgerbucketwindow.LedgerInfo{
+				Sequence:  latestLedgerSeqCache,
+				CloseTime: latestLedgerCloseTimeCache,
+			},
+		}, nil
+	}
+
```
2opremio, line 110: Did you measure whether it's worth doing this? It's probably as expensive to query the `MIN` as `MIN+MAX`.

2opremio:
I am missing a migration to remove the entry from the meta table.

cmd/soroban-rpc/internal/db/ledger.go
2opremio: In fact, we should be looking at the uses of `GetLedgerRange` to see if all callers are using the oldest seq+closeTime.

2opremio: From what I see `getLatestLedgerSequence` is one of those. It shouldn't be obtaining the oldest seq since it's not using it.

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -128,33 +128,37 @@ func getMetaValue(ctx context.Context, q db.SessionInterface, key string) (strin
...
-	if err != nil {
-		return 0, err
+func getLatestLedgerSequence(ctx context.Context, ledgerReader LedgerReader, cache *dbCache) (uint32, error) {
+	cache.RLock()
+	latestLedgerSeqCache := cache.latestLedgerSeq
+	cache.RUnlock()
+
+	if latestLedgerSeqCache != 0 {
+		return latestLedgerSeqCache, nil
 	}
```
2opremio, line 143: Are you sure this can be done now? 

I am not saying it can't. It's just that I haven't looked at this code for a while and from our latest conversation there were failing tests and races when enabling this. I would like to know what changed to make this possible.

cmd/soroban-rpc/internal/db/ledgerentry.go
tamirms: > Do you have a specific race condition in mind?

assume the read only transaction is created and then, right before we read lock the cache, the ingestion transaction on a separate goroutine commits and updates the latest ledger to n+1. After we acquire the read lock, we set `latestLedgerSeqCache` to n+1. However, since the read only transaction was created before the ingestion transaction was comitted, all the queries we do on that transaction will be consistent with ledger n instead of ledger n+1

2opremio: Ah, true, then we should lock while creating the transaction.

cmd/soroban-rpc/internal/db/db.go
aditya1702: @2opremio The failing tests earlier were happening because I did not lock the cache when reading the value:
https://github.com/stellar/soroban-rpc/pull/256/files#diff-2b645f26f8eb502d1988698186482a5910b37509e606f968a3a60c927c397cfeR137

After adding the `Rlock()` function, there are no races encountered

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: @2opremio If you look at the benchmark I poseted here - https://github.com/stellar/soroban-rpc/pull/256#issuecomment-2271592530 - it has reduced the `GetLedgerRange` time by almost half just by querying only the `MIN`. So yes, there is an advantage to using the cache here.

Technically, the `GetLedgerRange` will only be called once in `getLatestLedgerSequence` when the cache is initially empty. On all subsequent requests, it will just return the cached value.

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -364,22 +365,24 @@ func (r ledgerEntryReader) NewCachedTx(ctx context.Context) (LedgerEntryReadTx,
...
 }
 
 func (r ledgerEntryReader) NewTx(ctx context.Context) (LedgerEntryReadTx, error) {
 	txSession := r.db.Clone()
+	r.db.cache.RLock()
 	if err := txSession.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
 		return nil, err
 	}
-	r.db.cache.RLock()
 	defer r.db.cache.RUnlock()
```
tamirms, line 380: I think the `defer r.db.cache.RUnlock()` needs to be moved up as well. if there is an error in `txSession.BeginTx()` that will result in the the cache being permanently read locked

tamirms: also, now that `NewTx()` and `NewCachedTx()` are almost identical I think we can refactor the code so that they use a common helper function to reduce code duplication

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -341,21 +341,29 @@ func NewLedgerEntryReader(db *DB) LedgerEntryReader {
...
 
 func (r ledgerEntryReader) GetLatestLedgerSequence(ctx context.Context) (uint32, error) {
-	return getLatestLedgerSequence(ctx, r.db, r.db.cache)
+	return getLatestLedgerSequence(ctx, NewLedgerReader(r.db), r.db.cache)
+}
+
+func (r ledgerEntryReader) newBaseTx(ctx context.Context) (db.SessionInterface, error) {
+	txSession := r.db.Clone()
+	r.db.cache.RLock()
+	defer r.db.cache.RUnlock()
```
tamirms, line 350: the lock is released right after `newBaseTx()` returns, which means that the lock is not protecting anything in NewCachedTx() or NewTx()

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -355,7 +362,8 @@ func (r ledgerEntryReader) NewCachedTx(ctx context.Context) (LedgerEntryReadTx,
 	// in between, updating the cache.
 	r.db.cache.RLock()
 	defer r.db.cache.RUnlock()
-	if err := txSession.BeginTx(ctx, &sql.TxOptions{ReadOnly: true}); err != nil {
+	txSession, err := r.newBaseTx(ctx)
```
tamirms, line 365: I think we can get rid of the `newBase()` function and instead you can implement `NewCachedTx()` by calling `NewTx()` and then setting the `ledgerEntryCacheReadTx` and `stmtCache` fields on the *ledgerEntryReadTx instance before returning it

tamirms (approved):

PR #262, opened 2024-08-08
Add events table migration

### LINKED ISSUE #257: Add migration for events and optimize initStorage in daemon.go
We need to add migration similar to the transactions table.

aditya1702:
### What

Closes #257 

- Add migration for events table.
- Move db transaction logic for migration out of `guardedMigration`.

### Why

- We are moving events from in-memory to db, hence the migration support.
- Sqlite DB cannot handle two concurrent db transactions, so we need to have a common transaction for both migrations instead of individual ones. Having per migration db tx leads to `database is locked` error.

### Known limitations

NA

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -39,6 +39,7 @@ const (
 	defaultReadTimeout                    = 5 * time.Second
 	defaultShutdownGracePeriod            = 10 * time.Second
 	inMemoryInitializationLedgerLogPeriod = 1_000_000
+	firstLedger                           = 2
 )
 
 type Daemon struct {
@@ -310,7 +311,7 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 		d.logger.WithError(err).Fatal("failed to get latest ledger sequence: %w", err)
 	}
 	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
-	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: 2, LastLedgerSeq: latestLedger}
+	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: firstLedger, LastLedgerSeq: latestLedger}
 	if latestLedger > maxFeeRetentionWindow {
 		ledgerSeqRange.FirstLedgerSeq = latestLedger - maxFeeRetentionWindow
 	}
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -317,44 +317,60 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 	}
 	ledgerSeqRange = ledgerSeqRange.Merge(applicableRange)
 
-	dataMigrations, err := db.BuildMigrations(readTxMetaCtx, d.logger, d.db, cfg)
+	dataMigrations, err := db.BuildMigrations(readTxMetaCtx, d.logger, d.db, cfg.NetworkPassphrase, ledgerSeqRange)
 	if err != nil {
 		d.logger.WithError(err).Fatal("could not build migrations")
 	}
 
-	err = db.NewLedgerReader(d.db).StreamLedgerRange(
-		readTxMetaCtx,
-		ledgerSeqRange.FirstLedgerSeq,
-		ledgerSeqRange.LastLedgerSeq,
-		func(txmeta xdr.LedgerCloseMeta) error {
-			currentSeq = txmeta.LedgerSequence()
-			if initialSeq == 0 {
-				initialSeq = currentSeq
-				d.logger.WithFields(supportlog.F{
-					"seq": currentSeq,
-				}).Info("initializing in-memory store")
-			} else if (currentSeq-initialSeq)%inMemoryInitializationLedgerLogPeriod == 0 {
-				d.logger.WithFields(supportlog.F{
-					"seq": currentSeq,
-				}).Debug("still initializing in-memory store")
-			}
+	for _, migrationFactory := range dataMigrations {
+		// create guarded migration
+		guardedMigration, err := db.NewGuardedDataMigration(
+			readTxMetaCtx,
+			migrationFactory.MigrationName,
+			migrationFactory.Logger,
+			migrationFactory.Factory,
+			migrationFactory.DB,
+		)
+		if err != nil {
+			d.logger.WithError(err).Fatal("could not create guarded migration for: %s",
+				migrationFactory.MigrationName)
+		}
 
-			if err := feewindows.IngestFees(txmeta); err != nil {
-				d.logger.WithError(err).Fatal("could not initialize fee stats")
-			}
-
-			if applicableRange.IsLedgerIncluded(currentSeq) {
-				if err := dataMigrations.Apply(readTxMetaCtx, txmeta); err != nil {
-					d.logger.WithError(err).Fatal("could not run migrations")
+		err = db.NewLedgerReader(d.db).StreamLedgerRange(
+			readTxMetaCtx,
+			ledgerSeqRange.FirstLedgerSeq,
+			ledgerSeqRange.LastLedgerSeq,
+			func(txmeta xdr.LedgerCloseMeta) error {
+				currentSeq = txmeta.LedgerSequence()
+				if initialSeq == 0 {
+					initialSeq = currentSeq
+					d.logger.WithFields(supportlog.F{
+						"seq": currentSeq,
+					}).Info("initializing in-memory store")
+				} else if (currentSeq-initialSeq)%inMemoryInitializationLedgerLogPeriod == 0 {
+					d.logger.WithFields(supportlog.F{
+						"seq": currentSeq,
+					}).Debug("still initializing in-memory store")
 				}
-			}
-			return nil
-		})
-	if err != nil {
-		d.logger.WithError(err).Fatal("could not obtain txmeta cache from the database")
-	}
-	if err := dataMigrations.Commit(readTxMetaCtx); err != nil {
-		d.logger.WithError(err).Fatal("could not commit data migrations")
+
+				if err := feewindows.IngestFees(txmeta); err != nil {
+					d.logger.WithError(err).Fatal("could not initialize fee stats")
+				}
+
+				if applicableRange.IsLedgerIncluded(currentSeq) {
+					if err := guardedMigration.Apply(readTxMetaCtx, txmeta); err != nil {
+						d.logger.WithError(err).Fatal("could not run migrations")
+					}
+				}
+				return nil
+			})
+		if err != nil {
+			d.logger.WithError(err).Fatal("could not obtain txmeta cache from the database")
+		}
+		if err := guardedMigration.Commit(readTxMetaCtx); err != nil {
+			d.logger.WithError(err).Fatal("could not commit migration for: %s",
+				migrationFactory.MigrationName)
+		}
 	}
 
 	if currentSeq != 0 {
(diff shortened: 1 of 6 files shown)


cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -122,32 +128,20 @@ type guardedMigration struct {
...
-	}
-	if err := migrationDB.Begin(ctx); err != nil {
-		return nil, err
-	}
 	metaKey := "Migration" + uniqueMigrationName + "Done"
-	previouslyMigrated, err := getMetaBool(ctx, migrationDB, metaKey)
+	previouslyMigrated, err := getMetaBool(ctx, db, metaKey)
 	if err != nil && !errors.Is(err, ErrEmptyDB) {
-		err = errors.Join(err, migrationDB.Rollback())
+		err = errors.Join(err, db.Rollback())
```
tamirms, line 134: it seems strange that the migration is responsible for rolling back the transaction but it did not create the transaction. I think it would be more appropriate for the daemon to handle rolling back since it created the transaction in the first place

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,28 +300,49 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
 		d.logger.WithError(err).Fatal("failed to get latest ledger sequence: %w", err)
 	}
-
 	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
-	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: latestLedger - maxFeeRetentionWindow, LastLedgerSeq: latestLedger}
-	applicableRange := dataMigrations.ApplicableRange()
+	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: firstLedger, LastLedgerSeq: latestLedger}
+	if latestLedger > maxFeeRetentionWindow {
+		ledgerSeqRange.FirstLedgerSeq = latestLedger - maxFeeRetentionWindow
+	}
```
tamirms, line 317: it seems this ledger range calculation is equivalent to `db.GetMigrationLedgerRange()`

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,28 +300,49 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
+	// Start a common db transaction for the entire migration duration
+	err = d.db.Begin(readTxMetaCtx)
+	if err != nil {
+		d.logger.WithError(err).Fatal("could not commit database transaction: ", d.db.Rollback())
+	}
+	defer func() {
+		err = d.db.Commit()
+		if err != nil {
+			d.logger.WithError(err).Fatal("could not commit database transaction")
+		}
```
tamirms, line 331: this will end up the committing the transaction even if `mustInitializeStorage()` panics later on via a fatal log line. So, we need to make sure that the transaction is rolled back before the commit call whenever we intend to exit via an error

cmd/soroban-rpc/internal/db/migration.go
aditya1702: @tamirms Yeah this code was being used in the previous version where tx was being created in `guardedMigration`. It does not need to be here now.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,28 +300,44 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
-	dataMigrations, err := db.BuildMigrations(readTxMetaCtx, d.logger, d.db, cfg)
+	applicableRange, err := db.GetMigrationLedgerRange(readTxMetaCtx, d.db, cfg.HistoryRetentionWindow)
 	if err != nil {
-		d.logger.WithError(err).Fatal("could not build migrations")
+		d.logger.WithError(err).Fatal("could not get ledger range for migration")
 	}
 
-	// Merge migrations range and fee stats range to get the applicable range
-	latestLedger, err := db.NewLedgerEntryReader(d.db).GetLatestLedgerSequence(readTxMetaCtx)
+	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
```
tamirms, line 308: do you know why the soroban fee stats retention window is different from the history retention window?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -39,6 +39,7 @@ const (
 	defaultReadTimeout                    = 5 * time.Second
 	defaultShutdownGracePeriod            = 10 * time.Second
 	inMemoryInitializationLedgerLogPeriod = 1_000_000
+	firstLedger                           = 2
```
tamirms, line 42: is this variable still used?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,28 +300,44 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
-	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: latestLedger - maxFeeRetentionWindow, LastLedgerSeq: latestLedger}
-	applicableRange := dataMigrations.ApplicableRange()
-	ledgerSeqRange = ledgerSeqRange.Merge(applicableRange)
+	// Combine the ledger range for fees, events and transactions
+	ledgerSeqRange := feeStatsRange.Merge(applicableRange)
+
+	// Start a common db transaction for the entire migration duration
+	err = d.db.Begin(readTxMetaCtx)
+	if err != nil {
+		d.logger.WithError(err).Fatal("could not commit database transaction: ", d.db.Rollback())
```
tamirms, line 320: d.db.Rollback() is passed as an argument to the log message but the log message expects no place holders (e.g. `%v`)

cmd/soroban-rpc/internal/daemon/daemon.go
aditya1702: removed it.

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -179,46 +160,48 @@ func (g *guardedMigration) Commit(ctx context.Context) error {
...
+) (MultiMigration, error) {
+	migrations := make(MultiMigration, 0, numMigrations)
 
-	m1, err := newGuardedDataMigration(ctx, migrationName, logger, factory, db)
-	if err != nil {
-		return nil, fmt.Errorf("creating guarded transaction migration: %w", err)
+	migrationNameToFunc := map[string]migrationApplierF{
+		transactionsMigrationName: newTransactionTableMigration,
+		eventsMigrationName:       newEventTableMigration,
 	}
```
tamirms, line 189: I recommend moving the `migrations` declaration to below `migrationNameToFunc` . That way you can get rid of the `numMigrations` constant:

```
migrations := make(MultiMigration, 0, len(migrationNameToFunc))
```

cmd/soroban-rpc/internal/daemon/daemon.go
psheth9: I thought it was a part of requirement or something that we follow in Horizon for calculating fee stats to consider last 50 ledgers. @2opremio  would know it better tho.

2opremio: It’s because the soroban activity is much lower and a larger window will increase the chances of including actual soroban transactions. On the other hand we want to the metric to contain recent information so we want to include a short recent window (hence why it’s different and shorter for classic)

cmd/soroban-rpc/internal/daemon/daemon.go
aditya1702: @tamirms `Fatal` function does not have any placeholders. We can just chain strings and variables like I have done.

psheth9:
Looks good to me !!  @2opremio can you review it too as you had few initial comments on migration [logic](https://github.com/stellar/soroban-rpc/pull/215#issuecomment-2188698254)

cmd/soroban-rpc/internal/daemon/daemon.go
tamirms: I think this approach is a bit unintuitive (why should the rollback error be printed in the log message without any context?) and error prone (if we forget to call rollback as an argument to the fatal call we can accidentally commit the transaction). Consider this alternative:

after creating the transaction, refactor the remaining code in this function to be a separate function which returns an error. The separate function would no longer call `log...Fatal()` upon encountering an error. Instead it would return the error and the code which handles the return value would rollback (just in one place) and call `log...Fatal()` to terminate.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -332,22 +348,20 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 				}).Debug("still initializing in-memory store")
 			}
 
-			if err := feewindows.IngestFees(txmeta); err != nil {
-				d.logger.WithError(err).Fatal("could not initialize fee stats")
+			if err = feeWindows.IngestFees(txMeta); err != nil {
+				d.logger.WithError(err).Fatal("could not initialize fee stats: ", d.db.Rollback())
```
2opremio, line 352: I would be much cleaner and readable to make the `Rollback` independent of logging

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -332,22 +348,20 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
 			}
 
-			if applicableRange.IsLedgerIncluded(currentSeq) {
-				if err := dataMigrations.Apply(readTxMetaCtx, txmeta); err != nil {
-					d.logger.WithError(err).Fatal("could not run migrations")
-				}
+			if err := dataMigrations.Apply(readTxMetaCtx, txMeta); err != nil {
+				d.logger.WithError(err).Fatal("could not apply migration for ledger ", currentSeq, ": ", d.db.Rollback())
```
2opremio, line 356: Same here

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,28 +299,44 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
 
-	maxFeeRetentionWindow := max(cfg.ClassicFeeStatsLedgerRetentionWindow, cfg.SorobanFeeStatsLedgerRetentionWindow)
-	ledgerSeqRange := &db.LedgerSeqRange{FirstLedgerSeq: latestLedger - maxFeeRetentionWindow, LastLedgerSeq: latestLedger}
-	applicableRange := dataMigrations.ApplicableRange()
-	ledgerSeqRange = ledgerSeqRange.Merge(applicableRange)
+	// Combine the ledger range for fees, events and transactions
+	ledgerSeqRange := feeStatsRange.Merge(applicableRange)
+
+	// Start a common db transaction for the entire migration duration
+	err = d.db.Begin(readTxMetaCtx)
```
2opremio, line 317: I understand that a common migration needs to be used for all migrations. But it can be _hidden_ inside the migration code (my guess is multimigration would be the best place to place it). It would make this code more readable and would most probably make rolling back/commiting much simpler.

2opremio:
Can you add a test for the migrations? (both at once to make sure there are no transaction conflicts when applying multiple migrations at once)

cmd/soroban-rpc/internal/daemon/daemon.go
aditya1702: @2opremio Yeah I agree. Nice suggestion

aditya1702:
> Can you add a test for the migrations? (both at once to make sure there are no transaction conflicts when applying multiple migrations at once)

@2opremio Doesnt the current `TestMigrate` function already test that? Doing both transactions and events migrations together

2opremio:
@aditya1702 yaou are right!!

2opremio (approved):

PR #271, opened 2024-08-20
Correctly omit `transactionData` field when transaction simulation fails

Shaptic:
### What
Correctly handle the case where transaction data is empty. Also, add a bit of verbosity (the type name) to error messages coming from the conversion layer.

### Why
Previously, the handler would unconditionally try to encode an empty byte array (since `result.TransactionData` was empty) and fail with an internal panic:

```
"xdr_to_json() failed: called `Result::unwrap()` on an `Err` value: Io(Error { kind: UnexpectedEof, message: \"failed to fill whole buffer\" })",
```

Obviously this is not ideal. It was occurring because the simulation failed, but the code would still try to JSONify results. In the base64 case, `EncodeToString` will just return an empty string on an empty buffer.

This code modifies `xdr2json.ConvertBytes` to check its inputs like so:

```diff
 func ConvertBytes(xdr interface{}, field []byte) ([]byte, error) {
+       if len(field) == 0 {
+               return []byte(""), nil
+       }
+
```

to behave akin to `EncodeToString` and handle this case "in general."


sreuland:
> Alternatively, we could modify xdr2json.ConvertBytes to check its inputs

that approach of sanitizing the json lower in conversion to empty string seems to be a bit cleaner, results in less conditionals in upper code, I would 👍  for that.

Shaptic:
@sreuland agreed! Done :heavy_check_mark:

cmd/soroban-rpc/lib/xdr2json/src/lib.rs
```diff
@@ -50,12 +50,18 @@ pub unsafe extern "C" fn xdr_to_json(
...
+            Err(e) => panic!("couldn't match type {type_str}: {}", e),
+        };
 
         let xdr_bytearray = unsafe { from_c_xdr(xdr) };
         let mut buffer = xdr::Limited::new(xdr_bytearray.as_slice(), DEFAULT_XDR_RW_LIMITS.clone());
 
-        let t = xdr::Type::read_xdr_to_end(the_type, &mut buffer).unwrap();
+        let t = match xdr::Type::read_xdr_to_end(the_type, &mut buffer) {
+            Ok(t) => t,
+            Err(e) => panic!("couldn't read {type_str}: {}", e),
```
psheth9, line 63: I dont understand rust that much so wondering that how we use `panic` here is usually encouraged in rust world?

psheth9 (approved):
can you add few tests? overall looks good to me just one question on rust code

sreuland (approved):
lgtm, for consideration, echo the same as other reviewer commented, about adding unit test on conversion.go if feasible.

cmd/soroban-rpc/lib/xdr2json/src/lib.rs
Shaptic: No worries, I had to figure this out myself for this.

Since we're "crossing a language boundary" through C code, we need to make sure that errors in the Rust portion are handled correctly. There are two ways to do this: (a) make sure your code never panics, or (b) ["unwind"](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) all panics. 

I initially did (a) and returned a JSONified `{ "error": "<message>" }` field, but I scrapped it because it meant that the client-side would have to decode and inspect the JSON which meant an extra round of unmarshalling, whereas how it works today is just shoving the bytes returned from Rust directly into the HTTP response. That meant I had to do (b) which is already what we're doing for preflight so I adapted that code accordingly.

You can see that this whole method is wrapped in a call to `catch_json_to_xdr_panic`, which basically says "run this function and catch any panics," formatting the panic message into a returned error string (hence the need for a `ConversionResult` struct rather than just returning the JSON bytes).

Shaptic:
Good calls @sreuland and @psheth9! Done :)

PR #272, opened 2024-08-20
Simplify logging code: `WithFields` -> `WithField`

Shaptic:
### What
This small change replaces logging calls that only have a single field with the `WithField(key, value)` variant instead of the more-verbose `WithFields(supportlog.F{...})`.

### Why
Readability.

### Known limitations
n/a


psheth9 (approved):
🚀

Shaptic:
@psheth9 I'll wait for getEvents so you don't have to deal with merge conflicts :+1:

PR #282, opened 2024-08-29
Improve startup by eliminating unnecessary migration ranges

Shaptic:
### What
Speed up startup time in "already migrated" cases by eliminating unnecessary ledger range traversals.

Specifically, this includes the following changes:
 * Fee stats windows cannot exceed the history retention window, as this doesn't make sense.
 * Migrations that are already applied are not added to the list of "multi-migrations".
 * Fee stats window building conforms to the migration interface to simplify code.
 * `LedgerSeqRange` has been refactored to always use a value reference.

As a result, if all migrations have occurred, the traversal only occurs over the fee window.

This reduces startup times in a fully-loaded (24hr) database that needs no migrations **from ~1m40s to ~1s**.

### Why
Previously, startup would take a significant amount of time doing nothing because it would loop over a large ledger range and check whether or not a migration was necessary for that ledger. If migrations were already done, it would move on to the next ledger.

In this version, we instead appropriately calculate the ledger range to traverse by only expanding the ledger range if a migration has *not* already completed.

### Known limitations
n/a, tested at length locally

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -348,7 +348,7 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 	//    will return the *applicable* range for the incomplete set of migrations.
 	//    That means **it may be empty** if all migrations have occurred.
 	//
-	dataMigrations, applicableRange, err := db.BuildMigrations(
+	dataMigrations, err := db.BuildMigrations(
 		readTxMetaCtx, d.logger, d.db, cfg.NetworkPassphrase, retentionRange)
 	if err != nil {
 		d.logger.WithError(err).Fatal("could not build migrations")
@@ -364,7 +364,7 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 	if err != nil {
 		d.logger.WithError(err).Fatal("could not get ledger range for fee stats")
 	}
-	ledgerSeqRange := feeStatsRange.Merge(&applicableRange)
+	ledgerSeqRange := dataMigrations.ApplicableRange().Merge(feeStatsRange)
 
 	//
 	// 5. Apply migration for events & transactions, and perform fee stat analysis.
@@ -409,7 +409,9 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 	}
 
 	if currentSeq != 0 {
-		d.logger.WithField("seq", currentSeq).
+		d.logger.
+			WithField("first", retentionRange.First).
+			WithField("last", retentionRange.Last).
 			Info("Finished initializing in-memory store and applying DB data migrations")
 	}
 
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -34,11 +34,15 @@ import (
 )
 
 const (
-	prometheusNamespace                   = "soroban_rpc"
-	maxLedgerEntryWriteBatchSize          = 150
-	defaultReadTimeout                    = 5 * time.Second
-	defaultShutdownGracePeriod            = 10 * time.Second
-	inMemoryInitializationLedgerLogPeriod = 1_000_000
+	prometheusNamespace          = "soroban_rpc"
+	maxLedgerEntryWriteBatchSize = 150
+	defaultReadTimeout           = 5 * time.Second
+	defaultShutdownGracePeriod   = 10 * time.Second
+
+	// Since our default retention window will be 7 days (7*17,280 ledgers),
+	// choose a random 5-digit prime to have irregular logging intervals at each
+	// halfish-day of processing
+	inMemoryInitializationLedgerLogPeriod = 10_099
 )
 
 type Daemon struct {
@@ -289,6 +293,31 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
 
 // mustInitializeStorage initializes the storage using what was on the DB
 func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows {
+	//
+	// There is a lot of complex "ledger window math" here so it's worth
+	// clarifying as a comment beforehand.
+	//
+	// There are two windows in play here:
+	//  - the ledger retention window, which describes the range of txmeta
+	//    to keep relative to the latest "ledger tip" of the network
+	//  - the fee stats window, which describes a *subset* of the prior
+	//    ledger retention window on which to perform fee analysis
+	//
+	// If the fee window *exceeds* the retention window, this doesn't make any
+	// sense since it implies the user wants to store N amount of actual
+	// historical data and M > N amount of ledgers just for fee processing,
+	// which is nonsense from a performance standpoint. So we should prevent
+	// this config on startup.
+	//
+	maxFeeRetentionWindow := max(
+		cfg.ClassicFeeStatsLedgerRetentionWindow,
+		cfg.SorobanFeeStatsLedgerRetentionWindow)
+	if maxFeeRetentionWindow > cfg.HistoryRetentionWindow {
+		d.logger.Fatalf(
+			"Fee stat analysis window (%d) cannot exceed history retention window (%d).",
+			maxFeeRetentionWindow, cfg.HistoryRetentionWindow)
+	}
+
 	feeWindows := feewindow.NewFeeWindows(
 		cfg.ClassicFeeStatsLedgerRetentionWindow,
 		cfg.SorobanFeeStatsLedgerRetentionWindow,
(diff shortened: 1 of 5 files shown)


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -289,6 +293,31 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
 
 // mustInitializeStorage initializes the storage using what was on the DB
 func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows {
+	//
+	// There is a lot of complex "ledger window math" here so it's worth
```
aditya1702, line 297: Nice work adding the comments 👏🏻

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -165,55 +181,69 @@ func (g *guardedMigration) Commit(ctx context.Context) error {
...
 		)
 
 		guardedM, err := newGuardedDataMigration(ctx, migrationName, migrationLogger, factory, db)
 		if err != nil {
-			return MultiMigration{}, errors.Join(err, fmt.Errorf(
+			return MultiMigration{}, applicableRange, errors.Join(err, fmt.Errorf(
 				"could not create guarded migration for %s", migrationName), db.Rollback())
 		}
+
+		if guardedM == nil {
```
aditya1702, line 236: Since we added the check here, I think we can remove all the `if g.alreadyMigrated` conditions in `guardedMigration` interface - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/db/migration.go#L142

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -328,32 +378,39 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
-					Debug("still initializing in-memory store")
+				d.logger.
+					WithField("seq", currentSeq).
+					WithField("last", ledgerSeqRange.Last).
+					Debug("Still initializing in-memory store")
 			}
 
-			if err = feeWindows.IngestFees(txMeta); err != nil {
-				d.logger.WithError(err).Fatal("could not initialize fee stats")
+			if feeStatsRange.IsLedgerIncluded(currentSeq) { // skip irrelevant ledgers
```
aditya1702, line 392: Ah nice optimisation.

aditya1702 (approved):
Just one nit comment. Otherwise looks good.

cmd/soroban-rpc/internal/db/migration.go
Shaptic: Technically yes, since we won't run that code path the way it's written now, but I'd rather leave it in to keep each abstraction layer fully functional within itself.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -328,32 +378,39 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 			currentSeq = txMeta.LedgerSequence()
 			if initialSeq == 0 {
 				initialSeq = currentSeq
-				d.logger.WithField("seq", currentSeq).
-					Info("initializing in-memory store")
+				d.logger.
+					WithField("first", initialSeq).
+					WithField("last", ledgerSeqRange.Last).
+					Info("Initializing in-memory store")
```
tamirms, line 384: is this comment still relevant? do we have any memory data structures at this point?

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -27,10 +27,10 @@ func (mlr *LedgerSeqRange) IsLedgerIncluded(ledgerSeq uint32) bool {
 }
 
 func (mlr *LedgerSeqRange) Merge(other *LedgerSeqRange) *LedgerSeqRange {
```
tamirms, line 29: I know this is a pre-existing issue but I don't think `other` should be a pointer because that gives the impression that `Merge()` will modify `other` but we don't actually want to mutate `other` in `Merge()`

tamirms: furthermore, I think `Merge()` should return a `LedgerSeqRange` instead of a pointer. all of the LedgerSeqRange functions can have value receivers except for `MergeInPlace()` which modifies the ledger range struct. But I think we can get rid of `MergeInPlace()` (I'll elaborate on that in a separate comment)

cmd/soroban-rpc/internal/db/migration.go
```diff
@@ -165,55 +181,69 @@ func (g *guardedMigration) Commit(ctx context.Context) error {
...
 	}, nil
 }
 
-func BuildMigrations(ctx context.Context, logger *log.Entry, db *DB, networkPassphrase string,
-	ledgerSeqRange *LedgerSeqRange,
-) (MultiMigration, error) {
+func BuildMigrations(
+	ctx context.Context, logger *log.Entry, db *DB, networkPassphrase string,
+	ledgerSeqRange LedgerSeqRange,
+) (MultiMigration, LedgerSeqRange, error) {
```
tamirms, line 202: I don't think it's necessary to return the applicableRange because MultiMigration already has a `ApplicableRange()` function which calculates the applicable range in the same way you are doing here. So, if we get rid of the applicableRange return value here we can also get rid of `MergeInPlace()` because that's the only place where it is used

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -328,32 +378,39 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
+				d.logger.
+					WithField("seq", currentSeq).
+					WithField("last", ledgerSeqRange.Last).
+					Debug("Still initializing in-memory store")
 			}
 
-			if err = feeWindows.IngestFees(txMeta); err != nil {
-				d.logger.WithError(err).Fatal("could not initialize fee stats")
+			if feeStatsRange.IsLedgerIncluded(currentSeq) { // skip irrelevant ledgers
+				if err = feeWindows.IngestFees(txMeta); err != nil {
```
tamirms, line 393: you could implement the `Migration` interface for `feeWindows` and then you can expose an `Append()` function on `MultiMigration` so that the `feeWindows` migration can be included in dataMigrations.  The benefit of refactoring the code this way is that you can get rid of the fee stats special case here because everything would be encapsulated in `dataMigrations.Apply()`.  That would also allow you to get rid of:

`ledgerSeqRange := feeStatsRange.Merge(&applicableRange)`

because the multi migration applicable range would include the fee stats range

tamirms:
@Shaptic the optimization is great! In summary, I have 3 pieces of feedback:

1. LedgerSeqRange should be an immutable type
2. There is no need for `db.BuildMigrations()` to return a LedgerSeqRange because it already has a ApplicableRange() function which can be used to obtain the applicable ledger range
3. We can implement a Migration instance for the fee stats and then append it to the MultiMigration returned by `db.BuildMigrations()` which would allow us to remove the special case handling for fee stats in daemon.go

cmd/soroban-rpc/internal/daemon/daemon.go
aditya1702: @tamirms we still have the fee stats in memory

cmd/soroban-rpc/internal/db/migration.go
Shaptic: Yeah, you're right! This was the way I originally approached it but I kept getting an infinite loop for some reason. I tried it again and it worked as intended in [a5bfc3b](https://github.com/stellar/soroban-rpc/pull/282/commits/a5bfc3b4d4270396a036850daea6226d421320ec) :+1:

cmd/soroban-rpc/internal/db/migration.go
Shaptic: Yeah I definitely agree that these should be value receivers. Fixed in [a5bfc3b](https://github.com/stellar/soroban-rpc/pull/282/commits/a5bfc3b4d4270396a036850daea6226d421320ec)!

cmd/soroban-rpc/internal/daemon/daemon.go
Shaptic: Actually that's not true since you just need the `db.Migration` interface. Done in dce46af, is that what you had in mind?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -299,27 +326,42 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
...
 	if err != nil {
-		d.logger.WithError(err).Fatal("could not build migrations")
+		d.logger.WithError(err).Fatal("could not get ledger range for fee stats")
 	}
 
-	// Apply migration for events, transactions and fee stats
+	// Additionally, by treating the fee window *as if* it's a migration, we can
+	// make the interface here really clean.
+	dataMigrations.Append(feeWindows.AsMigration(feeStatsRange))
+	ledgerSeqRange := dataMigrations.ApplicableRange()
```
tamirms, line 359: if ledgerSeqRange is empty we can return here. having said that, I don't think this condition will ever be met because the fee stats migration will always be run

tamirms (approved):
looks great!

cmd/soroban-rpc/internal/daemon/daemon.go
tamirms: @Shaptic yeah, that's exactly what I had in mind, thanks!

cmd/soroban-rpc/internal/daemon/daemon.go
Shaptic: True! Plus the stream will just be a no-op and return, anyway.

PR #291, opened 2024-09-12
Update `simulateTransaction` fields to properly omit fields when empty

Shaptic:
### What
Update response fields to only be omitted if the pointers are absent rather than if the `string`/`array` are absent.

### Why
With the introduction of JSON mode for responses (`xdrFormat: "json"`, https://github.com/stellar/soroban-rpc/pull/249), it became necessary to omit empty fields for responses that didn't previous have omission behavior.

This is because if someone wants JSON responses, it is misleading to see a non-empty `authJson: [ ... ]` array alongside an empty `auth: []` array (which would have been filled out in the base64 mode).

Unfortunately, this means if someone simulates a transaction that has no auth requirements (which, admittedly, should be rare because most invocations will at least have source account authorization), the field will be omitted. The [documentation indicates](https://developers.stellar.org/docs/data/rpc/api-reference/methods/simulateTransaction) that the `auth` field is required to be present, so this behavior is technically incorrect.

This updates the code to use pointers to control `omitempty` behavior rather than an empty structure (e.g., an empty array or empty string).

### Known limitations
Using pointers is generally dangerous since it can lead to panics from `nil` pointer dereferences. However, there is no cleaner way to selectively provide results for different fields like this while maintaining backwards compatibility.


leighmcculloch (approved):

leighmcculloch (approved):
Because this change fixes a break that was introduced into the last RPC release, can we fast track a release with the fix? It's blocking e2e system tests passing in other repos.

We should also consider running e2e tests with RPC too on PRs.

aditya1702:
> Because this change fixes a break that was introduced into the last RPC release, can we fast track a release with the fix? It's blocking e2e system tests passing in other repos.
> 
> We should also consider running e2e tests with RPC too on PRs.

@leighmcculloch Is the e2e tests you are referring to different from the one we run on RPC PRs - `soroban tools e2e`?

aditya1702 (approved):

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -345,9 +345,11 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 				})
 
 			default:
+				rv := base64.StdEncoding.EncodeToString(result.Result)
+				auth := base64EncodeSlice(result.Auth)
 				results = append(results, SimulateHostFunctionResult{
-					ReturnValueXDR: base64.StdEncoding.EncodeToString(result.Result),
-					AuthXDR:        base64EncodeSlice(result.Auth),
+					ReturnValueXDR: &rv,
+					AuthXDR:        &auth,
```
tamirms, line 352: Is there a code path where we create a `SimulateHostFunctionResult{}` instance with a nil `ReturnValueXDR` field and/or a nil `AuthXDR` field? Looking through this code, I could not find one.

Shaptic: Yes, in the `case FormatJSON` above this one, the `SimulateHostFunctionResult` only fills out the `ReturnValueJSON` and `AuthJSON` fields.

tamirms (approved):

PR #295, opened 2024-09-18
simulate-transaction: remove confusing Cost field in response

2opremio:
### What

Remove confusing Cost field in simulate-transaction's response.

### Why

Closes https://github.com/stellar/soroban-rpc/issues/66

### Known limitations

It's a breaking change, which should be released with protocol 22


Shaptic (approved):

janewang:
Thanks for cleaning up. It was confusing. Thanks for fixing 🙏

PR #296, opened 2024-09-19
Add `build-stellar-rpc` in makefile

psheth9:
We need this so that op can update package builder[ pipeline](https://github.com/stellar/pipelines/blob/master/soroban-rpc/Jenkinsfile-soroban-rpc-package-builder)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,10 @@ endif
 # (libpreflight.a is put at target/release-with-panic-unwind/ when not cross compiling)
 CARGO_BUILD_TARGET ?= $(shell rustc -vV | sed -n 's|host: ||p')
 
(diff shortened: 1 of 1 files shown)


Shaptic (approved):

Makefile
```diff
@@ -77,11 +81,19 @@ clean:
 	cargo clean
 	go clean ./...
 
+# DEPRECATED - please use build-stellar-rpc instead
```
Shaptic, line 84: Can you make sure we aren't referencing this anywhere if it's deprecated?

psheth9: it is soon to be deprecated actually, we will deprecate it fully after p22 just to reduce the chaos. Also can't do it now as we are still planning to issue 2 binaries [`soroban-rpc, stellar-rpc`]

PR #303, opened 2024-09-26
Add `getLedgers` implementation

### LINKED ISSUE #111: add getLedgers endpoint

aditya1702:
### What

Closes #111 

Adding new endpoint - `getLedgers`. It has the following request structure:

```
type GetLedgersRequest struct {
	StartLedger uint32                   `json:"startLedger"`
	Pagination  *LedgerPaginationOptions `json:"pagination,omitempty"`
	Format      string                   `json:"xdrFormat,omitempty"`
}
```

and the response structure:

```
// LedgerInfo represents a single ledger in the response.
type LedgerInfo struct {
	Hash            string `json:"hash"`
	Sequence        uint32 `json:"sequence"`
	LedgerCloseTime int64  `json:"ledgerCloseTime,string"`

	LedgerHeader     string          `json:"headerXdr"`
	LedgerHeaderJSON json.RawMessage `json:"headerJson,omitempty"`

	LedgerMetadata     string          `json:"metadataXdr"`
	LedgerMetadataJSON json.RawMessage `json:"metadataJSON,omitempty"`
}

// GetLedgersResponse encapsulates the response structure for getLedgers queries.
type GetLedgersResponse struct {
	Ledgers               []LedgerInfo `json:"ledgers"`
	LatestLedger          uint32       `json:"latestLedger"`
	LatestLedgerCloseTime int64        `json:"latestLedgerCloseTime"`
	OldestLedger          uint32       `json:"oldestLedger"`
	OldestLedgerCloseTime int64        `json:"oldestLedgerCloseTime"`
	Cursor                string       `json:"cursor"`
}
```

Pagination functionality is similar to `getTransactions` and `getEvents`.

### Why

This endpoint allows users to access ledger-level info more easily and enables common use-cases: examining ledger sequence numbers, time of close/finality, fees, or getting a snapshot of the ledger at that point in time. For more info refer to the [discussion](https://github.com/stellar/soroban-rpc/issues/111#issuecomment-2218331764) on the original issue ticket.

### Benchmarking Details

- `BatchGetLedgers`: Fetches ledgers in batches from the DB. Note that I have used a batch size of 200 for benchmarking as that is the max ledgers request limit we have set for `getLedgers` endpoint.
```
goos: darwin
goarch: arm64
BenchmarkBatchGetLedgers
BenchmarkBatchGetLedgers-12    	    3655	    337446 ns/op
```

- `GetLedgers`: The RPC endpoint to get ledgers.
```
goos: darwin
goarch: arm64
BenchmarkGetLedgers
BenchmarkGetLedgers-12    	    1225	    973161 ns/op
```


cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -103,6 +104,25 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
...
+	batchSize uint,
+) ([]xdr.LedgerCloseMeta, error) {
+	sql := sq.Select("meta").
+		From(ledgerCloseMetaTableName).
+		Where(sq.And{
+			sq.GtOrEq{"sequence": sequence},
+			sq.LtOrEq{"sequence": sequence + uint32(batchSize) - 1},
+		})
+
+	var results []xdr.LedgerCloseMeta
```
Shaptic, line 118: prealloc here for sure

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	return IsValidFormat(req.Format)
+}
+
+// LedgerInfo represents a single ledger in the response.
+type LedgerInfo struct {
+	Hash            string `json:"hash"`
+	Sequence        uint32 `json:"sequence"`
+	LedgerCloseTime int64  `json:"ledgerCloseTime"`
```
Shaptic, line 58: don't forget 64-bit js quirks!
```suggestion
	LedgerCloseTime int64  `json:"ledgerCloseTime,string"`
```

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,251 @@
...
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+
+	ledgers, err := h.fetchLedgers(ctx, start, limit, request.Format)
+	if err != nil {
+		return GetLedgersResponse{}, err
+	}
+	cursor := strconv.FormatUint(uint64(ledgers[len(ledgers)-1].Sequence), 10)
```
Shaptic, line 121: keep it simple :laughing: 
```suggestion
	cursor := fmt.Sprintf("%d", ledgers[len(ledgers)-1].Sequence)
```

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,251 @@
...
+
+	headerB, err := ledger.LedgerHeaderHistoryEntry().MarshalBinary()
+	if err != nil {
+		return LedgerInfo{}, fmt.Errorf("error marshaling ledger header: %w", err)
+	}
+
+	// Format the data according to the requested format (JSON or XDR)
+	switch format {
+	case FormatJSON:
+		closeMetaJSON, headerJSON, convErr := ledgerToJSON(closeMetaB, headerB)
```
Shaptic, line 239: You can actually get away with not marshaling ahead of time in the JSON case by using `xdr2json.ConvertInterface`, instead and just passing `ledger` and `ledger.LedgerHeaderHistoryEntry()` directly.

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,251 @@
...
+		if convErr != nil {
+			return LedgerInfo{}, fmt.Errorf("could not convert ledger metadata and "+
+				"header to JSON: %w", convErr)
```
Shaptic, line 242: the default error is very verbose so you can just return it directly tbh

cmd/soroban-rpc/internal/db/ledger.go
aditya1702: Ohh yeah I already had this change done locally. That is why I had the WIP label still :p

cmd/soroban-rpc/internal/methods/get_ledgers.go
aditya1702: @Shaptic I remembered why I used strconv.FormatUint. Its much faster than fmt.Sprintf - https://medium.com/swlh/bad-go-frivolous-sprintf-2ad28fedf1a0

cmd/soroban-rpc/internal/methods/get_ledgers.go
aditya1702: Oooh thats nice! Thx

aditya1702: @Shaptic On another look, I still need to marshal them to encode them for non-JSON response.
```
ledgerInfo.LedgerCloseMeta = base64.StdEncoding.EncodeToString(closeMetaB)
ledgerInfo.LedgerHeader = base64.StdEncoding.EncodeToString(headerB)
```
So I can just keep it

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/creachadair/jrpc2"
+
+	"github.com/stellar/go/xdr"
+
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/db"
+	"github.com/stellar/soroban-rpc/cmd/soroban-rpc/internal/ledgerbucketwindow"
+)
+
+type LedgerPaginationOptions struct {
```
aditya1702, line 19: Note for reviewers: We have the same exact pagination struct for transactions - https://github.com/stellar/soroban-rpc/blob/main/cmd/soroban-rpc/internal/methods/get_transactions.go#L25. So, we can remove the duplication and just use a single `PaginationOptions` struct for both. However, that adds a lot of changes to this PR which are not related, so my suggestion is to do it in a separate PR.

cmd/soroban-rpc/internal/methods/json.go
```diff
@@ -58,3 +58,20 @@ func transactionToJSON(tx db.Transaction) (
...
+		return nil, nil, err
+	}
+
+	headerJSON, err = xdr2json.ConvertInterface(header)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return closeMetaJSON, headerJSON, nil
+}
```
Shaptic, line 77: A few comments here:

1. These can be quite large, it might be worth passing a pointer to avoid the stack copy for every ledger?
2. The header is a subset of the meta, so you shouldn't need to pass that in. 
3. If you don't pass it in, you can either (a) convert it separately like you're doing here or (b) actually extract it from the `closeMetaJSON` and return that, instead. The latter is probably more efficient but it's error-prone since it relies on the JSON structure not changing.

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,243 @@
...
+	switch format {
+	case FormatJSON:
+		closeMetaJSON, headerJSON, convErr := ledgerToJSON(ledger, ledgerHeader)
+		if convErr != nil {
+			return LedgerInfo{}, convErr
+		}
+		ledgerInfo.LedgerCloseMetaJSON = closeMetaJSON
+		ledgerInfo.LedgerHeaderJSON = headerJSON
+	default:
+		closeMetaB, err := ledger.MarshalBinary()
```
Shaptic, line 229: Isn't there a `MarshalBinaryBase64`?

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Format the data according to the requested format (JSON or XDR)
+	switch format {
+	case FormatJSON:
+		closeMetaJSON, headerJSON, convErr := ledgerToJSON(ledger, ledgerHeader)
+		if convErr != nil {
+			return LedgerInfo{}, convErr
+		}
+		ledgerInfo.LedgerCloseMetaJSON = closeMetaJSON
+		ledgerInfo.LedgerHeaderJSON = headerJSON
```
Shaptic, line 227:
```suggestion
		var convErr error
		ledgerInfo.LedgerCloseMetaJSON, ledgerInfo.LedgerHeaderJSON, convErr = ledgerToJSON(ledger, ledgerHeader)
		if convErr != nil {
			return ledgerInfo, convErr
		}
```

Shaptic:
Based on [this thread](https://stellarfoundation.slack.com/archives/CGY4VS8SZ/p1727118439438509), should we rename `LedgerCloseMeta` -> `LedgerMetadata`?

Shaptic:
Should we cherry-pick this onto the `protocol22` branch? I don't think it makes sense to target main given how close that is to merging upstream.

cmd/soroban-rpc/internal/methods/get_ledgers.go
aditya1702: @Shaptic That method is part of `xdr.LedgerKey` and not the ledger metadata xdr

aditya1702:
> Based on [this thread](https://stellarfoundation.slack.com/archives/CGY4VS8SZ/p1727118439438509), should we rename `LedgerCloseMeta` -> `LedgerMetadata`?

@Shaptic Yeah that is a good idea. Will make the change

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,241 @@
...
+		return GetLedgersResponse{}, &jrpc2.Error{
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+
+	ledgers, err := h.fetchLedgers(ctx, start, limit, request.Format)
+	if err != nil {
+		return GetLedgersResponse{}, err
+	}
```
2opremio, line 124: `GetLedgerRange` and `fetchLedgers` should be run within the same transaction in order to avoid races (e.g. AFAIU a new ledger could be ingested between `GetLedgerRange` and `fetchLedgers`). I don't know if we are doing the same in other endpoints, but we should. If not, please create a ticket for it.

2opremio:
@aditya1702 The PR looks really good! We should resolve the problem of the response ledger range potentially being inconsistent with the response.

cmd/soroban-rpc/internal/methods/get_ledgers.go
2opremio: BTW, I am happy to be proven wrong if you think that the race I am mentioning cannot happen.

tamirms: getEvents addressed this concern without a transaction by reducing `limit` in the case that `start + limit` is greater than `ledgerRange.LastLedger.Sequence`

2opremio: Fair enough, then we should do that here too

aditya1702: @2opremio Yeah nice catch. I think @tamirms suggestion would work here. Will add it and also add a test for it

aditya1702: Actually I just discussed this with @tamirms and just validating the `start + limit` value is not enough. It would still lead to a race in the following scenario:
1. `start` = `oldestLedger` when the `GetLedgerRange` is called.
2. Before `fetchLedgers` is called, a new ledger is ingested and `start` is trimmed off.
3. `fetchLedgers` for the `start` ledger will fail since it is not present in the db.

So I think a common transaction approach will need to be implemented here. Looks like `getEvents` also suffers from the same problem 

cc @2opremio @Shaptic

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -23,20 +25,70 @@ type LedgerReader interface {
...
+	GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error)
+	BatchGetLedgers(ctx context.Context, sequence uint32, batchSize uint) ([]xdr.LedgerCloseMeta, error)
+	Done() error
 }
 
 type LedgerWriter interface {
 	InsertLedger(ledger xdr.LedgerCloseMeta) error
 }
 
+type ReadDB interface {
```
tamirms, line 41: this interface does not need to be exported. I would make it private because there is no need for external packges to depend on this.

cmd/soroban-rpc/internal/db/ledger.go
```diff
@@ -23,20 +25,70 @@ type LedgerReader interface {
 	StreamAllLedgers(ctx context.Context, f StreamLedgerFn) error
```
tamirms, line 25: is `BatchGetLedgers()` ever called outside of `LedgerReaderTx`? If not then it seems like you can remove `LedgerReaderTx()` from the `LedgerReader` interface

aditya1702: @tamirms Right now its not. I was going to remove it but thought that it could be used in other places in future without the need of a tx so kept it in `LedgerReader` too

tamirms: I think it's better to remove unused code and add it back when it becomes necessary

cmd/soroban-rpc/internal/methods/get_ledgers.go
```diff
@@ -0,0 +1,252 @@
...
+			Code:    jrpc2.InvalidParams,
+			Message: err.Error(),
+		}
+	}
+
+	ledgers, err := h.fetchLedgers(ctx, start, limit, request.Format, readTx)
+	if err != nil {
+		return GetLedgersResponse{}, err
+	}
+	cursor := strconv.Itoa(int(ledgers[len(ledgers)-1].Sequence))
```
tamirms, line 136: is it possible for ledgers to be empty ? Consider the case where you query the latest ledgers and use the cursor to query the next page. In that scenario, you have exhausted all the ledgers

tamirms (approved):

PR #316, opened 2024-10-11
protocol22: Bump Rust dependencies

2opremio:
Unfortunately I have had to disable the dependecy-check script since, after [adding multi-protocol support](https://github.com/stellar/soroban-rpc/pull/264), it completely falls apart when using RC versions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,32 +7,30 @@ members = [
 ]
 
 [workspace.package]
-rust-version = "1.74.0"
+rust-version = "1.81.0"
 
 [workspace.dependencies.soroban-env-host-prev]
 package = "soroban-env-host"
-version = "=21.2.0"
+version = "=21.2.1"
 
 [workspace.dependencies.soroban-env-host-curr]
 package = "soroban-env-host"
-version = "=22.0.0"
-git = "https://github.com/stellar/rs-soroban-env"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -7,32 +7,30 @@ members = [
 ]
 
 [workspace.package]
-rust-version = "1.74.0"
+rust-version = "1.81.0"
```
Shaptic, line 10: The Core team is still using 1.74 because of some WASM incompatibilities I guess (e.g. [thread](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1727811768128959?thread_ts=1727760416.953479&cid=C030Z9EHVQE)), so we should keep the same one here too to be extra safe.

Shaptic (approved):

Cargo.toml
2opremio: env is pointing at 1.79: https://github.com/stellar/rs-soroban-env/blob/cf469d2d4be82f83c24b75aeb8a133cfb82a23cd/Cargo.toml#L24

willemneal:
Can confirm that locally when using this PR with quickstart the CLI is passing tests and deploying and constructing contracts!

2opremio:
Awesome!!

2opremio:
I will make an official release on Monday

PR #333, opened 2024-11-26
Fix e2e tests after soroban-rpc -> stellar-rpc rename

2opremio:
Also, fix a few remaining references to soroban-rpc

### DIFF
--- a/cmd/stellar-rpc/internal/config/options.go
+++ b/cmd/stellar-rpc/internal/config/options.go
@@ -33,15 +33,17 @@ func (cfg *Config) options() Options {
 	defaultStellarCoreBinaryPath, _ := exec.LookPath("stellar-core")
 	cfg.optionsCache = &Options{
 		{
-			Name:      "config-path",
-			EnvVar:    "STELLAR_RPC_CONFIG_PATH",
+			Name: "config-path",
+			// TODO: deprecate and rename to STELLAR_RPC_
+			EnvVar:    "SOROBAN_RPC_CONFIG_PATH",
 			TomlKey:   "-",
 			Usage:     "File path to the toml configuration file",
 			ConfigKey: &cfg.ConfigPath,
 		},
 		{
-			Name:         "config-strict",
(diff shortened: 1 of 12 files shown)


cmd/stellar-rpc/internal/config/options.go
```diff
@@ -200,7 +200,7 @@ func (cfg *Config) options() Options {
 			Name:         "db-path",
 			Usage:        "SQLite DB path",
 			ConfigKey:    &cfg.SQLiteDBPath,
-			DefaultValue: "soroban_rpc.sqlite",
+			DefaultValue: "stellar_rpc.sqlite",
```
2opremio, line 203: I am not sure we should be changing this just yet

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -34,14 +34,14 @@ func (cfg *Config) options() Options {
...
-			EnvVar:    "SOROBAN_RPC_CONFIG_PATH",
+			EnvVar:    "STELLAR_RPC_CONFIG_PATH",
 			TomlKey:   "-",
 			Usage:     "File path to the toml configuration file",
 			ConfigKey: &cfg.ConfigPath,
 		},
 		{
 			Name:         "config-strict",
-			EnvVar:       "SOROBAN_RPC_CONFIG_STRICT",
+			EnvVar:       "STELLAR_RPC_CONFIG_STRICT",
```
2opremio, line 44: We probably shouldn't be changing these options either :S

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	prometheusNamespace          = "soroban_rpc"
+	prometheusNamespace          = "stellar_rpc"
```
2opremio, line 38: And ... maybe we shouldn't change the prometheus namespace either

tamirms (approved):

PR #337, opened 2024-12-12
Add additional details to getLatestLedger response

### LINKED ISSUE #79: Add more ledger details to getLedger/getLatestLedger
The getLatestLedger endpoint should return all information about the ledger, such as the ledger header, and any other details the rpc knows about the ledger.

Currently it only reports the following:
```json
❯ http POST :8000/soroban/rpc jsonrpc=2.0 id=0 method=getLatestLedger
HTTP/1.1 200 OK
Accept-Post: application/json
Connection: keep-alive
Content-Length: 148
Content-Type: application/json
Date: Sat, 24 Feb 2024 00:03:04 GMT
Server: nginx/1.18.0 (Ubuntu)
Vary: Origin
Vary: Origin

{
    "id": "0",
    "jsonrpc": "2.0",
    "result": {
        "id": "171f5f42a7375f17a10b9633c063ec35291a6253e9f309d13d31a55933f537ed",
        "protocolVersion": 20,
        "sequence": 283919
    }
}
```

Related to:
- https://github.com/stellar/soroban-rpc/issues/78

aditya1702:
### What

Closes #79 

- Add additional information about a ledger to the `GetLatestLedger` response.
- Unittests

### Why

The current info lacks basic ledger information. This change brings it in sync with the information returned by `GetLedgers` endpoint - https://github.com/stellar/stellar-rpc/pull/303#issue-2551105711

### Known limitations

NA

### DIFF
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -25,7 +25,9 @@ type latestLedgerHandler struct {
 	ledgerReader      db.LedgerReader
 }
 
-func (h latestLedgerHandler) getLatestLedger(ctx context.Context, request GetLatestLedgerRequest) (GetLatestLedgerResponse, error) {
+func (h latestLedgerHandler) getLatestLedger(ctx context.Context,
+	request GetLatestLedgerRequest,
+) (GetLatestLedgerResponse, error) {
 	latestSequence, err := h.ledgerEntryReader.GetLatestLedgerSequence(ctx)
 	if err != nil {
 		return GetLatestLedgerResponse{}, &jrpc2.Error{
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -2,45 +2,83 @@ package methods
 
 import (
 	"context"
+	"encoding/base64"
+	"fmt"
 
 	"github.com/creachadair/jrpc2"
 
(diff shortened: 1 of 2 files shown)


2opremio:
I am not sure why this is needed. You can always obtain the latest ledger and query getLedger with the provided sequence.

I understand it saves one API call (which can be an advantage), but it also complicates the API. Is there a usecase justiying this?

If we really want to do this, can we at least try to factor out the obtention of `LedgerInfo` ? (maybe it ends up complicating the code further, but we should at least consider it)

leighmcculloch:
> I am not sure why this is needed. You can always obtain the latest ledger and query getLedger with the obtained sequence.

The `getLatestLedger` and `getLedger` both impl they'll return a "ledger", and yet both return different sets of data. The better developer experience would be to return the same resource. I understand this is an RPC API and not a RESTful API oriented around resources, but I think the concepts are still useful.

leighmcculloch:
> Is there a usecase justiying this?

There are no critical use cases, as you say you can make two calls, it's just inconvenient, or makes for an inconsistent developer experience. 

There are potential issues if operators horizontally scale RPC and return inconsistent results, such as latest refers to a ledger seq that can't be retrieved, but that's just theoretical, I'm not aware of this occurring.

aditya1702:
Based on the discussion we had and @leighmcculloch 's comments, it looks like we dont actually have a strong reason for doing this change right now. Closing this for now but we can always revisit in the future.

leighmcculloch:
> The better developer experience would be to return the same resource.

I think this ☝🏻 is a strong reason. 

> Based on the discussion we had

Could you link to the discussion?

PR #338, opened 2024-12-13
Use single DB transaction for fetching ledger range and ledgers in `getTransactions`

### LINKED ISSUE #334: Use a single DB transaction within `getTransactions`
Currently, both these methods involve a call to `GetLedgerRange` to get the current RPC range and then subsequent calls to the DB to get different ledger close meta. This introduces a race condition where ingestion can happen between these 2 calls and change the ledger range snapshot in the db.

To mitigate this, we want to use a single DB transaction throughout the entire lifecycle of the request.

aditya1702:
### What

Closes #334 

### Why

We want to get the same view of the db and avoid any race conditions when accessing without a db txn.

### Known limitations

NA

### DIFF
--- a/cmd/stellar-rpc/internal/db/ledger.go
+++ b/cmd/stellar-rpc/internal/db/ledger.go
@@ -28,6 +28,7 @@ type LedgerReader interface {
 }
 
 type LedgerReaderTx interface {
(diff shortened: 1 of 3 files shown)


aditya1702:
> looks good to me but we should investigate the integration test failures

@tamirms The integration tests and e2e test failures are being investigated by @2opremio here - https://github.com/stellar/stellar-rpc/pull/339

PR #343, opened 2024-12-18
Improve bound and cache update checks

2opremio:


cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -142,6 +128,24 @@ func (g *GetEventsRequest) Valid(maxLimit uint) error {
...
 
+	if g.Pagination != nil {
+		if g.Pagination.Cursor != nil && (g.StartLedger != 0 || g.EndLedger != 0) {
+			return errors.New("ledger ranges and cursor cannot both be set")
+		}
+		if g.Pagination.Limit > maxLimit {
+			return fmt.Errorf("limit must not exceed %d", maxLimit)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
2opremio: Good point. I changed it. PTAL

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -142,6 +128,23 @@ func (g *GetEventsRequest) Valid(maxLimit uint) error {
...
 	}
 
+	if g.Pagination != nil { //nolint:nestif
+		if g.Pagination.Cursor != nil && (g.StartLedger != 0 || g.EndLedger != 0) {
+			return errors.New("ledger ranges and cursor cannot both be set")
+		}
+		if g.Pagination.Limit > maxLimit {
+			return fmt.Errorf("limit must not exceed %d", maxLimit)
+		}
+	} else {
```
Shaptic, line 138: I think this misses the case where there is a `limit` set but also `startLedger`/`endLedger`, which is why that check was separate in the previous implementation.

cmd/stellar-rpc/internal/db/cursor.go
```diff
@@ -130,8 +130,8 @@ var (
 	//nolint:gochecknoglobals
 	MaxCursor = Cursor{
```
Shaptic, line 131: Way more reliable, nice

Shaptic (commented):
Something is up with event request validation, I think

cmd/stellar-rpc/internal/methods/get_events.go
2opremio: Then, removing the else should be enough. I will do that

2opremio: Done. PTAL!

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -142,6 +128,21 @@ func (g *GetEventsRequest) Valid(maxLimit uint) error {
...
+		if g.Pagination.Cursor != nil && (g.StartLedger != 0 || g.EndLedger != 0) {
+			return errors.New("ledger ranges and cursor cannot both be set")
+		}
+		if g.Pagination.Limit > maxLimit {
+			return fmt.Errorf("limit must not exceed %d", maxLimit)
+		}
+	}
+	if g.StartLedger <= 0 {
+		return errors.New("startLedger must be positive")
+	}
```
Shaptic, line 141: Now this will fail whenever `pagination` is set by itself, since `startLedger == 0` in that case, right? lol. This tells me we need better unit testing on this validation path.

2opremio: I have restored the original validation logic. I have actually realized that not setting the end of the range means _unlimited_, so we shouldn't be validating that end >= start (since end could be 0).

PR #347, opened 2025-01-16
Fix Misuse of ServerConfig.PublicKeyCallback may cause authorization bypass

lamcodeofpwnosec:
Applications and libraries which misuse the ServerConfig.PublicKeyCallback callback may be susceptible to an authorization bypass. ServerConfig.PublicKeyCallback says that "A call to this function does not guarantee that the key offered is in fact used to authenticate." Specifically, the SSH protocol allows clients to inquire about whether a public key is acceptable before proving control of the corresponding private key. PublicKeyCallback may be called with multiple keys, and the order in which the keys were provided cannot be used to infer which key the client successfully authenticated with, if any. Some applications, which store the key(s) passed to PublicKeyCallback (or derived information) and make security relevant determinations based on it once the connection is established, may make incorrect assumptions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,14 +86,14 @@ require (
 	go.opentelemetry.io/otel/metric v1.28.0 // indirect
 	go.opentelemetry.io/otel/trace v1.28.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/crypto v0.24.0 // indirect
+	golang.org/x/crypto v0.31.0 // indirect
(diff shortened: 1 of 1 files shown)


paliwangtel (approved):

Shaptic (commented):
This seems to be a copy-paste of [CVE-2024-45337](https://www.cvedetails.com/cve/CVE-2024-45337/)? This does not apply to this repository, as it only uses `crypto/sha256` for hashing which is entirely unaffected by this CVE.

Issue #269, opened 2024-08-13
Replace `getLedgerEntries()` with a proxy to Core's `/getledgerentry`.


Shaptic:
# Protocol 23: Replacing Endpoints

**Epic**: https://github.com/stellar/go/issues/5433
**References**: Platform's issues for Protocol 23: [projects/37#31](https://github.com/orgs/stellar/projects/37/views/32)

---------------------------

Stellar Core will have support a high-performance HTTP endpoint for retrieving ledger entries as part of their v22 in stellar/stellar-core#4350. Once stellar/go#5426 is complete, we can completely remove the implementation of getLedgerEntries and its database backing and replace it with a proxy to the Core client. 

### Acceptance Criteria 
The database tables are dropped.
The implementation is a proxy to Core.
The returned JSON schema stays the same.
The performance is still acceptable.

2opremio:
@Shaptic did you start working on this? Otherwise I will take it

Shaptic:
@2opremio this is blocked on Core actually providing this endpoint, but you can take it up if you want to assume the existence of Core's endpoint. I'm working on client support in https://github.com/stellar/go/pull/5428 but again, no real Core yet.

Also note that we will still need entry ingestion + db snapshotting until P23.

Shaptic:
Closed by https://github.com/stellar/stellar-rpc/pull/353

PR #353, opened 2025-01-28
`getLedgerEntries`: optionally use high-performance Core server

Shaptic:
### What
This introduces `STELLAR_CAPTIVE_CORE_HTTP_QUERY_PORT`, enabling a high-performance HTTP server on the Captive Core subprocess for fetching ledger entries.

This new endpoint is plumbed all the way through the `getLedgerEntries` handler, removing those database queries entirely.

This also updates the `stellar/go` monorepo reference and its cascading dependencies.

If `STELLAR_CAPTIVE_CORE_HTTP_QUERY_PORT` is set to 0 (the default value) RPC will use it's old and proven DB implementation of `getLedgerEntries` instead of querying Core.

### Why

With [CAP-66](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0066.md#getledgerentry-endpoint) arriving in Protocol 23 and its high-performance `/getledgerentry` endpoint (whose client was implemented in https://github.com/stellar/go/pull/5542), we don't need to store ledger entries in RPC anymore.

Today, we still need to provide snapshots to the simulation library, but tomorrow (in the opaque sense) we won't need to, so this is part of the transition work.

Related: #269.

### Known limitations
TODO:
- [x] Ensure that the correct captive core configuration is generated to enable the server, per details in https://github.com/stellar/stellar-core/pull/4623. 
- [x] Integration testing


Shaptic:
@2opremio this should probably target a new feature branch e.g. `protocol-23` :pray:

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -62,6 +61,7 @@ func (cfg *Config) options() Options {
 			Usage:     "Admin endpoint to listen and serve on. WARNING: this should not be accessible from the Internet and does not use TLS. \"\" (default) disables the admin server",
 			ConfigKey: &cfg.AdminEndpoint,
 		},
+		// TODO: should be gave a similar stellar-core-query-url parameter or should we assume that all queries will happen to the local captive core?
```
Shaptic, line 64: You can't enable it on a validator and we def don't want another core running, so yeah it should be on the same captive core :+1:

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -122,6 +122,16 @@ func (d *Daemon) Close() error {
 
 // newCaptiveCore creates a new captive core backend instance and returns it.
 func newCaptiveCore(cfg *config.Config, logger *supportlog.Entry) (*ledgerbackend.CaptiveStellarCore, error) {
+	var queryServerParams *ledgerbackend.HTTPQueryServerParams
+	if cfg.CaptiveCoreHTTPPort != 0 {
```
Shaptic, line 126: wrong one, should be `CaptiveCoreHTTPQueryPort`

2opremio: I wish I had read this before debugging it :)

cmd/stellar-rpc/internal/config/options.go
2opremio: I didn't know it couldn't be enabled it on validators.

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -84,7 +85,25 @@ func (cfg *Config) options() Options {
...
+			Name:         "stellar-captive-core-http-query-port",
+			Usage:        "HTTP port for Captive Core to listen on for high-performance queries like /getledgerentry (0 disables the HTTP server, must not conflict with CAPTIVE_CORE_HTTP_PORT)",
+			ConfigKey:    &cfg.CaptiveCoreHTTPQueryPort,
+			DefaultValue: uint16(0), // Disabled by default, although it normally uses 11628
+		},
+		{
+			Name:         "stellar-captive-core-http-query-thread-pool-size",
+			Usage:        "Number of threads to use by Captive Core's high-performance query server",
+			ConfigKey:    &cfg.CaptiveCoreHTTPQueryThreadPoolSize,
+			DefaultValue: uint16(runtime.NumCPU()), //nolint:gosec
```
Shaptic, line 100: Is this a safe default? We want to leave CPUs for RPC itself.. possible suggestion,

```suggestion
			DefaultValue: uint16(min(1, runtime.NumCPU() - 2)), //nolint:gosec
```

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -84,7 +85,25 @@ func (cfg *Config) options() Options {
...
+		},
+		{
+			Name:         "stellar-captive-core-http-query-snapshot-ledgers",
+			Usage:        "Size of ledger history in Captive Core's high-performance query server (don't touch unless you know what you are doing)",
+			ConfigKey:    &cfg.CaptiveCoreHTTPQuerySnapshotLedgers,
+			DefaultValue: uint16(4),
```
Shaptic, line 106: @SirTyson 0 is a better default, right?

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -235,6 +250,17 @@ func createStellarCoreClient(cfg *config.Config) stellarcore.Client {
 	}
 }
 
+func createHighperfStellarCoreClient(cfg *config.Config) interfaces.FastCoreClient {
+	// It doesn't make sense to create a client if the local server is not enabled
+	if cfg.CaptiveCoreHTTPQueryPort == 0 {
+		return nil
+	}
+	return &stellarcore.Client{
+		URL:  fmt.Sprintf("http://localhost:%d", cfg.CaptiveCoreHTTPQueryPort),
```
Shaptic, line 259: This should use `cfg.StellarCoreURL` over localhost, right?

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -70,15 +72,22 @@ type TestConfig struct {
...
 	CoreHTTPPort    uint16
 	CoreArchivePort uint16
-	// This only needs to be an unconflicting port
-	captiveCorePort uint16
+
+	// These only need to be unconflicting ports
+	captiveCorePort          uint16
+	captiveCoreHTTPQueryPort uint16
+	// TODO: this is only needed due to a quirk (bug) in the implementation of the
+	//       core's query server, which requires the http port to be set to non-zero for the query server to be spawned.
```
Shaptic, line 88: Just to clarify, this means the "regular" HTTP server **must** be enabled for the "query" server to be enable-able, right?

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -136,6 +146,11 @@ func NewTest(t *testing.T, cfg *TestConfig) *Test {
 			shouldWaitForRPC = !cfg.OnlyRPC.DontWait
 		}
 		parallel = !cfg.NoParallel
+		i.enableCoreHTTPQueryServer = cfg.EnableCoreHTTPQueryServer
+	}
+
+	if i.enableCoreHTTPQueryServer && GetCoreMaxSupportedProtocol() < 22 {
```
Shaptic, line 152: We should change this to 23 once there's a testing build of 23

cmd/stellar-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -19,8 +22,112 @@ var ErrLedgerTTLEntriesCannotBeQueriedDirectly = "ledger ttl entries cannot be q
...
+) ([]db.LedgerKeyAndEntry, uint32, error) {
+	latestLedger, err := c.latestLedgerReader.GetLatestLedgerSequence(ctx)
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not get latest ledger: %w", err)
+	}
+	// Pass latest ledger here in case Core is ahead of us (0 would be Core's latest).
+	resp, err := c.coreClient.GetLedgerEntries(ctx, latestLedger, keys...)
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not query captive core: %w", err)
+	}
```
Shaptic, line 46: @SirTyson just wanna double-check that this logic makes sense: we want Core to use the latest ledger that RPC is aware of

cmd/stellar-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -146,3 +201,82 @@ func NewGetLedgerEntriesHandler(logger *log.Entry, ledgerEntryReader db.LedgerEn
 		return response, nil
 	})
 }
+
+type keyEntriesAndOrdering struct {
```
Shaptic, line 205: Is this necessary because the query server doesn't return the keys in request order?

Shaptic (commented):
Looks really solid! Coupla questions but otherwise pretty sure this can get a :heavy_check_mark:

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
2opremio: Yep!

cmd/stellar-rpc/internal/methods/get_ledger_entries.go
2opremio: Yep, in fact, I don’t think there are any guarantees on the DB either.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
2opremio: Yes. Feel free to rephrase it!

cmd/stellar-rpc/internal/daemon/daemon.go
2opremio: Not really. Because:
1. The url includes the port
2. The user may choose a different host for the normal (non-querry) http requests

cmd/stellar-rpc/internal/config/options.go
2opremio: I checked with him. It needs to be as high as the number of ledgers captive core can be ahead of RPC. 4 seemed like a generous value.

cmd/stellar-rpc/internal/config/options.go
2opremio: Uhm. What if there is a single or 2 CPUs. As far as I understand, the query server is IO bound, so the threads will be blocked waiting for disk most of the time.

Shaptic: Sorry should've been `max` not `min` in the diff :facepalm: - so at least 2, but otherwise `CpuCount - 2`. 

Good point re: disk-bound, but not true for live state (live state will be wholly in memory). So I do think we should certainly leave a CPU completely free for RPC itself. Maybe for the rest of Captive Core, as well? A big `getLedgerEntries` stalling ingestion by hogging CPU would not be good.

@SirTyson opinions here?

cmd/stellar-rpc/internal/config/options.go
Shaptic: Good call, thanks :+1:

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
Shaptic: No need, it's clear but I just wanted to make sure I understood! :+1:

cmd/stellar-rpc/internal/methods/get_ledger_entries.go
Shaptic: Oof, subtle. Nice job :)

2opremio (approved):

cmd/stellar-rpc/internal/config/options.go
SirTyson: I highly suspect this endpoint to be IO bound, even with in-memory state. Tbh I haven't tested with a ton of cores yet, that sort of config setting is probably something you all are better equipped to do, but I'd assume these are IO threads and not compute threads.

cmd/stellar-rpc/internal/methods/get_ledger_entries.go
SirTyson: Yep, looks good!

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -70,15 +72,22 @@ type TestConfig struct {
...
 
 type TestCorePorts struct {
 	CorePort        uint16
 	CoreHTTPPort    uint16
 	CoreArchivePort uint16
-	// This only needs to be an unconflicting port
-	captiveCorePort uint16
+
+	// These only need to be unconflicting ports
+	captiveCorePort          uint16
```
tamirms, line 85: @2opremio @Shaptic do you still need this `captiveCorePort` field given that we have introduced `captiveCoreHTTPPort`? It seems like `captiveCorePort` and `captiveCoreHTTPPort` are both trying to configure the same port

Issue #344, opened 2024-12-23
Allow users to enable non-root auth in simulation


dmkozh:
### What problem does your feature solve?

Currently simulation by default disallows 'non-root' authorization in recording mode, i.e. disallows potentially 'frontrunnable' authorizations. This is a very sensible default setting as most of the time it's correct to make all the calls non-frontrunnable. That said, there is still a range of cases where non-root auth is intentional and thus we need to provide users with a way of enabling this.

### What would you like to see?

There should be some request parameter that enables non-root auth. It's fine for this to be not easily discoverable or accessible as this is meant for the advanced users. But there needs to be at least _some_ way of enabling this.

### What alternatives are there?

Not providing the users with an option. This basically makes auth simulation unusable for a fraction of the users, which is not desirable.

dmkozh:
This requires an env release as well. I'll work on this and assign this to myself for the time being. It would be great if someone else could help with the RPC interface part when the env change is released.

dmkozh:
The env-side change has been released in 22.1.3 release (https://github.com/stellar/rs-soroban-env/pull/1503) and it can be now implemented in soroban-rpc at any time now. 

It had to be guarded by `unstable-next-api` feature in order to pass the semver checks, which should be fine to enable in soroban-rpc as we'll be aware of the dependency and will make sure to not accidentally break it (should there be any more API-breaking features before v23 release).

The alternative is to wait until the stable v23 release, though it might be a good idea to at least prepare the change before that.

Shaptic:
@leighmcculloch @dmkozh is there a way to programmatically determine whether or not non-root authorization is present from the authorization entries themselves? My gut says "no" but I just want to make sure - it would help us avoid introducing + propagating an additional request parameter. My gut also says "this should be opt-in" since per Dima this only applies to a small subset of simulations.

Otherwise, I'm proposing the following:

```typescript
interface SimulateTransactionRequest { 
   // ...
  authMode?: "enforce" | "record" | "nonroot"
}
```

where

* **omission** (default): enforcement if auth entries are included in the invocation operation, root recording otherwise (the behavior today)
* **enforce**: enforcement always, with an empty list if auth entries are omitted
* **record**: root record always, ignoring the list if auth entries are included
* **nonroot**: record with non-root authorization, ignoring the list if auth entries are included (naming? could be `record-nonroot`)

Thoughts? Is this overkill? A simpler `nonroot?: boolean` would be a direct parallel to the simulation library.

[...]

PR #356, opened 2025-01-30
Upgrade Soroban env to latest patch

Shaptic:
### What
Bump dependencies to their latest versions and run `cargo update`.

### Why
Self-explanatory.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ rust-version = "1.79.0"
 
 [workspace.dependencies.soroban-env-host-prev]
 package = "soroban-env-host"
-version = "=21.2.1"
+version = "=21.2.2"
 
 [workspace.dependencies.soroban-env-host-curr]
 package = "soroban-env-host"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ rust-version = "1.79.0"
 
 [workspace.dependencies.soroban-env-host-prev]
 package = "soroban-env-host"
-version = "=21.2.2"
+version = "=21.2.1"
 
 [workspace.dependencies.soroban-env-host-curr]
 package = "soroban-env-host"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ version = "=21.2.1"
 
 [workspace.dependencies.soroban-env-host-curr]
 package = "soroban-env-host"
-version = "=22.0.0"
+version = "=22.1.3"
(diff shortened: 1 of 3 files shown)


dmkozh (approved):

Cargo.toml
```diff
@@ -11,11 +11,11 @@ rust-version = "1.79.0"
 
 [workspace.dependencies.soroban-env-host-prev]
 package = "soroban-env-host"
-version = "=21.2.1"
+version = "=21.2.2"
```
dmkozh, line 14: CI seems to be complaining about this change; there is really no need for it as we no longer need to simulate protocol 21. So maybe just revert it and hopefully make CI pass?

Shaptic: that's what I get for trying to get the latest version

Shaptic:
@dmkozh looks like https://github.com/stellar/rs-soroban-env/commit/5a28b696b2fe0dcca08d1dd47322473b0cc6f818 was a breaking change to `soroban-simulation`: what's the appropriate resolution in [this preflight call](https://github.com/stellar/stellar-rpc/blob/main/cmd/stellar-rpc/lib/preflight/src/shared.rs#L151-L160)?

Or is it more correct to remove `unstable-next-api` from our [toml](https://github.com/stellar/stellar-rpc/blob/main/cmd/stellar-rpc/lib/preflight/Cargo.toml#L21-L22)?

dmkozh:
>  was a breaking change to soroban-simulation

Yes, it's expected to be breaking, which is why`unstable-next-api` guard is used. I'm surprised this feature is enabled by default in RPC. FWIW the particular change is done in order to allow addressing https://github.com/stellar/stellar-rpc/issues/344 in RPC.

I think it would make sense to remove this feature from RPC (as it has to be added intentionally), but it also makes sense to update to just 22.1.0.

dmkozh:
> Edit: Went with the second option, which changes the question: what should it be in the future when we're ready to use the next API?

I think a separate PR should be created for adaptation to the API changes. Basically the idea behind `unstable-next-api` is to clean it up on every major release and I think it makes sense for clients to remove it when bumping the major version of env. Then for the work in-between the releases it can be enabled in order to adopt the API changes without waiting for the major release.

Shaptic:
Yep makes sense! We'll refactor for the new API as part of P23 work.

Armss9936 (approved):

Armss9936 (approved):

PR #366, opened 2025-02-19
Stop allocating a port for the HTTP server when the query server is enabled

Shaptic:
### What

### Why
This workaround is no longer necessary (see [Slack thread](https://stellarfoundation.slack.com/archives/C07ET021CA2/p1739921403806869?thread_ts=1739921155.418029&cid=C07ET021CA2)).

### Known limitations
n/a


tamirms:
@Shaptic there are some integration test failures and it's not clear if they are related to this change

2opremio (approved):

Shaptic:
Merging now as the failures (dep checker and e2e) are known and understood (Go is using P23 XDR but Rust is not, e2e needs everything to work w/ p23).

Issue #370, opened 2025-02-26
Stop storing ledger entries and defer to Core


2opremio:
More context: https://github.com/stellar/stellar-rpc/pull/353

Protocol 23 comes with complicated state archival changes for which we will be delegating all state management to Captive Core. In particular, all the state querying will happen through HTTP calls to Captive Core.

As a side effect, we no longer need to manage state (i.e. ledger entry values). Thus we need to remove:

1. All state book-keeping code from the ingestion code.
2. The state table (`ledger_entries`) from the database. This will require a SQL migration. We should confirm the migration works with a full DB using pubnet to ensure there are no issues.
3. The (unreleased) `--stellar-captive-core-http-query-port` flag (currently defaulting to 0) since Captive Core's query port needs to be enabled by default for RPC to work. Alternatively, we can leave the flag but make it default to a non-zero value (note that even if we remove the flag, the value is still configurable by the captive core config file).

2opremio:
(3) is done by #412

2opremio:
One important thing to take into account about this  this ticket is upgrading from p22 to p23.

It may be that the Captiver Core instance RPC is running against when the p23 upgrade happens doesn't yet provide ledger entries.

We should check with Core if they will provide a release both capable of p22, p23 whilst providing the new HTTP server endpoints.

@SirTyson can you confirm this will be the case?

On our side, as usual, we should ensure to release and deploy such RPC version in advance.

SirTyson:
Ya, the endpoint won't be protocol gated and will work for both p22 and p23 once we cut the package

PR #377, opened 2025-03-01
Filter non-temporary entries from eviction

Shaptic:
### What
Filter non-temporary entries from ledger entry eviction routine.

### Why
The name was not updated, but the semantic meaning of the field has changed. See 

https://github.com/stellar/stellar-xdr/blob/5d7aebdd9f550707b482037e75c692e3e70aac27/Stellar-ledger.x#L561-L564

This means we need to filter out all persistent entries that are being evicted, only evicting temporary ones from our ledger entry store. This effectively means that RPC's ledger entry store will contain all valid entries (as if state archival didn't exist), which is exactly what we need during the transition period before Core's `/getledgerentry` fuels simulation.

### Known limitations
This will need to change later, but allows behavior to be correct for now.

### DIFF
--- a/cmd/stellar-rpc/internal/ingest/service.go
+++ b/cmd/stellar-rpc/internal/ingest/service.go
@@ -295,12 +295,30 @@ func (s *Service) ingest(ctx context.Context, sequence uint32) error {
 		return err
 	}
 
-	// EvictedTemporaryLedgerKeys will include both temporary ledger keys which
-	// have been evicted and their corresponding ttl ledger entries
-	evictedTempLedgerKeys, err := ledgerCloseMeta.EvictedTemporaryLedgerKeys()
+	// EvictedTemporaryLedgerKeys will, in this completely borked up version of
+	// XDR, actually include ALL evicted ledger keys (including persistent and
+	// code entries).
+	//
+	// In order to maintain the facade for simulation that eviction isn't
+	// happening yet (simulation isn't ready for state archival yet), we will
+	// continue to only evict temporary entries from our state.
+	//
+	// Tomorrow, when ledger state isn't managed by RPC at all, this code can be
+	// removed entirely and we can rely on Core to maintain ledger entries for
+	// simulation.
+	evictedLedgerKeys, err := ledgerCloseMeta.EvictedTemporaryLedgerKeys()
 	if err != nil {
 		return err
 	}
+
+	evictedTempLedgerKeys := make([]xdr.LedgerKey, 0, len(evictedLedgerKeys))
+	for _, key := range evictedLedgerKeys {
+		if key.Type == xdr.LedgerEntryTypeContractData &&
+			key.MustContractData().Durability == xdr.ContractDataDurabilityTemporary {
+			evictedTempLedgerKeys = append(evictedTempLedgerKeys, key)
+		}
+	}
+
 	if err := s.ingestTempLedgerEntryEvictions(ctx, evictedTempLedgerKeys, tx); err != nil {
 		return err
 	}


cmd/stellar-rpc/internal/ingest/service.go
```diff
@@ -295,12 +295,30 @@ func (s *Service) ingest(ctx context.Context, sequence uint32) error {
 		return err
 	}
 
-	// EvictedTemporaryLedgerKeys will include both temporary ledger keys which
-	// have been evicted and their corresponding ttl ledger entries
-	evictedTempLedgerKeys, err := ledgerCloseMeta.EvictedTemporaryLedgerKeys()
+	// EvictedTemporaryLedgerKeys will, in this completely borked up version of
```
tamirms, line 298: good catch, I forgot about this change. I think I'll need to rename `EvictedTemporaryLedgerKeys()` to `EvictedLedgerKeys()` in the go monorepo

cmd/stellar-rpc/internal/ingest/service.go
```diff
@@ -295,12 +295,30 @@ func (s *Service) ingest(ctx context.Context, sequence uint32) error {
...
 		return err
 	}
+
+	evictedTempLedgerKeys := make([]xdr.LedgerKey, 0, len(evictedLedgerKeys))
+	for _, key := range evictedLedgerKeys {
+		if key.Type == xdr.LedgerEntryTypeContractData &&
+			key.MustContractData().Durability == xdr.ContractDataDurabilityTemporary {
+			evictedTempLedgerKeys = append(evictedTempLedgerKeys, key)
+		}
+	}
```
tamirms, line 320: I just realized that the `EvictedPersistentLedgerEntries` in ledger close meta will be empty from protocol 23 onwards. The ingestion library has used `EvictedPersistentLedgerEntries()` to include evicted persistent ledger entry changes (representing them as deletions) in the change stream. However, In protocol 23 we won't have access to the ledger entry which was evicted (regardless of whether it was a temporary or persistent ledger entry).

I will need to update the ingestion library so that we don't merge `EvictedPersistentLedgerEntries()` into the change stream. However, rpc needs to evict all the ledger keys in `evictedLedgerKeys` including the persistent ledger keys. So, I think you can remove the filtering you have here but I think you'll need to rename `ingestTempLedgerEntryEvictions()` and call it `ingestLedgerEntryEvictions()` so that it handles evictions for both temporary and persistent ledger entries.

Shaptic: Acknowledged! I'll wait to merge this PR until that monorepo work is done so I can update the reference here. For this, though,

> rpc needs to evict all the ledger keys in evictedLedgerKeys including the persistent ledger keys. 

I don't think this is entirely right. We want to keep around a ledger entry database that simulation can use, which means that it needs to know about archived entries. Thus we don't want to _actually_ evict persistent entries so that simulation can know about them. 

Once we move over to the Core-backed variant (#370), we can remove this code entirely, but for now, my understanding was that we want to be "Core-like" in our entry storage and hence pretend like persistent entries haven't been archived yet. This is backed by the fact that `EvictedPersistentLedgerEntries` doesn't appear in the RPC ingestion code at all. My understanding of the simulation flow is a little vague, though.

cc @2opremio

tamirms: > This is backed by the fact that EvictedPersistentLedgerEntries doesn't appear in the RPC ingestion code at all

I think that's because evicted persistent ledger entries were included in the transaction change reader as removed ledger entries.

In any case, like you said, I don't think this matters much because all this code will be deleted in https://github.com/stellar/stellar-rpc/issues/370

Shaptic: Ahhh of course, good point, so those changes were being processed implicitly :brain: bless `ingest` :pray: I think that means I need to update my understanding of simulation, then, since I have no clue how simulation is handling archived ledger keys if RPC isn't keeping them around.

tamirms: > since I have no clue how simulation is handling archived ledger keys if RPC isn't keeping them around

archived ledger entries are still kept in the rpc db . it's only when they are evicted that they would get removed from the db. But since nothing got evicted before protocol 23, rpc has never had to deal with this problem in simulation. From protocol 23 onwards I think the only solution is to use core's endpoint during simulation.

If for some reason we're not able to use core's endpoint then we need to look into how we should maintain evicted state in rpc's db and that will also require figuring out how to bootstrap evicted state as well when ingesting from scratch. But hopefully we will not have to go down that road

tamirms (approved):

Issue #367, opened 2025-02-26
Measure performance of new core-based state querying


2opremio:
Context: https://github.com/stellar/stellar-rpc/pull/353

We need to ascertain whether the performance of the Core implementation is good enough. In particular, we should test it in AWS, using EBS storage, which has had a negative performance impact in the past.

PR #380, opened 2025-03-05
Add benchmark for getLedgerentries method

2opremio:
Part of https://github.com/stellar/stellar-rpc/issues/367

The benchmark tests the two backends for the implementation:

* Using the new Captive Core HTTP endpoint
* Using the DB-based internal soroban-rpc implementation

### DIFF
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -33,8 +33,8 @@ jobs:
           make build-libs
 
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@58eda26a511c265ee35b3ee4b101fb8adfd76480 # version v6.1.1
+        uses: golangci/golangci-lint-action@v7
         with:
-          version: v1.63.4 # this is the golangci-lint version
+          version: v2.0.2 # this is the golangci-lint version
           github-token: ${{ secrets.GITHUB_TOKEN }}
           only-new-issues: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,125 +1,100 @@
-linters-settings:
-  dupl:
-    threshold: 100
-
-  funlen:
-    lines: 100
-    statements: 50
-
-  misspell:
-    locale: US
-
-  gci:
-    # Section configuration to compare against.
-    # Section names are case-insensitive and may contain parameters in ().
-    # The default order of sections is `standard > default > custom > blank > dot > alias > localmodule`,
-    # If `custom-order` is `true`, it follows the order of `sections` option.
-    # Default: ["standard", "default"]
-    sections:
-      - standard # Standard section: captures all standard packages.
-      - default # Default section: contains all imports that could not be matched to another section type.
-      - prefix(github.com/stellar/) # Custom section: groups all imports with the specified Prefix.
-      - localmodule # Local module section: contains all local packages. This section is not present unless explicitly enabled.
-    skip-generated: false
-    # Enable custom order of sections.
-    # If `true`, make the section order the same as the order of `sections`.
-    # Default: false
-    custom-order: true
-
-  dogsled:
-    # Checks assignments with too many blank identifiers.
-    # Default: 2
-    max-blank-identifiers: 3
-
-  cyclop:
-    # The maximal code complexity to report.
-    # Default: 10
-    max-complexity: 15
-
-  wrapcheck:
-    # An array of strings that specify substrings of signatures to ignore.
-    # If this set, it will override the default set of ignored signatures.
-    # See https://github.com/tomarrell/wrapcheck#configuration for more information.
-    # Default: [".Errorf(", "errors.New(", "errors.Unwrap(", "errors.Join(", ".Wrap(", ".Wrapf(", ".WithMessage(", ".WithMessagef(", ".WithStack("]
-    ignoreSigs:
-      - .Errorf(
-      - errors.New(
-      - errors.Unwrap(
-      - errors.Join(
-      - .Wrap(
-      - .Wrapf(
-      - .WithMessage(
-      - .WithMessagef(
-      - .WithStack(
-    # An array of strings that specify regular expressions of signatures to ignore.
-    # Default: []
-    ignoreSigRegexps:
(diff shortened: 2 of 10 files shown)


SirTyson:
> Testing on a network with history would be much harder and I don't even think we can use @SirTyson 's code on pubnet.
> 
> BTW, running it in Kubernetes required quite a lot of manual handling (since docker-compose doesn't work inside pods, and integration tests rely on it). I will document the process soon.

Wrt testing, what AWS sku are you using for local storage? I've found when testing stellar-core a lot of our default SKUs use networked back storage for root, so "local storage" may still be backed by an EBS like drive, leading to no difference.
When building BucketListDB, I've found that perf testing on a laptop is basically meaningless. The reason being that BucketListDB has defined memory usage, whereas SQL generally grows caches depending on the amount of memory available. On my laptop, SQL also blows BucketListDB out of the water because I have like 64 GB so all SQL state is cached. This is not the case on Validator SKUs, where BucketListDB outperforms SQL due to memory limits.

I'm curious why this won't work on pubnet though? Speeds under kubernetes seem similar. By "history" do you mean the requirement to store 7 days worth of ledger state snapshots?

2opremio:
As to how I managed to measure it in k8s. This is what I did:

1. Scale down RPC to liberate a PVC we can attach to for testing.
2. Create a and deploy a testing POD based on a custom Docker container which attaches to the PVC from (1)

`Dockerfile` for `2opremio/pvc-test`:
```Dockerfile
FROM ubuntu:focal

RUN apt-get update
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y git emacs-nox wget curl postgresql-client sqlite3 apt-transport-https gnupg2 build-essential

#install stellar-core
RUN wget -qO - https://apt.stellar.org/SDF.asc | apt-key add -
RUN echo "deb https://apt.stellar.org focal unstable" | tee -a /etc/apt/sources.list.d/SDF-unstable.list
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y stellar-core=22.1.1-2291.a50f3f919.focal~do~not~use~in~prd

# install rust
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
ENV PATH=$PATH:/root/.cargo/bin/

#install go
RUN wget https://go.dev/dl/go1.24.1.linux-amd64.tar.gz && tar -C /usr/local -xzf go1.24.1.linux-amd64.tar.gz && rm go1.24.1.linux-amd64.tar.gz
ENV PATH=$PATH:/usr/local/go/bin

RUN git clone https://github.com/2opremio/soroban-rpc.git && cd soroban-rpc && git checkout benchmark-core-getledgerentries && make build && go test -c -o integrationtest.test github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/integrationtest

ENV STELLAR_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN=/usr/bin/stellar-core
ENV STELLAR_RPC_INTEGRATION_TESTS_CORE_MAX_SUPPORTED_PROTOCOL=22
ENV STELLAR_RPC_INTEGRATION_TESTS_ENABLED=true



# STELLAR_RPC_BENCH_BASE_STORAGE_DIR=/var/lib/stellar/foo
ENV STELLAR_RPC_BENCH_CORE_HOST_PORTS=core:11625;core:1570;core:11626

# Example:

# curl -v 'core:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=22'
# integrationtest.test -test.v -test.paniconexit0 -test.bench '^\QBenchmarkGetLedgerEntries\E$/^\QWithCore\E$' -test.benchtime=100x -test.run '^$'
```

`pvc-inspector.yaml`:

```yaml
kind: Pod
apiVersion: v1
metadata:
  name: fons-test-rpc-issue-367
spec:
  volumes:
    - name: soroban-rpc-pubnet-dev-var-lib-stellar
      persistentVolumeClaim:
        claimName: soroban-rpc-pubnet-dev-var-lib-stellar-soroban-rpc-pubnet-dev-1
  containers:
    - name: debugger
      image: 2opremio/pvc-test
      imagePullPolicy: Always
      command: [ 'sleep', 'infinity' ]
      resources:
        requests:
          memory: "8Gi"
          cpu: "4000m"
        limits:
          memory: "8Gi"
          cpu: "4000m"
      volumeMounts:
        - mountPath: /var/lib/stellar
          name: soroban-rpc-pubnet-dev-var-lib-stellar
```

3. Convert intergation tests files from docker compose to K8s with https://kompose.io by doing `kompose -f cmd/stellar-rpc/internal/integrationtest/infrastructure/docker/docker-compose.yml convert`. It is worth noting that this required two extra modifications

    3.1 Add explicit resources to core and postgres:

```yaml
      resources:
        requests:
          memory: "8Gi"
          cpu: "4000m"
        limits:
          memory: "8Gi"
          cpu: "4000m"
```

    3.2 Make the start script from the Core container executable:
    
```yaml
        - configMap:
            items:
              - key: core-start.sh
                path: start
            name: core-cm1
            defaultMode: 0777
```

4. Deploy the resources created by kompose

5. Exec into the pvc tester pod e.g. `kubectl exec --namespace soroban-rpc-pubnet-dev -ti fons-test-rpc-issue-367  -- bash`

6. Run a benchmark by:
    6.1. Upgrading core: `curl -v 'core:11626/upgrades?mode=set&upgradetime=1970-01-01T00:00:00Z&protocolversion=22'`
    6.2 Running the benchmark command e.g. `cd soroban-rpc && integrationtest.test -test.v -test.paniconexit0 -test.bench '^\QBenchmarkGetLedgerEntries\E$/^\QWithCore\E$' -test.benchtime=100x -test.run '^$'`
    
    
After every test the core kompose resorces should be recreated to remove state.

Shaptic (approved):
Nice, LGTM. I wonder if we should add a test that somehow expires entries to see if there’s a significant difference when state archival kicks in, but we can discuss that separately. This is great skeleton work for adding a bunch more perf tests!

PR #383, opened 2025-03-21
Add operation index to event result

### LINKED ISSUE #371: Add operation index in `getEvents` response
Due to changes in protocol 23 ( @tamirms  : can you edit and include the exact reason? ) events can refer to operations in multi-operation transactions. Thus, we need to specify what operation the event is referring to.

Shaptic:
### What
See #371. 

### Why
Closes #371.

### Known limitations
TODO: Fix tests

### DIFF
--- a/protocol/get_events.go
+++ b/protocol/get_events.go
@@ -25,6 +25,7 @@ type EventInfo struct {
 	LedgerClosedAt string `json:"ledgerClosedAt"`
 	ContractID     string `json:"contractId"`
 	ID             string `json:"id"`
+	OpIndex        int32  `json:"operationIndex"`
 
 	// Deprecated: PagingToken field is deprecated, please use Cursor at top level for pagination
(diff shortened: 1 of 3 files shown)


protocol/get_events.go
```diff
@@ -25,6 +25,7 @@ type EventInfo struct {
 	LedgerClosedAt string `json:"ledgerClosedAt"`
 	ContractID     string `json:"contractId"`
 	ID             string `json:"id"`
+	OpIndex        uint32 `json:"operationIndex"`
```
tamirms, line 28: I think we should also include the transaction index . With the ledger sequence, transaction index, and operation index I think we have all the information needed to sort events

Shaptic: Might as well! Done in 49fd1e8

tamirms (approved):

Issue #394, opened 2025-04-04
Race condition in go RPC client


2opremio:
Due to https://github.com/creachadair/jrpc2/issues/118 , we reopen the client every time a call fails as follows:

https://github.com/stellar/stellar-rpc/blob/49af4430cbcef8bb9995e4b266a36c619da7650f/client/main.go#L29-L50

Modifying the client is racy (a goroutine could read it while another modifies it):

https://github.com/stellar/stellar-rpc/blob/main/client/main.go#L40

Even worse, another goroutine could be accessing the client before the one refreshing it is done, leading to an inmediate error.

It could be fixed with a mutex, but ... we would need to lock every time we access the cli. Maybe a RWMutex would be fine, but it may lead to contention.

PR #395, opened 2025-04-04
Fix CLI refresh race conditions

2opremio:
### What

Fix https://github.com/stellar/stellar-rpc/issues/394 with a mutex on the client.

### Why

We don't like races.

### Known limitations

It can cause contention if the client needs to be refreshed often due to errors. But this is better than a race, at least until a better solution is offered by the maintainer at https://github.com/stellar/stellar-rpc/issues/394

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"net/http"
+	"sync"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/creachadair/jrpc2/jhttp"
@@ -13,6 +14,7 @@ import (
 type Client struct {
 	url        string
 	cli        *jrpc2.Client
+	mx         sync.RWMutex // to protect cli writes in refreshes
 	httpClient *http.Client
 }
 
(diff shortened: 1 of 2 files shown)


client/main.go
```diff
@@ -23,25 +25,33 @@ func NewClient(url string, httpClient *http.Client) *Client {
...
+	defer c.mx.Unlock()
+	if c.cli != nil {
+		c.cli.Close()
+	}
+	c.cli = cli
 }
 
 func (c *Client) callResult(ctx context.Context, method string, params, result any) error {
+	c.mx.RLock()
 	err := c.cli.CallResult(ctx, method, params, result)
```
urvisavla, line 53: Do you need to check for c.cli != nil here?

urvisavla (approved):
lgtm!

client/main.go
2opremio: I think that the consumers should make sure that the client isn't explicitly closed. After using the mutex, the client can only be found to be nil when it's closed explicitly

Issue #390, opened 2025-03-28
Simulation fails when contract removes persistent entry that doesn't exist


leighmcculloch:
### What version are you using?

```
22.1.2-755d81f1c738943f55ba6d61cbd72dc65de0661f
```

### What did you do?

Deployed this contract:
```rust
#![no_std]
use soroban_sdk::{contract, contractimpl, symbol_short, Env};

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn remi(env: Env) {
        env.storage().instance().remove(&symbol_short!("noexist"));
    }

    pub fn remp(env: Env) {
        env.storage().persistent().remove(&symbol_short!("noexist"));
    }
}
```

Removed from instance storage when it doesn't exist is fine, which worked fine.
```
$ stellar contract invoke \
    --id CCLEEVOSGIJQEGORUJU5YZRKGHQG6GQMZIQZ2JW77MIWBI37JSPFNKO3 \
    -- \
    remi
ℹ️ Signing transaction: 1a6d47bce30f0940fa2a5744100791f503e263d63a9642d1b6fd8634722b10f4
```

Removed from persistent storage that fails simulation:
```
$ stellar contract invoke \
    --id CCLEEVOSGIJQEGORUJU5YZRKGHQG6GQMZIQZ2JW77MIWBI37JSPFNKO3 \
    -- \
    remp
❌ error: transaction simulation failed: missing before and after
```

### What did you expect to see?

I expected to see the invocation of a function removing data from persistent storage to succeed.

### What did you see instead?

The transaction fail to simulate, and therefore be unable to submit.

---

This bug was first reported here:
- https://discord.com/channels/897514728459468821/1354901427482394786/1354901427482394786

leighmcculloch:
The error message is generated by the code that does XDR diffs. On the surface it looks like it doesn't know what to do when there is no before or after in the diff. There being no before or after makes sense, since the contract data entry didn't exist, and the only interaction with it was to remove it.

https://github.com/stellar/stellar-rpc/blob/49af4430cbcef8bb9995e4b266a36c619da7650f/cmd/stellar-rpc/internal/methods/simulate_transaction.go#L52

2opremio:
Related https://github.com/stellar/rs-soroban-env/issues/1448

2opremio:
Related: https://github.com/stellar/stellar-rpc/issues/239

ElliotFriend:
[...]

PR #396, opened 2025-04-07
Ignore state change entries which don't incur an actual diff

Shaptic:
### What
Even though the key was touched (but left unchanged), it won't appear in the list of state changes. This is semantically fine, since no state changed.

### Why
Closes https://github.com/stellar/stellar-rpc/issues/390 until https://github.com/stellar/rs-soroban-env/issues/1448 is done. Simulation should still be able to succeed in these cases.

### Known limitations
A bit hacky, but it's a temporary fix until the "proper" one is done in P23, where we'll include the key itself and allow `null`/`null` for before/after.

### DIFF
--- a/cmd/stellar-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/stellar-rpc/internal/methods/simulate_transaction.go
@@ -23,6 +23,8 @@ type PreflightGetter interface {
 	GetPreflight(ctx context.Context, params preflight.GetterParameters) (preflight.Preflight, error)
 }
 
+var errMissingDiff = errors.New("no ledger difference found")
+
 func LedgerEntryChangeFromXDRDiff(diff preflight.XDRDiff, format string) (protocol.LedgerEntryChange, error) {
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> A bit hacky, but it's a temporary fix until the "proper" one is done in P23, where we'll include the key itself and allow `null`/`null` for before/after.

I think the key here is that while this change will stop rpc from erroring, rpc's response won't be usable by clients to create the footprint required to submit the tx because the client won't know which ledger entry the null, null refers to, in the same way that rpc doesn't know.

Is that correct?

Shaptic:
@leighmcculloch thankfully, no. This only affects the `stateChanges` field in [simulateTransaction](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/simulateTransaction), which is purely there as an informative "here's what would change as part of this invocation." The `transactionData` which contains the footprint is completely unaffected.

tamirms (approved):

2opremio (approved):

Issue #387, opened 2025-03-25
Load testing for `getEvents` endpoint via TTP


Shaptic:

2opremio:
As mentioned elsewhere:

The cap67 simulation ( #403 ) causes soroban-rpc's DB to occupy about 18GB more than the non-cap67 DB (174GB vs 156GB):
```
% kubectl exec --namespace soroban-rpc-pubnet-dev -ti soroban-rpc-pubnet-dev-0  -- ls -lh  /var/lib/stellar/soroban-rpc-db.sqlite
-rw-r--r--. 1 root root 156G May  6 22:16 /var/lib/stellar/soroban-rpc-db.sqlite
% kubectl exec --namespace soroban-rpc-pubnet-dev -ti soroban-rpc-pubnet-dev-cap67-0  -- ls -lh  /var/lib/stellar/soroban-rpc-db.sqlite
-rw-r--r--. 1 root root 174G May  6 22:16 /var/lib/stellar/soroban-rpc-db.sqlite
```

That's about 2.5 GB extra per day (there is a 7 day retention window), which I don't think is that bad.

Also, `getEvent` requests have a similar latency (or the difference is negligible) 

e.g:

```
% kubectl exec --namespace soroban-rpc-pubnet-dev -ti soroban-rpc-pubnet-dev-0  -- time curl -X POST -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 8675309, "method": "getEvents", "params": { "startLedger": 56977314, "filters": [ { "type": "contract" } ],  "pagination": { "limit": 2 } } }'  localhost:8000


0.00user 0.00system 0:00.13elapsed 3%CPU (0avgtext+0avgdata 10540maxresident)k
```

vs

```
% kubectl exec --namespace soroban-rpc-pubnet-dev -ti soroban-rpc-pubnet-dev-cap67-0  -- time curl -X POST -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 8675309, "method": "getEvents", "params": { "startLedger": 56977314, "filters": [ { "type": "contract" } ],  "pagination": { "limit": 2 } } }'  localhost:8000


0.00user 0.00system 0:00.12elapsed 4%CPU (0avgtext+0avgdata 10588maxresident)k
```

PR #403, opened 2025-04-10
Emulate CAP 67 events for perf assesment

2opremio:
### What

Emulate CAP 67 classic events by using the token transfer processor (enabled through flag)

### Why

It's required for https://github.com/stellar/stellar-rpc/issues/387

### Known limitations

The events generated are probably not accurate with respect to CAP 67, but it will help give us a performance impact ballpark.


cmd/stellar-rpc/internal/db/event.go
```diff
@@ -43,14 +44,174 @@ type EventReader interface {
...
+	}
+	return result, nil
+}
+
+func ttpEventToDiagnosticEvent(event *token_transfer.TokenTransferEvent) xdr.DiagnosticEvent {
+	topics := []xdr.ScVal{}
+	var data xdr.ScVal
+	switch event.GetEvent().(type) {
+	case *token_transfer.TokenTransferEvent_Mint:
+		burn := event.GetMint()
```
karthikiyer56, line 87: nit - improper variable name.
shud be `mint :=...`

cmd/stellar-rpc/internal/db/event.go
```diff
@@ -43,14 +44,174 @@ type EventReader interface {
...
+	case *token_transfer.TokenTransferEvent_Burn:
+		burn := event.GetBurn()
+		sym := xdr.ScSymbol("burn")
+		// We cannot fully use addresses yet, since we would need cap67 for that
+		from := xdr.ScSymbol(burn.From)
+		// TODO: parse the amount to i128 (we are just simulating though)
+		amount := xdr.ScSymbol(burn.Amount)
+		topics = append(topics,
+			xdr.ScVal{
+				Type: xdr.ScValTypeScvSymbol,
```
karthikiyer56, line 116: all this logic is still missing the parsing of the `Asset` from the proto.
It might be a good idea to parse the asset from the string as well and convert it into a sep-11-asset string and include it in the `topics` variable.
To convert Asset from proto to Sep-11-compliant string, you can use this helper
```
assetStr := event.GetAsset().ToXdrAsset().StringCanonical()
```
and then you can convert it into an ScVal

2opremio: Will CAP-67 include the asset as a topic in all events?

karthikiyer56: For classic events - always

2opremio:
@karthikiyer56 PTAL. I have also changed the target base so that we don't pollute main with this.

karthikiyer56 (approved):

cmd/stellar-rpc/internal/db/event.go
```diff
@@ -43,14 +44,180 @@ type EventReader interface {
...
+	}
+	assetStr := event.GetAsset().ToXdrAsset().StringCanonical()
+	assetSym := xdr.ScSymbol(assetStr)
+	topics = append(topics, xdr.ScVal{
+		Type: xdr.ScValTypeScvSymbol,
+		Sym:  &assetSym,
+	})
+
+	return xdr.DiagnosticEvent{
+		InSuccessfulContractCall: false,
```
Shaptic, line 208: Should be `true` actually I think?

PR #405, opened 2025-04-14
Drop soroban-rpc build support

urvisavla:
### What

Remove `BINARY_NAME` build argument as we are discontinuing support for building `soroban-rpc` image. Only `stellar-rpc` will be built going forward.

### Why
We renamed `soroban-rpc` to `stellar-rpc` before protocol 22 and kept both builds to give users time to switch. Now that sufficient time has passed, we’re discontinuing the `soroban-rpc` build

### Known limitations
Users relying on the `soroban-rpc` image will need to migrate to `stellar-rpc`.


2opremio (approved):

2opremio:
🎆 

I think this breaks the Jenkins pipeline though, so that should be adjusted ASAP.

PR #406, opened 2025-04-15
Remove Ubuntu 20 from CI as it's no longer supported

urvisavla:
### What

Remove Ubuntu 20 from CI. 

### Why

RPC CI failing because Ubuntu 20 is now fully unsupported (see: https://github.com/actions/runner-images/issues/11101).

### DIFF
--- a/.github/workflows/stellar-rpc.yml
+++ b/.github/workflows/stellar-rpc.yml
@@ -18,7 +18,7 @@ jobs:
     name: Unit tests
     strategy:
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04 ]
+        os: [ ubuntu-22.04 ]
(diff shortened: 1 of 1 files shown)


Shaptic (approved):

.github/workflows/stellar-rpc.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     name: Unit tests
     strategy:
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04 ]
+        os: [ ubuntu-22.04, ubuntu-24.04 ]
```
Shaptic, line 21: I don't think core releases for 24 yet, so we can change the matrix to just be 22!

urvisavla: Yes, you're right.

Issue #391, opened 2025-03-28
getLedgers returns inaccurate error message when requested ledger range outside of retention window


sreuland:
### What version are you using?

22.1.x

### What did you do?

requested a ledger range that was not present in the RPC's current retention window:
```
curl --location 'https://soroban-testnet.stellar.org' \
--header 'Content-Type: application/json' \
--data '{
  "jsonrpc": "2.0",
  "id": 8675309,
  "method": "getLedgers",
  "params": {
    "startLedger": 1000,
    "pagination": {
      "limit": 2
    }
  }
}'
```

### What did you expect to see?
an error response which sets `error.code` to standardized json-rpc code for client request error, [-32602](https://docs.trafficserver.apache.org/en/latest/developer-guide/jsonrpc/jsonrpc-node-errors.en.html#standard-errors) and a `error.message` which describes the reason, should state requested ledger range is not in retention window data range of RPC. 

### What did you see instead?
a generic error message stating it was due to internal aspect:
```
{
    "jsonrpc": "2.0",
    "id": 8675309,
    "error": {
        "code": -32003,
        "message": "[-32003] request failed to process due to internal issue"
    }
}
```

PR #407, opened 2025-04-15
Unify request validation methods for `getLedgers` and `getTransactions`

Shaptic:
### What
Use a single method to ensure request parameters are correct. Specifically, follow the documented set of rules:

 * If `pagination` is set:
   - If the `cursor` is set, the `startLedger` is **not** set.
   - If the `limit` is set, it does not exceed the maxLimit.
   - If the `cursor` is **not** set, the `startLedger` is in the RPC's range of known ledgers.
 * Otherwise,
   - The `startLedger` is set and is in the RPC's range of known ledgers.

This also improves the error messages in failure cases by, for example, including the `startLedger` that was provided in the message itself.

### Why
Several issues have cropped up related to this. 

For example, https://github.com/stellar/stellar-rpc/issues/391 is caused by a lack of `startLedger` validation due to the presence of `pagination`. A recent [Slack thread](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1744727144784449) by @fnando also highlights problematic validation.

### Known limitations
n/a

### DIFF
--- a/protocol/get_ledgers.go
+++ b/protocol/get_ledgers.go
@@ -58,9 +58,9 @@ type GetLedgersResponse struct {
 	Cursor                string       `json:"cursor"`
 }
 
-// IsStartLedgerWithinBounds checks whether the request start ledger/cursor is
-// within the max/min ledger for the current RPC instance.
-func IsStartLedgerWithinBounds(startLedger uint32, ledgerRange LedgerSeqRange) bool {
+// IsLedgerWithinRange checks whether the request start ledger/cursor is within
+// the max/min ledger for the current RPC instance.
+func IsLedgerWithinRange(startLedger uint32, ledgerRange LedgerSeqRange) bool {
 	return startLedger >= ledgerRange.FirstLedger && startLedger <= ledgerRange.LastLedger
 }
 
@@ -93,13 +93,13 @@ func ValidatePagination(
 				return fmt.Errorf("startLedger (%d) and cursor (%s) cannot both be set",
 					startLedger, pagination.Cursor)
 			}
-		} else if !IsStartLedgerWithinBounds(startLedger, ledgerRange) { // xor startLedger
+		} else if !IsLedgerWithinRange(startLedger, ledgerRange) { // xor startLedger
 			return errBadPage
 		}
 		if pagination.Limit > maxLimit {
 			return fmt.Errorf("limit must not exceed %d", maxLimit)
 		}
-	} else if !IsStartLedgerWithinBounds(startLedger, ledgerRange) {
+	} else if !IsLedgerWithinRange(startLedger, ledgerRange) {
 		return errBadPage
 	}
 
--- a/protocol/get_ledgers.go
+++ b/protocol/get_ledgers.go
@@ -88,12 +88,12 @@ func ValidatePagination(
 	)
 
 	if pagination != nil {
-		if pagination.Cursor != "" {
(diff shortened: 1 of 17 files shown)


urvisavla:
It looks like getEvents has its own checks for pagination and ledger range. Should that updated it to use this as well for consistent error handling?

Shaptic:
@urvisavla `getEvents` has a more complicated series of checks (e.g. it has `endLedger`, needs to validate topics, etc.) and uses a different cursor (`*Cursor` vs. `string`), so it has to have its own validation. It could, I think, be unified in theory, but I didn't want to make this a massive PR so I figured this would be a quick n' easy win.

protocol/get_ledgers.go
```diff
@@ -74,3 +57,51 @@ type GetLedgersResponse struct {
 	OldestLedgerCloseTime int64        `json:"oldestLedgerCloseTime"`
 	Cursor                string       `json:"cursor"`
 }
+
+// IsStartLedgerWithinBounds checks whether the request start ledger/cursor is
+// within the max/min ledger for the current RPC instance.
+func IsStartLedgerWithinBounds(startLedger uint32, ledgerRange LedgerSeqRange) bool {
+	return startLedger >= ledgerRange.FirstLedger && startLedger <= ledgerRange.LastLedger
```
leighmcculloch, line 64: Not required, but I think this function in its name and description describe more of a use case it is used for, rather than the simple primitive it provides, and it could be reduced to:
```suggestion
// IsLedgerWithinBounds checks whether the request ledger is within the range specified.
func IsLedgerWithinRange(ledger uint32, ledgerRange LedgerSeqRange) bool {
	return ledger >= ledgerRange.FirstLedger && ledger <= ledgerRange.LastLedger
```

leighmcculloch (approved):

urvisavla:
> @urvisavla `getEvents` has a more complicated series of checks (e.g. it has `endLedger`, needs to validate topics, etc.) and uses a different cursor (`*Cursor` vs. `string`), so it has to have its own validation. It could, I think, be unified in theory, but I didn't want to make this a massive PR so I figured this would be a quick n' easy win.

I missed that GetEvents also includes the endLedger.  I agree there’s no need to expand the scope of this PR. The changes look good overall. There are some failing integration tests that you'll want to fix.

protocol/get_ledgers.go
Shaptic: Yeah good call, done

protocol/get_ledgers.go
```diff
@@ -33,21 +27,10 @@ type GetLedgersRequest struct {
...
-		return fmt.Errorf(
-			"start ledger must be between the oldest ledger: %d and the latest ledger: %d for this rpc instance",
-			ledgerRange.FirstLedger,
-			ledgerRange.LastLedger,
-		)
-	}
+	return errors.Join(
+		ValidatePagination(req.StartLedger, req.Pagination, maxLimit, ledgerRange),
+		IsValidFormat(req.Format),
+	) // nils will coalesce
```
leighmcculloch, line 33: Thanks for adding the nils comment. The semantics of the `error` APIs are rarely obvious to me.

leighmcculloch (approved):

protocol/get_ledgers.go
Shaptic: Yep, a little too subtle for my taste but it does make this a clean call!

PR #408, opened 2025-04-15
Return `[]` rather than `null` when `getTransactions` is empty

### LINKED ISSUE #398: getTransactions should always return transactions as an array instead of null
### What version are you using?

22.1.2-755d81f1c738943f55ba6d61cbd72dc65de0661f

### What did you do?

```
$ stellar container start
$ curl http://localhost:8000/rpc \
    --header 'Content-Type: application/json' \
    --data \
    jq '{"jsonrpc":"2.0","id":1,"method":"getTransactions","params":{"startLedger":1366}}'
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "transactions": null,
    "latestLedger": 1452,
    "latestLedgerCloseTimestamp": 1744155679,
    "oldestLedger": 8,
    "oldestLedgerCloseTimestamp": 1744154230,
    "cursor": "6236292513793"
  }
}
```

Transactions as an empty array `[]`.

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "transactions": [],
    "latestLedger": 1452,
    "latestLedgerCloseTimestamp": 1744155679,
    "oldestLedger": 8,
    "oldestLedgerCloseTimestamp": 1744154230,
    "cursor": "6236292513793"
  }
}
```

It makes coding clients simpler if the data type is consistent.

### What did you see instead?

Transactions as `null`.

See above.

### Related

- https://github.com/stellar/js-stellar-sdk/issues/1161
- https://github.com/stellar/js-stellar-sdk/pull/1162

Shaptic:
### What
Using an uninitialized array causes the result to be `null` if it's never `append`ed to.

### Why
This is way more intuitive, closes #398.

### Known limitations
n/a

### DIFF
--- a/cmd/stellar-rpc/internal/methods/get_transactions.go
+++ b/cmd/stellar-rpc/internal/methods/get_transactions.go
@@ -216,7 +216,7 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 
 	// Iterate through each ledger and its transactions until limit or end range is reached.
 	// The latest ledger acts as the end ledger range for the request.
-	var txns []protocol.TransactionInfo
+	txns := make([]protocol.TransactionInfo, 0, limit)
 	var done bool
 	cursor := toid.New(0, 0, 0)
 	for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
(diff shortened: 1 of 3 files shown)


cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -216,7 +216,7 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 
 	// Iterate through each ledger and its transactions until limit or end range is reached.
 	// The latest ledger acts as the end ledger range for the request.
-	var txns []protocol.TransactionInfo
+	txns := make([]protocol.TransactionInfo, 0, limit)
```
urvisavla, line 219: Should we add a unit test to catch this if it breaks in future?

Shaptic: Good point, might as well! 0d72b91

leighmcculloch (approved):

PR #415, opened 2025-04-23
Fix initialization and encoding of cap67 emulation

2opremio:
Some fixes for cap67 emulation (which explains why the experiment didn't result in any performance or DB size changes)


karthikiyer56:
@2opremio ,
the `contractId` for classic events, including all fee events, is the hash of the asset.
The hash will always be present in the `Meta` field of the `TokenTransferEvent`
```
{
        "type": "system",
        "ledger": 56752984,
        "ledgerClosedAt": "2025-04-23T20:37:55Z",
        "contractId": "",
.......
 ```

PR #417, opened 2025-04-23
Remove support for storing ledger entries in local db

urvisavla:
### What

Remove ingestion of ledger entries, including the initial ledger entry ingestion from history archives when db is empty.

Added a new migration to delete the ledger_entries table. It's tested locally with a limited amount of data. One thing I noticed is that the db size on disk didn’t change after dropping the table. It decreased only after I explicitly ran VACUUM command.

### Why
Addressed item `1.` from https://github.com/stellar/stellar-rpc/issues/370.

### Known limitations
N/A


cmd/stellar-rpc/internal/db/ledgerentry.go
```diff
@@ -1,403 +1,11 @@
...
-type LedgerEntryReader interface {
-	GetLatestLedgerSequence(ctx context.Context) (uint32, error)
-	NewTx(ctx context.Context, cacheTx bool) (LedgerEntryReadTx, error)
-}
-
 type LedgerKeyAndEntry struct {
 	Key                xdr.LedgerKey
 	Entry              xdr.LedgerEntry
 	LiveUntilLedgerSeq *uint32 // optional live-until ledger seq, when applicable.
 }
```
2opremio, line 11: It may be more appropriate to have this in the `ledgerentries` package

cmd/stellar-rpc/internal/ingest/service.go
```diff
@@ -211,11 +209,6 @@ func (s *Service) fillEntriesFromCheckpoint(ctx context.Context, archive history
 	ctx, cancel = context.WithTimeout(ctx, s.timeout)
 	defer cancel()
 
-	reader, err := ingest.NewCheckpointChangeReader(ctx, archive, checkpointLedger)
```
urvisavla, line 214: Since we're no longer reading changes, can we remove fillEntriesFromCheckpoint and simplify maybeFillEntriesFromCheckpoint?

urvisavla: This has been removed now.

2opremio (approved):

2opremio:
Looks great!

PR #419, opened 2025-04-28
Add support for flexible topics length matching using wildcard in getEvents

### LINKED ISSUE #358: bug: getEvents filter matches on number of topics not just topic content
### What did you do?

Queried events and filtering by events with a first topic of "transfer".

```
$ curl -X POST \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":0,"method":"getEvents","params":{"startLedger":997500,"xdrFormat":"json","filters":[{"type":"contract","topics":[["AAAADwAAAAh0cmFuc2Zlcg=="]]}]}}' \    
https://soroban-testnet.stellar.org | jq
```

Results. In the case of the above query from ledger 997500, I expected to see one result:

```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": {
    "events": [
      {
        "type": "contract",
        "ledger": 997500,
        "ledgerClosedAt": "2025-02-06T12:08:56Z",
        "contractId": "CBIELTK6YBZJU5UP2WWQEUCYKLPU6AUNZ2BQ4WWFEIE3USCIHMXQDAMA",
        "id": "0004284229877780480-0000000003",
        "pagingToken": "0004284229877780480-0000000003",
        "inSuccessfulContractCall": true,
        "txHash": "c7e476df1befffb9c2fb584110a6bdbbfd476037ec039328fcb7d6f7ba2c82c1",
        "topicJson": [
          {
            "symbol": "transfer"
          },
          {
            "address": "CDIGPGARVOUWAS2ZXTUXQMPS7D5SZ6Q2HBAT4OCJDHURRSDQDEKSCRHP"
          },
          {
            "address": "CDBIFK6KMO5EJD4LNN6GQP3N27XJRTZCJVYEV2LMUUD6SCSH2VE37QXU"
          },
          {
            "string": "USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5"
          }
        ],
        "valueJson": {
          "i128": {
            "hi": 0,
            "lo": 10000000
          }
        }
      }
    ],
    "latestLedger": 997885,
    "cursor": "0004285887735132160-4294967294"
  }
}
```

### What did you see instead?

No results.

It appears that the topic filter matches not only on the topic, but also on the number of topics. So to filter on transfers from a SAC token, you have to specify three `"*"` additional topics so that the filter matches.

When filtering on transfers for a custom token, you have to specify two `"*"` additional topics because those tokens don't emit a fourth topic.

For example:

```
$ curl -X POST \
-H 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","id":0,"method":"getEvents","params":{"startLedger":997500,"xdrFormat":"json","filters":[{"type":"contract","topics":[["AAAADwAAAAh0cmFuc2Zlcg==","*","*","*"]]}]}}' \    
https://soroban-testnet.stellar.org | jq
```

### Discussion

It's an unintuitive developer experience to need to match on the exact topic count.

If a developer is filtering on a subset of topics, they should be able to see all events that match those initial topics irrespective of how many topics are being published.

This is particularly problematic when filtering events for SEP-41 where the SAC implements SEP-41 but chooses to emit an additional topic.

cc @stellar/platform-eng

urvisavla:
### What

A new wildcard `"**"` is introduced in the `getEvents` RPC endpoint for flexible topic length matching of zero or more topics.

This eliminates the need for manual padding in filters. For eg, previously, to match events where the first topic is `"X"` followed by any number of subsequent topics, you'd need to specify multiple filters like `["X"], ["X", "*"], ["X", "*", "*"]` and so on. The new "**" wildcard simplifies this to a single filter like `["X", "**"]`.

Rules for the new wildcard
1. `"**"` wildcard can only appear as the last segment in a filter
Invalid: `["X", "**", "*"]` ❌ 
Valid: `["X", "*", "**"]` ✅ -  matches events with at least two topics, where the first topic is "X".

2. The existing limit of max of 4 filter segments per filter is unchanged. The "**" wildcard is exempt from this count.
Invalid ["A","B","C","D","*"] ❌ 
Valid: ["A","B","C","D","**"] ✅  

### Why

Fixes #358

### DIFF
--- a/cmd/stellar-rpc/internal/methods/get_events_test.go
+++ b/cmd/stellar-rpc/internal/methods/get_events_test.go
@@ -381,7 +381,7 @@ func TestGetEvents(t *testing.T) {
 	t.Run("filtering by topic, flexible length matching", func(t *testing.T) {
 		doubleStar := "**"
 		dbx := newTestDB(t)
-		ctx := context.TODO()
+		ctx := t.Context()
 		log := log.DefaultLogger
 		log.SetLevel(logrus.TraceLevel)
 
@@ -439,7 +439,7 @@ func TestGetEvents(t *testing.T) {
 		require.NoError(t, err)
 
 		// strict topic length matching by default
-		results, err := handler.getEvents(context.TODO(), protocol.GetEventsRequest{
+		results, err := handler.getEvents(t.Context(), protocol.GetEventsRequest{
 			StartLedger: 1,
 			Filters: []protocol.EventFilter{
 				{
@@ -448,7 +448,8 @@ func TestGetEvents(t *testing.T) {
 							{ScVal: &xdr.ScVal{Type: xdr.ScValTypeScvSymbol, Sym: &counter}},
 							{ScVal: &xdr.ScVal{Type: xdr.ScValTypeScvU64, U64: &number}},
 						},
-					}},
+					},
+				},
 			},
 		})
 		require.NoError(t, err)
@@ -465,7 +466,7 @@ func TestGetEvents(t *testing.T) {
 		)
 
 		// strict topic length matching
-		results, err = handler.getEvents(context.TODO(), protocol.GetEventsRequest{
+		results, err = handler.getEvents(t.Context(), protocol.GetEventsRequest{
 			StartLedger: 1,
 			Filters: []protocol.EventFilter{
 				{
@@ -474,7 +475,8 @@ func TestGetEvents(t *testing.T) {
 							{ScVal: &xdr.ScVal{Type: xdr.ScValTypeScvSymbol, Sym: &counter}},
 							{ScVal: &xdr.ScVal{Type: xdr.ScValTypeScvU64, U64: &number}},
 						},
-					}},
+					},
+				},
 			},
 		})
 		require.NoError(t, err)
@@ -502,7 +504,8 @@ func TestGetEvents(t *testing.T) {
 							{ScVal: &xdr.ScVal{Type: xdr.ScValTypeScvU64, U64: &number}},
 							{Wildcard: &doubleStar},
 						},
-					}},
+					},
+				},
 			},
 		})
 		require.NoError(t, err)
--- a/cmd/stellar-rpc/internal/methods/get_events_test.go
+++ b/cmd/stellar-rpc/internal/methods/get_events_test.go
@@ -379,6 +379,7 @@ func TestGetEvents(t *testing.T) {
 	})
 
 	t.Run("filtering by topic, flexible length matching", func(t *testing.T) {
+		doubleStar := "**"
 		dbx := newTestDB(t)
 		ctx := context.TODO()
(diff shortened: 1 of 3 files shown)


leighmcculloch:
> 2\. At least one segment is still required (existing behavior)
>     Filters must contain at least one non-** segment.
>     Invalid: `["**"]` ❌
>     Valid: `["*", "**"]` ✅ - matches events with one or more topics.

This seems like an unnecessary restriction given that you can get `["**"]` functionality by omitting the topic filter. Could we support this case so that the field works consistently for all positions it can be specified in?

leighmcculloch:
> 3\. The existing limit of max of 4 filter segments remains the same. The `"**"` wildcard does not count toward this 4 segment limit.

Does this mean that `["*", "*", "*", "**"]` won't match an event that has 5 topics?

urvisavla:
> Does this mean that `["*", "*", "*", "**"]` won't match an event that has 5 topics?

No, it will match any event that  >= 3 topics.

> Or does `["*", "*", "*", "*"]` already match events with 4+ topics?

No, it does not.

> 3. The existing limit of max of 4 filter segments remains the same. The "**" wildcard does not count toward this 4 segment limit.

By this I meant that, the existing limit of max of 4 filter segments per filter is unchanged. The "**" wildcard is exempt from this count. so `["A","B","C","D","**"]` is a valid filter where as `["A","B","C","D","*"]` is not.

Also, per the [docs](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getEvents#:~:text=List%20of%20topic%20filters.%20If%20omitted%2C%20query%20for%20all%20events.%20If%20multiple%20filters%20are%20specified%2C%20events%20will%20be%20included%20if%20they%20match%20any%20of%20the%20filters.%20Maximum%205%20filters%20are%20allowed%20per%20request.
) the max limit for filter segments is 5 while the code currently enforces a limit of 4. So either the docs are wrong or it's been implemented incorrectly.

urvisavla:
> > 2. At least one segment is still required (existing behavior)
> > Filters must contain at least one non-** segment.
> > Invalid: `["**"]` ❌
> > Valid: `["*", "**"]` ✅ - matches events with one or more topics.
> 
> This seems like an unnecessary restriction given that you can get `["**"]` functionality by omitting the topic filter. Could we support this case so that the field works consistently for all positions it can be specified in?

yes, we can do that for the sake of consistency but using [**] is effectively the same as no filter at all.

Shaptic (approved):
Something is wonky with the commit history here but otherwise this LGTM! I'm surprised there were no query changes necessary but I guess the fact that it's an `OR` clause for topic filters helps!

urvisavla:
> Something is wonky with the commit history here but otherwise this LGTM!

Yikes! Fixed it.

PR #427, opened 2025-05-06
Allow segment filters to be serializable

### LINKED ISSUE #404: GetEventsRequest doesn't serialize SegmentFilter struct into RPC friendly string
### What version are you using?
Installed go package using `go get github.com/stellar/stellar-rpc@rpcclient-v0.1.0`

### What did you do?
Here's a simple test to call `getEvents` using a filter where I'm trying to filter events with first topic as `Symbol("EventSymbol")`. This request errors out with `[-32602] invalid parameters` error. Calling `Valid()` on the request params passes so it's not an issue with malformed params.

```
	t.Run("GetEvents Test", func(t *testing.T) {
		// Create the ScVal for the event symbol
		sym := xdr.ScSymbol("EventSymbol")
		eventScVal := xdr.ScVal{
			Type: xdr.ScValTypeScvSymbol,
			Sym:  &sym,
		}

		getEventsRequest := protocol.GetEventsRequest{
			StartLedger: 390799,
			EndLedger:   390800,
			Filters: []protocol.EventFilter{
				{
					EventType: protocol.EventTypeSet{
						protocol.EventTypeContract: nil,
					},
					Topics: []protocol.TopicFilter{
						{
							protocol.SegmentFilter{
								ScVal: &eventScVal,
							},
						},
					},
				},
			},
			Pagination: &protocol.PaginationOptions{
				Limit: 10000,
			},
		}

		err := getEventsRequest.Valid(10000)
		require.NoError(t, err)

		events, err := client.GetEvents(context.Background(), getEventsRequest)
		require.NoError(t, err)
		require.NotEmpty(t, events, "Should have found events")
	})
```

I tried to debug what's happening and I think it's a serialisation issue with the SegmentFilter.

 I placed a breakpoint here and the value of `bits` string looks like 

```
"{\"startLedger\":390799,\"endLedger\":390800,\"filters\":[{\"type\":\"contract\",\"topics\":[[{\"Wildcard\":null,\"ScVal\":{\"Type\":15,\"B\":null,\"Error\":null,\"U32\":null,\"I32\":null,\"U64\":null,\"I64\":null,\"Timepoint\":null,\"Duration\":null,\"U128\":null,\"I128\":null,\"U256\":null,\"I256\":null,\"Bytes\":null,\"Str\":null,\"Sym\":\"EventSymbol\",\"Vec\":null,\"Map\":null,\"Address\":null,\"NonceKey\":null,\"Instance\":null}}]]}],\"pagination\":{\"limit\":10000}}"
```

I think it's because of how `SegmentFilter` is being marshalled as struct instead of a base64 XDR string?

Expected to get back Events. The call works perfectly if I remove Topic filters. If the Request params were indeed malformed, the `Valid` function should've errored

### What did you see instead?

Invalid parameters error by the RPC.

Shaptic:
### What
Allow `SegmentFilter`s to be serializable for client SDK usage.

### Why
Closes #404

### Known Limitations
This will have to be updated for Protocol 23 due to https://github.com/stellar/stellar-rpc/pull/419, but I'm targeting `main` so we can tag that off faster to unblock clients.

### DIFF
--- a/protocol/get_events.go
+++ b/protocol/get_events.go
@@ -267,8 +267,8 @@ func (t TopicFilter) Matches(event []xdr.ScVal) bool {
 }
 
 type SegmentFilter struct {
-	Wildcard *string
-	ScVal    *xdr.ScVal
+	Wildcard *string    `json:"-"`
+	ScVal    *xdr.ScVal `json:"-"`
 }
 
 func (s *SegmentFilter) Matches(segment xdr.ScVal) bool {
@@ -317,6 +317,19 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
 		s.ScVal = &out
 	}
 	return nil
+}
+
+func (s SegmentFilter) MarshalJSON() ([]byte, error) {
+	if err := s.Valid(); err != nil {
+		return nil, err
+	}
+
+	if s.Wildcard != nil {
+		return []byte("\"*\""), nil
+	}
+
+	scv, err := xdr.MarshalBase64(s.ScVal)
+	return []byte(fmt.Sprintf("\"%s\"", scv)), err
 }
 
 type PaginationOptions struct {
--- a/protocol/get_events_test.go
+++ b/protocol/get_events_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/xdr"
 )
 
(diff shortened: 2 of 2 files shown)


protocol/get_events.go
```diff
@@ -267,8 +267,8 @@ func (t TopicFilter) Matches(event []xdr.ScVal) bool {
 }
 
 type SegmentFilter struct {
-	Wildcard *string
-	ScVal    *xdr.ScVal
+	Wildcard *string    `json:"-"`
+	ScVal    *xdr.ScVal `json:"-"`
```
leighmcculloch, line 271: I believe adding the "-" is unnecessary. Doing so tells the default marshal implementation to skip marshalling the field. However, the type has a custom marshal function, so the default impl does not read these tags. Is that correct or do they still get used somehow?

Ref: https://pkg.go.dev/encoding/json#Marshal

Shaptic: Yeah, I think you’re right. I added these while first developing prior to the interface implementation just to hide ‘em. I personally think it’s helpful as a reader to see that and think “okay lemme find how this is actually serialized,” but also happy to remove it due to its redundancy.

protocol/get_events.go
```diff
@@ -319,6 +319,19 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
 	return nil
 }
 
+func (s SegmentFilter) MarshalJSON() ([]byte, error) {
+	if err := s.Valid(); err != nil {
+		return nil, err
+	}
+
+	if s.Wildcard != nil {
+		return []byte(`"*"`), nil
```
urvisavla, line 328: If you change this to `return []byte(fmt.Sprintf("%s", *s.Wildcard)), nil` then it should work with #419 without any additional changes, right?

protocol/get_events.go
```diff
@@ -319,6 +319,19 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+	}
+
+	scv, err := xdr.MarshalBase64(s.ScVal)
+	return []byte(fmt.Sprintf(`"%s"`, scv)), err
```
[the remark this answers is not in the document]
Shaptic: It needs to be wrapped in `"` marks to be a JSON string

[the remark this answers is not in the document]
urvisavla: How about using `json.Marshal(scv)` instead?

protocol/get_events.go
urvisavla: actually, how about using `json.Marshal(*s.Wildcard)` instead?

protocol/get_events.go
```diff
@@ -319,6 +319,19 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
...
+	if err := s.Valid(); err != nil {
+		return nil, err
+	}
+
+	if s.Wildcard != nil {
+		return fmt.Appendf(nil, `"%s"`, *s.Wildcard), nil
+	}
+
+	scv, err := xdr.MarshalBase64(s.ScVal)
+	return fmt.Appendf(nil, `"%s"`, scv), err
```
urvisavla, line 332: How about using json.Marshal(scv) instead?

protocol/get_events.go
```diff
@@ -319,6 +319,19 @@ func (s *SegmentFilter) UnmarshalJSON(p []byte) error {
 	return nil
 }
 
+func (s SegmentFilter) MarshalJSON() ([]byte, error) {
+	if err := s.Valid(); err != nil {
+		return nil, err
+	}
+
+	if s.Wildcard != nil {
+		return fmt.Appendf(nil, `"%s"`, *s.Wildcard), nil
```
urvisavla, line 328: How about using json.Marshal(*s.Wildcard) instead?

protocol/get_events.go
Shaptic: Brilliant :brain: the linter should've suggested that instead of `fmt.Appendf` :roll_eyes: done in 5f0a009

urvisavla (approved):

Issue #413, opened 2025-04-22
p32: API/documentation change in the getLedgerEntries response of archived entries


2opremio:
In protocol 23 the TTL of archived entries won't be kept anymore. Thus, it's not possible to return it in the response of `getLedgerEntries`. In fact, in the new HTTP API of Core, the  TTL is not returned for archived entreis.

For now, we are returning a TTL placeholder of 0 for archived entries in `getLedgerEntries`.

However, we should consider either documenting that or refining the API to reflect archived entries explicitly.

Shaptic:
We discussed this later, iirc, btw, that it should be `state: string` over `archived: bool` in case we need to expand it (e.g., `"live" | "archived" | "new"`).

I don't think there's a defined user flow besides just bubbling up accurate information from Core. We could artificially omit archived entries entirely, I guess, but that seems intentionally limiting.

leighmcculloch:
> We discussed this later, iirc, btw, that it should be `state: string` over `archived: bool` in case we need to expand it (e.g., `"live" | "archived" | "new"`).

Makes sense. Could you share more of what was designed?

Would the "new" state ever be visible? "new" means the entry can't be returned. Entries not found are currently not returned from this API. For example:

```console
$ http https://soroban-testnet.stellar.org jsonrpc=2.0 id=1 method=getLedgerEntries 'params[keys][0]=AAAABgAAAAHMA/50/Q+w3Ni8UXWm/trxFBfAfl6De5kFttaMT0/ACwAAABAAAAABAAAAAgAAAA8AAAAHQ291bnRlcgAAAAASAAAAAAAAAAAg4dbAxsGAGICfBG3iT2cKGYQ6hK4sJWzZ6or1C5v6GAAAAAE='

{
    "id": "1",
    "jsonrpc": "2.0",
    "result": {
        "entries": [],
        "latestLedger": 956372
    }
}
```

Or would this API expose a limited set of states, such as `live` and `archived`?

leighmcculloch:
> We could artificially omit archived entries entirely

I think the RPC should include archived entries. Some use cases come to mind:

- Looking up balances of infrequently used accounts that just don't make sense to restore.

- Fork testing in the contract SDK so that someone could simulate in a forked test using as much ledger data as possible that would be available through restores.

leighmcculloch:
> I don't think there's a defined user flow

I think it would be reasonable to not add any new field, either `archived` or `state`, and to serve both live and archived state without communicating which it is.

[...]

PR #430, opened 2025-05-13
Add `archived` field to getLedgerEntries response

leighmcculloch:
### What
  Add archived boolean field to getLedgerEntries response for entries supporting archival. When true, the entry is archived and has no liveUntilLedgerSeq. When false, the entry is live with a liveUntilLedgerSeq. If omitted, the entry does not support archival.

  ### Why
  Reflect in the API that a ledger entry is not live and not available to contracts without being restored either automatically or manually.


leighmcculloch:
Closing because of https://github.com/stellar/stellar-rpc/issues/413#issuecomment-2878526113.

PR #432, opened 2025-05-13
Add support for non-root authorization in simulation

Shaptic:
### What
* Bumps `soroban-env-host-prev` to P22 and `-curr` to P23
* Splits the `GoLedgerStorage` trait implementations for Protocol 22 (which uses `SnapshotSourceWithArchive`) and Protocol 23 (which just uses `SnapshotSource`)
* Introduces a new, optional parameter to `simulateTransaction`: 
```typescript
interface SimulateTransactionRequest {
  authMode?: "" | "enforce" | "record" | "record_allow_nonroot"
}
```
which explicitly allows users to opt-in to non-root authorization in recording mode.

The behavior rules are:
* default (omitted): `enforce` if auth entries are present, `record` otherwise (the behavior today)
* `enforce`: always enforcement mode, even with an empty list
* `record`: always recording mode, failing if any auth exists
* `record_allow_nonroot`: like `record` but allowing non-root authorization

### Why
Handle the breaking changes across the Protocol 22 and 23 simulation libraries cleanly.
Add non-root simulation support for both protocols per #344.

### Known limitations
Integration testing isn't possible without an accompanying Core build.


cmd/stellar-rpc/internal/methods/simulate_transaction.go
```diff
@@ -285,17 +285,60 @@ func NewSimulateTransactionHandler(logger *log.Entry,
...
+					request.AuthMode = protocol.AuthModeRecord
+				}
+			} else if hasAuth && (request.AuthMode == protocol.AuthModeRecord ||
+				request.AuthMode == protocol.AuthModeRecordAllowNonroot) {
+				return protocol.SimulateTransactionResponse{
+					Error: fmt.Sprintf(
+						"cannot set authMode to '%s' with an auth footprint",
+						request.AuthMode,
+					),
+				}
```
leighmcculloch, line 325: The PR description says:
> * `record`: always recording mode, ignoring the auth entry list

But it appears that in this code it doesn't ignore the auth entry list, it errors if an auth entry list is provided with a request for recording mode.

I prefer the implementation error in this case, because there has been some discussion ([slack](https://stellarfoundation.slack.com/archives/C0766BREE6S/p1746498480916609?thread_ts=1746487456.265639&cid=C0766BREE6S)) about changing recording mode in the future to support identifying additional auths in the presence of some auths being provided.

If this mode didn't error in the has auth and recording case, someone might depend on recording mode discarding and recreating auths, so I think it's good that the impl errors in this case. In the future if the above idea for supporting finding additional auths gets implemented, this error can be removed, or a new auth mode can be added.

Shaptic: Oops yeah, you're right - I forgot to update the description after writing https://github.com/stellar/stellar-rpc/issues/344#issuecomment-2881335866, 100% agreed on all counts :+1:

.github/workflows/stellar-rpc.yml
```diff
@@ -66,6 +66,11 @@ jobs:
           C:/msys64/usr/bin/pacman.exe -S mingw-w64-x86_64-gcc --noconfirm
           echo "CC=C:/msys64/mingw64/bin/gcc.exe" >> $GITHUB_ENV
           echo "C:/msys64/mingw64/bin" >> $GITHUB_PATH
+          CONFIG=$(cat << EOF >> .cargo/config.toml
+          [target.windows-gnu]
+          linker = "gcc"
+          EOF
+          )
```
2opremio, line 73: Can you revert this? The windows build is being addressed at #443

.github/workflows/stellar-rpc.yml
```diff
@@ -77,19 +82,25 @@ jobs:
...
+
       - uses: stellar/actions/rust-cache@main
       - run: make build-libs
         env:
           CARGO_BUILD_TARGET: ${{ matrix.rust_target }}
 
       - name: Build Stellar RPC reproducible build
         run: |
-          go build -trimpath -buildvcs=false ./cmd/stellar-rpc
+          go build -trimpath -ldflags="-v -linkmode=external" -buildvcs=false ./cmd/stellar-rpc
```
2opremio, line 97: why is this needed?

.github/workflows/stellar-rpc.yml
```diff
@@ -77,19 +82,25 @@ jobs:
...
           ls -lh stellar-rpc
           file stellar-rpc
         env:
           CGO_ENABLED: 1
+          CGO_LDFLAGS: -v
+          GOTOOLCHAIN: local
```
2opremio, line 103: Is this for the windows build?

protocol/simulate_transaction_test.go
```diff
@@ -0,0 +1,20 @@
+package protocol
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSimulatingNonRootAuth(t *testing.T) {
```
2opremio, line 10: Nice!

cmd/stellar-rpc/lib/preflight/src/lib.rs
```diff
@@ -35,23 +35,84 @@ extern crate soroban_simulation_prev;
...
+                    // Store the internal error in the storage as the info won't
+                    // be propagated from simulation.
+                    if let Ok(mut err) = self.internal_error.try_borrow_mut() {
+                        *err = Some(e);
+                    }
+                    // Errors that occur in storage are not recoverable, so we
+                    // force host to halt by passing it an internal error.
+                    Err((ScErrorType::Storage, ScErrorCode::InternalError).into())
+                }
+            }
```
2opremio, line 74: Can you factor this out? (since it's identical in for `soroban_simulation::SnapshotSourceWithArchive` as well)

cmd/stellar-rpc/lib/preflight/src/shared.rs
```diff
@@ -299,14 +310,25 @@ fn ledger_entry_diff_vec_to_c(modified_entries: &[LedgerEntryDiff]) -> CXDRDiffV
     CXDRDiffVector { array, len }
 }
 
+impl From<u32> for AuthMode {
```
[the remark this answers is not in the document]
Shaptic: good idea, but I did want to make sure _only_ the cross-boundary type (`uint32_t`) would convert like this, just in case

.github/workflows/stellar-rpc.yml
Shaptic: yeah, buncha debugging commits - reverted

cmd/stellar-rpc/lib/preflight/src/lib.rs
Shaptic: I thought I could, but it doesn't work because `ScErrorType` and `ScErrorCode` come from different XDR libraries (`prev::xdr` and `curr::xdr`) for the two implementations :cry:

Shaptic: jk I figured it out :sunglasses: see 9492eb4

2opremio: Nice!

2opremio (approved):

PR #433, opened 2025-05-14
Add test for get ledger entry without liveUntilLedgerSeq

leighmcculloch:
### What
  Add a test for retrieving account ledger entry which does not have liveUntilLedgerSeq field.

  ### Why
  When working on https://github.com/stellar/stellar-rpc/pull/430 I couldn't see a test case for getting ledger entries that don't have a liveUntilLedgerSeq.


tamirms (approved):

cmd/stellar-rpc/internal/integrationtest/get_ledger_entries_test.go
```diff
@@ -120,7 +132,7 @@ func TestGetLedgerEntriesSucceeds(t *testing.T) {
 	require.Equal(t, xdr.LedgerEntryTypeContractCode, firstEntry.Type)
 	require.Equal(t, infrastructure.GetHelloWorldContract(), firstEntry.MustContractCode().Code)
 
-	require.Positive(t, result.Entries[1].LastModifiedLedger)
```
tamirms, line 123: the linter is suggesting to use `require.Positive()` instead of `require.Greater()`

leighmcculloch: Yup, and if I use require.Positive, it says to use require.Greater.

leighmcculloch: Example is here: https://stellarfoundation.slack.com/archives/C02B04RMK/p1747313010371909?thread_ts=1747312845.342179&cid=C02B04RMK

.github/workflows/golang.yml
```diff
@@ -35,6 +35,6 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v7
         with:
-          version: v2.0.2 # this is the golangci-lint version
+          version: v2.1.6 # this is the golangci-lint version
```
leighmcculloch, line 38: Updating golangci-lint to get testifylint v1.6.1 that contains this fix:
- https://github.com/Antonboom/testifylint/pull/236

The version fixes an issue where testifylint was reporting false-positives on this PR.

PR #435, opened 2025-05-15
Include liveUntilLedgerSeq only for entries that have one

leighmcculloch:
### What
  Include liveUntilLedgerSeq only for entries with a TTL. Remove the zero TTL for archived entries.

  ### Why
  The code was incorrectly using a zero placeholder for archived entries, but at @Shaptic points out in https://github.com/stellar/stellar-rpc/issues/413#issuecomment-2880926910 that doesn't make much sense. Instead of returning a value, since archived entries have no TTL entry, omit the field.


leighmcculloch:
There wasn't an integration test for the archived case, and it's not obvious to me how to add an integration test for that case. Do we have testing infra / functions to help setup an entry in an archived state? I haven't added a test for the moment.

2opremio:
@leighmcculloch do you plan to revive this for protocol 23? The time is now! :)

leighmcculloch:
This change had unintended consequences that are incompatible with the `soroban-simulation` library.

The simulation library requires that all entries being restored have a live until, even though core exposes archived entries to rpc without a live until:
- https://github.com/stellar/rs-soroban-env/blob/7dd8a1c45692aa154e95b95f918b4a7af0560db9/soroban-simulation/src/resources.rs#L207-L209
- https://github.com/stellar/rs-soroban-env/blob/7dd8a1c45692aa154e95b95f918b4a7af0560db9/soroban-simulation/src/snapshot_source.rs#L90-L93

I am pretty confident the requirement in the simulation library does not need to be preserved. Until now the requirement has had no consequence because until now rpc always had a live until for an entry, and in the p23 branch was passing around `0` for a live until instead of omitting the value.

I've opened these two changes on the simulation library to allow for simulating restore and auto-restore with an archived entry that has no live until:
- https://github.com/stellar/rs-soroban-env/pull/1565
   - Would need to be released in a new rc of env (would not require a new rc of core)
- https://github.com/stellar/rs-soroban-env/pull/1567
   - Would need to be released in: https://github.com/stellar/rs-soroban-env/pull/1566

---

I think modifying the simulation library addresses the problem at the source. However, this area of code and the assumptions around the presence of the live-until even though it was optional makes it difficult to know if we're missing anything. I'd like @dmkozh / @sisuresh eyes on it. If they agree with the change then we can release the two versions of the simulation lib to merge this PR into the RPC.

There are alternative ways to address this issue in rpc only, without making a change to the simulation library:

Rpc could store `nil` (this PR) but translate `nil` into `0` (new change) when passing the ledger entry to the simulation library.

Rpc could store `0` (discard this PR), and translate `0` into `nil` at the api boundary (a new change).

cc @2opremio @Shaptic

leighmcculloch:
Closing because Core is going to start emitting `0` as the live until value for archived entries:
- https://github.com/stellar/stellar-core/issues/4766

In the long run Core and RPC APIs should converge, so there's really nothing else to do here since RPC already exposes the `0` in the case of an archived entry.

It's not what I would describe as an optimal API to have magic zero values, but it is reasonable.

Issue #425, opened 2025-05-06
Enable `getLedgers` to proxy to Galexie for requests outside it's retention window


mollykarcher:
### What problem does your feature solve?

From https://github.com/orgs/stellar/discussions/1717:

> The concept of a "full archive node" is commonplace in other ecosystems. Typically when dealing with partners that support many different chains, they expect something like this to be available. At present, RPC is not architected in a way that would support retaining full history in it's database, but by-and-large people want to use the same API for historical and current data. So ultimately, we'd like to provide access to historical data through RPC's API (or an RPC-like API) to ease the integration burden as much as possible.

While we explore a different, potentially longer-term solution in https://github.com/stellar/stellar-rpc/issues/424 this option (#1 from the above referenced github discussion) would be significantly faster to implement and give us a better solution for partners to utilize in the shorter-term cases. Given that goldsky is currently working on exposing a public data lake and plans to launch within the next 4-8 weeks, this would make this solution particularly attractive as people would not need to run any additional services/infrastructure.

### What would you like to see?

- When `getLedgers` requests are made for ledgers outside of RPC's retention window, it proxies the request to a Datastore (if configured) and returns the result
- The `getLedgers` response is modified to return both the transactions and events as top-level response fields
- TBD what we do with `latestLedger` and `oldestLedger` response fields when a Datastore is configured/available

### What alternatives are there?

See options 2-5 on https://github.com/orgs/stellar/discussions/1717

urvisavla:
This is complete and merged though there are some follow-up items listed in #453.

PR #437, opened 2025-05-16
Add support for external datastore as source of historical ledger data

urvisavla:
### What

This PR addresses [#425](https://github.com/stellar/stellar-rpc/issues/425) -  adding support for handling getLedgers requests that fall outside the rpc's local ledger retention window by proxies out of range requests to the datastore.

- Adds support for serving historical ledgers from a datastore when the requested range isn’t available locally.
- Uses BufferedStorageBackend to read from the datastore.
- The `gen-config-file` command now includes default `datastore_config` and `buffered_storage_backend_config` sections (commented out by default).
```
# Buffered storage backend configuration for reading ledgers from the datastore.
# [buffered_storage_backend_config]
  # buffer_size = 100
  # num_workers = 10
  # retry_limit = 3
  # retry_wait = "30s"

# External datastore configuration including type, bucket name and schema.
# [datastore_config]
  # type = "GCS"

  # [datastore_config.params]
    # destination_bucket_path = "path_to_bucket"

  # [datastore_config.schema]
    # files_per_partition = 64000
    # ledgers_per_file = 1
```

- Adds a new flag `serve_ledger_from_datastore`. The flag used instead of just relying on the presence of a datastore config because same datastore config can in future be also be used for ingestion from a datastore, so this flag control over which behavior is enabled.

- Includes basic unit tests. 

### Why
https://github.com/stellar/stellar-rpc/issues/425

### Known limitations

1. Currently, this assumes that the datastore has all ledgers starting from genesis. This is because there's no way yet to query datastore what range of ledgers it actually has. That limitation will be fixed once this [stellar/go#5498](https://github.com/stellar/go/issues/5498)) is done. We will then also use the real range to fill in the latestLedger and oldestLedger fields more accurately.

2. The `datastore_config` and `buffered_storage_backend_config`  can only be set through the config file and not with command-line flags. That’s because these settings can be complex or different depending on the type of datastore being used (like S3 or GCS). Supporting them in the config file keeps things simpler and more flexible.

### DIFF
--- a/cmd/stellar-rpc/internal/config/options.go
+++ b/cmd/stellar-rpc/internal/config/options.go
@@ -544,17 +544,7 @@ func (cfg *Config) options() Options {
 			ConfigKey: &cfg.BufferedStorageBackendConfig,
 			Usage:     "Buffered storage backend configuration for reading ledgers from the datastore.",
 			CustomSetValue: func(option *Option, i interface{}) error {
-				tree, ok := i.(*toml.Tree)
-				if !ok {
-					return fmt.Errorf("expected TOML table for buffered_storage_backend_config, got %T", i)
-				}
-
-				tomlBytes, err := toml.Marshal(tree.ToMap())
-				if err != nil {
-					return fmt.Errorf("failed to marshal buffered_storage_backend_config: %w", err)
-				}
-
-				return toml.Unmarshal(tomlBytes, option.ConfigKey)
+				return unmarshalTOMLTree(i, option.ConfigKey, "buffered_storage_backend_config")
 			},
 			MarshalTOML: func(_ *Option) (interface{}, error) {
 				tomlBytes, err := toml.Marshal(defaultBufferedStorageBackendConfig())
@@ -569,17 +559,7 @@ func (cfg *Config) options() Options {
 			ConfigKey: &cfg.DataStoreConfig,
 			Usage:     "External datastore configuration including type, bucket name and schema.",
 			CustomSetValue: func(option *Option, i interface{}) error {
-				tree, ok := i.(*toml.Tree)
-				if !ok {
-					return fmt.Errorf("expected TOML table for datastore_config, got %T", i)
-				}
-
-				tomlBytes, err := toml.Marshal(tree.ToMap())
-				if err != nil {
-					return fmt.Errorf("failed to marshal datastore_config: %w", err)
-				}
-
-				return toml.Unmarshal(tomlBytes, option.ConfigKey)
+				return unmarshalTOMLTree(i, option.ConfigKey, "datastore_config")
 			},
 			MarshalTOML: func(_ *Option) (interface{}, error) {
 				tomlBytes, err := toml.Marshal(defaultDataStoreConfig())
@@ -611,6 +591,20 @@ func defaultDataStoreConfig() datastore.DataStoreConfig {
 			FilesPerPartition: 64000,
 		},
 	}
+}
+
+func unmarshalTOMLTree(tree interface{}, out interface{}, configName string) error {
+	t, ok := tree.(*toml.Tree)
+	if !ok {
+		return fmt.Errorf("expected TOML table for %s, got %T", configName, tree)
+	}
+
+	tomlBytes, err := toml.Marshal(t.ToMap())
+	if err != nil {
+		return fmt.Errorf("failed to marshal TOML tree for %s: %w", configName, err)
+	}
+
+	return toml.Unmarshal(tomlBytes, out)
 }
 
 type missingRequiredOptionError struct {
--- a/cmd/stellar-rpc/internal/config/options.go
+++ b/cmd/stellar-rpc/internal/config/options.go
@@ -2,6 +2,7 @@
 package config
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -9,9 +10,10 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/pelletier/go-toml"
 	"github.com/sirupsen/logrus"
-
 	"github.com/stellar/go/network"
+	"github.com/stellar/go/support/datastore"
 	"github.com/stellar/go/support/strutils"
 )
 
@@ -529,6 +531,75 @@ func (cfg *Config) options() Options {
 			Usage:        "The maximum duration of time allowed for processing a getFeeStats request. When that time elapses, the rpc server would return -32001 and abort the request's execution",
 			ConfigKey:    &cfg.MaxGetFeeStatsExecutionDuration,
 			DefaultValue: 5 * time.Second,
+		},
+		{
+			Name:         "serve-ledgers-from-datastore",
+			TomlKey:      strutils.KebabToConstantCase("serve-ledgers-from-datastore"),
+			Usage:        "Fetch historical ledgers from the datastore if they're not available locally.",
+			ConfigKey:    &cfg.ServeLedgersFromDatastore,
+			DefaultValue: false,
+		},
+		{
+			TomlKey:   "buffered_storage_backend_config",
+			ConfigKey: &cfg.BufferedStorageBackendConfig,
+			Usage:     "Buffered storage backend configuration for reading ledger data from the datastore.",
+			CustomSetValue: func(option *Option, i interface{}) error {
+				tree, ok := i.(*toml.Tree)
+				if !ok {
+					return fmt.Errorf("invalid type for buffered_storage_backend_config:"+
+						" expected toml table, got %T", i)
+				}
+
+				tomlBytes, err := toml.Marshal(tree.ToMap())
+				if err != nil {
+					return fmt.Errorf("unable to serialize buffered_storage_backend_config: %w", err)
+				}
+
+				return toml.Unmarshal(tomlBytes, option.ConfigKey)
+			},
+			MarshalTOML: func(_ *Option) (interface{}, error) {
+				tomlBytes, err := toml.Marshal(cfg.BufferedStorageBackendConfig)
+				if err != nil {
+					return nil, fmt.Errorf("unable to marshal buffered_storage_backend_config to toml: %w", err)
+				}
+				return toml.LoadBytes(tomlBytes)
+			},
+		},
+		{
+			TomlKey:   "datastore_config",
+			ConfigKey: &cfg.DataStoreConfig,
+			Usage:     "External datastore configuration including type, bucket name and schema.",
+			CustomSetValue: func(option *Option, i interface{}) error {
+				tree, ok := i.(*toml.Tree)
+				if !ok {
+					return fmt.Errorf("invalid type for datastore_config: expected toml table, got %T", i)
+				}
+
+				tomlBytes, err := toml.Marshal(tree.ToMap())
+				if err != nil {
+					return fmt.Errorf("unable to serialize datastore_config: %w", err)
+				}
+
+				return toml.Unmarshal(tomlBytes, option.ConfigKey)
+			},
+			MarshalTOML: func(_ *Option) (interface{}, error) {
+				tomlBytes, err := toml.Marshal(cfg.DataStoreConfig)
+				if err != nil {
+					return nil, fmt.Errorf("unable to marshal datastore_config: %w", err)
+				}
+				return toml.LoadBytes(tomlBytes)
+			},
+			Validate: func(_ *Option) error {
+				if cfg.ServeLedgersFromDatastore {
+					// performs a basic check to verify credentials, bucket name and access.
+					dataStore, err := datastore.NewDataStore(context.Background(), cfg.DataStoreConfig)
+					if err != nil {
+						return fmt.Errorf("failed to initialize datastore: %w", err)
+					}
+					defer dataStore.Close()
+				}
+				return nil
+			},
 		},
 	}
 	return *cfg.optionsCache
--- a/cmd/stellar-rpc/internal/config/options.go
+++ b/cmd/stellar-rpc/internal/config/options.go
@@ -2,7 +2,6 @@
 package config
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -13,7 +12,6 @@ import (
 	"github.com/pelletier/go-toml"
 	"github.com/sirupsen/logrus"
 	"github.com/stellar/go/network"
-	"github.com/stellar/go/support/datastore"
 	"github.com/stellar/go/support/strutils"
 )
 
@@ -588,17 +586,6 @@ func (cfg *Config) options() Options {
 					return nil, fmt.Errorf("unable to marshal datastore_config: %w", err)
 				}
 				return toml.LoadBytes(tomlBytes)
-			},
-			Validate: func(_ *Option) error {
-				if cfg.ServeLedgersFromDatastore {
-					// performs a basic check to verify credentials, bucket name and access.
-					dataStore, err := datastore.NewDataStore(context.Background(), cfg.DataStoreConfig)
-					if err != nil {
-						return fmt.Errorf("failed to initialize datastore: %w", err)
-					}
-					defer dataStore.Close()
-				}
-				return nil
 			},
 		},
 	}
--- a/cmd/stellar-rpc/internal/config/options.go
+++ b/cmd/stellar-rpc/internal/config/options.go
@@ -11,7 +11,9 @@ import (
 
 	"github.com/pelletier/go-toml"
 	"github.com/sirupsen/logrus"
+	"github.com/stellar/go/ingest/ledgerbackend"
 	"github.com/stellar/go/network"
+	"github.com/stellar/go/support/datastore"
 	"github.com/stellar/go/support/strutils"
 )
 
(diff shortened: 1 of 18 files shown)


cmd/stellar-rpc/internal/config/options.go
```diff
@@ -530,6 +532,75 @@ func (cfg *Config) options() Options {
 			ConfigKey:    &cfg.MaxGetFeeStatsExecutionDuration,
 			DefaultValue: 5 * time.Second,
 		},
+		{
+			Name:         "serve-ledgers-from-datastore",
```
Shaptic, line 536: Isn't this implied by the presence of a datastore config?

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -530,6 +532,75 @@ func (cfg *Config) options() Options {
...
+			TomlKey:      strutils.KebabToConstantCase("serve-ledgers-from-datastore"),
+			Usage:        "Fetch historical ledgers from the datastore if they're not available locally.",
+			ConfigKey:    &cfg.ServeLedgersFromDatastore,
+			DefaultValue: false,
+		},
+		{
+			TomlKey:   "buffered_storage_backend_config",
+			ConfigKey: &cfg.BufferedStorageBackendConfig,
+			Usage:     "Buffered storage backend configuration for reading ledger data from the datastore.",
```
Shaptic, line 545: wait is this supposed to be a filename or an actual toml string?

Shaptic: we could provide reasonable defaults across the board instead of requiring thisto be user-set

cmd/stellar-rpc/internal/datastore/ledger_reader.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, fmt.Errorf("failed to create data store: %w", err)
+	}
+	defer dataStore.Close()
+
+	// Initialize the buffered storage backend
+	bufferedBackend, err := ledgerbackend.NewBufferedStorageBackend(lr.storageBackendConfig, dataStore)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create buffered storage backend: %w", err)
+	}
+	defer bufferedBackend.Close()
```
Shaptic, line 49: Do these need to be re-initialized every time a fetch occurs? I guess it's not super costly but it does seem weird to do every time.

cmd/stellar-rpc/internal/datastore/ledger_reader.go
```diff
@@ -0,0 +1,68 @@
...
+	if err := bufferedBackend.PrepareRange(ctx, ledgerRange); err != nil {
+		return nil, err
+	}
+
+	// Fetch each ledger in the range
+	ledgers := make([]xdr.LedgerCloseMeta, 0, end-start+1)
+	for sequence := ledgerRange.From(); sequence <= ledgerRange.To(); sequence++ {
+		ledger, err := bufferedBackend.GetLedger(ctx, sequence)
+		if err != nil {
+			return nil, err
```
Shaptic, line 62: suggestion: we could return `ledgers, err` to at least give back whatever progress we made

Shaptic (commented):
Just a lil' drive by review :car:

urvisavla:
> Just a lil' drive by review 🚗

Thanks @Shaptic these are helpful comments. This work is still in progress so I'll address them once it's more complete.

cmd/stellar-rpc/internal/config/options.go
urvisavla: done.

cmd/stellar-rpc/internal/config/options.go
urvisavla: reason
>Adds a new flag serve_ledger_from_datastore. The flag used instead of just relying on the presence of a datastore config because same datastore config can in future be also be used for ingestion from a datastore, so this flag control over which behavior is enabled.

cmd/stellar-rpc/internal/datastore/ledger_reader.go
urvisavla: Good point. We could return the ledgers and let the caller decide.

In our case though partial results may not be ideal because it can potentially create gaps. For eg, if someone asks for ledgers 5-14 and we only have local data for 10-14, the data store partially returns 5-7 there will be a gap between 7-10. But I'm open to suggestions.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -64,6 +66,7 @@ type Daemon struct {
 	closeError          error
 	done                chan struct{}
 	metricsRegistry     *prometheus.Registry
+	dataStore           datastore.DataStore
```
sreuland, line 69: The daemon could maintain instance of `rpcdatastore.LedgerReader`, and `MustNew` takes care of creating the datastore and the reader together. `createJSONRPCHandler` wouldn't need to handle the datastore reader creation separately.

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -32,7 +32,7 @@ type LedgerReader interface {
 type LedgerReaderTx interface {
 	GetLedger(ctx context.Context, sequence uint32) (xdr.LedgerCloseMeta, bool, error)
 	GetLedgerRange(ctx context.Context) (ledgerbucketwindow.LedgerRange, error)
-	BatchGetLedgers(ctx context.Context, sequence uint32, batchSize uint) ([]xdr.LedgerCloseMeta, error)
+	BatchGetLedgers(ctx context.Context, sequence uint32, end uint32) ([]xdr.LedgerCloseMeta, error)
```
sreuland, line 35: for consistency, is it worthwhile to rename `sequence` to `start`? `ledgerReaderTx` implementation names this parameter in that way also.

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -45,36 +52,54 @@ func (h ledgersHandler) getLedgers(ctx context.Context, request protocol.GetLedg
...
+	if h.datastoreLedgerReader != nil {
+		dsRange, err := h.datastoreLedgerReader.GetAvailableLedgerRange(ctx)
+		if err != nil {
+			// log error but continue using local ledger range
+			h.logger.WithError(err).Error("failed to get available ledger range from datastore")
+		} else {
+			// extend available range to include datastore
+			availableLedgerRange.FirstLedger = min(dsRange.FirstLedger, availableLedgerRange.FirstLedger)
+		}
+	}
```
sreuland, line 76: Noticing a design choice for consideration. Rather than the handler layer carrying the know-how of when to use internal db vs. datastore here in `getLedgers` and `fetchLedgers` could this logic be encapsulated lower in the data access Reader layer?

datastore_layer

ideally this makes it transparent at the handler and would not have to change.

cmd/stellar-rpc/internal/rpcdatastore/ledger_reader.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return ledgers, nil
+}
+
+// GetAvailableLedgerRange returns the assumed available ledger range.
+// TODO: Support fetching the actual range from the datastore.
+func (r *ledgerReader) GetAvailableLedgerRange(_ context.Context) (protocol.LedgerSeqRange, error) {
+	return protocol.LedgerSeqRange{
+		FirstLedger: 2, // Assume datastore holds all ledgers from genesis.
```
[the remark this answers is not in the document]
urvisavla: Are you suggesting something along these lines https://github.com/stellar/go/issues/5498?

cmd/stellar-rpc/internal/daemon/daemon.go
urvisavla: doing it in createJSONRPCHandler just to keep things consistent since the db.LedgerReader is also created there. Also I thought having the datastore in the daemon might be useful later for something like health checks. But if you think it's cleaner, we can create and store rpcdatastore.LedgerReade in the daemon.

cmd/stellar-rpc/internal/rpcdatastore/ledger_reader.go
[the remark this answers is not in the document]
Shaptic: ah yes exactly lol :)

cmd/stellar-rpc/internal/daemon/daemon.go
sreuland: Ok, what stood out is the repetitive check on `cfg.ServeLedgersFromDatastore` to create datastore and reader separately at different times. The suggestion would consolidate to `mustCreateDataStore` which is used by `GetDB`. 
This is minor suggestion for consideration. I see the existing creation pattern you referenced and seems fine either way. Also, correction on my prior comment, I mentioned `MustNew` but meant `mustCreateDataStore`.

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -107,39 +132,104 @@ func (h ledgersHandler) initializePagination(request protocol.GetLedgersRequest,
...
+		if err != nil {
+			return nil, &jrpc2.Error{
+				Code:    jrpc2.InternalError,
+				Message: fmt.Sprintf("error fetching ledgers from datastore: %v", err),
+			}
+		}
+		return ledgers, nil
+	}
+
+	var ledgers []xdr.LedgerCloseMeta
```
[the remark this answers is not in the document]
urvisavla: done

cmd/stellar-rpc/internal/rpcdatastore/ledger_reader.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+// GetLedgers retrieves a contiguous batch of ledgers in the range [start, end] (inclusive)
+// from the configured datastore using a buffered storage backend.
+// Returns an error if any ledger in the specified range is unavailable.
+func (r *ledgerReader) GetLedgers(ctx context.Context, start, end uint32) ([]xdr.LedgerCloseMeta, error) {
+	bufferedBackend, err := r.ledgerBackendFactory.NewBufferedBackend(r.storageBackendConfig, r.dataStore)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create buffered storage backend: %w", err)
```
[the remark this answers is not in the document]
urvisavla: I don't think you can join a string and an error using errors.Join.

cmd/stellar-rpc/internal/db/ledger.go
urvisavla: done

cmd/stellar-rpc/internal/datastore/ledger_reader.go
urvisavla: Single instance of BufferedStorageBackend works with only one PrepareRange. It does not support multiple PrepareRange calls with different ranges so a new instance must be created for each range

cmd/stellar-rpc/internal/datastore/ledger_reader.go
sreuland: +1 for not retuning partial ledgers when an error occurs. It implies more complexity on the caller to handle and it seems like an error implies data won't be deterministic in any case.

cmd/stellar-rpc/internal/methods/get_ledgers.go
urvisavla: I tried to move the logic for getting available ledger ranges (from both the local db and datastore) and the fetchLedgers logic into a new `UnifiedLedgerReader` data access layer but it's turning out to be more complex than I expected.

Here's the proposed UnifiedLedgerReader interface:
```
type UnifiedLedgerReader interface {
    // get available range combining local db and datastore
    GetAvailableRange(ctx context.Context) (protocol.LedgerSeqRange, error)

    // get ledgers in the requested range
    GetLedgers(ctx context.Context, start, end uint32) ([]xdr.LedgerCloseMeta, error)
}
```
-------------------------------------------------------------------------------
The first problem is how we manage db Tx along with request validation.
The handler needs to:
- Get the combined ledger range (local db + datastore) for validation (e.g. pagination, cursor etc). 
- Then call GetLedgers to fetch the actual data.

The problem is that fetching the ledger range and then the ledgers themselves must be in the same db Tx. If the handler is responsible for maintaining the Tx and passing it to the UnifiedReader, it defeats the purpose of encapsulating data access to UnifiedLedgerReader.

To address this, we could  add Tx management methods (like `NewTx` and `CloseTx`) directly to UnifiedLedgerReader to keep the handler free from Tx management.

```
    // manages the lifecycle of the db transaction internally
    NewTx(ctx context.Context) error
    CloseTx() error
```

-------------------------------------------------------------------------------------------------------

Another bigger Issue is that we haven't yet implemented the logic to fetch the ledger range (including oldest/latest sequence and timestamps) from the data store.

Right now, we're juggling two different ranges:
- One  from the local db for response fields (like OldestLedgerCloseTime and LatestLedgerCloseTime).
- Another combined range (local + data store) used for request validation and determining the ledger range to return.

This means UnifiedLedgerReader would either return both ranges separately or return a combined range but without CloseTime info from the datastore and then we’d have to use hacks like hardcoding it or leaving LedgerCloseTime empty. Neither approach feels quite right.

Given these complexities, especially datastore range fetching, I think it might be better to postpone this refactor until we have a proper way to retrieve the accurate oldest/latest ledger and timestamps directly from the datastore. That will allow for a much cleaner and more robust implementation of UnifiedLedgerReader.  
Open to other suggestions.

tamirms:
Now that there are several stellar-core flags (diagnostic events, classic cap-67 events, etc) which affect the contents of ledger close meta, I think we should consider the scenario where ledgers in the datastore are inconsistent with the ledgers served locally from rpc.

This issue can be addressed separate from this PR but I think we'll need to introduce some way to determine which features are enabled in the ledgers found from the data lake vs the rpc captive core. https://github.com/stellar/go/issues/5507 could be relevant

cmd/stellar-rpc/internal/methods/get_ledgers.go
sreuland: I see, yeah, the tx ensures read consistency between range check and data retrieval from db in relation to sliding retention window. 

Related to the tx aspect, also noticed the current change effectively inlines the remote datastore ledgers request as synchronous task during existing open db reader tx session from the handler. Is there concern for this potentially leading to long(er) running tx and cache locks due to remote datastore request latency?

Could try to mitigate this in current changeset if concerned such as move the `LedgerReaderTx.Done()` from deferred to instead be right after [GetBatchLedgers invocation](https://github.com/stellar/stellar-rpc/pull/437/files#diff-1a02c0b74383b9a669f91c41ddbc9dea98fd6e5e8d9c993d3659d549afd082e4R165)?

Understood if want to postpone `UnifiedLedgerReader`. The concept could avoid the tx conflation also since it would internally orchestrate the db.LedgerReaderTX usage and close the tx asap before doing the datastore.GetLedgers. 

One other design idea that may work and get around the tx juggling could be to allow handler to pass range validation as a call back so UnifiedReader can just internally (1)create db/datastore readers, (2)calc the ranges, (3)run validations, (4)query for ledgers and lastly manage LedgerReaderTX lifecycle(call `Done()` asap):
```
type LedgerReaderPrepareRange func (protocol.LedgerSeqRange) (start int, limit int, err Error)

type UnifiedLedgerReader interface {
    // get ledgers using the requested range returned from prepareRangeCb
    GetLedgers(ctx context.Context, prepareRangeCb LedgerReaderPrepareRange ) ([]xdr.LedgerCloseMeta, error)
}
```

Composite Ledger Reader, multiple datastores (2)

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -45,36 +52,54 @@ func (h ledgersHandler) getLedgers(ctx context.Context, request protocol.GetLedg
...
+	ledgers, err := h.fetchLedgers(ctx, start, end, request.Format, readTx, ledgerRange.ToLedgerSeqRange())
 	if err != nil {
 		return protocol.GetLedgersResponse{}, err
 	}
 	cursor := strconv.Itoa(int(ledgers[len(ledgers)-1].Sequence))
 
 	return protocol.GetLedgersResponse{
-		Ledgers:               ledgers,
+		Ledgers: ledgers,
+		//	TODO: update these fields using ledger range from datastore
```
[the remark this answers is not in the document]
sreuland: Can you describe a client use case which benefits from the breakout of ledgers? 

This pushes forward an internal aspect of RPC, totally understand it given the possible difference of meta content [mentioned](https://github.com/stellar/stellar-rpc/pull/437#issuecomment-2916511754). It seems like this introduces additional work for clients to handle and possibly some confusion. 

Ideally, the outcome from sourcing other backends for ledger metadata by RPC should be transitive at the client API layer, i.e. data provided by RPC is from the network, nothing more?

One thought..Can the potential difference of ledger close meta between RPC or datastores be mitigated if Galexie were to always enable all captive core flags that affect ledger close meta? In that way a datastore would always provide the superset of meta avoiding missing data cases when interleaved with meta from any other source.

mollykarcher:
> Now that there are several stellar-core flags (diagnostic events, classic cap-67 events, etc) which affect the contents of ledger close meta, I think we should consider the scenario where ledgers in the datastore are inconsistent with the ledgers served locally from rpc.

This is a really good point. Another way to solve this might be, that if you have a datastore configured as your backend, RPC's ledger retention window gets automatically set to 0, so RPC is effectively just acting as a lightweight facade/API layer on top of the data lake. I think I would be more comfortable with doing that if `getTransactions`/`getEvents` were also being served from the data lake though. Though there is precedence in other ecosystems (at least, ethereum) for some endpoints not working on when a node is running in archival mode.

PR #442, opened 2025-05-21
Update Golang XDR definitions to protocol 23

2opremio:


cmd/stellar-rpc/internal/ingest/service_test.go
```diff
@@ -98,12 +98,11 @@ func createTestLedger(t *testing.T) xdr.LedgerCloseMeta {
...
-			TxSet:                          createTransactionSet(),
-			TxProcessing:                   createTransactionProcessing(t),
-			UpgradesProcessing:             []xdr.UpgradeEntryMeta{},
-			EvictedTemporaryLedgerKeys:     []xdr.LedgerKey{createEvictedTempLedgerKey()},
-			EvictedPersistentLedgerEntries: []xdr.LedgerEntry{createEvictedPersistentLedgerEntry()},
+			LedgerHeader:       createLedgerHeader(),
+			TxSet:              createTransactionSet(),
+			TxProcessing:       createTransactionProcessing(t),
+			UpgradesProcessing: []xdr.UpgradeEntryMeta{},
+			EvictedKeys:        []xdr.LedgerKey{createEvictedTempLedgerKey()},
```
Shaptic, line 105: We should still add a `createEvictedPersistentLedgerEntry()`, right?

cmd/stellar-rpc/internal/preflight/preflight_test.go
```diff
@@ -98,24 +98,41 @@ var mockLedgerEntriesWithoutTTLs = []xdr.LedgerEntry{
...
-					TxMaxWriteLedgerEntries:        100,
-					TxMaxWriteBytes:                100,
-					FeeReadLedgerEntry:             100,
-					FeeWriteLedgerEntry:            100,
-					FeeRead1Kb:                     100,
-					BucketListTargetSizeBytes:      100,
-					WriteFee1KbBucketListLow:       1,
-					WriteFee1KbBucketListHigh:      1,
-					BucketListWriteFeeGrowthFactor: 1,
+					// TODO(fons): adjust
```
Shaptic, line 105: Is this TODO for this PR or later?

cmd/stellar-rpc/internal/methods/simulate_transaction.go
```diff
@@ -385,8 +385,16 @@ func getBucketListSizeAndProtocolVersion(
 	if !ok {
 		return 0, 0, fmt.Errorf("missing meta for latest ledger (%d)", latestLedger)
 	}
-	if closeMeta.V != 1 {
+	switch closeMeta.V {
+	case 1:
+		return uint64(closeMeta.V1.TotalByteSizeOfLiveSorobanState),
+			uint32(closeMeta.V1.LedgerHeader.Header.LedgerVersion),
+			nil
+	case 2:
```
Shaptic, line 393: Should we also handle ingestion of `TransactionMetaV4` in this PR or separately?

cmd/stellar-rpc/internal/ingest/service_test.go
2opremio: I don't think so, since the field is not used anymore

cmd/stellar-rpc/internal/preflight/preflight_test.go
2opremio: Ahh, I should address that before merging. Thanks

cmd/stellar-rpc/internal/methods/simulate_transaction.go
2opremio: Uh, separately, I would think. Once we add integration tests for protocol 23

cmd/stellar-rpc/internal/ingest/service_test.go
Shaptic: I thought the semantics have changed: that is, `EvictedKeys` will have both temporary and persistent entries now.

cmd/stellar-rpc/internal/methods/simulate_transaction.go
Shaptic: Yep makes sense :+1:

cmd/stellar-rpc/internal/ingest/service_test.go
2opremio: I don't think the entries were verified in the test.

Issue #414, opened 2025-04-22
Design: Determine how to handle event interleaving


Shaptic:
In [CAP-67](https://stellar.org/protocol/cap-67), events will occur at two cadences:

* Operation-level: these will be present in an array of arrays, `OperationMetaV2`. Each array corresponds to the events generated by the respective operation in the transaction. For Soroban transactions, this will be an array of length 1.
* Transaction-level: this will be an array with one or two events. In the Soroban case, there will be an additional "refunded fee" event when there is a refund.

In RPC, we made an assumption about the event ID that latched onto the 1:1 notion of Soroban transaction:operations. Namely, the ID / cursor that is used for pagination and querying looks like this:

```
<toid>-<event index>
```

where the `<event-index>` iterates directly through the `DiagnosticEvent[]` array. However, in order to enable event consumers to see a chronologically accurate picture of value movement in CAP-67, we need to process events in a particular, different order:

* tx-level fee charge
* operation-level events
* tx-level fee refund

This means iteration order needs to change and the ID needs to support paginating in this fashion.

Relevant discussion: https://github.com/stellar/stellar-protocol/discussions/1553#discussioncomment-12914103

(cc @tamirms: please edit with any missing details!)

leighmcculloch:
> Namely, the ID / cursor that is used for pagination and querying looks like this: `<toid>-<event index>` where the `<event-index>` iterates directly through the `DiagnosticEvent[]` array.

Does this mean that the presence of diagnostic events in the `DiagnosticEvent[]` array affects the index of non-diagnostic contract events? Diagnostic events are not guaranteed to be consistent across any core release, so it will be possible for diagnostic events to be added anywhere or removed that would change the indexes of events if there index is relative to other diagnostic events.

leighmcculloch:
Relevant to this discussion is this proposal to remove diagnostic events from the `getEvents` method output, which would also affect IDs generated:
- https://github.com/stellar/stellar-rpc/issues/402

Shaptic:
> Diagnostic events are not guaranteed to be consistent across any core release, so it will be possible for diagnostic events to be added or removed that would change the indexes of contract events if the contract events index is relative to other diagnostic events.

[...]

PR #447, opened 2025-05-27
Bump Core image and test protocol 23

2opremio:


cmd/stellar-rpc/internal/db/event.go
```diff
@@ -89,12 +89,29 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
 
-		txEvents, err := tx.GetDiagnosticEvents()
+		allEvents, err := tx.GetTransactionEvents()
 		if err != nil {
 			return err
 		}
+		diagEvents := allEvents.DiagnosticEvents
+
+		// Transform the other events to diagnostic events for processing
+		for _, opEvents := range allEvents.OperationEvents {
```
karthikiyer56, line 99: doing this blanket check will result in double counting.
**For smart contract transactions/operations**, txMetaV3 and txMetaV4 will already include contract events in the diagnostic events.
so perhaps, some check like so  `if len(diagEvents) = 0`  is warranted before running this loop - `for _, opEvents := range allEvents.OperationEvents` ?

2opremio: Is that the case only for operation events? How about transaction events?

2opremio: OK, it should be fixed now

2opremio: PTAL though

cmd/stellar-rpc/internal/db/transaction.go
```diff
@@ -235,17 +235,25 @@ func ParseTransaction(lcm xdr.LedgerCloseMeta, ingestTx ingest.LedgerTransaction
...
+
+	tx.Events = make([][]byte, 0, len(diagEvents))
+	for i, event := range diagEvents {
+		bytes, ierr := event.MarshalBinary()
+		if ierr != nil {
+			return tx, fmt.Errorf("couldn't encode transaction DiagnosticEvent %d: %w", i, ierr)
 		}
-	} else {
-		return tx, fmt.Errorf("couldn't encode transaction DiagnosticEvents: %w", diagErr)
+		tx.Events = append(tx.Events, bytes)
```
karthikiyer56, line 256: is this right?
you are preallocating tx.Events to be of size `len(diagEvents)`
So assuming there are 5 diagEvents, this `append` will add them to the end and make the eventual size of txEvents to be 10, no?
this shud be `tx.Events[i] = bytes` perhaps?

cmd/stellar-rpc/internal/db/event.go
```diff
@@ -53,6 +53,32 @@ func NewEventReader(log *log.Entry, db db.SessionInterface, passphrase string) E
...
+			for _, event := range opEvents {
+				diagEvents = append(diagEvents, xdr.DiagnosticEvent{
+					InSuccessfulContractCall: true,
+					Event:                    event,
+				})
+			}
+		}
+	}
+
+	for _, event := range allEvents.TransactionEvents {
```
karthikiyer56, line 72: JFYI, this ordering is not chronological (as we want it to be)
There will perhaps need to be some code added to account for that, which will have to check TxMetaVersion and the `TransactionEventStage` and do some reordering of events in this function

2opremio: That will be handled in #414

cmd/stellar-rpc/internal/db/event.go
```diff
@@ -53,6 +53,32 @@ func NewEventReader(log *log.Entry, db db.SessionInterface, passphrase string) E
 	return &eventHandler{log: log, db: db, passphrase: passphrase}
 }
 
+func transactionEventsIntoDiagnosticEvents(allEvents ingest.TransactionEvents) []xdr.DiagnosticEvent {
+	diagEvents := allEvents.DiagnosticEvents
+
+	// For smart contract transactions/operations, txMetaV3 and txMetaV4 will already include
+	// contract events in the diagnostic events.
+	if len(diagEvents) == 0 {
```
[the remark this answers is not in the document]
urvisavla: got it. could you please point me to the thread? thanks!

cmd/stellar-rpc/internal/db/event.go
```diff
@@ -89,12 +115,14 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
 		transactionHash := tx.Result.TransactionHash[:]
 
-		txEvents, err := tx.GetDiagnosticEvents()
+		allEvents, err := tx.GetTransactionEvents()
 		if err != nil {
 			return err
 		}
 
-		if len(txEvents) == 0 {
+		diagEvents := transactionEventsIntoDiagnosticEvents(allEvents)
```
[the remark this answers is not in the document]
2opremio: I don't think it's worth to test this since it's going to change with #414

cmd/stellar-rpc/internal/feewindow/feewindow.go
```diff
@@ -167,10 +167,21 @@ func (fw *FeeWindows) IngestFees(meta xdr.LedgerCloseMeta) error {
 		if len(ops) == 1 {
 			switch ops[0].Body.Type { //nolint:exhaustive
```
karthikiyer56, line 168: Just a callout. Dont need to refactor in this PR.

There is a function now that tells you if a tx is soroban or not
`LedgerTranaction.IsSorobanTx`
Perhaps that can be leveraged here and in other places instead of the `len(ops)` check + `InvokeHostFunction` check

cmd/stellar-rpc/internal/integrationtest/infrastructure/docker/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     # Note: Please keep the image pinned to an immutable tag matching the Captive Core version.
     #       This avoids implicit updates which break compatibility between
     #       the Core container and captive core.
-    image: ${CORE_IMAGE:-stellar/unsafe-stellar-core:22.1.1-2291.a50f3f919.focal-do-not-use-in-prd}
+    image: ${CORE_IMAGE:-stellar/unsafe-stellar-core:22.3.1-2509.c2e465a3e.focal-do-not-use-in-prd}
```
karthikiyer56, line 16: I know this is picked up from here: https://stellarfoundation.slack.com/archives/C02B04RMK/p1748924162965139?thread_ts=1748537685.780399&cid=C02B04RMK
isnt the name supposed to have `23.xxx` ?

PR #449, opened 2025-05-27
Fix json parsing bug for "**" wildcard in topic filters

urvisavla:
### What

### Why

### Known limitations

[TODO or N/A]


2opremio:
Wouldn't it make more sense to target the main branch? (since it's not protocol-23 specific AFAICT)

urvisavla:
> Wouldn't it make more sense to target the main branch? (since it's not protocol-23 specific AFAICT)
> 
> (we will merge the protocol-23 branch into main before releasing anyways)

The original issue https://github.com/stellar/stellar-rpc/issues/358 was tagged protocol-23 so I added the feature in protocol-23 branch. This feature doesn't exist in main yet.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Added
+
+- Added "**" wildcard to the `getEvents` endpoint, enabling flexible topic matching without manual padding.
```
sreuland, line 7: should this briefly mention the `**` usage is only allowed as last or only topic?

urvisavla: updated.

PR #461, opened 2025-06-24
Reset event indices at ledger level over transaction level

Shaptic:
### What
The transaction-level event indices (`beforeIndex` and `afterIndex`) should never be reset, because this would create conflicts across cursors from transaction to transactions. Namely, notice the way the cursor is built:

```go
insertedEvent.Cursor = protocol.Cursor{
	Ledger: lcm.LedgerSequence(),
	Tx:     0, // min value
	Op:     0,
	Event:  beforeIndex,
}.String()
```

This would make the cursor identical for every event in the `TransactionEventStageTransactionEventStageBeforeAllTxs` for a ledger because `beforeIndex` would get reset for each transaction. However, these events are at the ledger level, so they need to be preserved in the outer loop. The same is the case for `...AfterAllTxs`.

### Why
Ingestion would halt on the uniqueness constraint of events otherwise.

This did not appear in integration testing because the ledger only contained a single transaction :facepalm:

### DIFF
--- a/cmd/stellar-rpc/internal/db/event.go
+++ b/cmd/stellar-rpc/internal/db/event.go
@@ -104,8 +104,9 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
 	//  - Post-application events have a TOID with { ledger seq, -1, 0 }
 	// where -1 is actually the largest possible uint32.
 	//
+	var beforeIndex, afterIndex uint32
 	insertableEvents := []dbEvent{}
-
+	
 	for {
 		var tx ingest.LedgerTransaction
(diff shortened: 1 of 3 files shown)


urvisavla (commented):
Fix lgtm! Could we update this [unit test]( https://github.com/stellar/stellar-rpc/blob/a21a7c36fef2e9d5fe65df6d7ef46237a037b614/cmd/stellar-rpc/internal/db/event_test.go#L138) to cover multiple transactions with events just to be safe?

Shaptic:
@urvisavla good call! Fixed in 7b79c1d, this wasn't showing up in unit tests because they were using meta v3 :facepalm:

urvisavla (approved):

PR #463, opened 2025-06-25
Enforce autorestore for simulations in protocols > 22

2opremio:
### What

Enforce autorestore for simulations in protocols > 22, by including the required `archivedEntries` vector in the `SorobanResourcesExtV0` of the `SorobanTransactionData` returned by the `simulateTransaction` method.

### Why

Autorestore wasn't working in the simulation of protocol 23

### Known limitations

The restore preamble won't be generated in the corner cases in which autorestore isn't enough. (See https://github.com/stellar/stellar-rpc/issues/464 )

### DIFF
--- a/cmd/stellar-rpc/lib/preflight/src/shared.rs
+++ b/cmd/stellar-rpc/lib/preflight/src/shared.rs
@@ -180,6 +180,7 @@ pub(crate) fn preflight_invoke_hf_op_or_maybe_panic(
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn preflight_invoke_hf_op_post_autorestore_or_maybe_panic(
     go_storage: &Rc<GoLedgerStorage>,
     network_config: &NetworkConfig,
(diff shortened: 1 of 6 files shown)


cmd/stellar-rpc/lib/preflight/src/shared.rs
```diff
@@ -156,25 +153,118 @@ pub(crate) fn preflight_invoke_hf_op_or_maybe_panic(
...
+    go_storage: &Rc<GoLedgerStorage>,
+    network_config: &NetworkConfig,
+    adjustment_config: &SimulationAdjustmentConfig,
+    ledger_info: &LedgerInfo,
+    hf: HostFunction,
+    auth_mode: RecordingInvocationAuthMode,
+    source_account: &AccountId,
+    enable_debug: bool,
+) -> Result<CPreflightResult> {
+    // TODO: A restore preamble should be generated when network limits are surpassed.
```
dmkozh, line 194: nit: I think this is better suited as an issue and todo could just refer to it.

dmkozh (commented):
Rust part LGTM

cmd/stellar-rpc/lib/preflight/src/shared.rs
2opremio: Done! https://github.com/stellar/stellar-rpc/issues/464

PR #469, opened 2025-07-09
Expose all diagnostic events in the events object

Shaptic:
### What
Do not filter out contract-type events from the diagnostic event list.

### Why
[Slack thread.](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1751916530430629)

### Known limitations
n/a


urvisavla (approved):
lgtm!

PR #471, opened 2025-07-10
v23.0.0 Release Candidate 2

Shaptic:
Holding release branch for rc.2 changes.


overcat:
Hi @Shaptic, I would like to ask if we will update `stellar/go` to the latest version when we release RPC v23?

2opremio (approved):

Shaptic:
@overcat ah good call! I opened https://github.com/stellar/go/pull/5751.

Shaptic:
@overcat we're in the process of making changes to the S3 backend so that it can access read-only buckets plus add testing for robustness, etc. This improvement would only make it into the next release candidate(s). However, I can make RPC's rc.2 point to the above PR's branch for this RC if the feature is important enough to get out :call_me_hand:

overcat:
> @overcat we're in the process of making changes to the S3 backend so that it can access read-only buckets plus add testing for robustness, etc. This improvement would only make it into the next release candidate(s). However, I can make RPC's rc.2 point to the above PR's branch for this RC if the feature is important enough to get out 🤙

This is not urgent; I think we can postpone it.

Issue #464, opened 2025-06-26
simulateTransaction: re-enable emitting restore preambles after protocol 23


2opremio:
Protocol 23 introduced autorestore, which, for most cases, makes explicit restore operations (and the restore preamble returned by `simulateTransaction`)  redundant.

https://github.com/stellar/stellar-rpc/pull/463 enabled autorestore for `simulateTransaction` for protocol > 22 but disabled restore preambles.

However, there are corner cases (when network limits are surpassed) for which autorestore isn't enough. In those cases we need to split restoring into autorestore+restore preamble.

To implement this we should include this changes in the `libpreflight` Rust library:

1. run simulation using regular (non-auto-restore) snapshot source. the result will contain auto-restored entries (if any) in `SorobanTransactionData` . The result will also have proper fees and limits for autorestoration.
2. verify the output `SorobanTransactionData` against the network limits.
      a. if it is below limits, return the result from step 1
      b. if it is above limits, re-run simulation using `AutoRestoringSnapshotSource` and populate the restore preamble. now you'll have an invoke transaction without any auto-restored entries, and a restore transaction.

note, that step 2b doesn't guarantee that the original transaction without autorestore, or the restore transaction don't exceed the network limits. which is why I suggest to postpone that until after we have a more coherent design for handling the network limits in simulation. thus the algorithm can be shortcut to just step 1.

PR #481, opened 2025-07-25
Upgrade Soroban limits instead of using configuration parameters

Shaptic:
### What
This removes the usage of Core configuration flags related to testing, namely:

```toml
TESTING_MINIMUM_PERSISTENT_ENTRY_LIFETIME=10
TESTING_SOROBAN_HIGH_LIMIT_OVERRIDE=true
```

in favor of actually upgrading network limits via Core (see `upgradeLimits`). The settings (both `testnet` and `unlimited`) explicitly set the `min_persistent_ttl` setting to 10. By default, it will set unlimited limits, but we need limits to be enforced for a pending test case.

### Why
We need to enforce limits for the edge case of autorestore where the restore+invoke transaction exceeds limits and needs to be separated into restore then invoke (pending) as part of #464.

### Known limitations
n/a

### DIFF
--- a/cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
+++ b/cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
@@ -155,14 +155,25 @@ func NewTest(t testing.TB, cfg *TestConfig) *Test {
 		i.protocolVersion = cfg.ProtocolVersion
 		i.sqlitePath = cfg.SQLitePath
 		i.captiveCoreStoragePath = cfg.CaptiveCoreStoragePath
+		parallel = !cfg.NoParallel
+		i.datastoreConfigFunc = cfg.DatastoreConfigFunc
+
 		if cfg.OnlyRPC != nil {
 			i.onlyRPC = true
 			i.testPorts.TestCorePorts = cfg.OnlyRPC.CorePorts
 			shouldWaitForRPC = !cfg.OnlyRPC.DontWait
+
+			// If we're running RPC in a container, we don't apply limits.
+			if cfg.ApplyLimits != nil && *cfg.ApplyLimits != "" {
+				i.t.Logf("ApplyLimits ('%s') isn't allowed with OnlyRPC", *cfg.ApplyLimits)
+				i.t.FailNow()
+			}
+
+			noUpgrade := ""
+			i.limitFile = &noUpgrade
+		} else {
+			i.limitFile = cfg.ApplyLimits
 		}
-		parallel = !cfg.NoParallel
-		i.datastoreConfigFunc = cfg.DatastoreConfigFunc
-		i.limitFile = cfg.ApplyLimits
 	}
 
 	if i.sqlitePath == "" {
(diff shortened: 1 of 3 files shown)


Shaptic:
```
running command stellar [contract build] 

stellar: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by stellar)
stellar: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by stellar)
stellar: /lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.15' not found (required by stellar)
```

This is some weird CI bs that's gonna take a year and a day to investigate, so we'll just have to let it slide for now as this is higher priority.

cmd/stellar-rpc/internal/integrationtest/migrate_test.go
```diff
@@ -14,10 +14,14 @@ import (
...
-// Test that every Stellar RPC version (within the current protocol) can migrate cleanly to the current version
-// We cannot test prior protocol versions since the Transaction XDR used for the test could be incompatible
+// Test that every Stellar RPC version (within the current protocol) can migrate
+// cleanly to the current version. We cannot test prior protocol versions since
+// the Transaction XDR used for the test could be incompatible
+//
 // TODO: find a way to test migrations between protocols
 func TestMigrate(t *testing.T) {
+	t.Skip("Idk why it doesn't work but lemme check others")
+
```
tamirms, line 24: it looks like this still needs to be resolved

cmd/stellar-rpc/internal/xdr2json/conversion.go
```diff
@@ -63,16 +63,16 @@ func convertAnyBytes(xdrTypeName string, field []byte) (json.RawMessage, error)
 	// scope just added to show matching alloc/frees
 	{
```
tamirms, line 64: I think the scope is not necessary anymore now that you're using defer to free the memory

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -719,13 +756,141 @@ func (i *Test) CreateHelloWorldContract() (protocol.GetTransactionResponse, [32]
...
 func (i *Test) InvokeHostFunc(
 	contractID xdr.ContractId, method string, args ...xdr.ScVal,
 ) protocol.GetTransactionResponse {
 	op := CreateInvokeHostOperation(i.MasterAccount().GetAccountID(), contractID, method, args...)
 	return i.PreflightAndSendMasterOperation(op)
 }
 
+//nolint:funlen // it's a test bro relax
+func (i *Test) upgradeLimits() {
+	helper := func(limitFile string) string {
```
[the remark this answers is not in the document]
Shaptic: laziness :laughing: fixed!

cmd/stellar-rpc/internal/integrationtest/infrastructure/docker/upgrades/enable.json
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
Shaptic: brilliant! done

cmd/stellar-rpc/internal/integrationtest/migrate_test.go
Shaptic: Fixed it: when we restart RPC as part of a test (e.g. migration), we need to be sure to (a) not apply a settings upgrade again and (b) not use the Core config overrides since they're incompatible with doing the upgrade "for real."

PR #486, opened 2025-08-08
Bump CI to stable Core packages and prepare changelog

Shaptic:
self-explanatory: bump to stable and add our stable changelog


sreuland:
looks like dependency check issue a case of the v23 tag on stellar-xdr having grabbed one more later commit than the go protocol-23 has been built with:

https://github.com/stellar/stellar-xdr/commits/curr/

I put up [go/5779](https://github.com/stellar/go/pull/5779) to rebuild xdr off the newer commit, once that passes, would then need to update the go.mod here for stellar/go dependency to pick that up?

Shaptic:
@sreuland ty! I'll `go get` as soon as that branch merges.

PR #498, opened 2025-08-22
Add more fine-grained ingestion metrics

tamirms:
### What

### Why

### Known limitations

[TODO or N/A]


cmd/stellar-rpc/internal/db/db.go
```diff
@@ -269,7 +277,7 @@ func (rw *readWriter) NewTx(ctx context.Context) (WriteTx, error) {
 
 type writeTx struct {
 	globalCache            *dbCache
-	postCommit             func() error
+	postCommit             func(durationMetrics map[string]time.Duration) error
```
sreuland, line 280: can `durationMetrics` be a new member of writeTx struct instead of passing to the commit methods with side effects? Looks like [caller only creates one.instance per tx](https://github.com/stellar/stellar-rpc/pull/498/files#diff-3d9a88a5df59f56b1038c6920642ee764f653e2d07ce6b03dc8243748ae5c71aR209) and could be passed through `NewTx`?

sreuland: yes, it exposes the caller to duration metrics more so than if it were encapsulated in the writer. i don't feel strongly about it, it's compositional choice.

sreuland (approved):
lgtm!

PR #505, opened 2025-08-27
Optimize `getLedgers` queries by only decoding the header

Shaptic:
### What
Instead of decoding the raw binary data directly into an `xdr.LedgerCloseMeta`, this modifies the code to only decode the `xdr.LedgerHeaderHistoryEntry` out of the data which comes first in the layout. The rest of the meta is passed around directly as a `[]byte` until the final step, either converting it to base64 directly or to JSON.

Rudimentary benchmarks show a ~40% improvement over the baseline:

```
$ git switch main
$ go test -failfast -run=None -bench=BenchmarkBatchGetLedgers ./cmd/stellar-rpc/internal/db
goos: linux
goarch: amd64
pkg: github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db
cpu: 13th Gen Intel(R) Core(TM) i7-1360P
BenchmarkBatchGetLedgers-16    	git switch -    1788	    773248 ns/op
PASS
ok  	github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db	7.652s

$ git switch getLedgers_depthCheck
$ go test -failfast -run=None -bench=BenchmarkBatchGetLedgers ./cmd/stellar-rpc/internal/db
goos: linux
goarch: amd64
pkg: github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db
cpu: 13th Gen Intel(R) Core(TM) i7-1360P
BenchmarkBatchGetLedgers-16    	    1939	    530915 ns/op
PASS
ok  	github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db	3.325s
```

### Why
We have observed performance issues with `getLedgers` due to their density and would like to improve them.

### Known limitations
This needs to be evaluated against real load to see the improvement.

### DIFF
--- a/cmd/stellar-rpc/internal/methods/get_ledgers.go
+++ b/cmd/stellar-rpc/internal/methods/get_ledgers.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/creachadair/jrpc2"
 	"github.com/stellar/go/support/log"
-	"github.com/stellar/go/xdr"
 
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db"
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcdatastore"
@@ -37,8 +36,10 @@ func NewGetLedgersHandler(ledgerReader db.LedgerReader, maxLimit, defaultLimit u
 	}).getLedgers)
 }
 
-// getLedgers fetch ledgers and relevant metadata from DB and falling back to the remote rpcdatastore if necessary.
-func (h ledgersHandler) getLedgers(ctx context.Context, request protocol.GetLedgersRequest,
+// getLedgers fetch ledgers and relevant metadata from DB and falling back to
+// the remote rpcdatastore if necessary.
+func (h ledgersHandler) getLedgers(
+	ctx context.Context, request protocol.GetLedgersRequest,
 ) (protocol.GetLedgersResponse, error) {
 	readTx, err := h.ledgerReader.NewTx(ctx)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -159,99 +161,104 @@ func (h ledgersHandler) parseCursor(cursor string, ledgerRange protocol.LedgerSe
...
 		if err != nil {
-			return nil, &jrpc2.Error{
+			return &jrpc2.Error{
 				Code:    jrpc2.InternalError,
 				Message: fmt.Sprintf("error fetching ledgers from db: %v", err),
 			}
 		}
-		return ledgers, nil
+
+		return addToResult(ledgers, []xdr.LedgerCloseMeta{})
```
sreuland, line 212: nit, could pass nil instead of []xdr.LedgerCloseMeta{}, same for the empty slice allocation on line 230.

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -159,99 +161,104 @@ func (h ledgersHandler) parseCursor(cursor string, ledgerRange protocol.LedgerSe
...
-
-		ledgerInfo, err := h.parseLedgerInfo(ledger, format)
-		if err != nil {
-			return nil, &jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: fmt.Sprintf("error processing ledger %d: %v", ledger.LedgerSequence(), err),
-			}
-		}
-		result = append(result, ledgerInfo)
+		err = errors.Join(
```
sreuland, line 246: pulling out all the tricks!

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -159,99 +161,104 @@ func (h ledgersHandler) parseCursor(cursor string, ledgerRange protocol.LedgerSe
...
-	fetchFromDatastore := func(start uint32, end uint32) ([]xdr.LedgerCloseMeta, error) {
+	fetchFromDatastore := func(start, end uint32) error {
 		if h.datastoreLedgerReader == nil {
-			return nil, &jrpc2.Error{
+			return &jrpc2.Error{
 				Code:    jrpc2.InvalidParams,
 				Message: "datastore ledger reader not configured",
 			}
 		}
 		ledgers, err := h.datastoreLedgerReader.GetLedgers(ctx, start, end)
```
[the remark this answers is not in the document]
sreuland: That might be a neat feature request to consider at some point for `LedgerBackend` interface..something like `GetRawLedger(ctx, seq) (LedgerMetadataChunk, error)`

PR #515, opened 2025-09-30
Prepare v23.0.4 for release

Shaptic:
### What
Bumps soroban-env.

### Why
Bug fixes.

### Known limitations
n/a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,12 +15,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -31,15 +25,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.98"
+version = "1.0.100"
```
leighmcculloch, line 28: More deps are upgrading than I would have expected. Are they required or was that accidental when using the env dep?

leighmcculloch (approved):
One concern otherwise lgtm

Cargo.lock
Shaptic: Uhhh I just modified `soroban-env` then ran `cargo update`, so probably side-effects but everything still works :shrug:

Issue #503, opened 2025-08-26
Run rpc ingestion load tests with io2 EBS volume


tamirms:
The [preliminary results](https://stellarfoundation.slack.com/archives/C02B04RMK/p1755812711390049) from the last rpc ingestion load test shows that RPC cannot keep up with ingestion rates of 500 TPS.

However, these initial results were obtained using a [GP3 EBS volume](https://docs.aws.amazon.com/ebs/latest/userguide/general-purpose.html#gp3-ebs-volume-type) for the sqlite db. We have recently added support for provisioning io2 EBS volumes in our kubernetes cluster. So, we should work with ops to rerun the ingestion load tests with an io2 EBS volume to see if the performance improves.

PR #518, opened 2025-10-06
Add configuration options to enable load testing

Shaptic:
### What
This introduces a toml configuration parameter enabling passing a file with ledgers bundled in to perform load testing instead of Captive Core-based ingestion.

### Why
We need to evaluate ingestion performance with ledgers that are denser than what is currently available on pubnet, see #503 et al.

### DIFF
--- a/cmd/stellar-rpc/internal/config/main.go
+++ b/cmd/stellar-rpc/internal/config/main.go
@@ -89,7 +89,9 @@ type Config struct {
 	BufferedStorageBackendConfig ledgerbackend.BufferedStorageBackendConfig
 	DataStoreConfig              datastore.DataStoreConfig
 
-	LoadTestFile string
+	LoadTestFile           string
+	LoadTestMergingEnabled bool
+	LoadTestFrequency      time.Duration
 
 	// We memoize these, so they bind to pflags correctly
 	optionsCache *Options
--- a/cmd/stellar-rpc/internal/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/daemon/daemon.go
@@ -299,11 +299,20 @@ func createIngestService(cfg *config.Config, logger *supportlog.Entry, daemon *D
 			WithField("path", cfg.LoadTestFile).
 			Warnf("Ingestion will run with load testing")
 
-		backend = loadtest.NewLedgerBackend(loadtest.LedgerBackendConfig{
+		config := loadtest.LedgerBackendConfig{
 			NetworkPassphrase:   cfg.NetworkPassphrase,
 			LedgersFilePath:     cfg.LoadTestFile,
-			LedgerCloseDuration: time.Second * 5000,
-		})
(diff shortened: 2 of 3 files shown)


cmd/stellar-rpc/internal/config/main.go
```diff
@@ -64,22 +70,26 @@ type Config struct {
...
+	MaxGetLedgersExecutionDuration          time.Duration
+	MaxSendTransactionExecutionDuration     time.Duration
+	MaxSimulateTransactionExecutionDuration time.Duration
+	MaxGetFeeStatsExecutionDuration         time.Duration
+
+	ServeLedgersFromDatastore    bool
+	BufferedStorageBackendConfig ledgerbackend.BufferedStorageBackendConfig
+	DataStoreConfig              datastore.DataStoreConfig
+
+	LoadTestFile string
```
Shaptic, line 92: This is the only addition - the other changes are just grouping the parameters based on functionality rather than in a monolithic, aligned list.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -292,6 +293,19 @@ func createIngestService(cfg *config.Config, logger *supportlog.Entry, daemon *D
...
+	var backend ledgerbackend.LedgerBackend = daemon.core
+	if cfg.LoadTestFile != "" {
+		daemon.Logger().
+			WithField("path", cfg.LoadTestFile).
+			Warnf("Ingestion will run with load testing")
+
+		backend = loadtest.NewLedgerBackend(loadtest.LedgerBackendConfig{
+			NetworkPassphrase:   cfg.NetworkPassphrase,
+			LedgersFilePath:     cfg.LoadTestFile,
+			LedgerCloseDuration: time.Second * 5000,
```
tamirms, line 305: in horizon we have 2 more configuration parameters:

https://github.com/stellar/go/blob/master/services/horizon/cmd/ingest.go#L88-L112

- `merge` - configures whether we merge real ledgers with the synthetic ledgers in the load test
- `close-duration` - configures the close time. This is pretty important because we can use it to test if we're able to handle a faster close duration. We configure this value to 2 seconds for the horizon load tests

tamirms:
In horizon we introduced a separate `load-test` command which was useful because we could restore the db back to its previous state after the load test completed:

https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/main.go#L650-L669

I would recommend implementing the same pattern here.

There is also a `load-test-restore` command which is useful in case the load test crashes at some point in the middle of its run:

https://github.com/stellar/go/blob/master/services/horizon/cmd/ingest.go#L270

cmd/stellar-rpc/internal/daemon/daemon.go
Shaptic: Done! Added both flags and am targeting a feature branch now. Restoration will take more effort and I'll do it in a separate PR :+1:

PR #519, opened 2025-10-11
refactor: use strings.Builder to improve performance

stemlaud:
### What

use strings.Builder to improve performance

### Why

strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)

### Known limitations

N


stemlaud:
@Shaptic Hi, Could you please review this PR at your convenience? Thank you very much.

PR #540, opened 2025-11-19
Add `--network` parameter to prefill parameters with sane defaults per-network

cjonas9:
### What

The PR introduces a new optional `NETWORK` configuration parameter in RPC, which allows users to specify a preset Stellar network, `pubnet` or `testnet.` When `NETWORK` is set, RPC automatically adjusts the relevant parameters in the deployed instance's configuration and generates a corresponding captive core config file. The `NETWORK` parameter may be specified through the command line (usage: --network [preset]), the `.toml `config file, or ENV variables.

If the `NETWORK` parameter is specified, these fields are automatically populated based on the chosen network:

- `HISTORY_ARCHIVE_URLS`
- `NETWORK_PASSPHRASE`
- `CAPTIVE_CORE_CONFIG_PATH`: The default configuration file for pubnet and testnet is embedded in the Horizon binary and used to generate the default configuration.

If the `NETWORK` parameter isn't specified, `HISTORY_ARCHIVE_URLS` and `NETWORK_PASSPHRASE` are both mandatory. An error message is displayed if they're not supplied.

If the `NETWORK` parameter is specified but is not `"pubnet"` or `"testnet"`, an error message is displayed.

If the `NETWORK` parameter is specified and (`HISTORY_ARCHIVE_URLS` or `NETWORK_PASSPHRASE`) is supplied, an error message is displayed. This error message is not displayed if `HISTORY_ARCHIVE_URLS` and `NETWORK_PASSPHRASE` exactly match the values the `NETWORK` parameter would supply.

This PR also introduces a test in `main_test.go` that validates this behavior. Specifically, the test will verify that when the `network` option is provided, the `HISTORY_ARCHIVE_URLS` and `NETWORK_PASSPHRASE` fields are populated with their correct values. Further, it verifies that supplying the `network` option with `NETWORK_PASSPHRASE` errors.

### Why

This simplifies the user experience and enables users to rapidly choose a default network configuration without manually configuring `HISTORY_ARCHIVE_URLS` or `NETWORK_PASSPHRASE`. Users also no longer have to manually install a captive core configuration file for `pubnet` and `testnet`.

### Known limitations

[N/A]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
-
+### Added
+- Added support for the `network` configuration parameter, which allows users to specify a default Stellar network (`testnet` or `pubnet`) ([#540](https://github.com/stellar/stellar-rpc/pull/540)).
 
 ## [v24.0.0](https://github.com/stellar/stellar-rpc/compare/v23.0.4...v24.0.0): Protocol 24
 


cmd/stellar-rpc/internal/config/main_test.go
```diff
@@ -85,3 +86,63 @@ func TestConfigLoadFlagsDefaultValuesOverrideExisting(t *testing.T) {
...
+	networkFlagOptions := generateNetworkParameters()
+
+	for _, networkFlagOption := range networkFlagOptions {
+		var cfg Config
+
+		// Part confirming network option writes historyArchiveURLs and networkPassphrase
+		cmd := &cobra.Command{}
+		require.NoError(t, cfg.AddFlags(cmd))
+		require.NoError(t, cmd.ParseFlags([]string{
+			"--stellar-core-binary-path", "/usr/overridden/stellar-core",
```
Shaptic: It's ideal to omit anything from the test that isn't relevant to what's being tested, so you can drop this line (and again below).

cmd/stellar-rpc/internal/config/options.go
```diff
@@ -666,3 +706,44 @@ func positive(option *Option) error {
 	}
 	return nil
 }
+
+func setForNetwork(cfg *Config, networkParams networkConfig) error {
+	if slices.Equal(cfg.HistoryArchiveURLs, networkParams.historyArchiveURLs) && cfg.NetworkPassphrase == networkParams.networkPassphrase {
+		return nil
+	}
+	if len(cfg.HistoryArchiveURLs) != 0 || cfg.NetworkPassphrase != "" {
+		return errors.New("both network and (HistoryArchiveURLs or NetworkPassphrase) specified (only one may be chosen)")
```
Shaptic: I'm really not a fan of the phrasing of the error but it's hard to figure out a better way to say it without ambiguity... what about:

```
network option conflicts with values provided for passphrase and/or history archive URLs
```

urvisavla:
@Shaptic @cjonas9 Did we consider using the default configs from the [ingest](https://github.com/stellar/go/tree/master/ingest/ledgerbackend/configs) library? They were added specifically to avoid duplicating configuration, so updates to quorum sets etc need to be made only in one place.

Issue #42, opened 2022-11-01
support version parameter on command line


sreuland:
### What problem does your feature solve?
find out the build version of soroban-rpc

### What would you like to see?
`soroban-rpc version` prints a brief summary on command line:
* build or release label version, git commit hash, show whichever combo of this is available.
* version of XDR 
* version of embedded rs-soroban-env library.

PR #550, opened 2025-11-26
Added rs-soroban-env version info to version command, patch command for MacOS

cjonas9:
### What

This PR expands the `./stellar-rpc version` command by adding information about the rs-soroban-env versions (both prev, which is supported, and current). This information is sourced from `cargo.toml`. Additionally, due to a bug in how a patch for Darwin/MacOS was applied, the version command would never work on MacOS (all Go linker flags are overwritten on MacOS). This PR fixes that.

### Why

This enables rapid access to versioning information for RPC components. Re: https://github.com/stellar/stellar-rpc/issues/42

### Known limitations

N/A.


Makefile
```diff
@@ -26,7 +30,7 @@ GOLDFLAGS :=	-X 'github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/config.
 # we instruct the go compiler to produce binaries compatible with version 13.0.
 # this is a mac-only limitation.
 ifeq ($(shell uname -s),Darwin)
-	MACOS_MIN_VER = -ldflags='-extldflags -mmacosx-version-min=13.0'
+	GOLDFLAGS += -extldflags '-mmacosx-version-min=13.0'
```
cjonas9: this is the overwriting bug referenced in my PR

Makefile
```diff
@@ -13,11 +13,15 @@ ifeq ($(strip $(REPOSITORY_VERSION)),)
 	override REPOSITORY_VERSION = "$(shell git describe --tags --always --abbrev=0 --match='v[0-9]*.[0-9]*.[0-9]*' 2> /dev/null | sed 's/^.//')"
 endif
 REPOSITORY_BRANCH := "$(shell git rev-parse --abbrev-ref HEAD)"
+RS_ENV_VERSION_PREV := "$(shell grep -A2 "soroban-env-host-prev" Cargo.toml | grep "version =" | head -1 | sed 's/version = "=\(.*\)"/\1/')"
+RS_ENV_VERSION_CURR := "$(shell grep -A2 "soroban-env-host-curr" Cargo.toml | grep "version =" | head -1 | sed 's/version = "=\(.*\)"/\1/')"
```
leighmcculloch: One downside of reading this out of Cargo.toml is that the Cargo.toml only specifies the version requirement, not the selected version. As it happens today the version in the Cargo.toml is an exact match `=X.Y.Z` with an equal sign, so it will be the same, but if someone changed the Cargo.toml to remove that `=`, or builds from a git commit, then this would no longer capture the correct version or sufficient to identify the code.

The `soroban-env-host` expenses a `VERSION` constant that contains all of the version info for it, including the xdr it is using, and git revs as well. Most of the rust crates we build contain this, and it is what other rust crates and tools we build use to pull more information. Such as the `stellar version` and `stellar-core version` commands. This is the preferred method because it contains more information.

Shaptic: > The `soroban-env-host` expenses a `VERSION` constant

I don't want to overload @cjonas9 with needing to understand Rust :left_right_arrow: Go FFI to extract that field for this particular, but I agree that in principle we'd want to enhance this with more details later.

>  if someone changed the Cargo.toml to remove that `=`, or builds from a git commit, then this would no longer capture the correct version or sufficient to identify the code

True, this is a good call-out. It's an edge case (if you're building from source you don't need this command), but def would be nice to cover it. And per @sreuland's comment we should be sure to handle the failure case gracefully.

Makefile
```diff
@@ -13,11 +13,15 @@ ifeq ($(strip $(REPOSITORY_VERSION)),)
 	override REPOSITORY_VERSION = "$(shell git describe --tags --always --abbrev=0 --match='v[0-9]*.[0-9]*.[0-9]*' 2> /dev/null | sed 's/^.//')"
 endif
 REPOSITORY_BRANCH := "$(shell git rev-parse --abbrev-ref HEAD)"
+RS_ENV_VERSION_PREV := "$(shell grep -A2 "soroban-env-host-prev" Cargo.toml | grep "version =" | head -1 | sed 's/version = "=\(.*\)"/\1/')"
```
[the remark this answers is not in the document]
cjonas9: Failure case amended to this, cargo metadata leveraged!

Shaptic (approved):
LGTM! Just add a bit of explanation to the parsing

PR #563, opened 2025-12-08
Expand getNetwork endpoint response with core supported protocol version info (P25)

### LINKED ISSUE #17: Add Soroban interface version to getNetwork response
https://github.com/stellar/soroban-tools/blob/967351b064bac431f0d6f9ae61ba6e4f0b417a06/cmd/soroban-rpc/internal/methods/get_network.go#L31C26-L31C26

It would be very useful for the RPC method `getNetwork` to return the Soroban contract interface version alongside everything else it already returns.

Being able to fetch what Soroban version is deployed on the network exposed through the RPC would allow for a simple check before trying to deploy a new WASM. Makes the endpoint more useful.

cjonas9:
### What

This PR adds `coreSupportedProtocolVersion` information to the `getNetwork` endpoint to provide the protocol version that core supports.

See [this PR]( https://github.com/stellar/go-stellar-sdk/pull/5877) for changes made to `go-stellar-sdk` that must be merged before this is.

<details>
<summary> ** sample query/response ** </summary>

``` curl -X POST http://localhost:8000 \
  -H "Content-Type: application/json" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getNetwork"
  }' | jq 
```

```  {
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "passphrase": "Test SDF Network ; September 2015",
    "protocolVersion": 24,
    "coreSupportedProtocolVersion": 24
  }
}
```
</details>

### Why

The `getNetwork` endpoint provides less information than what's exposed to us by core. See discussion at [this issue](https://github.com/stellar/stellar-rpc/issues/17).

### Known limitations

Version info extraction is fragile. Specifically, to get this information, because it's not included in any endpoint but rather in the `stellar-core version` command, a regex is used to capture what comes after "ledger protocol version:" in the output of that command. If core changes the wording of the output of `stellar-core version`, this endpoint will fail gracefully.


leighmcculloch:
The linked issue is quite old. I'm posted some thoughts on whether it is still relevant and asking about what the use cases are for exposing the different versions here: https://github.com/stellar/stellar-rpc/issues/17#issuecomment-3634338047

sisuresh:
the `sorobaninfo` http command in core has a couple different formats it can return. From 
[commands.md](https://github.com/stellar/stellar-core/blob/master/docs/software/commands.md).
```
sorobaninfo sorobaninfo?[format=basic,detailed,upgrade_xdr] Retrieves the current Soroban settings in different formats.

    **basic** is the default if the format parameter is not specified. It will dump a subset of the Soroban settings in an easy to read format.

    **detailed** will insert every setting into a ConfigUpgradeSet and dump it in the same format as the dumpproposedsettings command, which lets a user easily compare the existing settings against a proposal.

    **upgrade_xdr** will insert the current upgradeable settings into a ConfigUpgradeSet and dump it as base64 xdr. This can be used along with the stellar-xdr command line tool to dump the current settings in the same format as the JSON file we use for upgrades. This is helpful if you want to make settings changes off of the current settings. Ex. curl -s "127.0.0.1:11626/sorobaninfo?format=upgrade_xdr" | stellar-xdr decode --type ConfigUpgradeSet --output json-formatted
```

I believe the `basic` format was initially added for exposing a subset of the settings externally for Horizon and Supercluster. It's a manual [mapping](https://github.com/stellar/stellar-core/blob/586809d048676b20f2efcd98d340504c0747ffee/src/main/CommandHandler.cpp#L764). The `upgrade_xdr` format dumps the upgrade-able settings in base64 xdr, which we use with the `stellar-xdr decode` to get the format in quickstart. Can this format be used instead in RPC?

leighmcculloch:
Yeah that's fair, if we can't maintain the format, let's not do it. A better developer experience is only achieved if we can commit to maintaining it, and in lieu of that exposing the raw XDR 👍🏻, and as you say, RPC could just omit the two long cost type lists if the use case limits are being added for don't need them.

@Shaptic you mentioned at https://github.com/stellar/stellar-rpc/issues/17#issuecomment-3634510011:

> And `limits` weren't covered by the issue but they're nice to have imo.

What's the use case we're exposing the limits for?

Limits are already collectable via the `getLedgerEntries` method ([example code](https://github.com/stellar/stellar-cli/blob/861af9b0a131ccef61ea216fe3ef6257ec4e831a/cmd/soroban-cli/src/commands/network/settings.rs#L68-L98)), albeit requiring more effort and not as discoverable. But, if the use case we're adding limits for is for an advanced use case, then that's probably okay.

Shaptic:
@leighmcculloch tbh adding the limits just seemed like a nice-to-have: a convenient, human-readable place to check the network's limits instead of perusing [stellar.expert](https://stellar.expert/explorer/public/protocol-history) or relying on [human docs](https://developers.stellar.org/docs/networks/resource-limits-fees): I wasn't aware of `stellar network settings`, which does introduce redundancy, but it's full of fluff (`contract_cost_params_cpu_instructions` et al.) unless you do the `jq` voodoo in your comment. The idea was to have programmatic access to the important subset covered in the docs (much like `basic`).

However, I just realized that since `/sorobaninfo` uses the old single-threaded HTTP server (right, @sisuresh?), someone may be able to slam `getNetwork` and jam up the Core process with HTTP requests. Fetching ledger entries is a better way to do this, but then you run into a very similar maintenance burden of API upkeep parsing said entries. Furthermore, if we do it that way, RPC would need non-trivial updates for new host functions which we want to avoid.

So I think I agree with the general sentiment here that maybe we shouldn't do the same thing in two places in different ways and just point people to the CLI command (which maybe can just filter out the verbose cost functions).

leighmcculloch:
> point people to the CLI command (which maybe can just filter out the verbose cost functions)

Opened:
- https://github.com/stellar/stellar-cli/issues/2322

urvisavla:
@cjonas9 
1. I read through the comments on #17 and the consensus seems to be (which I agree with) that the core *supported* version does not belong in the getNetwork response, but rather in getVersionInfo. And this is already possible via getVersionInfo, since the supported protocol can be easily derived from the core version (it’s always the core major version). Based on that, should this PR be closed?

2.  In case it is decided to proceed with this ,I think you might be able to use the existing `CoreProtocolVersionFunc` helper instead of writing a new one.
https://github.com/stellar/go-stellar-sdk/blob/main/ingest/ledgerbackend/stellar_core_version.go

Issue #203, opened 2024-06-06
Implement offline ingestor


2opremio:
As part of https://github.com/stellar/soroban-rpc/issues/196 , the wanted history window may be smaller than the existing window.

As part of this task we should create an offline ingestor (e.g. not real time) which backfills range of ledgers. The orchestration of what ledgers to ingest should be part of https://github.com/stellar/soroban-rpc/issues/202

The ingestor should (together with the orchestrator) to make sure no inconsistencies are left on the database.

In particular, it should ingest ledgers backwards, to ensure that no gaps are left on the database:

![image](https://github.com/stellar/soroban-rpc/assets/2362916/0e418c27-4435-4b27-a504-ec6e8ff8e321)

2opremio:
Reopening as part of the backfill history

mollykarcher:
See https://github.com/stellar/stellar-rpc/discussions/323

cjonas9:
See my [design doc](https://docs.google.com/document/d/14bUy9plo8MLpge-ED4DwpeA0IPnwUkiuDi4q3KLE268/edit?usp=sharing) for this to provide input. Thanks!

PR #571, opened 2025-12-15
Synchronous history backfilling

cjonas9:
### What

This PR adds synchronous history backfilling of ledgers through the new configuration parameter `backfill`. This will fill the local SQL database with the most recent `HISTORY_RETENTION_WINDOW` ledgers, fetched from CDP. Usage: `./stellar-rpc --backfill`, or one can specify this as a config parameter in the TOML or as an environmental variable.

Notes:
- After verifying the local DB state, this will backfill the specified number of ledgers from CDP backwards/"leftward", then backfill forwards/"rightward" up to the new most current ledger. By termination, we are guaranteed a database filled with the most recent `HISTORY_RETENTION_WINDOW` ledgers.
    - Leftward/backwards backfill: This backfills the range [oldest ledger we want in our DB <- oldest ledger in local DB-1] into the local DB. This phase is skipped if the DB is empty and this is generally slower than forwards backfill.
    - Rightward/forwards backfill: This backfills the range (max(newest ledger currently in local DB, oldest ledger we want in our DB) -> current tip of datastore] into the local DB. This phase is run twice if the DB is empty (on the second run, it just refreshes the current DB tip).
- After completion, it will start RPC to begin filling in the ledgers live through captive core. [See my design document for details](https://docs.google.com/document/d/14bUy9plo8MLpge-ED4DwpeA0IPnwUkiuDi4q3KLE268/edit?tab=t.0) for how this is done.
- This design **guarantees** that, regardless of faults, the database will never be left with gaps in it. The client may supply a partially filled database, but backfill will error if this database initially has gaps in it. Note that (due to how the BufferedStorageBackend prevents backwards iteration) supplying a partially filled DB may be slower than the empty DB case if the partially filled database is 1.) minimally filled and/or 2.) close to the datastore current tip ledger (i.e. the larger the backwards backfilling run is, the slower this can be).
- The backfilled ledgers are ingested via CDP through the `bufferedStorageBackend`. It's datastore agnostic and has been tested on GCS and S3.
- Using the (rate-limited) public S3 datastore in the k8s `soroban-rpc-pubnet-dev` deployment, to ingest one week of ledgers (120,960 ledgers, ~150Gb), this takes about 3 hours.
    - Of three runs backfilling from no local DB, we see the following timing: 3h6m, 3h7m, 2h41m, 2h48m. I suspect that if **any** ledgers are in the DB (and a backwards phase runs), we would see longer runtimes. 
    - In a truly pessimistic case (e.g. the backfill backwards phase is maximal), the runtime for this scenario is around ~3h45m.

### Why

There is no way to backfill RPC's local database with historical ledgers, let alone a time-conscious method for this. This PR provides one.
See issues/discussions on this: #203, [1718](https://github.com/orgs/stellar/discussions/1718)

### Known limitations

[N/A]

### DIFF
--- a/cmd/stellar-rpc/internal/ingest/backfill.go
+++ b/cmd/stellar-rpc/internal/ingest/backfill.go
@@ -192,8 +192,8 @@ func (b *BackfillMeta) runBackfill(ctx context.Context, bounds fillBounds) (fill
 			return fillBounds{}, errors.Wrap(err, "backfill failed")
 		}
 		b.dbInfo.sequences.First = bounds.backfill.First
-		b.logger.Infof("Backfill of old ledgers complete")
 	}
+	b.logger.Infof("Backfill of old ledgers complete")
 	return bounds, nil
 }
 


cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+		if err := fillChunk(callerCtx, dsInfo, tx, nil, lChunkBound, rChunkBound); err != nil {
+			return fmt.Errorf("couldn't fill chunk [%d, %d]: %w", lChunkBound, rChunkBound, err)
+		}
+		logger.Infof("Forwards backfill: committed ledgers [%d, %d]", lChunkBound, rChunkBound)
+
+	}
+	return nil
+}
+
+func fillChunk(callerCtx context.Context, dsInfo DatastoreInfo, tx db.WriteTx, tempBackend *ledgerbackend.LedgerBackend, left uint32, right uint32) error {
```
Shaptic: this should just be an interface

```suggestion
func fillChunk(callerCtx context.Context, dsInfo DatastoreInfo, tx db.WriteTx, tempBackend ledgerbackend.LedgerBackend, left uint32, right uint32) error {
```

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/config"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db"
+)
+
+const (
+	// OneDayOfLedgers is (roughly) a 24 hour window of ledgers.
+	OneDayOfLedgers   = config.OneDayOfLedgers
+	SevenDayOfLedgers = config.OneDayOfLedgers * 7
+	// Number of ledgers to read/write at a time during backfill
+	ChunkSize uint32 = OneDayOfLedgers / 4 // 6 hours. Takes X minutes to process
```
Shaptic: X minutes?

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+// This function backfills the local database with n ledgers from the datastore
+// It is called by daemon.go if cfg.Backfill is true
+func RunBackfill(cfg *config.Config, logger *supportlog.Entry, localDbConn *db.DB, localDbRW db.ReadWriter, dsInfo DatastoreInfo) error {
+	logger.Infof("Beginning backfill process")
+	var (
+		ctx              context.Context = context.Background()
```
Shaptic: We should probably have a context with a reasonably excessive timeout for this process based on the ledger quantity.

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	dsInfo.backend = backend
+	defer backend.Close()
+
+	// Determine what ledgers have been written to local DB
+	// Note GetLedgerRange assumes lexicographical ordering of ledger files in datastore
+	ledgerRange, err := localDbReader.GetLedgerRange(ctx)
+	if err != nil && err != db.ErrEmptyDB {
+		return fmt.Errorf("error getting ledger range from local DB: %w", err)
+	} else if err == db.ErrEmptyDB {
```
Shaptic: CI is correct here: `errors.Is` is better

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+		dbIsEmpty = true
+	}
+	maxWrittenLedger, minWrittenLedger := ledgerRange.LastLedger.Sequence, max(ledgerRange.FirstLedger.Sequence, 1)
+
+	// Phase 1: precheck to ensure no gaps in local DB
+	if !dbIsEmpty && (maxWrittenLedger >= minWrittenLedger) {
+		logger.Infof("Starting precheck for backfilling the database at %s, phase 1 of 4", localDbPath)
+		if err = verifyDbGapless(ctx, localDbReader, minWrittenLedger, maxWrittenLedger); err != nil {
+			return fmt.Errorf("backfill precheck failed: %w", err)
```
Shaptic: in general you should use `errors.Wrap` instead of formatting a new error with the old error inside of it as a string, because the latter method breaks the `errors.Is`ability of a returned value

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+
+	}
+	return nil
+}
+
+func fillChunk(callerCtx context.Context, dsInfo DatastoreInfo, tx db.WriteTx, tempBackend *ledgerbackend.LedgerBackend, left uint32, right uint32) error {
+	var ledger xdr.LedgerCloseMeta
+	var err error
+
+	if tempBackend == nil {
```
Shaptic: why would this be `nil`?

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+	ctx, cancelPrecheck := context.WithTimeout(callerCtx, 4*time.Minute)
+	defer cancelPrecheck()
+
+	tx, err := reader.NewTx(ctx)
+	if err != nil {
+		return fmt.Errorf("db verify: failed to begin read transaction: %w", err)
+	}
+	defer tx.Done()
+
+	ct, err := tx.CountLedgersInRange(ctx, minLedgerSeq, maxLedgerSeq)
```
Shaptic: really no reason to shorten it here
```suggestion
	count, err := tx.CountLedgersInRange(ctx, minLedgerSeq, maxLedgerSeq)
```

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+			_ = tx.Rollback()
+			return fmt.Errorf("couldn't get ledger %d from backend: %w", seq, err)
+		}
+		if err = tx.LedgerWriter().InsertLedger(ledger); err != nil {
+			_ = tx.Rollback()
+			return fmt.Errorf("couldn't write ledger %d to local db: %w", seq, err)
+		}
+		processed = true
+	}
+	if processed {
```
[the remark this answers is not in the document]
cjonas9: good catch!

cmd/stellar-rpc/internal/ingest/backfill.go
cjonas9: pending timing data collection outside of my machine

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,304 @@
...
+// Backfills the local DB with ledgers in [lBound, rBound] from the cloud datastore
+// Used to fill local DB backwards towards older ledgers
+func (metaInfo *BackfillMeta) runBackfillBackwards(ctx context.Context, lBound uint32, rBound uint32) error {
+	for rChunkBound := rBound; rChunkBound >= lBound; {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		// Create temporary backend for backwards-filling chunks
+		// Note monotonicity constraint of the ledger backend
+		tempBackend, err := makeBackend(metaInfo.dsInfo)
```
Shaptic: Seems like you can create/close this outside of the loop, yeah?

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,403 @@
...
+	ledgerRange, err := backfill.dbInfo.reader.GetLedgerRange(ctx)
+	if err != nil {
+		return 0, 0, errors.Wrap(err, "db verify: could not get ledger range")
+	}
+	// Get sequence number of highest/lowest ledgers in local DB
+	minDbSeq, maxDbSeq := ledgerRange.FirstLedger.Sequence, ledgerRange.LastLedger.Sequence
+	backfill.logger.Infof("DB verify: checking for gaps in [%d, %d]",
+		minDbSeq, maxDbSeq)
+	expectedCount := maxDbSeq - minDbSeq + 1
+	sequences, err := backfill.dbInfo.reader.GetLedgerSequencesInRange(ctx, minDbSeq, maxDbSeq)
```
Shaptic: If this is a large range, this could be a big chunk of memory. I'm wondering if it should be something more like `GetLedgerCountInRange` where it returns `min`/`max`/`count` (computed SQL-side) rather than the entire list of values.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -190,10 +198,41 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
...
+		}
+		if err := backfillMeta.RunBackfill(cfg); err != nil {
+			logger.WithError(err).Fatal("failed to backfill ledgers")
+		}
+		// Clear the DB cache and fee windows so they re-populate from the database
+		daemon.db.ResetCache()
+		feewindows.Reset()
+	}
+	// Start ingestion service only after backfill is complete
+	ingest.StartService(daemon.ingestService, ingestCfg)
```
Shaptic: this is a weird line to read to me - you pass the ingest service attached to the daemon to the ingest module? I'd expect something more like `daemon.ingestService.Start(ingestCfg)`, instead.

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,377 @@
...
+	minSeq uint32
+	// Note maxSeq is excluded because it goes stale every 6 seconds
+	// it is replaced by `currentTipLedger` in RunBackfill
+}
+
+// This struct holds the local database read/write constructs and metadata initially associated with it
+type databaseInfo struct {
+	reader  db.LedgerReader
+	minSeq  uint32
+	maxSeq  uint32
```
Shaptic: seems like you could/should use `db.LedgerSeqRange` here since that's what you asign it to later

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,377 @@
...
+	if err := backfill.runPrecheck(ctx); err != nil {
+		return err
+	}
+	bounds, nBackfill, err := backfill.setBounds(ctx, cfg.HistoryRetentionWindow)
+	if err != nil {
+		return errors.Wrap(err, "could not set backfill bounds")
+	}
+
+	// If DB isn't empty, backfill backwards from local DB tail to the left edge of retention window
+	if !bounds.skipBackwards {
```
Shaptic: isn't this implied by `backfill.dbInfo.isEmpty`?

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,278 @@
...
+			return fmt.Errorf("couldn't commit range [%d, %d]: %w", left, right, err)
+		}
+	}
+	return nil
+}
+
+// Creates a buffered storage backend for the given datastore
+// Sets it in the DatastoreInfo struct
+func makeBackend(dsInfo DatastoreInfo) (ledgerbackend.LedgerBackend, error) {
+	backend, err := ledgerbackend.NewBufferedStorageBackend(
```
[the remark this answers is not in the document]
Shaptic: still true heh

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,377 @@
...
+
+	// If DB isn't empty, backfill backwards from local DB tail to the left edge of retention window
+	if !bounds.skipBackwards {
+		if err := backfill.runBackfillBackwards(ctx, bounds); err != nil {
+			return err
+		}
+	}
+
+	// Backfill from local DB head (or left edge of retention window, if empty) to current tip of datastore
+	if err := backfill.runBackfillForwards(ctx, &bounds, cfg.CheckpointFrequency); err != nil {
```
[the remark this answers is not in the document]
cjonas9: i can see that being clearer, will do!

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -85,6 +85,14 @@ func (d *Daemon) GetEndpointAddrs() (net.TCPAddr, *net.TCPAddr) {
 	return *addr, adminAddr
 }
 
+func (d *Daemon) GetIngestService() *ingest.Service {
```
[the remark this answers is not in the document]
cjonas9: good catch, it's not, this and StopIngestion() were part of my initial integration testing efforts

cmd/stellar-rpc/internal/daemon/daemon.go
cjonas9: absolutely true

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,408 @@
...
+			RetryLimit: 3,
+			RetryWait:  5 * time.Second,
+		},
+		dsInfo.ds,
+		dsInfo.schema,
+	)
+}
+
+// Gets the latest ledger number stored in the cloud Datastore/datalake and updates datastoreInfo.sequences.Last
+func (dsInfo *datastoreInfo) getLatestSeqInCDP(callerCtx context.Context) error {
```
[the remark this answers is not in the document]
cjonas9: Totally true; this function ended up getting pruned and replaced to reduce the statefulness of keeping a quickly expiring frontfill.Last/dsInfo.sequences.Last and because it was basically just a wrapper

cmd/stellar-rpc/internal/ingest/backfill.go
```diff
@@ -0,0 +1,408 @@
...
+		b.logger.Infof("Backfilling to the left edge of retention window, ledgers [%d <- %d]",
+			bounds.backfill.First, bounds.backfill.Last)
+		if bounds, err = b.backfillChunks(ctx, bounds); err != nil {
+			return backfillBounds{}, errors.Wrap(err, "backfill failed")
+		}
+		b.dbInfo.sequences.First = bounds.backfill.First
+		b.logger.Infof("Backfill of old ledgers complete")
+	} else {
+		b.logger.Infof("No backfill needed, local DB tail already at retention window edge")
+	}
```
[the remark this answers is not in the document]
cjonas9: removed else clause, but kept "complete" log within if statement -- I think it's cleaner to output backfill phase messages only if there's a backfill phase that took place

PR #608, opened 2026-02-20
batch event inserts to avoid SQLite bind variable limit

karthikiyer56:
A single Soroban transaction emitting >3,276 events would exceed SQLite's SQLITE_MAX_VARIABLE_NUMBER limit (32,767) since each event uses 10 bind variables in the INSERT statement. This caused crashes for transactions with ~4K events.

Batch insertableEvents into chunks of 1,000 per INSERT (10,000 bind variables) to stay well within the limit. All batches complete before InsertEvents returns.

Add TestInsertEventsBatchingExceedsLimit to verify a single transaction with 5,000 operation events inserts successfully.

### DIFF
--- a/cmd/stellar-rpc/internal/db/event_test.go
+++ b/cmd/stellar-rpc/internal/db/event_test.go
@@ -1,7 +1,6 @@
 package db
 
 import (
-	"context"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/stellar-rpc/internal/db/event.go
```diff
@@ -195,26 +195,36 @@ func (eventHandler *eventHandler) InsertEvents(lcm xdr.LedgerCloseMeta) error {
...
-				"event_data",
-				"ledger_close_time",
-				"transaction_hash",
-				"topic1", "topic2", "topic3", "topic4",
-			)
+		// Batch inserts to avoid exceeding SQLite's SQLITE_MAX_VARIABLE_NUMBER
+		// limit (32,767 by default). With 10 bind variables per event, we cap
+		// each INSERT at 1000 events (10,000 bind variables) to stay well
+		// within the limit.
+		const maxEventsPerBatch = 1000
```
karthikiyer56: shud this be 3k is the qeueston

PR #616, opened 2026-02-25
Fix failure to check error result from BuildEventsJSONFromTransaction()

cjonas9:
### What

Fixes a `nil == nil` error checking bug.

### Why

An error value is assigned to `convErr`, but the stale/confirmed-nil variable `err` is checked instead. This leads to never checking the return status of `BuildEventsJSONFromTransaction()` in the `getTransactions` endpoint, which can plausibly return an error.

### Known limitations

N/A


Shaptic:
Nice catch!

PR #619, opened 2026-03-03
Fetch transaction prior to checking latest ledger cache

Shaptic:
### What
Move the transaction fetch prior to the "latest ledger" cache fetch.

### Why
There is a race where the cache isn't updated but the ledger is ingested, causing occasional test failures on tx-fetch-after-sub behavior (specifically in client.go on this line):

```golang
	require.GreaterOrEqual(t, response.LatestLedger, response.Ledger)
```

### Known limitations
n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 ## Unreleased
 
+### Added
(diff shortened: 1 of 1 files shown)


Shaptic:
@tamirms in theory, yes, but I don't think we should for a few reasons: significantly increased complexity and refactoring to couple the two queries together for something that only affects tests, and, more importantly, there's no reason that `LatestLedger` should _match_ (or be coupled in any way to) `Transaction.Ledger`. It should just be `>=`, so switching the order is sufficient to mitigate.

tamirms (approved):

PR #633, opened 2026-03-18
Add packfile library design doc

tamirms:
## Summary
- Adds a design document for a compact, immutable file format that provides O(1) random access to ordinal-indexed data (events, bitmaps, ledgers) with minimal I/O
- This is a building block for a full history implementation of RPC (v2)
- The doc covers the file format, read/write paths, integrity model, and API reference
- The goal is to align on this design before proceeding with implementation — please flag any concerns or blockers so we can address them during the design stage

**Note:** This PR targets `main` for now but will be retargeted to the full history RPC / v2 branch once we align on that before merging.

## Further reading
- Reference implementation: [`packfile`](https://github.com/tamirms/event-analysis/tree/main/packfile), [`intpack`](https://github.com/tamirms/event-analysis/tree/main/intpack)
- [Benchmarks](https://github.com/tamirms/event-analysis/blob/main/BENCHMARKS.md) — packfile vs RocksDB sstable across write throughput, sequential/random reads, compression efficiency, and bitmap indexing
- [Format comparison](https://github.com/tamirms/event-analysis/blob/main/FORMAT-COMPARISON.md) — technical evaluation of packfile vs RocksDB sstable across performance, code complexity, dependencies, and suitability for remote storage

### DIFF
--- a/design-docs/packfile-library.md
+++ b/design-docs/packfile-library.md
@@ -2,16 +2,39 @@
 
 ## Problem
 
-We need to store large collections of variable-length items (events, compressed bitmaps, ledgers) in immutable files and read individual items by ordinal index with minimal I/O. The files are written once and read many times over their lifetime.
+We need to store large collections of variable-length **items** (opaque byte blobs) in immutable files. Items are appended sequentially during a write phase, then the file is sealed and never modified. Reads need to fetch individual items efficiently — given an item's position in the sequence (0, 1, 2, ...), return its bytes.
 
-A flat file is the natural fit: write items sequentially, keep an offset table at the end, look up any item by index. This is simple, fast, and well-suited to immutable data. General-purpose storage engines like RocksDB or SQLite add key management, transactions, and mutable write paths that we don't need — overhead without benefit for immutable, ordinal-indexed data.
+The two main read patterns are scattered point lookups (hundreds or thousands of potentially non-contiguous items per query) and sequential scans (all items in order). The storage may have a limited I/O budget (e.g. EBS volumes with a fixed number of operations per second) or high per-request latency (e.g. S3 range GETs). Minimizing the number of I/O operations per query is critical.
 
-Packfile is an implementation of this approach. It handles the details that a naive implementation gets wrong or leaves to the caller: record grouping for compression, compact index encoding, parallel block processing, SHA-256 content hashing, CRC32C integrity checks, and safe concurrent reads.
+Packfile is designed around this constraint:
 
-Two packages:
+- **Minimize I/O operations per query.** Opening a file and looking up any item should require as few I/O operations as possible. Ideally, one I/O on open, then one I/O per item lookup.
+- **Concurrent reads.** All read methods must be safe for parallel use. On storage with limited I/O budgets or high per-request latency, issuing reads sequentially is far too slow. Parallel reads are required to keep query latency manageable.
+- **Immutable after write.** Files are written once, sealed, then read-only. No updates, no deletes.
 
-- **`packfile`** — item-level random access to immutable files. Handles record grouping, zstd compression, CRC32C integrity, content hashing, and parallel I/O. Uses a thin CGo wrapper (`zstd/`) for compression — the pure-Go zstd implementation is 4x slower on our workloads.
-- **`intpack`** — integer compression using Frame of Reference (FOR) encoding. Pure Go, no external dependencies. General-purpose codec — not packfile-specific.
+Non-goals:
+- **Key-based lookup.** Items are accessed by their position in the sequence, not by key.
+- **Mutability.** No append-after-finalize.
+- **Chunk management.** Directory layout, rotation, and tiering are caller concerns.
+
+## Concepts
+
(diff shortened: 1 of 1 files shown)


design-docs/packfile-library.md
```diff
@@ -0,0 +1,482 @@
...
+slot := mphf.Lookup("USD")
+r.ReadItem(slot, func(data []byte) error {
+    bitmap.UnmarshalBinary(data) // data is borrowed — copy if needed
+    return nil
+})
+```
+
+## Goals
+
+- **O(1) random access by ordinal index.** Every `ReadItem` call maps index to record via arithmetic, then reads and decodes a single record.
```
graydon: This is .. depending on how you look at it "not achieved" by the design in this doc. The index is delta-encoded which means the entire index needs decoding on file-open (which is O(n) despite having a fairly healthy divisor on n).

Once open and decoded, it's O(1), but that's not guaranteed to be easy or cheap (eg. if you have lots of files you might be constantly reading/decoding/dropping their indexes from memory). I'd recommend just making your index be seekable itself, so you:
 - open the file
 - readAt a position in the index you calculate from the ordinal, to get the offset of the variable-sized record
 - readAt the variable-sized record

graydon (commented):
I got curious so I had claude spit out a simpler program (https://github.com/graydon/xdrpack -- both go and rust though I only tested the rust version) that just does a zstd train on an xdr stream's frames to build a shared dictionary, and then compresses each frame with that dictionary and indexes the offsets of each compressed frame a single table, with offsets sized to minimum to span the file. seems to work, though who knows if it's comparable to the benchmark numbers you're after. I guess I'm just a bit unsure whether the complexity of the structure you've built here buys you much, but perhaps it does -- happy to talk more about it!

design-docs/packfile-library.md
```diff
@@ -0,0 +1,482 @@
...
+- **Mutability.** No append-after-finalize, no updates, no deletes.
+- **Chunk management.** Directory layout, chunk-to-sequence mapping, rotation are caller concerns.
+- **Caching.** No built-in LRU. Callers manage their own pool of open `Reader` instances.
+
+## How It Works
+
+Items are grouped into fixed-size **records** (default 128 items per record). Small items like events don't compress well individually, but 128 of them together do. Large items like ledgers are stored one per record since they compress well on their own. Each record is compressed and written as a contiguous block on disk.
+
+An **offset index** at the end of the file maps record numbers to byte offsets. On open, the entire index is decoded into a flat `[]int64` array. Looking up item `i` is arithmetic: `offsets[i / RecordSize]` gives the record's byte offset, then a single disk read + decode extracts the item.
+
```
karthikiyer56: Can you include 1-2 exampes here?

one for events (RecordSize=128) and one for ledgers (RecordSize=1)?
something like - "here are 5 items (item being an individual event or lcm) of these sizes, heres what the records look like on disk, heres what the offset index looks like, heres the FOR encoding step by step."

i have read through this a few times and I am still struggling to understand the format here.
There are like 6-7 concepts/terms here - `item, record, offset index, FOR group, delta, W, min` - all in abstract terms across different sections.  i had to read it three times to piece together how they relate and I still have some visualization issues 🙈 

especially useful would be showing the two modes side by side — 
events with RecordSize=128 where you get the per-record FOR_index AND the file-level offset index
VS 
ledgers with RecordSize=1 where the per-record FOR_index disappears entirely. that difference is non-obvious from the current text.

design-docs/packfile-library.md
```diff
@@ -0,0 +1,482 @@
...
+When `ContentHash: true`, the writer computes a chunked SHA-256 over the logical item stream:
+
+```
+chunkDigest_i = SHA-256([4B len][item_{i*K}] ... [4B len][item_{i*K+K-1}])
+finalHash     = SHA-256(chunkDigest_0 || ... || chunkDigest_M)
+K = RecordSize
+```
+
+The hash is independent of compression and format — same items in the same order with the same RecordSize always produce the same hash. Note that changing RecordSize changes the chunk boundaries and therefore the hash.
```
karthikiyer56: `independent of compression and format` reads like the hash depends only on the data itself. but the very next sentence says changing RecordSize changes the hash.

Can we rephrase it to something like:
"""
The hash is independent of compression and format, but depends on RecordSize.
This means you can't use the content hash to verify "these two packfiles contain the same data" unless they were written with the same RecordSize
"""

design-docs/packfile-library.md
```diff
@@ -0,0 +1,482 @@
...
+defer w.Close() // removes file if Finish was not called
+
+for _, event := range events {
+    w.Append(event)
+}
+
+w.Finish(nil) // flushes partial record, writes index + trailer, fsyncs
+```
+
+Items are appended in order. `Finish` flushes any partial record, writes the offset index, optional app data, and a 64-byte trailer, then fsyncs. `Close` after `Finish` is a no-op; `Close` without `Finish` removes the incomplete file.
```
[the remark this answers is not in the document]
tamirms: I'm not sure what you're referring to. Are you taking about the use case of calling `Close()` / `Finish()` ?

karthikiyer56 (approved):

PR #649, opened 2026-04-01
Add intpack FOR encoding and packfile offset index codec

tamirms:
## Summary

First in a series of PRs checking in the packfile implementation for full history RPC v2. See #633 for the full design doc.

This PR adds two packages:

**intpack** (`intpack.go`) — A standalone [Frame of Reference](https://lemire.me/blog/2012/02/08/effective-compression-using-frame-of-reference-and-delta-coding/) integer codec. Encodes groups of uint32 values by subtracting the group minimum and bit-packing residuals at the minimum bit width needed. Each encoded group ends with a 1-byte width and 4-byte minimum, so callers can locate the metadata from the tail without knowing the packed size. Used by the packfile offset index (this PR), per-record item size indexes (later PR), and the event store app data encoding.

**Offset index codec** (`index.go`) — Builds on intpack to encode/decode the file-level index that maps record numbers to byte positions on disk. Rather than storing absolute offsets (which grow with file size), the index stores record byte sizes as FOR groups of up to 128 values with a trailing CRC32C. On open, all groups are decoded into a flat `[]int64` offset table for O(1) lookup. Includes an OOM guard that validates record count against index size before allocating, and a structural invariant check (prefix-sum must equal the known index base offset).

## Test plan
- [x] `intpack_test.go` — roundtrip encoding/decoding, full uint32 range (width 32), error cases (invalid width, truncated payload, n<=0), buffer reuse, binary layout verification, decoding with prefix data, crafted width-0 input
- [x] `index_test.go` — roundtrip with various sizes (single record, multi-group, partial last group), CRC corruption detection, implausible record count rejection, index base mismatch detection, negative record count, buffer too small, non-monotonic offsets, zero records, empty offsets, non-zero start, delta exceeding uint32, large deltas near MaxUint32

### DIFF
--- a/cmd/stellar-rpc/internal/intpack/intpack.go
+++ b/cmd/stellar-rpc/internal/intpack/intpack.go
@@ -52,8 +52,8 @@ func DecodeGroup(buf []byte, n int, dst []uint32) ([]uint32, int, error) {
 	}
 
 	width := uint64(buf[len(buf)-footerSize])
-	if width > 32 {
-		return dst, 0, fmt.Errorf("intpack: invalid FOR width %d (max 32)", width)
+	if width == 0 || width > 32 {
+		return dst, 0, fmt.Errorf("intpack: invalid FOR width %d (must be 1..32)", width)
 	}
 
 	groupMin := binary.LittleEndian.Uint32(buf[len(buf)-footerSize+1:])
@@ -103,18 +103,11 @@ func packResiduals(buf []byte, values []uint32, minVal uint32, width uint8) {
 // Safe to call with any packed length — elements near the boundary where an
 // 8-byte read would exceed len(packed) are decoded using a zero-padded copy.
 //
+// Caller must ensure w is in [1, 32] — DecodeGroup validates this.
 // The uint64-to-int and uint64-to-uint32 conversions are safe:
 //   - bytePos is bounded by len(packed) which fits in int
 //   - (raw>>shift)&mask is bounded by width <= 32 bits, fits in uint32
 func unpackResiduals(packed []byte, n int, w uint64, groupMin uint32, values []uint32) {
-	if w == 0 {
-		for i := range values {
-			values[i] = groupMin
-		}
-
-		return
-	}
-
 	mask := uint64((1 << w) - 1)
 
 	// When len(packed) < 7, safeLimit is negative — all reads use the
--- a/cmd/stellar-rpc/internal/intpack/intpack.go
+++ b/cmd/stellar-rpc/internal/intpack/intpack.go
@@ -6,8 +6,10 @@
 //
 //	[packed residuals][1-byte width][4-byte minimum (little-endian)]
 //
-// Width and minimum are always the final 5 bytes, so callers can locate
-// metadata from the tail of any buffer without knowing the packed size upfront.
+// Width and minimum are always the final 5 bytes, so the group can be located
+// from the end of a buffer without a separate length field. This is useful when
+// a FOR group is appended after variable-length data — the decoder can strip
+// the group from the tail, then recover the preceding data's boundary.
 package intpack
 
 import (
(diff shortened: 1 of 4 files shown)


cmd/stellar-rpc/internal/packfile/index.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	if len(buf) < indexSize {
+		return nil, fmt.Errorf("%w: buffer length %d < indexSize %d", ErrCorrupt, len(buf), indexSize)
+	}
+
+	buf = buf[:indexSize]
+
+	// Sanity-check recordCount against indexSize to prevent OOM from crafted trailers.
+	// Each FOR group of up to 128 records requires at least 6 bytes
```
urvisavla: nit: Isn't 6 bytes is actually the minimum for a group of any size (not just 128 records group)

cmd/stellar-rpc/internal/intpack/intpack.go
```diff
@@ -0,0 +1,143 @@
+// Package intpack provides Frame-of-Reference (FOR) integer encoding for
+// groups of uint32 values. Values are encoded as bit-packed residuals relative
+// to the group minimum, requiring only ceil(log2(max-min)) bits per value.
+//
+// On-disk layout per group:
+//
+//	[packed residuals][1-byte width][4-byte minimum (little-endian)]
+//
+// Width and minimum are always the final 5 bytes, so callers can locate
+// metadata from the tail of any buffer without knowing the packed size upfront.
```
urvisavla: Curious about the design choice here - the comment states that tail-first metadata enables "locate without upfront packSize, but a head-first [width][min][residuals] layout would have the same property from the other side. Was tail chosen for consistency with the packfile's overall tail pattern (trailer, CRC) or something else? Not asking for a change, just interested.

cmd/stellar-rpc/internal/intpack/intpack.go
```diff
@@ -0,0 +1,143 @@
...
+
+// unpackResiduals unpacks n bit-packed residuals from packed and adds groupMin.
+// Safe to call with any packed length — elements near the boundary where an
+// 8-byte read would exceed len(packed) are decoded byte-by-byte.
+//
+// The uint64-to-int and uint64-to-uint32 conversions are safe:
+//   - bytePos is bounded by len(packed) which fits in int
+//   - (raw>>shift)&mask is bounded by width <= 32 bits, fits in uint32
+func unpackResiduals(packed []byte, n int, w uint64, groupMin uint32, values []uint32) {
+	if w == 0 {
```
urvisavla: Minor inconsistency: rangeWidth clamps width to ≥1 so w=0 never appears on disk from this encoder, but unpackResiduals still has a w==0 branch.

cmd/stellar-rpc/internal/intpack/intpack.go
```diff
@@ -0,0 +1,143 @@
...
+
+// EncodeGroup FOR-encodes values into one group.
+// Panics if len(values) == 0.
+func EncodeGroup(values []uint32) []byte {
+	minVal, width := rangeWidth(values)
+	packSize := (int(width)*len(values) + 7) / 8
+
+	// 5 bytes for the footer (1-byte width + 4-byte minimum),
+	// 7 bytes of overshoot for safe 8-byte writes during bit-packing.
+	buf := make([]byte, packSize+5+7)
```
urvisavla: nit: 5 (footer size) and 7 (overshoot pad) appear as magic numbers in a few places. Might be worth named constants.

cmd/stellar-rpc/internal/intpack/intpack.go
```diff
@@ -0,0 +1,143 @@
...
+		bytePos := bitPos / 8
+		shift := bitPos % 8
+
+		var raw uint64
+		if int(bytePos) < safeLimit { //nolint:gosec // bytePos bounded by packed length
+			raw = binary.LittleEndian.Uint64(packed[bytePos:])
+		} else {
+			for k := 0; k < 8 && int(bytePos)+k < len(packed); k++ { //nolint:gosec // bytePos bounded by packed length
+				raw |= uint64(packed[int(bytePos)+k]) << (k * 8) //nolint:gosec // bytePos bounded by packed length
+			}
```
urvisavla: Can this byte-shift loop could be replaced with something like for better readability?
  ```
  var tail [8]byte
  copy(tail[:], packed[bytePos:])
  raw = binary.LittleEndian.Uint64(tail[:])
  ```

urvisavla (approved):
Looks great, nice work. Left a few minor comments.

cmd/stellar-rpc/internal/intpack/intpack.go
tamirms: There's no inherent advantage of one layout over the other — they're symmetric. The choice is driven by what the FOR group sits next to:

- If a FOR group has variable-length data **before** it (`[data][FOR]`), only tail-first works — the decoder finds W/min at a fixed offset from the end, since the start position is unknown.
- If a FOR group has variable-length data **after** it (`[FOR][data]`), only head-first works — the decoder finds W/min at a fixed offset from the start, since the end position is unknown.
- If there's no variable-length data on either side (fixed-size surroundings, or just the group alone), both layouts work.

Packfile's per-record layout is `[compressed payload][FOR index][CRC]`. The compressed payload has variable size and sits before the FOR index, which forces tail-first.

The deeper question is why the record is ordered payload-first. It's because the writer produces bulk data first and derives the metadata from it (item sizes). Writing the FOR index before the payload would mean either buffering the whole payload in memory or seeking back to rewrite the header after the payload is done.

So tail-first is downstream of that layout choice, not intrinsic to the FOR encoding itself. The offset index doesn't have variable-length data around it, so either would work there — it uses tail-first for code reuse with the per-record case.

PR #675, opened 2026-04-12
refactor: modernize sync/atomic usage with typed atomic apis

chuanshanjida:
### What

Modernize sync/atomic usage with typed atomic APIs

Replace legacy sync/atomic function-based API (AddInt32, LoadInt32, etc.) 
with Go 1.19+ typed atomic types (atomic.Int32, atomic.Int64, etc.).

This change:
- Improves code readability by removing explicit address-of operators
- Enhances type safety at compile-time
- Provides better API ergonomics with method-based access

The transformation is source-compatible and requires Go 1.19 or later.

### Why

More info https://github.com/golang/go/issues/50860

### Known limitations

N

### DIFF
--- a/cmd/stellar-rpc/internal/network/backlogQ.go
+++ b/cmd/stellar-rpc/internal/network/backlogQ.go
@@ -24,9 +24,9 @@ type gauge interface {
 
 type backlogQLimiter struct {
 	limit        uint64
-	pending      uint64
+	pending      atomic.Uint64
 	gauge        gauge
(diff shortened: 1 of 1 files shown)


box4wangjing:
I made the same observation, but I saw that it has already been submitted.
Good fixes. @chuanshanjida 

CC @urvisavla @tamirms

chuanshanjida:
@urvisavla @tamirms friendly ping~

PR #677, opened 2026-04-15
Add streaming workflow design doc, renumber design docs

### LINKED ISSUE #704: RPC Full History Streaming
Details to folow

### LINKED ISSUE #722: Streaming workflow design (catch-up, live ingestion, freeze & prune)
## TL;DR
Design the full-history daemon's streaming workflow — one daemon, one mode that catches up from backfill on startup, ingests live ledgers, and maintains the retention window by freezing completed data and pruning what's aged out. Query serving is the only thing out of scope (#770).

## What the design covers
The daemon's lifecycle:
- **Catch up** on startup by running backfill as a subroutine, up to the live tip.
- **Ingest** live ledgers into hot storage.
- **Freeze** completed chunks/indexes to immutable files and **prune** what's past the retention window.

…and staying correct across crashes and restarts.

## Out of scope
- Reader routing / query serving across hot and cold — #770.

karthikiyer56:
This PR introduces the streaming-workflow design doc for the unified `stellar-rpc` service and reshapes the backfill design to describe backfill as an internal subroutine of that service.

## `02-streaming-workflow.md` (new)

End-to-end design for the unified RPC service. Covers:

- Startup phases 1–4 (Phase 1 catchup → Phase 2 `.bin` hydration → Phase 3 reconcile → Phase 4 live ingestion).
- Configuration (`[SERVICE]`, `[CAPTIVE_CORE]`, `[ACTIVE_STORAGE]`, `[HISTORY_ARCHIVES]`, `[BSB]`), validation rules, the three operator profiles (archive / pruning-history / tip-tracker).
- Meta-store key schema, including the `hot:*` active-store tracking keys that make Phase 3 fully meta-store-driven (no filesystem scans for recovery), and the `pruning:index:*` marker family for crash-recoverable atomic prune.
- Active-store architecture (per-chunk ledger + events RocksDB stores; per-tx-index txhash store with 16 column families).
- Ingestion loop, freeze transitions (LFS, events, RecSplit) with per-kind single-flight gates.
- Pruning lifecycle (whole-index granularity, marker bookends the work).
- Query contract (`service_ready` flips after Phase 3 + `compute_resume_ledger` and BEFORE Phase 4's captive-core spinup, so historical queries don't wait the 4–5 minute spinup window).
- Resilience: crash recovery via flag-after-fsync, file-before-flag-delete, and the `pruning:index:*` marker; concurrent-access guards; error handling.

## `01-backfill-workflow.md` (reshaped vs `origin/feature/full-history`)

Substantive changes:

- **CLI removed.** No `stellar-rpc full-history-backfill` subcommand and no per-run flags (`--start-ledger`, `--end-ledger`, etc.). Backfill is now an internal subroutine invoked by Phase 1 (catchup) of the RPC service; operators never invoke it directly.
- **`run_backfill(config, range_start_chunk_id, range_end_chunk_id)`** — signature accepts a chunk range only; the caller (Phase 1) is responsible for computing the range from BSB tip and retention.
- **BSB-only.** Captive core is never a backfill source. Captive core belongs to Phase 4 (live ingestion); when `[BSB]` is absent, Phase 1 is a no-op and Phase 4's captive core archive-catches-up from a retention-aligned resume ledger.
- **Per-chunk BSB.** Each `process_chunk` constructs its own `BSBSource(config.bsb)` at task entry, scoped to its chunk's 10_000 ledgers. No shared source state across tasks.
- **Retention removed from backfill scope.** `RETENTION_LEDGERS` validation lives in the service's `validate_config`; backfill only sees the resolved chunk range.
- **DAG handles cross-iteration tx-index completion.** A trailing-partial tx_index whose last chunk lies past `range_end_chunk_id` defers the `build_txhash_index` task; a subsequent `run_backfill` invocation (or streaming Phase 4) completes it.
- **Flat TOML config.** No nested `[BACKFILL]` / `[STREAMING]` sections — operational sections (`[BSB]`, `[CAPTIVE_CORE]`, `[ACTIVE_STORAGE]`, `[IMMUTABLE_STORAGE.*]`, `[META_STORE]`, `[LOGGING]`, `[HISTORY_ARCHIVES]`) are top-level under `[SERVICE]`.
- **Pseudocode uses named geometry helpers** (`chunk_id_of_ledger`, `chunks_for_tx_index`, `last_chunk_in_tx_index`, etc.) instead of inline `GENESIS_LEDGER` / `LEDGERS_PER_CHUNK` arithmetic.

## `README.md`

Reduced to a reading-order guide with pointers into the two design docs.

## Naming-only changes

- File rename: `03-backfill-workflow.md` → `01-backfill-workflow.md`.
- UPPER_SNAKE_CASE for all TOML keys and section names (matches existing stellar-rpc convention).
- Full-name identifiers (`tx_index_id`, `chunk_id`) replacing single-letter placeholders throughout pseudocode.
- 1-arg path-helper signatures (`ledger_pack_path(chunk_id)`, etc.).
- "RPC service" / "service" naming throughout — never "streaming service" or bare "daemon".
- Internal DAG pipeline renamed `Phase 1/2/3/4` → `Stage 1/2/3/4` (the word "Phase" is reserved for service startup phases 1–4).


tamirms:
While reviewing I used Claude to investigate how Erigon v3 handles the transition from catch-up to live streaming. Its Staged Sync architecture treats catch-up and live ingestion as phases of a single continuous loop driven by one state machine — there's no handoff, no separate command, no operator-sequenced phase transition. The daemon ticks, reads its state, and picks up whatever work is eligible, whether that's backfilling genesis or streaming a new ledger.

**The key insight that makes this possible is that users don't need to backfill arbitrary ranges.** The range to backfill is fully determined by the retention config: either full history, or the last N ledgers. This is a common pattern in other nodes — Erigon v3's `--prune.mode` (archive/full/blocks/minimal presets), Bitcoin Core's `-prune=<N>` (MiB budget), Polkadot's `--state-pruning=<N>` / `--blocks-pruning=<N>` (block windows). In each case, retention is a single declarative knob on the node, and the node itself decides what to fetch and what to prune — not a pair of commands the operator sequences by hand.

Given that, the backfill/streaming split stops being a useful distinction — the daemon just keeps the persisted history matching the configured retention window, ingesting at the tip and pruning behind it, both governed by the same config. I asked Claude to propose an alternative design applying that shape to our problem: https://gist.github.com/tamirms/d18ae6481bf5a9cd9c79606adb090d69

Highlights:

- **One loop, no handoff** — backfill and streaming are phases of the same process.
- **Meta-store is the single source of truth** — the daemon's view of what exists, what's in progress, and what to do next is derived entirely from RocksDB, not from a mix of filesystem inspection and persisted state.
- **Crash recovery falls out of the design** — every operation is idempotent, so the next tick picks up wherever a crash left off.
- **Builds on the current backfill design** with minor changes, not a rewrite.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
+# Streaming Workflow
+
+## Overview
```
karthikiyer56, line 3: @tamirms 

 I could not think of a clean place to post this big comment, so posting here.
~~This comment is about comparsion b/w your gist vs this version of the streaming-doc - what's retained, what changed, where each scenario lands~~
(I am attempting to write individual comments at relevant places with the title "DIFFERING FROM GIST" to draw your attention to the relevant divergences between your gist and this design. I only hope that it is informative and not excessice/noise/difficult to read)

This design is built on top of your gist's framing rather than against it. A lot carried over verbatim.

Salient features:

- One binary, one invocation, no operator-visible "backfill" subcommand or per-run flags. (Your Overview.)
- Catchup-on-startup via the previously-designed backfill subroutine, scoped to the retention window.
- Meta-store-driven recovery — every startup decision reads meta-store keys; no filesystem inspection.
- File-before-flag-delete on every cleanup path (hot DBs in your design; elevated here to a universally-applied rule in the Flag Semantics section).
- Captive core via `LedgerBackend` for live ingestion; chunk + tx-index boundary handlers as the cadence for freeze + prune.
- Per-chunk RocksDB hot store for live events queries — converged on this after backing out an earlier in-memory-deltas draft.
- Validate-config-immutability for `chunks_per_txhash_index` (and `RETENTION_LEDGERS`) on every start.


----


----

full-history/design-docs/01-backfill-workflow.md
```diff
@@ -0,0 +1,588 @@
+# Backfill Workflow
+
+## Overview
+
+Backfill is the RPC service's historical-ingestion subroutine — it pulls ledgers from a configured remote object store (GCS or S3) and writes them as **immutable, query-ready artifacts** on local disk.
```
karthikiyer56, line 5: there are a couple of things going on here.
 - I have renamed this file to be `01`
 - The diff here is large enough that the entire file is shown as changed, but I assure you that the desing has not changed by much, if at all.
 - ***The design is structurally identical to the prior version*** — same DAG with the same three task types (`process_chunk`, `build_txhash_index`, `cleanup_txhash`), same per-task BSB connection per `process_chunk`, same per-flag idempotency on each chunk's three outputs, same all-or-nothing RecSplit recovery, same trailing-partial tx_index pattern, same `cleanup_txhash` separation as a downstream DAG task, same `MAX_RETRIES`-bounded task retries, same flag-after-fsync invariant.

None of those mechanics changed.

---


### What did happen:

- **Tighter framing.** Backfill was previously framed as a standalone subcommand with its own CLI surface and operator-facing ledger-range flags (`flags.start_ledger`, `flags.end_ledger`).

  It's now framed as an internal subroutine of the unified RPC service's Phase 1 — `phase1_catchup` hands it a `(config, range_start_chunk_id, range_end_chunk_id)` tuple and reads back nothing.

  The doc opens with that framing rather than treating "what's a chunk?" before "why does this thing exist?".
- **Seamless flow with the streaming doc.** Geometry, configuration, ledger-source description, crash-recovery framing — each lives in one place, split between `01-backfill-workflow.md` and `02-streaming-workflow.md` with cross-doc anchors instead of being repeated.

  Reading both docs end-to-end no longer hits the same definitions twice.
- **Pruned unnecessary language.** Pre-pseudocode bullet lists that restated what the function showed in code; post-pseudocode prose that summarized the function's own comment; multi-paragraph operator-facing rationale (alignment trade-offs, ledger-range arbitrariness justifications) that no longer applies now that backfill is caller-driven.


----

Two substantive contract changes. Both downstream of "backfill is now a subroutine of Phase 1 instead of a standalone subcommand" — and both get a focused inline comment below:

1. ***`run_backfill` signature.*** `(config, flags)` → `(config, range_start_chunk_id, range_end_chunk_id)`.

   The flags object carrying operator-visible ledger ranges is gone; the caller passes integer chunk ids directly.

   Backfill no longer translates between operator-language (ledgers) and internal-language (chunks).

   See the inline comment at `def run_backfill`.
2. ***Validation surface shrunk.*** The prior `validate(config, flags)` enforced four things: ledger-range bounds, BSB-presence, cpi-immutability, and (implicitly) tx-index alignment via `configured_indexes(...)`.

   The new `validate(range_start_chunk_id, range_end_chunk_id)` is two `assert` lines — argument sanity.

   The other three checks moved to the caller (Phase 1 / service-startup `validate_config`); backfill trusts what it gets.

   See the inline comments at `def validate` and at `build_dag`'s tx-index-aligned-start invariant.

----

### Naming-only changes (no behavior delta — these account for most of the diff line count):

- File rename `03-backfill-workflow.md` → `01-backfill-workflow.md`.
- UPPER_SNAKE_CASE TOML keys (`DEFAULT_DATA_DIR`, `CHUNKS_PER_TX_INDEX`, `IMMUTABLE_STORAGE.LEDGERS.PATH`, etc.).
- `tx_index_id` / `chunk_id` full names replacing single-letter `N` / `C` placeholders.
- 1-arg path-helper signatures (`ledger_pack_path(chunk_id)` instead of `ledger_pack_path(config, chunk_id)`).
- Pseudocode uses named geometry helpers (`chunk_id_of_ledger`, `chunks_for_tx_index`, `last_chunk_in_tx_index`, `first_ledger_in_chunk`, etc.) instead of inline `// LEDGERS_PER_CHUNK + GENESIS_LEDGER` arithmetic.
- Internal RecSplit pipeline rename `Phase 1/2/3/4` → `Stage 1/2/3/4` inside `build_txhash_index` ("Phase" reserved for service-startup phases).
- "RPC service" / "service" naming throughout — never "daemon", never "streaming service".

Net: most of the inline review surface here is line-count noise from renames + structural reshape. The actual contract has two changes, both flowing from the same underlying decision.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+
+### Operator Profiles
+
+Three profiles emerge from config combinations. No separate profile flag.
+
+| Profile | `RETENTION_LEDGERS` | `[BSB]` | Use case | Backfill behavior |
+|---|---|---|---|---|
+| Archive | `0` | present | Public archive node; full history. | Backfill over full history (chunks `[0, bsb_tip_chunk]`) |
+| Pruning-history | `N × LEDGERS_PER_TX_INDEX`, N ≥ 1 | present | Windowed history with bulk initial catchup. | Backfill over retention window (start aligned to first chunk of the tx index containing the retention floor) |
+| Tip-tracker | `N × LEDGERS_PER_TX_INDEX`, N ≥ 1 | absent | App developer; short retention; no object-store dep. | **No-op.** Phase 4 (live ingestion)'s captive core archive-catches-up from a `resume_ledger` aligned to the retention-aligned tx-index boundary |
```
karthikiyer56, line 129: ### DIFFERING FROM YOUR GIST — no-BSB tip-tracker profile

Your gist's Phase 1 catch-up loop opens with:

> ```python
> T = bsb_network_tip(config)
> range_start = chunk_id(retention_floor(T, R))
> ```

implicitly requiring `[BSB]` to be configured. App developers who want last 30 days of transaction lookups without provisioning an object store have no path forward in your design.

This row of the Operator Profiles table is the divergence: `[BSB]` absent + retention > 0 is supported as a third profile.

`phase1_catchup` early-returns; `compute_resume_ledger` Branch 3 falls into `retention_aligned_resume_ledger`, which samples tip from `HISTORY_ARCHIVES.URLS` (existing stellar-rpc plumbing). Phase 4's captive core archive-catches-up from there.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+    # Phases 1-3 + compute_resume_ledger: bring on-disk state into consistency.
+    # No query traffic during this window.
+    last_phase1_chunk_id = phase1_catchup(config, meta_store)
+    phase2_hydrate_txhash(config, meta_store, last_phase1_chunk_id)
+    phase3_reconcile(config, meta_store)
+    resume_ledger = compute_resume_ledger(config, meta_store)
+
+    # Frozen-artifact queries don't need captive core; flip service_ready here
+    # (not after spinup) to avoid an unnecessary 4-5 min outage per restart.
+    set_service_ready()
```
karthikiyer56, line 229: ### DIFFERING FROM YOUR GIST — recovery completes before queries open

Your gist's `start_streaming` epilogue:

> ```python
> notify_lifecycle()
> core = start_captive_core(config, resume_ledger)
> run_in_background(lifecycle_loop, config, meta_store)
> run_ingestion_loop(config, core, hot_dbs, meta_store)
> ```

starts captive core, ingestion, and the lifecycle goroutine concurrently. The lifecycle's first scan completes in some unbounded window after queries become serveable; within that window queries can hit half-recovered state.

This `set_service_ready()` line is the divergence: it flips the in-memory `service_ready` to `true` ONLY after Phases 1–3 finish. Until then, `getLedger` / `getTransaction` / `getEvents` 4xx; only `/getHealth` is served (with `status = "catching_up"`).

Cost: 4xx-during-recovery instead of inconsistent-data-during-recovery. Deeper walk-through in the dedicated thread at `### Phase 3 — Reconcile`

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+Phase 1's backfill range almost always ends mid-tx-index — BSB tip lands wherever the live network is, rarely on an index boundary. 
+For that trailing partial tx index, per-chunk `.bin` files are on disk but `index:N:txhash` is not yet written (RecSplit waits until every chunk of N is complete).
+Phase 2 loads each surviving `.bin` into the active txhash store, then deletes the `.bin` and `chunk:{chunk_id:08d}:txhash` flag.
+After Phase 2: no `.bin` files and no `:txhash` chunk flags remain.
+
+```python
+def phase2_hydrate_txhash(config, meta_store, last_phase1_chunk_id):
+    # Both sweeps: file-before-flag-delete (see Flag Semantics).
+
+    # Sweep 1: clean leftover .bin from completed indexes (cleanup_txhash crashed mid-pair).
```
karthikiyer56, line 330: ### DIFFERING FROM YOUR GIST — Phase 2 has two sweeps, not one

Your gist's phase 2:

> **2. Hydrate the txhash hot DB for the resume index.** ... If backfill left a partial index at the tail of its range (common, since the network tip rarely aligns to an index boundary), streaming needs the partial index's already-frozen `.bin` files loaded into the hot DB so live queries can resolve them.

only hydrates the resume index. It doesn't address tx indexes whose `cleanup_txhash` crashed mid-run — chunks of N that still carry `:txhash` flags + `.bin` files even though the tx index is already RecSplit-complete.

This Sweep 1 (the loop just below) is the divergence: walk every chunk that still carries `:txhash` flags + `.bin` files (across all tx indexes, not just the resume index) and finish the per-pair delete (file-before-flag-delete invariant).

Sweep 2 below then hydrates the trailing partial — the case your gist's phase 2 already covers.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+        # advances the marker.
+        current_tip = meta_store.get("streaming:last_committed_ledger")
+    if current_tip is None:
+        log.warn("phase3 pass1: no tip reference available (archive unreachable + marker absent); skipping past-retention cleanup")
+        return
+
+    floor_ledger = retention_floor_ledger(current_tip, retention_ledgers)
+    floor_chunk  = chunk_id_of_ledger(floor_ledger)
+
+    # Category 1 — lifecycle-unreachable orphans (hot keys + chunks of incomplete prior-run tx indexes).
```
karthikiyer56, line 472: ### DIFFERING FROM YOUR GIST — past-retention orphans your design can't reach

Your gist's `cleanup_past_retention_orphans`:

> ```python
> for key in meta_store.prefix_scan("hot:chunk:"):
>     ...
> for key in meta_store.prefix_scan("hot:index:"):
>     ...
> ```

walks `hot:*` keys only. Chunks-with-`:lfs+:events`-flags from a tx index whose `index:N:txhash` was never written (a prior run that crashed mid-tx-index) are unreachable: the pruning lifecycle's eligibility scan keys off `index:N:txhash = "frozen"` so it can't find them either — they sit on disk forever.

This Category 1 chunk-loop is where this design covers them: `scan_prefix("chunk:")` finds the chunk-level keys, and the sibling blanket-delete (further down) catches partial files left by a crashed `process_chunk` where one kind completed (flag set) but the next hadn't (no flag, partial file with no key).

Deeper walk-through in the dedicated thread at `### Phase 3 — Reconcile`

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+
+    # Branch 2 — staleness check. Skipped when retention=0 (no floor) or
+    # when the archive is unreachable (best-effort fall-through to Branch 1).
+    if config.service.retention_ledgers > 0:
+        current_tip = try_sample_network_tip(config)
+        if current_tip is not None:
+            floor = retention_floor_ledger(current_tip, config.service.retention_ledgers)
+            if marker < floor:
+                log.info(f"compute_resume_ledger: marker ({marker}) below retention floor ({floor}); skipping forward")
+                return retention_aligned_resume_ledger_with_tip(config, current_tip)
```
karthikiyer56, line 614: ### DIFFERING FROM YOUR GIST — BSB-stale-but-configured

In your gist's `start_streaming`, when `T = bsb_network_tip(config)` returns a stale BSB tip (operator stopped populating galexie after an initial backfill of some several million ledgers; BSB outage; ...) the resume cursor lands at BSB-stale tip:

> ```python
> L = max(L, chunk_last_ledger(range_end))
> resume_ledger = L + 1
> ```

Captive core then archive-catches-up from there to the live tip — re-ingesting weeks of ledgers the prune lifecycle is about to delete.

The line above is where this design avoids it: Branch 2 samples live tip from the history archive and leapfrogs to a tx-index-aligned ledger at the retention floor when the marker has fallen below it.

Deeper walk-through with concrete numbers in the dedicated thread at `def compute_resume_ledger`

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+### Keys Introduced by Streaming
+
+| Key | Value | Written when |
+|--|-|--|
+| `streaming:last_committed_ledger` | uint32 | Monotonic progress marker; written via `advance_progress_marker`. Two writers: Phase 1 (post-catchup) and the live ingestion loop (per ledger). |
+| `config:retention_ledgers` | uint32 | First run (stored); enforced on subsequent starts. |
+| `hot:chunk:{chunk_id:08d}:lfs` | `"1"` | Set BEFORE active ledger store dir is created; cleared AFTER dir is removed by freeze. Presence ⇒ dir exists or its lifecycle is incomplete. |
+| `hot:chunk:{chunk_id:08d}:events` | `"1"` | Same pattern, active events store dir. |
+| `hot:index:{tx_index_id:08d}:txhash` | `"1"` | Same pattern, active txhash store dir. Per-index cadence (one per tx index). |
+| `pruning:index:{tx_index_id:08d}` | `"1"` | Set by `prune_tx_index` BEFORE any file delete; cleared AFTER everything else (artifacts + `index:{N}:txhash`). QueryRouter returns 4xx while present; `prunable_tx_index_ids` re-enqueues N if the marker survives a crash. |
```
karthikiyer56, line 147: ### DIFFERING FROM YOUR GIST — schema-level binary keys

Your gist uses multi-valued keys for both artifact state and hot DB state:

> | `chunk:{C:08d}:lfs` | `"frozen"` \| `"pruning"` | Per-chunk pack file state. |
> | `hot:chunk:{C:08d}:lfs` | `"creating"` \| `"ready"` \| `"deleting"` | Per-chunk LFS hot DB. |

Each value carries state semantics ("which state is this thing in?") inside the key value.

This row is the divergence — and the load-bearing example for it: ***every meta-store key in this design is binary*** (`"1"` or absent across all key families).

State distinctions that need persistence get a separate key family instead of overloading an existing key's value. The `pruning:index:N` marker on this row is exactly that pattern: it's its own key, not a `"pruning"` value on `index:N:txhash`.

Reader code stays uniform across all key families — `meta_store.has(...)` is the only primitive needed; no string compares against `"frozen"` / `"pruning"` / `"creating"` / etc.

**Why multi-valued would actively hurt:**
 more transitions = more crash-window combinations to reason about. The `"creating"` → `"ready"` path in particular has a wasted-rebuild window after `rocksdb.open` completes but before the second meta-store put. The `"deleting"` state is redundant under file-before-flag-delete (once the dir is gone, key absence already means "done").

Deeper walk-through of the hot-DB-specific crash-window tradeoff in the dedicated thread ⬆️  at the hot-keys row

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+
+Single RocksDB instance, WAL always enabled. Authoritative source for every startup decision. Reference for the schema and lifecycle below; reads more naturally after [Startup Sequence](#startup-sequence) defines the phases that write and consume these keys.
+
+### Keys Introduced by Streaming
+
+| Key | Value | Written when |
+|--|-|--|
+| `streaming:last_committed_ledger` | uint32 | Monotonic progress marker; written via `advance_progress_marker`. Two writers: Phase 1 (post-catchup) and the live ingestion loop (per ledger). |
+| `config:retention_ledgers` | uint32 | First run (stored); enforced on subsequent starts. |
+| `hot:chunk:{chunk_id:08d}:lfs` | `"1"` | Set BEFORE active ledger store dir is created; cleared AFTER dir is removed by freeze. Presence ⇒ dir exists or its lifecycle is incomplete. |
```
karthikiyer56, line 144: ### Hot DB state keys — binary value catches every crash window your three-state model catches

Under **Data Model → Hot DB state keys** the gist proposes three values per key:

> | `hot:chunk:{C:08d}:lfs` | `"creating"` \| `"ready"` \| `"deleting"` | Per-chunk LFS hot DB. |
> | `hot:chunk:{C:08d}:events` | `"creating"` \| `"ready"` \| `"deleting"` | Per-chunk events hot DB. |
> | `hot:index:{N:08d}:txhash` | `"creating"` \| `"ready"` \| `"deleting"` | Per-index txhash hot DB. |

with the lifecycle:

> ```
> absent  ──►  creating  ──►  ready  ──►  deleting  ──►  absent
> ```

This design adopts the **premise** — active-store directory state belongs in the meta store, not on the filesystem (your gist drove that convergence).

The value is binary: `"1"` if creation or freeze cleanup is in flight, key absent if everything's done.

One put + one delete per active-store lifecycle, instead of three puts + one delete.

Walking the nine distinguishable crash points in the active-store lifecycle, binary catches every recovery scenario three-state catches — except one window where binary is *strictly better*:

***Crash point 5: after `rocksdb.open` completes, before the second `put("ready")` writes.*** In your three-state model, on restart `state == "creating"` triggers `delete_dir_if_exists(path)` and `create_fn(path)` — wiping a fully-functional RocksDB and rebuilding from scratch.

Wasted work, no correctness gain.

In binary, there's no second put after `rocksdb.open`, so this crash window doesn't exist.

The other eight crash points (before any write, mid-mkdir, mid-rocksdb.open, mid-ingestion, post-freeze pre-`delete_dir`, mid-rmdir, post-rmdir pre-clear-key, etc.) recover equivalently — different state values, same outcome.

Three reasons this design landed on binary:

1. **Same recovery coverage**, one fewer wasted-rebuild crash window.
2. **Halfs the meta-store op count** per active-store lifecycle (4 ops → 2 ops). Over a 6_000-chunk archive backfill that's ~24K ops vs ~12K — small absolute cost in either case, but the simpler op count means fewer crash-window combinations to reason about.
3. **Schema consistency** — every meta-store key here is binary (`"1"` or absent). `chunk:{C}:lfs / :events / :txhash`, `index:{N}:txhash`, `hot:*`, `pruning:index:*` — all binary. Three-state for `hot:*` would split the schema; uniform binary keeps reader code free of state-name string compares.

Phase 3 Pass 2 (`pass2_recover_in_flight_transitions`) is a single `for hot_key in scan_prefix("hot:")` loop with four branches dispatched off `(scope_id vs resume)` and freeze-flag-presence — A (resume target → Phase 4 reopens), B (flag-is-truth → finish cleanup), C (id < resume → finish-interrupted-freeze), D (id > resume → log + defensive cleanup).

Each branch is a 2-3 line idempotent body.

The three-state `"creating"` vs `"ready"` distinction would either fold into "is the dir openable?" — RocksDB itself answers that idempotently on its own — or add a fifth branch that a binary scheme doesn't need.

that being said...
`"creating"` / `"ready"` / `"deleting"` does have one advantage: schema-self-documentation in a meta-store dump.
But I dont feel that its worth the simplicity cost.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+- `chunk:{chunk_id:08d}:events`
+- `chunk:{chunk_id:08d}:txhash`
+- `index:{tx_index_id:08d}:txhash`
+
+All values are binary (`"1"` or absent); prune-in-progress is tracked via the separate `pruning:index:*` key family rather than overloading the value space.
+
+### Flag Semantics
+
+- **Flag-after-fsync** — a freeze flag is set only AFTER its artifact is fsynced. Present ⇒ artifact durable.
+- **File-before-flag-delete** — cleanup paths delete the file/dir FIRST, clear the key LAST. Reverse order would orphan a file with no meta-store record on a crash mid-pair, recoverable only by filesystem scan.
```
karthikiyer56, line 164: ### Cleanup-ordering invariant — convergence, applied universally

The gist's hot DB lifecycle uses the file-before-flag-delete ordering:

> Hot DB keys are put before the dir is created (`"creating"` state) and deleted only after rmdir completes (`"deleting"` state).

Same ordering principle here, but elevated from a hot-DB-specific convention to a universally-applied rule across every cleanup path:

- Phase 2 hydration (both sweeps): `.bin` file deleted before `chunk:C:txhash` flag cleared.
- Backfill `cleanup_txhash`: per-chunk `.bin` deleted before per-chunk flag cleared.
- Streaming `prune_tx_index`: pack files, events cold segments, `.bin` files, RecSplit `.idx` files deleted before `chunk:*` and `index:*` flags cleared.
- Active-store directory cleanup at freeze-task end: `delete_dir_if_exists(...)` before `meta_store.delete("hot:...")`.

The reverse order (flag cleared before file deleted) would orphan a file with no meta-store record on a crash mid-pair, recoverable only by a filesystem scan — which violates the meta-store-driven recovery principle this design shares with your gist:

> Both lifecycles are observable purely from meta-store keys — no filesystem inspection needed.

Codifying it in one place (the Flag Semantics section) makes it a single rule to verify in TDD across every cleanup path, rather than a per-call-site convention.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+def open_active_ledger_store(config, meta_store, chunk_id) -> RocksDBStore:
+    ...  # same shape: put hot:chunk:{C}:lfs (idempotent), mkdir, open RocksDB.
+
+def open_active_events_store(config, meta_store, chunk_id) -> RocksDBStore:
+    ...  # same shape: put hot:chunk:{C}:events (idempotent), mkdir, open RocksDB.
+```
+
+Pure-streaming restarts (no recent Phase 1 output) never see `.bin` files; the live path writes txhash directly to the active store. Phase 2 is a no-op.
+
+### Phase 3 — Reconcile
```
karthikiyer56, line 404: ### Recovery shape — explicit Phase 3 + service_ready gate vs lifecycle-loop scans

Your **Crash Recovery** section frames recovery as a consequence of the design:

> Recovery is a consequence of the design, not a separate specification:
>
> - **Eligibility is meta-store-driven.** The goroutine's four scans read the meta store and emit the operations needed to make state match the invariants. A crash mid-op leaves a state that the next scan detects and corrects.

Same outcome here — recovery is meta-store-driven, idempotent, no filesystem scan — but with an explicit one-shot Phase 3 (reconcile) at startup, split into two passes:

- **Pass 1 (`pass1_drop_past_retention_state`)** — drops everything below the retention floor in three categories: lifecycle-unreachable orphans (hot:* keys + chunks of an incomplete prior-run tx index, with sibling blanket-delete to catch partial files left by a crashed `process_chunk`); past-retention complete state (drained eagerly, even though the pruning lifecycle could reach it); and below-floor `pruning:index:*` markers from a prior crash.
- **Pass 2 (`pass2_recover_in_flight_transitions`)** — single `for hot_key in scan_prefix("hot:")` loop with four-branch classification: A (resume target → Phase 4 reopens), B (flag-is-truth → finish cleanup), C (id < resume → finish-interrupted-freeze), D (id > resume → log + defensive cleanup).

Both passes run to completion before captive core starts. `set_service_ready()` flips the in-memory `service_ready` after Phase 3 returns and `compute_resume_ledger` settles. Until then, `getLedger` / `getTransaction` / `getEvents` 4xx (only `/getHealth` is served).

Two specific advantages of explicit one-shot:

1. **Recovery completes before query-serving** — clients never see half-recovered state, only "service is catching up" (HTTP 4xx + `getHealth.status = "catching_up"`).

   In your design the lifecycle goroutine wakes from `notify_lifecycle()` and runs eligibility scans concurrently with captive-core startup and ingestion; during the window between captive-core start and the first scan completing, queries can in principle hit state that hasn't been reconciled.

2. **Pass 1 has a path the lifecycle-loop pattern doesn't.** Your `cleanup_past_retention_orphans` walks `hot:*` keys, but it can't reach a chunk-with-`:lfs+:events`-flags from an incomplete prior-run tx index whose `index:N:txhash` was never written — the pruning lifecycle's eligibility scan keys off `index:N:txhash = "frozen"`.

   Pass 1's chunk loop (Category 1 in the docstring) finds these via `scan_prefix("chunk:")` and the sibling blanket-delete; the same key in your design would sit on disk forever.

Cost: one prefix scan at startup (sub-second at cpi=1_000 with a handful of orphans). The lifecycle goroutine's scans during steady-state are still per-chunk-boundary, but no longer carry initial-recovery responsibility.

Your design treats recovery as eventually consistent. This design treats it as a precondition for serving. Both are internally consistant; this is the more conservative side of the tradeoff on the query-correctness axis.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+    set_service_ready()
+    phase4_live_ingest(config, meta_store, resume_ledger)
+```
+
+See [Query Contract](#query-contract) for the query-gating contract.
+
+### Phase 1 — Catchup
+
+```python
+def phase1_catchup(config, meta_store) -> Optional[int]:
```
karthikiyer56, line 238: ### Phase 1 / `compute_resume_ledger` — separation of concerns

Under **Startup → Phase 1** the gist's catch-up loop tracks a local `L` that's later used as `resume_ledger = L + 1`:

> ```python
> L = low_water    # = streaming:last_committed_ledger
> ...
> while True:
>     ...
>     run_backfill(config, range_start, range_end)
>     L = max(L, chunk_last_ledger(range_end))
>     ...
> ```

Two distinct concerns share `L` in your design: the catchup loop (which advances `L` as backfill makes progress) and the resume-cursor derivation (which reads `L + 1` after the loop exits).

The `max(L, chunk_last_ledger(range_end))` is the line that bridges them — it accumulates information from both `low_water` (the prior shutdown's progress) and `chunk_last_ledger(range_end)` (this run's catchup progress) into one variable, so a single `L + 1` works for both fresh-start and same-tip-restart paths.

This design splits those two concerns into separate functions. The only durable signal between them is the meta-store progress marker.

***`phase1_catchup` does only catchup.*** Its body is a `while True:` loop calling `run_backfill` and `advance_progress_marker`. Returns `Optional[int]` — the highest chunk_id it completed — but the value is consumed by Phase 2 to find the trailing partial tx index, *not* by Phase 4 to compute resume.

***`compute_resume_ledger` does only resume derivation.*** It reads the meta-store progress marker and dispatches off three named branches:

```python
def compute_resume_ledger(config, meta_store) -> int:
    marker = meta_store.get("streaming:last_committed_ledger")

    if marker is None:                                   # Branch 3
        return retention_aligned_resume_ledger(config)

    if config.service.retention_ledgers > 0:             # Branch 2
        current_tip = try_sample_network_tip(config)
        if current_tip is not None:
            floor = retention_floor_ledger(current_tip, config.service.retention_ledgers)
            if marker < floor:
                return retention_aligned_resume_ledger_with_tip(config, current_tip)

    return marker + 1                                    # Branch 1
```

What the separation buys:

- ***Local readability.*** A developer reading `compute_resume_ledger` can verify all three branches without modeling Phase 1's loop semantics. A developer reading `phase1_catchup` can verify catchup correctness without thinking about Phase 4's resume math. Neither function needs to encode the other's invariants.
- ***Independent change surface.*** Adding Branch 4 to `compute_resume_ledger` doesn't touch Phase 1. Refactoring the catchup loop doesn't touch resume math. The inter-phase contract is the meta-store progress marker — a durable, schema-described signal rather than a local variable's value at a particular point in a loop.
- ***Independent tests.*** Phase 1 tests use a mock BSB and assert on the progress marker after `run_backfill` returns. `compute_resume_ledger` tests use meta-store state fixtures (marker present / absent / below floor) and assert on the returned ledger. Neither test has to stand up the other layer.

Tradeoff: `compute_resume_ledger`'s Branch 2 (the `try_sample_network_tip` HTTP GET against the history archive) runs every startup when retention is configured, even when Branch 1 is the eventual hit. A few 10ish seconds of cost, paid once per startup for the separation.

full-history/design-docs/02-streaming-workflow.md
```diff
@@ -0,0 +1,945 @@
...
+### Readiness Signal
+
+- In-memory boolean. Flipped `true` by `set_service_ready()` after Phases 1–3 complete and `compute_resume_ledger` returns, BEFORE Phase 4's captive-core spinup. Frozen-artifact queries don't need captive core; flipping after spinup would add an unnecessary 4–5 minute outage per restart.
+- Not persisted; every startup begins `false`. Clients see 4xx on every startup until Phase 4 is reached, regardless of prior runs.
+- HTTP server binds before Phase 1, so `/getHealth` is always servable.
+
+### Behavior During Phases 1–3
+
+- `/getLedger`, `/getTransaction`, `/getEvents` → HTTP 4xx, no payload.
+- `/getHealth` → served. Response shape matches existing stellar-rpc: `status` (`catching_up` during Phases 1–3, `healthy` during Phase 4), `latestLedger` (= `streaming:last_committed_ledger`, or `0` if absent), `oldestLedger`, `ledgerRetentionWindow`.
```
karthikiyer56, line 928: ### `getHealth` semantics — pinning the existing stellar-rpc contract

Payload matches the existing stellar-rpc shape exactly: `status`, `latestLedger`, `oldestLedger`, `ledgerRetentionWindow` — same fields, same types.

The only new value is `status = "catching_up"` during Phases 1–3 (flips to `"healthy"` once Phase 4 is reached).

**No `drift` field, no `networkTip` field, no `bsbTip` field** — I was flitting with idea after talking to @Shaptic , but then explicitly rejected becuase:

- A drift field would require periodic network-tip sampling during Phase 4 steady state (HTTP GET to `HISTORY_ARCHIVES.URLS` every N seconds). Recurring overhead for a signal existing clients don't need; they already handle stale `latestLedger` gracefully via standard semantics.
- A `networkTip` field splits the "what's the current ledger" question across two fields. Existing stellar-rpc clients consume `latestLedger` as the authoritative cursor.
- A `bsbTip` field leaks BSB-specific implementation detail to clients. Doesn't apply at all to the tip-tracker profile (no BSB).
- We dont do it today in existing rpc anyways - idt

full-history/design-docs/01-backfill-workflow.md
```diff
@@ -0,0 +1,588 @@
...
+- All `process_chunk` tasks for a tx index must complete before `build_txhash_index` fires.
+- `cleanup_txhash` runs after `build_txhash_index` succeeds.
+- Cleanup deletes the raw `.bin` files and their `chunk:{chunk_id:08d}:txhash` meta keys.
+
+### Main Flow
+
+`run_backfill` is invoked by the service's [Phase 1 (catchup)](./02-streaming-workflow.md#phase-1--catchup) with an integer chunk range:
+
+```python
+def run_backfill(config, range_start_chunk_id, range_end_chunk_id):
```
karthikiyer56, line 302: ### `run_backfill` — chunk range, not ledger range

The signature changed from `run_backfill(config, flags)` (where `flags` carried operator-facing fields like `flags.start_ledger`, `flags.end_ledger`) to:

```python
def run_backfill(config, range_start_chunk_id, range_end_chunk_id):
    validate(range_start_chunk_id, range_end_chunk_id)
    dag = build_dag(config, range_start_chunk_id, range_end_chunk_id)
    dag.execute(max_workers=MAX_CPU_THREADS)
```

Two integers, both chunk ids, plus the invariant (documented in `build_dag`) that `range_start_chunk_id` is tx-index-aligned by the caller.

Why the shape change matters:

- ***No more ledger ↔ chunk translation inside backfill.*** The caller (`phase1_catchup`, defined in 02-streaming-workflow.md) samples BSB tip in ledgers, computes `retention_aligned_start_chunk(...)` and `bsb_latest_complete_chunk_id(...)`, and hands `run_backfill` already-resolved chunk ids.

  Backfill takes them at face value.
- ***No flags object means no operator-facing surface.*** The prior version had to defend against "operator passes an arbitrary ledger range" (unaligned, partial chunks, out-of-bounds, etc.).

  The new contract is two integers — anything satisfying `0 <= start <= end` is processed; everything else is the caller's problem.
- ***Crash recovery still works the same way.*** Phase 1's loop re-invokes `run_backfill` with whatever range BSB exposes on the next iteration; per-flag idempotency inside `process_chunk` skips already-frozen artifacts. The signature change doesn't affect resume behavior.

Internal mechanics — DAG construction, task dispatch, retries, all-or-nothing RecSplit recovery, trailing-partial deferred build, `cleanup_txhash` as a downstream task — unchanged from the prior version.

full-history/design-docs/01-backfill-workflow.md
```diff
@@ -0,0 +1,588 @@
...
+    # issue (upper or lower bound) surfaces at runtime as a per-task get_ledger failure.
+    assert range_start_chunk_id >= 0
+    assert range_end_chunk_id >= range_start_chunk_id
+```
+
+### DAG Setup
+
+```python
+def build_dag(config, range_start_chunk_id, range_end_chunk_id):
+    # Invariant: range_start_chunk_id is always tx-index-aligned
```
karthikiyer56, line 330: ### `build_dag` — tx-index-aligned start is the caller's contract, not backfill's check

The new comment block at the top of `build_dag`:

```python
# Invariant: range_start_chunk_id is always tx-index-aligned 
# Phase 1 (catchup) is the only caller of this function, and it aligns the start chunk ID
# to the nearest tx index boundary, which is why validate() doesn't check for that. 
# This means the first tx index in the range is always fully covered by the chunk range,
# and thus always buildable.
```

is the explicit articulation of a contract that's now upstream of backfill. The flow:

1. `phase1_catchup` (02-streaming-workflow.md) computes `retention_aligned_start_chunk(last_ledger_in_chunk(end_chunk), retention_ledgers)`. That helper aligns DOWN to a tx-index boundary.
2. `phase1_catchup` calls `run_backfill(config, range_start, end_chunk)` with the aligned start.
3. `build_dag` *trusts* alignment — no check, no fallback. The first tx_index in `range(first_index, last_index + 1)` is therefore guaranteed to be fully covered by the chunk range; `build_txhash_index` and `cleanup_txhash` are scheduled for it unconditionally.

The trade: alignment correctness lives in one place (the caller), `build_dag` is simpler. A future caller — say, a CLI tool reintroduced for ops — would need to either align externally or wrap `run_backfill` with its own alignment helper. Documented invariant, but no defensive check.

The trailing partial at the *end* of the range works the same way it always has: `if last_chunk_in_tx_index(tx_index_id) <= range_end_chunk_id:` skips `build_txhash_index` + `cleanup_txhash` for the trailing tx_index, leaving its `.bin` files on disk + per-chunk `:txhash` flags set.

A subsequent `run_backfill` (next Phase 1 iteration), Phase 2 hydration on next service start, or Phase 4 ingestion reaching `last_chunk_in_tx_index(tx_index_id)` completes it.

Same mechanic as before, now operating on chunk ids instead of ledger ranges.

full-history/design-docs/02-streaming-workflow.md
karthikiyer56: THIS IS NOT AN ISSUE ANYMORE...
if this were to be an isue, as in there were residual bin files that needed cleanup, that wud mean that backfill was broken.
thats not the case.
bacfkill still unconditionally removes bin files for al indexes that have been built.
this sweep 1 is no longer needed

karthikiyer56:
closing this out in favor of #787

PR #650, opened 2026-04-01
Add thin CGo zstd wrapper for packfile compression

tamirms:
## Summary

Second in a series of PRs checking in the packfile implementation for full history RPC v2. See #633 for the design doc and #649 for the first PR (FOR encoding + offset index).

This PR adds a thin CGo wrapper around system libzstd (>= 1.5.7) for compression and decompression, and updates build infrastructure to provide the required library version.

**Why a custom wrapper instead of existing Go bindings?** (see `zstd.go` package doc for details)
- **klauspost/compress (pure Go):** ~4x slower than system libzstd on our 28KB blocks at level 3
- **DataDog/zstd:** vendored C symbols cause ELF symbol interposition with RocksDB on Linux, making RocksDB 5x slower

**Why 1.5.7?** Benchmarked on representative production data (28KB blocks of 128 events):

| | libzstd 1.5.5 | libzstd 1.5.7 | Change |
|---|---|---|---|
| Compress | 717 MB/s | 766 MB/s | **+6.9%** (p=0.000, n=10) |
| Decompress | 3.43 GB/s | 3.65 GB/s | **+6.3%** (p=0.000, n=10) |

**Linking strategy:**
- Linux: statically linked via `#cgo linux LDFLAGS: -l:libzstd.a` — the binary carries libzstd 1.5.7 baked in, no runtime dependency. This applies to Docker builds, CI, and the Jenkins .deb package pipeline (which uses our Dockerfile).
- macOS: dynamically linked via pkg-config/homebrew for local development.

**Build infrastructure changes:**
- Dockerfile: switch build stage from `golang:1.25-bookworm` to `golang:1.25-trixie` (Debian 13 ships libzstd 1.5.7 in default repos), add `libzstd-dev` and `pkg-config`
- CI: build libzstd 1.5.7 from source in the `setup-go` action on Linux (with SHA256 verification); install via brew on macOS

## Test plan
- [x] Roundtrip compression/decompression
- [x] Corrupt data detection (truncated, header-corrupted, bit-flipped)
- [x] Empty/nil input handling (no SIGSEGV on nil C pointer)
- [x] Context reuse across multiple calls with varying payloads
- [x] Scratch buffer aliasing contract
- [x] dst buffer reuse (no unnecessary allocation)
- [x] Close idempotency (no double-free)
- [x] Checksum detection of single-bit corruption
- [x] WithoutChecksum option
- [x] Concurrent independent instances
- [x] Encode/Decode after Close returns error

[...]

PR #702, opened 2026-04-23
Add packfile writer

tamirms:
## Summary

Third in a series of PRs checking in the packfile implementation for full-history RPC v2.

This PR adds the writer:

- `packfile.go` — on-disk format constants (magic, version, 76-byte trailer size, content-hash flag bit, `ErrContentHashMismatch`)
- `writer.go` — `Writer` with a unified worker pipeline for compression/hashing; caller-supplied per-record `Compressor` and optional SHA‑256 content hashing
- `writeback_linux.go` / `writeback_other.go` — Linux-only `sync_file_range(SYNC_FILE_RANGE_WRITE)` hint for background page flushing; no-op elsewhere
- `writer_test.go` — validation, lifecycle, trailer / index format, passthrough, content-hash extract parity, and error propagation
- `writer_bench_test.go` — end-to-end write throughput across representative configurations

Also tightens `zstd.Compressor.Close` and `zstd.Decompressor.Close` to return `error` so they satisfy `io.Closer` directly (always return nil; the error return is purely for interface conformance).

Depends on the intpack + offset-index codec already merged in #649 and the zstd wrapper in #650.

Full roundtrip coverage (reader-side decode of record contents) arrives with the reader in the next PR.

## Key design points

- **Caller-supplied compression**: `WriterOptions.NewCompressor func() Compressor` returns a per-worker `Compressor` (`Encode([]byte) ([]byte, error)` + `io.Closer`). The interface's `Close` lets stateful codecs (e.g. zstd's CGo context) release native resources deterministically when a worker exits or the writer finalizes, rather than waiting on GC finalizers. If `nil`, records are written as-is (passthrough). This supports both zstd compression and the rocksdb-passthrough use case (where ledgers are pre-compressed and stored verbatim).
- **Opaque `Format` identifier**: `WriterOptions.Format` is a caller-assigned `uint32` written to the trailer. The library does not interpret it; readers use it to dispatch to the matching decompressor.
- **Independent content hashing**: `ContentHash bool` enables a chunked SHA‑256 over the logical item stream, independent of compression. `ContentHashExtract func([]byte) ([]byte, error)` lets callers normalize already-encoded items before hashing — so a packfile-built file and a rocksdb-built file produce the same content hash over the same logical ledger stream.
- **Unified worker pipeline**: workers spawn whenever the writer has CPU work (compression and/or content hashing). `Concurrency` sets the number of parallel worker goroutines (0 defaults to 1). Each worker compresses (if a compressor is supplied) and hashes (if `ContentHash` is enabled, via an inner hash sub-goroutine that overlaps with compress). Records are dispatched in ordinal order; `runWriter` reorders worker results and appends to the file in order. Pure passthrough writers (no compressor, no content hash) bypass the pipeline and write records directly from the main goroutine.
- **Error propagation**: the first fatal error is recorded in an `atomic.Pointer[error]` so both the pipeline goroutine and the main goroutine observe it with proper happens-before; first-error-wins. This prevents a data-loss window where `AppendItem` could keep enqueuing records after a background write/compress error. `Compressor.Close` errors from worker defers are funneled through the same atomic.

## Test plan

- [x] `TestCreateValidation` — input validation (Concurrency / BytesPerSync / ItemsPerRecord)
- [x] Lifecycle: `TestCreateFailsIfFileExists`, `TestCloseWithoutFinishRemovesFile`, `TestFinishTwice`, `TestAppendItemAfterFinish`, `TestCloseIdempotent`, `TestEmptyPackfile`, `TestAppendItemEdgeCases` (no-args, empty item, multi-part concatenation)
- [x] On-disk format: `TestTrailerFields`, `TestOffsetIndexDecodes`, `TestAppDataPreserved`, `TestItemsPerRecordOne`, `TestOverwriteReplaces`
- [x] Passthrough (rocksdb use case): `TestPassthroughStoresVerbatim` — items stored byte-identical when `NewCompressor` is nil; `TestNewCompressorReturnsNil` — caller's factory returning nil doesn't panic
- [x] Content hash: `TestContentHashParity` (different worker counts produce same hash), `TestContentHashExtract` (extract hook), `TestContentHashExtractConcurrent` (extract hook with multiple workers), `TestConcurrentContentHashExtractError` (extract errors surface from worker goroutines)
- [x] Compress error propagation: `TestSerialCompressErrorSurfaces`, `TestConcurrentCompressErrorSurfaces`, `TestCompressorCloseErrorSurfaces` (Close errors from worker defers surface from Finish; run under `-race`)
- [x] Write error propagation via `/dev/full` (Linux): `TestConcurrentWriteErrorSurfaces`, `TestSerialWriteErrorSurfaces` (run under `-race`)
- [ ] Roundtrip item-content verification lands with the reader in the next PR


cmd/stellar-rpc/internal/packfile/writer.go
```diff
@@ -0,0 +1,723 @@
...
+const defaultItemsPerRecord = 128
+
+// ErrWriterClosed is returned by AppendItem or Finish after the Writer has
+// been closed (successfully finalized or aborted).
+var ErrWriterClosed = errors.New("packfile: writer is closed")
+
+// Format is a caller-assigned identifier stored in the trailer. The packfile
+// library does not interpret Format values; readers use them to dispatch to
+// the matching decompressor. Callers agree on Format values out of band.
+type Format uint32
```
karthikiyer56, line 32: so, I read the PR summary - sepcificailly the first 2 bullets "Caller-supplied compression:" and "Opaque Format Identifier" - technically they go hand in hand right? 

This is a BYOC thing right — bring your own compression logic?
Pretty neat and makes the packfile even more generic.
packfile doesnt have to know about zstd or lz2 or anything else specifically anymore,

As in, the `Compressor` interface and the `Format` uint32 feel like two halves of one contract — caller brings the codec, caller picks the int that names it, packfile just writes the int into the trailer and forgets about it, yes?


This will need some packfile doc updates though...
[here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L138-L142)

Also, [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L361)  and [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L480) --->  these lines need to be changed/removed since packfile is codec-agnostic now.

can you also update with a  doc commit on this PR or fold into the reader PR before that one ships? otherwise the next person reading the spec is gonna get a very misleading picture of the architecture

tamirms: sure, I can update the design doc and make it up to date in a separate PR once the implementation is complete. I'll do that after this PR is merged and the next PR (which implements the reader and the remaining tests) are merged.

cmd/stellar-rpc/internal/packfile/packfile.go
```diff
@@ -0,0 +1,19 @@
+package packfile
+
+import "errors"
+
+const (
+	magic       = 0x48434C53 // "SLCH" in the on-disk (little-endian) byte order
+	version     = 1          // on-disk format version; bump on any breaking trailer/index change
+	trailerSize = 76
```
karthikiyer56, line 8: trailer was 64 right? In the pckfile doc?
and now you added format = uint32. So this shud be 68?
whats with the  76? 


and if it is relly 76, then there are quite a few places that the doc needs to be updated with.
I found 2 [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L293) and [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L448), there might be more

cmd/stellar-rpc/internal/packfile/packfile.go
```diff
@@ -0,0 +1,19 @@
+package packfile
+
+import "errors"
+
+const (
+	magic       = 0x48434C53 // "SLCH" in the on-disk (little-endian) byte order
```
karthikiyer56, line 6: the doc has this `magic` fielld  as `0x534C4348` [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L452)

cmd/stellar-rpc/internal/packfile/writer.go
```diff
@@ -0,0 +1,723 @@
...
+// at a time. Encode is called once per record before write. Close is called
+// when the writer no longer needs the compressor (Finish or Close), so
+// stateful codecs that hold non-Go resources (e.g. CGo zstd contexts) can
+// release them deterministically rather than waiting on GC finalizers.
+//
+// Encode's returned slice may alias an internal buffer of the compressor and
+// is valid until the next call on this Compressor; it must not alias the
+// input slice. A Compressor is not safe for concurrent use — the writer
+// creates one per worker goroutine via WriterOptions.NewCompressor.
+type Compressor interface {
```
karthikiyer56: this one is a slight doozy.
so..
pre-pivot the lib had three per-record integrity strategies baked in (Compressed used zstd's xxHash64, Uncompressed had an explicit 4B CRC32C, Raw had nothing). 
post-pivot it looks like that whole layer is gone — `Encode`'s output gets written verbatim and the lib doesnt add a CRC around the payload.

the only integrity thats still guaranteed is the FOR-group CRC32C inside multi-item records + the offset-index CRC + the trailer CRC.

iiuc the reasoning is — lib has no idea what `Encode` produced (could be zstd, lz4, raw, encrypted, anything), so it cant decide whether or how to wrap integrity around it.


if so its a real shift in what the lib guarantees vs what it delegates, and the doc at [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L334-L358)  and [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L480) need to be rewritten to make this explicit. right now they read as if the lib still owns per-record payload integrity.

cmd/stellar-rpc/internal/packfile/packfile.go
tamirms: Two intentional growth steps, not just the one for `Format`:

**64 → 72** (RecordFormat refactor): added `Format uint32` (+4) but also restructured the layout to make all `uint32` fields 4-byte aligned (the original had them at offsets 6, 10, 14, 18, 22 — works but messy) and reserved 4 bytes near the end for future flag/format extensions without breaking the format. Net +8:
- +4 for the new `Format` field
- +2 alignment padding after the flags byte (bytes 6-7)
- +4 reserved tail bytes (64-67) before CRC
- −2 from packing `itemsPerRecord` and `indexGroupSize` into a shared 4-byte slot as `uint16`s

**72 → 76** (ItemsPerRecord widening): the 72-byte layout packed `itemsPerRecord` as `uint16` (cap 65,535) to share a slot with `indexGroupSize`, but the API exposes `ItemsPerRecord int`, so the on-disk type didn't match the parameter type — felt like an accidental ceiling. I widened to `uint32` (cap matches int conceptually), shifted subsequent fields, +4 bytes.

Current 76-byte layout:
```
0-3   magic
4     version
5     flags
6-7   reserved
8-11  Format (uint32)
12-15 recordCount
16-19 totalItems
20-23 itemsPerRecord (uint32)
24-25 indexGroupSize (uint16)
26-27 reserved
28-31 indexSize
32-35 appDataSize
36-67 contentHash (32 bytes)
68-71 reserved
72-75 CRC32C(trailer[0:72])
```

So 76, not 68. Happy to tighten if you'd prefer. Will update the design doc in a follow-up PR.

cmd/stellar-rpc/internal/packfile/writer.go
tamirms: yeah, the compression algorithm is responsible for checksumming. this was the case before as zstd implements it internally. I'll update the design doc in a separate PR

cmd/stellar-rpc/internal/packfile/packfile.go
tamirms: The doc value is the original — `0x534C4348` — but it was wrong because `binary.LittleEndian.PutUint32(0x534C4348)` writes the bytes `0x48 0x43 0x4C 0x53` on disk, which spells "HCLS", not "SLCH".

The code value is now `0x48434C53` so that little-endian encoding produces `0x53 0x4C 0x43 0x48` = "SLCH" on disk, matching the intent. Copilot caught this earlier and we fixed the constant; the design doc just hasn't been updated yet. Will fold this into the doc-update follow-up PR.

cmd/stellar-rpc/internal/packfile/writer.go
tamirms: Sorry, I realized I forgot to answer your question:

> As in, the Compressor interface and the Format uint32 feel like two halves of one contract — caller brings the codec, caller picks the int that names it, packfile just writes the int into the trailer and forgets about it, yes?

yes, that's right. the caller encodes the format as an int and based on that it can determine how to decompress the record (if necessary) when reading the packfile.

cmd/stellar-rpc/internal/packfile/writer.go
tamirms: However, you're right that, in this PR, there is no direct support for the uncompressed with CRC case.  We do not have anything in full history rpc which requires that feature. However, it would be nice to support it if is easy to do so.

There is a workaround where you can implement a Compressor which just takes in the given payload and adds a CRC. But, I don't think it's super clear that you can do that. I'll try to brainstorm with claude to see if there is a better way to handle this case.

tamirms: ok, I renamed Compressor to RecordEncoder and update the docs to make the use case of uncompressed + CRC more apparent.

cmd/stellar-rpc/internal/packfile/packfile.go
karthikiyer56: > Happy to tighten if you'd prefer.

 not following. 
 Did you mean
(a) tighten it meaning, reduce from 76 to something else 
OR 
(b) tighten as in, in leave an inline comment displaying byte layout?

I am ok with it being 76. Its a trailer with some sub 100 bytes of meta information - no biggie

cmd/stellar-rpc/internal/packfile/packfile.go
karthikiyer56: While we are here on this topic of doc updates..

this comment is pretty much the only place that explicitly says `little-endian.`
the design doc only mentions LE inside the FOR-encoding section [here](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/packfile-library.md?plain=1#L312)
the trailer table + offset-index sections list field types like `uint32` / `uint16` without specifying endianness.
anyone trying to write a non-Go reader would have to reverse-engineer the convention.

perhaps when updating docs, add a line like so somewhere up top:
> **All multi-byte integer fields in the packfile metadata — trailer fields, offset-index FOR groups, per-record item-size FOR groups, CRC32C values, and the length prefixes inside the SHA-256 content-hash construction — are encoded little-endian. Item content, app data, and the 32-byte content-hash digest are stored byte-for-byte verbatim.**

cmd/stellar-rpc/internal/packfile/packfile.go
```diff
@@ -0,0 +1,19 @@
...
+	version     = 1          // on-disk format version; bump on any breaking trailer/index change
+	trailerSize = 76
+)
+
+// On-disk flag bits (uint8 at trailer offset 5). Only one flag is currently
+// defined; the remaining bits are reserved for future use.
+const (
+	flagContentHash uint8 = 1 << 0
```
karthikiyer56, line 14: I think this `flagContentHash` also might need a doc update.
Flaggin it for reference

cmd/stellar-rpc/internal/packfile/writer_test.go
```diff
@@ -0,0 +1,675 @@
...
+// TestPassthroughStoresVerbatim verifies that with NewRecordEncoder == nil, items
+// are written to disk exactly as received (the rocksdb-passthrough use case).
+func TestPassthroughStoresVerbatim(t *testing.T) {
+	items := [][]byte{
+		[]byte("hello world"),
+		[]byte("second item"),
+		[]byte("third"),
+	}
+	// ItemsPerRecord=1 + no encoder → record N is exactly item N's bytes.
+	path := writePackfile(t, WriterOptions{Format: 99, ItemsPerRecord: 1}, items)
```
karthikiyer56: Can you also test this for the case where `itemsPerRecord > 1`, say 2 and 3?  in the same test and assert correctness?

also minor nit:
it took me a minute to figure out that it is passthru coz `Encoder:nil`
Even though it is the default and redundant, can I still ask you to add it for clarity?

cmd/stellar-rpc/internal/packfile/writer.go
```diff
@@ -0,0 +1,728 @@
...
+	"io"
+	"math"
+	"os"
+	"sync"
+	"sync/atomic"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/intpack"
+)
+
+const defaultItemsPerRecord = 128
```
karthikiyer56, line 23: this const shud be packfile.go, perhaps?

cmd/stellar-rpc/internal/packfile/writer.go
```diff
@@ -0,0 +1,728 @@
...
+// library doesn't interpret the output bytes — readers dispatch on the
+// trailer's Format field to apply the matching decoder.
+//
+// Encode is called once per record before write. Close is called when the
+// writer no longer needs the encoder (Finish or Close), so stateful encoders
+// that hold non-Go resources (e.g. CGo zstd contexts) can release them
+// deterministically rather than waiting on GC finalizers.
+//
+// Encode's returned slice may alias an internal buffer of the encoder and is
+// valid until the next call on this RecordEncoder; it must not alias the
```
karthikiyer56: > it must not alias the
// input slice

this shud probably be more strict and say `Ecnoder must not modify the input string either` on the off chance that someone migh t use a custom encoder that does something funky and mutates input as well 🤷

karthikiyer56 (approved):

cmd/stellar-rpc/internal/packfile/packfile.go
tamirms: I meant (a) — I could have skipped the alignment padding and reserved bytes to keep it closer to 68 if you preferred a tighter footprint. Glad 76 is fine.

cmd/stellar-rpc/internal/packfile/packfile.go
tamirms: Good catch. I'll fold this into the design-doc update PR — the little-endian convention is consistent throughout (trailer fields, FOR groups, CRC32C, length prefixes for content hash). Easy to call out at the top.

cmd/stellar-rpc/internal/packfile/packfile.go
tamirms: Acked, will include `flagContentHash` semantics in the design-doc update PR.

cmd/stellar-rpc/internal/packfile/writer_test.go
tamirms: Good points — applying both: extending the test to cover `ItemsPerRecord = 2` and `3`, and adding explicit `NewRecordEncoder: nil` for readability.

cmd/stellar-rpc/internal/packfile/writer.go
tamirms: `defaultItemsPerRecord` is a writer-side behavior default (used by `resolveItemsPerRecord`), not an on-disk format constant — `packfile.go` currently holds magic / version / trailerSize / flag bits, all of which are part of the wire format. I'd lean toward keeping it in `writer.go` where it's used, but happy to move if you'd prefer the consolidation.

cmd/stellar-rpc/internal/packfile/writer.go
tamirms: Good point — I'll strengthen the contract to "must not alias OR modify the input slice". The writer's still using `work.data` for hashing in parallel with the encoder call, so any mutation would corrupt the hash too. Worth making explicit.

Issue #661, opened 2026-04-02
[events] Freeze (hot → cold)


urvisavla:
Implement the segment freeze process:
1. Start new hot segment immediately (new segment accepts writes)
2. Build cold segment from old hot segment using cold segment writer
   - If using LiveWriter: `Freeze()` in-place, then build index files
   - If using flat files/RocksDB: read events, produce cold segment
3. Old hot segment continues serving reads during freeze
4. Mark segment as frozen in DB atomically after all cold files are written
5. Discard old hot segment files and in-memory index

Add tests for: freeze produces valid cold segment, concurrent reads during freeze, two hot segments coexist briefly.

Issue #660, opened 2026-04-02
[events] Hot segment reader


urvisavla:
Implement the read primitives for the hot segment, against the storage backend used by the writer (#659).

**Scope:**
- `Lookup(termKey) → bitmap` of event IDs (returns `ErrTermNotFound` on miss).
- `FetchEvents([]eventID) → iter` of decoded payloads.
- `All() → iter` over every event in the chunk in event-id order.
- Metadata accessors: `ChunkID`, `EventCount`, `Offsets`.
- Concurrency: many concurrent readers; writer holds an exclusive mutex during ingest.

**Out of scope (separate issues):**
- Bitmap intersection, ledger-range trimming, and post-filter across multiple segments — that's the segment-coordination layer (#663).
- Wiring into the query handler (part of #663).

Add tests covering: lookup hits and misses, ordered streaming of all events, fetch-by-id round-trips, and concurrent-read correctness.

Issue #659, opened 2026-04-02
[events] Hot segment writer


urvisavla:
Implement the writer for the hot segment — the active-ingest store that owns one Chunk's events while it's being filled.

**Scope:**
- Per-ledger ingest: write events to storage, update in-memory bitmap mirror, update ledger offset array — all atomic in one batch.
- Up-front validation: ledger must be in the chunk's range AND must be the next-expected ledger (no duplicates, no gaps).
- Per-ledger commit with fsync and durability guarantees.
- Open / Close lifecycle; warmup on reopen to reconstruct the in-memory mirrors from the on-disk caches.

**Out of scope (separate issues):**
- Wiring the writer into the live ingestion pipeline (#664).
- Freeze (hot → cold transition) (#661).

Add tests for write correctness, per-ledger durability, sequence validation, and reopen-after-close recovery.

Issue #658, opened 2026-04-02
[events] Backfill workflow


urvisavla:
Implement the backfill workflow that reads historical LCM and produces cold segments. For each segment (10,000 ledgers):
- Extract events from each ledger's LCM
- Append events to packfile writer
- Build in-memory bitmaps and ledger offset array
- At segment completion: finalize using cold segment writer

No concurrent reads, no per-ledger commits, no crash recovery needed. Failed segments are discarded and restarted. Segments are independent so backfill can be parallelized across workers.

Add tests covering the end-to-end path: feed sample LCM → produce cold segments → verify queryable via cold segment reader.

PR #705, opened 2026-04-29
[events] In-memory bitmap index

urvisavla:
## Summary

Implements the in-memory bitmap index (issue #655) for the full-history event store. This is the foundational building block used by the hot segment write path (#659), query path (#660), freeze (#661), and backfill (#658).

- `TermKey` + `ComputeTermKey` — 16-byte term keys via xxh3-128
- `BitmapStore` interface — pluggable storage backend
- `memBitmaps` — in-memory store with list→roaring bitmap promotion at 64 entries (sparse term optimization)
- `EventIndex` — top-level interface that owns a BitmapStore
- `LedgerOffsets` — ledger range to event ID range conversion

`memBitmaps` supports an immutable lifecycle: after `Close()`, mutating operations return `ErrClosed`, `All()` iterates without lock contention, and `Get()` skips bitmap cloning. This matches the freeze workflow where a chunk transitions from mutable → immutable before serialization.

## Test plan

- [x] All unit tests pass with `-race`
- [x] 100% statement coverage for the events package
- [x] `BenchmarkEventIndex_10M` validates production-scale memory: ~577 MB heap for 9M events / 1.94M terms (matches design doc's measured ~369 MB at typical scale)
- [x] `BenchmarkSparseTermMemory` confirms list optimization saves ~4.5x heap vs always-bitmap for sparse terms
- [x] Full project tests still pass (no breakage outside this package)


cmd/stellar-rpc/internal/events/eventindex.go
```diff
@@ -0,0 +1,66 @@
+package events
+
+import (
+	"iter"
+
+	"github.com/RoaringBitmap/roaring/v2"
+)
+
+// EventIndex manages a chunk's term-to-bitmap index.
+type EventIndex interface {
```
tamirms, line 15: BitmapStore and EventIndex look really similar. When would it be appropriate to use BitmapStore instead of  EventIndex or vice versa?

cmd/stellar-rpc/internal/events/ledgeroffsets.go
```diff
@@ -0,0 +1,125 @@
...
+func NewLedgerOffsets(startLedger uint32) *LedgerOffsets {
+	return &LedgerOffsets{
+		offsets:     make([]uint32, 0, LedgersPerChunk),
+		startLedger: startLedger,
+	}
+}
+
+// Append records the cumulative event count after processing one ledger.
+// The ledger must be the next expected in sequence. Counts must be non-decreasing.
+func (m *LedgerOffsets) Append(ledger, cumulativeCount uint32) error {
```
tamirms: would it make sense for Append to take the event count for the current ledger instead of the cumulative count?

```
Append(ledger, eventCount uint32) error
```

from eventCount you can derive the cumulative count by adding it to the previous m.offsets value. This would allow the caller of LedgerOffsets to not have to maintain a cumulative count.

cmd/stellar-rpc/internal/events/termkey.go
```diff
@@ -0,0 +1,49 @@
...
+	var buf []byte
+	if n <= len(scratch) {
+		buf = scratch[:n]
+	} else {
+		buf = make([]byte, n)
+	}
+	buf[0] = byte(field)
+	copy(buf[1:], value)
+
+	// TODO: replace with streamhash.PreHashInPlace once #654 (MPHF integration) lands.
```
tamirms, line 42: 👍 

this is a good reminder that I should move streamhash into a stellar repo at some point before #654 so the work is not blocked

cmd/stellar-rpc/internal/events/ledgeroffsets.go
```diff
@@ -0,0 +1,125 @@
...
+		return fmt.Errorf("expected ledger %d, got %d", expected, ledger)
+	}
+	m.offsets = append(m.offsets, cumulativeCount)
+	return nil
+}
+
+// EventIDRange converts an inclusive absolute ledger range [startLedger, endLedger]
+// into a half-open event ID range [startEventID, endEventID).
+// If the range contains no events, startEventID == endEventID.
+func (m *LedgerOffsets) EventIDRange(startLedger, endLedger uint32) (uint32, uint32, error) {
```
tamirms: there is a case where a query could touch 2 adjacent segments. For that scenario, I wonder if a more useful API would be:

```go
// EventID returns the id of the first event in the given ledger
EventID(ledger uint32) (uint32, error)
```

This function could be used in the case where we have a range contained within a single segment (call EventID() on start and then again on end) and also in the case where we have a range spanning two adjacent segments (call EventID() on start using the first segment and then again on the next segment).

cmd/stellar-rpc/internal/events/ledgeroffsets.go
urvisavla: Done in 2c6d29bb. `Append(ledger, eventCount)` now maintains the cumulative internally.

cmd/stellar-rpc/internal/events/ledgeroffsets.go
urvisavla: Done in 2c6d29bb. Settled on `EventIDs(ledger uint32) (start, end uint32, err error)` — a middle ground that returns the half-open range for a single ledger. Multi-ledger queries compose by calling on first/last; cross-chunk queries call on each chunk's `LedgerOffsets`. Lets you avoid the +1 dance for the common single-ledger case while still composing across chunk boundaries.

cmd/stellar-rpc/internal/events/eventindex.go
urvisavla: Still thinking through this — will follow up. The current intent: `BitmapStore` is the storage backend interface (works with raw `TermKey`, swappable hot/cold/disk implementations), while `EventIndex` is the user-facing facade that handles term key computation from `(value, field)` pairs. Open to merging or restructuring if it's cleaner.

urvisavla: Collapsed into a single `EventIndex` interface in 362be195. Dropped the `BitmapStore` interface and the `eventIndex` wrapper; the `(value, field)` methods now live directly on `memBitmaps`. With a single backend today, the two-interface split was premature — we can reintroduce a storage-backend interface when a second impl (disk/cold-tier) actually shows up.

tamirms (approved):

PR #712, opened 2026-04-30
Add packfile reader

### LINKED ISSUE #719: Packfile 5: Generic packfile reader
## TL;DR

Tracker for the generic packfile reader — the type-agnostic core that decodes compressed, indexed pack files written by the generic writer (PR https://github.com/stellar/stellar-rpc/pull/702).

## Status

**In review.** PR https://github.com/stellar/stellar-rpc/pull/712 open.

## What this enables

This reader is the foundation that "Packfile 7: Decoder primitive" builds on. The downstream consumer-side readers (ledger query path in backfill epic, events cold-segment reader in events epic #665) depend on the decoder primitive, which in turn depends on this generic reader.

## Acceptance

- [ ] PR https://github.com/stellar/stellar-rpc/pull/712 merged.

## Notes

This issue exists for tracking purposes under the packfile epic. PR is the source of truth.

tamirms:
## Summary

Adds the reader for the packfile format whose writer shipped in #702. Lifted from the events-analysis prototype, adapted to match the writer's caller-supplied-codec architecture.

Public surface, mirror of `WriterOptions`/`Create`:
- `Open(path, ReaderOptions)` — non-blocking; lazy I/O on first read.
- `Reader.ReadItem(position, fn)` — single positional access.
- `Reader.ReadRange(start, count)` — sequential scan; coalesces consecutive records into batched `ReadAt` calls via a pooled 1 MiB buffer.
- `Reader.ReadItems(ctx, positions, fn)` — scattered reads with optional worker fan-out (off by default).
- `Reader.Verify(ctx)` — replays the chunked-SHA-256 content hash and compares to the stored digest.
- Metadata accessors: `ContentHash`, `AppData`, `Trailer`, `TotalItems`.
- `Close` — closes the underlying file. The caller-supplied `RecordDecoder` is not closed; it's typically a single shared instance across many Readers, so its lifecycle is the caller's concern.

Callers plug in the codec via `ReaderOptions.RecordDecoder` — a single concurrent-safe decoder instance shared across all reads (nil for passthrough). `*zstd.Decompressor` satisfies this directly: it is concurrent-safe via an internal `sync.Pool` of decompression contexts, so one Decompressor per process amortizes context construction across every Reader the app opens.

**Codec contract unified.** `Format`, `RecordEncoder`, and `RecordDecoder` now live together in `codec.go` (moved out of `writer.go` and the old `decoder.go`). Both interfaces use the same `Encode/Decode(dst, src []byte) ([]byte, error)` shape, so the writer and reader share a footgun-free aliasing contract: the returned slice always owns its bytes, no "valid until next call" caveat. `*zstd.Compressor` and `*zstd.Decompressor` satisfy these interfaces directly with no adapter; compile-time assertions in `writer_bench_test.go` lock the contract.

## Deviations from the design doc and events-analysis prototype

The reader matches the writer's caller-supplied-codec architecture, which means several documented behaviors changed. The deliberate ones worth flagging:

- **Default `Concurrency` is 1 (serial)**, not 8 as the doc says. Parallel scattered I/O is opt-in. Matches the writer; avoids silent goroutine fan-out.
- **`ReadRange` / `ReadItems` return `ErrPositionOutOfRange`** instead of panicking. Consistent with `ReadItem`; safer for runtime conditions like a stale view of file size. `ReadItems` also returns a distinct `ErrPositionsUnsorted` for duplicate/unsorted positions.
- **`RecordDecoder` is a single concurrent-safe instance, caller-owned.** The events-analysis prototype vended per-Reader decoders and leaked C contexts on `sync.Pool` GC eviction. The design here moves concurrency inside the decoder (`zstd.Decompressor` pools DCtxs internally with finalizers), so one decoder amortizes across the whole process and Reader lifecycle stops touching codec state — `Reader.Close` only closes the file.

The on-disk format changes (76-byte trailer, `Format` field, single `flagContentHash`) all came in with the writer in #702.

The design doc itself is now stale relative to both writer and reader; updating it is deferred to a separate PR per project convention.

## Test plan

- [x] `go build ./cmd/stellar-rpc/internal/packfile/... ./cmd/stellar-rpc/internal/intpack/...` clean
- [x] `go test -race -count=1 ./cmd/stellar-rpc/internal/packfile/... ./cmd/stellar-rpc/internal/intpack/...` passes
- [x] `golangci-lint run ./cmd/stellar-rpc/internal/packfile/... ./cmd/stellar-rpc/internal/intpack/...` 0 issues
- [x] Round-trip matrix tests cover {passthrough, xor codec} x {serial, c=4} x {hash on/off} x {ItemsPerRecord in {1, 128}}
- [x] Race-stress sub-tests for the worker fan-out path
- [x] Corruption tests for trailer CRC, index CRC, magic, unknown flags, content hash mismatch, app-data tamper (uncovered by design)
- [x] `Reader.Close` after failed Open returns the open error joined with the file-close error (`TestCloseAfterFailedOpen`)
- [x] Cross-Reader workspace-pool stress: alternates passthrough and decoder-mode Readers against the package-level pool (`TestWorkspaceSharedAcrossDecoderModes`)
- [x] Negative `Concurrency` surfaces an error on the first read call

### DIFF
--- a/cmd/stellar-rpc/internal/packfile/reader.go
+++ b/cmd/stellar-rpc/internal/packfile/reader.go
@@ -68,18 +68,17 @@ type ReaderOptions struct {
 	Concurrency int
 }
 
-// Trailer holds the parsed trailer fields of an open packfile.
-//
-// Flags is the raw on-disk flag byte; HasContentHash is the typed view of
-// flagContentHash for callers' convenience. Other flag bits are reserved.
+// Trailer holds the parsed trailer fields of an open packfile. Reserved /
+// internal-only on-disk fields (raw flags byte, FOR group size constant,
+// stored CRC) are intentionally not exposed: HasContentHash is the typed
+// view of the only currently-defined flag bit; IndexSize and AppDataSize
+// describe regions a caller may want to inspect.
 type Trailer struct {
 	Version        uint8
-	Flags          uint8
 	Format         Format
 	RecordCount    uint32
 	TotalItems     uint32
 	ItemsPerRecord uint32
-	IndexGroupSize uint16
 	IndexSize      uint32
 	AppDataSize    uint32
 	ContentHash    [32]byte
@@ -330,14 +329,12 @@ func doOpen(path string) openResult {
 		file: f,
 		trailer: Trailer{
 			Version:        v,
-			Flags:          flags,
 			Format:         format,
 			RecordCount:    uint32(recordCount),    //nolint:gosec // bounded by trailer field width
 			TotalItems:     uint32(totalItems),     //nolint:gosec // bounded by trailer field width
 			ItemsPerRecord: uint32(itemsPerRecord), //nolint:gosec // bounded by trailer field width
-			IndexGroupSize: uint16(indexGroupSize),
-			IndexSize:      uint32(indexSize),   //nolint:gosec // bounded by trailer field width
-			AppDataSize:    uint32(appDataSize), //nolint:gosec // bounded by trailer field width
+			IndexSize:      uint32(indexSize),      //nolint:gosec // bounded by trailer field width
+			AppDataSize:    uint32(appDataSize),    //nolint:gosec // bounded by trailer field width
 			ContentHash:    contentHash,
 			HasContentHash: hasContentHash,
 		},
@@ -616,10 +613,15 @@ func (r *Reader) ReadItems(ctx context.Context, positions []int, fn func(idx int
 	var firstErr error
 	var canceled atomic.Bool
 
-	wg.Add(numWorkers)
+	// setErr records the first error and signals workers to stop. Workers
+	// must still `return` after calling it; setErr does not unwind the goroutine.
+	setErr := func(err error) {
+		errOnce.Do(func() { firstErr = err })
+		canceled.Store(true)
+	}
+
 	for range numWorkers {
-		go func() {
-			defer wg.Done()
+		wg.Go(func() {
 			rd := r.getDecoder()
 			defer r.putDecoder(rd)
 			bp, _ := readBufPool.Get().(*[]byte)
@@ -631,9 +633,8 @@ func (r *Reader) ReadItems(ctx context.Context, positions []int, fn func(idx int
 				if bi >= len(batches) || canceled.Load() {
 					return
 				}
-				if ctx.Err() != nil {
-					errOnce.Do(func() { firstErr = ctx.Err() })
-					canceled.Store(true)
+				if err := ctx.Err(); err != nil {
+					setErr(err)
 					return
 				}
 				batch := batches[bi]
@@ -650,8 +651,7 @@ func (r *Reader) ReadItems(ctx context.Context, positions []int, fn func(idx int
 				}
 
 				if _, err := r.file.ReadAt(readBuf, readStart); err != nil {
-					errOnce.Do(func() { firstErr = err })
-					canceled.Store(true)
+					setErr(err)
 					return
 				}
 
@@ -662,20 +662,18 @@ func (r *Reader) ReadItems(ctx context.Context, positions []int, fn func(idx int
 						recOff := r.offsets[rec] - readStart
 						recEnd := r.offsets[rec+1] - readStart
 						if err := rd.Decode(readBuf[recOff:recEnd], rec); err != nil {
-							errOnce.Do(func() { firstErr = err })
-							canceled.Store(true)
+							setErr(err)
 							return
 						}
 						prevRec = rec
 					}
 					if err := fn(k, rd.Item(localIdx)); err != nil {
-						errOnce.Do(func() { firstErr = err })
-						canceled.Store(true)
+						setErr(err)
 						return
 					}
 				}
 			}
-		}()
+		})
 	}
 	wg.Wait()
 	return firstErr
@@ -716,6 +714,13 @@ func (r *Reader) Verify(ctx context.Context) error {
 
 // Close releases all resources. Safe to call multiple times.
 // Must always be called, even if no query methods were called.
+//
+// Close is NOT safe to call concurrently with in-flight read methods —
+// callers must ensure every ReadItem / ReadRange / ReadItems call has
+// returned before invoking Close, otherwise a worker mid-Decode may race
+// with a RecordDecoder.Close on the decoder it is currently using.
+// Read methods themselves are safe to call concurrently with one another;
+// only the Close vs read interaction needs serialization at the caller.
 func (r *Reader) Close() error {
 	r.closeOnce.Do(func() {
 		_ = r.waitOpen() // drain goroutine, populate all fields
--- a/cmd/stellar-rpc/internal/packfile/reader.go
+++ b/cmd/stellar-rpc/internal/packfile/reader.go
@@ -159,7 +159,7 @@ func Open(path string, opts ReaderOptions) *Reader {
 	r.concurrency = max(opts.Concurrency, 1)
 
 	r.recordPool.New = func() any {
-		rec := &record{}
+		rec := &record{reader: r}
 		if r.newRecordDecoder != nil {
 			rec.decoder = r.newRecordDecoder()
 		}
@@ -357,28 +357,27 @@ func doOpen(path string) openResult {
 	return res
 }
 
-// getRecord returns a pooled record-processing workspace configured for
-// this reader. Placed
-// here, between the lifecycle constructors above and the read methods below,
-// because every read path goes through it.
+// getRecord returns a pooled record-processing workspace. Placed here,
+// between the lifecycle constructors above and the read methods below,
+// because every read path goes through it. Records carry a back-pointer
+// to their owning Reader (set in recordPool.New) so per-call setup is
+// just the pool Get.
 //
 //nolint:funcorder // pool plumbing kept near callers; matches writer.go style
 func (r *Reader) getRecord() *record {
 	rec, _ := r.recordPool.Get().(*record)
-	rec.totalItems = r.totalItems
-	rec.itemsPerRecord = r.itemsPerRecord
 	return rec
 }
 
 // putRecord returns a record to the pool. Its RecordDecoder, if any, stays
-// bound for reuse — Reader.Close closes it on shutdown.
+// bound for reuse — Reader.Close closes it on shutdown. Reset only the
+// per-call scratch/payload/sizes/offsets state; the reader back-pointer
+// and the decoder field persist across pool cycles.
 //
 //nolint:funcorder // paired with getRecord
 func (r *Reader) putRecord(rec *record) {
-	rec.totalItems = 0
-	rec.itemsPerRecord = 0
 	rec.scratch = rec.scratch[:0]
-	rec.decompressed = rec.decompressed[:0]
+	rec.payload = rec.payload[:0]
 	rec.sizes = rec.sizes[:0]
 	rec.offsets = rec.offsets[:0]
 	r.recordPool.Put(rec)
--- a/cmd/stellar-rpc/internal/packfile/reader.go
+++ b/cmd/stellar-rpc/internal/packfile/reader.go
@@ -127,12 +127,12 @@ type Reader struct {
 
 	waitOpen func() error // blocks until background open completes
 
-	// All decoders ever created by decoderPool.New, used so Close can close
+	// All records ever created by recordPool.New, used so Close can close
 	// every caller-supplied RecordDecoder deterministically. sync.Pool
 	// itself doesn't expose a way to drain.
-	decoderMu   sync.Mutex
-	decoders    []*decoder
-	decoderPool sync.Pool
+	recordsMu  sync.Mutex
+	records    []*record
+	recordPool sync.Pool
 
 	closeOnce sync.Once
 	closeErr  error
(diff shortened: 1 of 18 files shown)


cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+		return nil
+	}
+
+	hasher := newContentHasher(r.itemsPerRecord)
+	i := 0
+	for item, err := range r.ReadRange(0, r.totalItems) {
+		if err != nil {
+			return err
+		}
+		hasher.Add(item)
```
urvisavla: The writer has an option of transforms items via `ContentHashExtract` before hashing, but the reader has no matching option. So Verify will fail on any file written with extract. Also trailer has no field saying which extract was used. Is there a planned follow up for this?

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+// goroutines (serially when Concurrency is 0 or 1) and in arbitrary order.
+// The idx argument identifies which element in positions the data corresponds to.
+//
+// positions must be sorted ascending with no duplicates. Returns
+// ErrPositionOutOfRange if any position is outside [0, TotalItems) or if
+// positions are not strictly sorted.
+//
+//nolint:gocognit,cyclop,funlen // batch partitioning + worker fan-out; splitting hurts readability
+func (r *Reader) ReadItems(ctx context.Context, positions []int, fn func(idx int, data []byte) error) error {
+	if len(positions) == 0 {
```
urvisavla: There's a discrepancy in how the three Read methods handle invalid input.
  
   - ReadItem(-1, fn), validates, returns ErrPositionOutOfRange.
   - ReadRange(-100, 0), silently succeeds in contradiction to what the doc which promises ErrPositionOutOfRange for negative start 
  - ReadItems(ctx, [], fn),  also silently succeeds and skips waitOpen() too.
  
Should we move the empty case early returns in ReadRange and ReadItems to happen after validation and waitOpen()?

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+	closeOnce sync.Once
+	closeErr  error
+}
+
+// Open returns a Reader immediately. File I/O runs in a background goroutine;
+// the first read call blocks until the file is ready. Open itself does not
+// return an error — option-validation and I/O failures are deferred to the
+// first method call that needs the result.
+// Close must always be called.
+func Open(path string, opts ReaderOptions) *Reader {
```
urvisavla, line 170: What’s the use case for async Open?
It seems to add a lot of machinery: channels, panic recovery, deferred errors, etc. The main benefit appears to be opening multiple readers in parallel without making the caller write a goroutine.
Is that benefit worth the added complexity, or could the parallelism be pushed to callers who need it?

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+
+// Trailer holds the parsed trailer fields of an open packfile. The fields
+// mirror the on-disk trailer: a caller can introspect the file's metadata
+// (e.g. for diagnostic dumps or for verifying a stored Checksum against an
+// independent recomputation).
+//
+// HasContentHash is the typed view of the only currently-defined flag bit;
+// the raw flags byte itself is not exposed because no caller can act on
+// unknown bits (Open rejects them via knownFlags).
+type Trailer struct {
```
urvisavla: The Trailer struct seems to be reader-only right now. The writer writes trailer fields directly into bytes. 
Common Trailer with marshal() / unmarshal() could keep the on-disk encoding in one place and reduce the chance of drift.

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+	} else {
+		rec.scratch = rec.scratch[:size]
+	}
+	if _, err := r.file.ReadAt(rec.scratch, start); err != nil {
+		return fmt.Errorf("packfile: read record %d: %w", recordIdx, err)
+	}
+	if err := rec.decode(rec.scratch, recordIdx); err != nil {
+		return err
+	}
+	return fn(rec.item(localIdx))
```
urvisavla, line 424: Should fn != nil be checked earlier? Right now a nil fn can panic after I/O has already happened. Same for ReadItems().

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+	var firstErr error
+	var canceled atomic.Bool
+
+	// setErr records the first error and signals workers to stop. Workers
+	// must still `return` after calling it; setErr does not unwind the goroutine.
+	setErr := func(err error) {
+		errOnce.Do(func() { firstErr = err })
+		canceled.Store(true)
+	}
+
```
urvisavla: Consider using errgroup.WithContext here. I think it would simplify the code. workers can return errors directly, the first error wins, and the derived context handles cancellation propagation.

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+	newRecordDecoder func() RecordDecoder
+
+	waitOpen func() error // blocks until background open completes
+
+	// All records ever created by recordPool.New, used so Close can close
+	// every caller-supplied RecordDecoder deterministically. sync.Pool
+	// itself doesn't expose a way to drain.
+	recordsMu  sync.Mutex
+	records    []*record
+	recordPool sync.Pool
```
urvisavla: I’m wondering if a bounded channel pool would simplify this. The current approach of using sync.Pool requires the list of records protected via mutex.  A chan *record sized to max concurrency and that way let Close drain the channel and free each decoder, avoiding the separate list.

cmd/stellar-rpc/internal/packfile/reader.go
```diff
@@ -0,0 +1,759 @@
...
+// Consecutive records are coalesced into single ReadAt calls using a pooled
+// 1MB buffer, minimizing I/O syscalls for large ranges.
+// Each yielded []byte is valid only until the next iteration — copy if you
+// need to retain it. Safe to break early. Thread-safe.
+//
+// Yields ErrPositionOutOfRange (one-shot) if start or count is negative or
+// the range falls outside [0, TotalItems).
+//
+//nolint:gocognit,cyclop // single batched-coalesce loop; splitting hurts readability
+func (r *Reader) ReadRange(start, count int) iter.Seq2[[]byte, error] {
```
urvisavla, line 446: Curious about use case for ReadRange?

If a caller wants to read many sequential records, should they use ReadRange, or enumerate the positions and use ReadItems? ReadItems supports concurrency, while ReadRange doesn’t, though ReadItems may have different ordering behavior under fan-out.

Verify seems to be the only caller of ReadRange today, so I’m trying to understand when callers are expected to use one vs. the other.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Real gap, fixed. The reader now has `ReaderOptions.ContentHashExtract` mirroring the writer-side option, and Verify applies it before hashing. Caller is responsible for plumbing the same extract on both sides — same shape as `NewRecordEncoder`/`RecordDecoder` configuration. Added `TestContentHashExtractVerifyRoundTrip` for the symmetric path.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Good catch — the ordering was wrong. ReadRange/ReadItems now do `waitOpen` → validate → empty-check, so `ReadRange(-100, 0)` correctly returns `ErrPositionOutOfRange` instead of silently succeeding. Also split `ErrPositionsUnsorted` out of `ErrPositionOutOfRange` while I was there, so duplicate/unsorted positions in ReadItems are a distinct sentinel from out-of-range. Doc updated.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Kept async, but simplified the machinery. The events query design relies on per-request parallel-open of multiple files — a single getEvents request opens `events.pack`, `index.pack`, `index.hash` per cold segment and overlaps their initial I/O. Sync Open with caller-side goroutines would push the deferred-error pattern back onto every caller.

That said, you were right that the implementation was over-machined. `doOpen` now runs inside the `sync.OnceValue` closure directly; the explicit goroutine + channel + panic-recovery wrapper is gone. Same async property, ~10 fewer lines.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Done. `Trailer.marshal(dst)` and `unmarshalTrailer(src)` now single-source the on-disk encoding in `packfile.go`; both writer and reader call them instead of hand-coding offsets. Same pattern as `tOff*` constants and `encodeForIndex`/`decodeForIndex` (also factored out in this pass — the FOR-index wire format had similar drift risk).

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: `fn` is a required argument; passing nil is a programmer bug, not a runtime condition. The natural panic at the call site is the right behavior — same as passing nil to any non-optional function parameter in Go.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Done. ReadItems now uses `errgroup.WithContext` — workers return errors directly, `ctx.Err()` replaces the canceled flag, `g.Wait()` returns the first non-nil error. Drops ~10 lines of hand-rolled state.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Tried that and then went further. The chan-based pool was a small improvement over sync.Pool + tracking list, but the fundamental tension was that decoder pooling was per-Reader scope, which wastes the decoder construction across the many short-lived Readers an events query opens.

The reworked design:
- Decoder is now a **single concurrent-safe instance** the caller passes in via `ReaderOptions.RecordDecoder`. `zstd.Decompressor` is internally concurrent-safe with its own `sync.Pool[*dctx]` + finalizers. One Decompressor per process, shared across all Readers — decoder construction amortizes across the whole app lifetime.
- `RecordDecoder` is now a single-method interface (no Close); decoder cleanup is the caller's concern.
- Per-Reader record pool dropped entirely. Replaced with a package-level `recordWorkspacePool` (just `*record` workspaces — slice headers, no Close, GC-friendly sync.Pool semantics).
- `Reader.Close` is 3 lines (closes the file; decoder is caller-owned).

So no records list, no mutex, no chan, no per-Reader scope at all. The complexity that motivated your comment is gone rather than reorganized.

cmd/stellar-rpc/internal/packfile/reader.go
tamirms: Verify is the only current caller, but ReadRange will also serve the getLedgers endpoint when returning multiple consecutive ledgers. The API shapes are distinct in ways that matter beyond either use:
- ReadRange: `iter.Seq2`, strictly in-order, break-early via the iterator, single-threaded (no concurrency overhead).
- ReadItems: callback-based, may yield out-of-order under concurrency, sorted-or-scattered positions.

Forcing a getLedgers caller through ReadItems would mean building the positions slice manually and accepting out-of-order yields under fan-out — losing the iterator ergonomics and the in-order guarantee. Updated the docs to make the choice explicit (`Use ReadRange for in-order streaming reads ... Use ReadItems for sorted-or-scattered with optional worker fan-out`).

urvisavla (commented):
Two follow-up comments on the polish commit (4bb30159):

**`packfile.go:55`** — Is `ContentHashExtract` meant to be agreed out-of-band? The trailer has no flag for it, so a reader with mismatched (or missing) extract gets `ErrContentHashMismatch` on an intact file — indistinguishable from corruption. The commit message mentions a `HasContentHashExtract` flag bit — was that intended to land?

**`reader.go:102`** — Now that `Trailer.marshal` / `unmarshalTrailer` live in `packfile.go`, the `Trailer` struct itself should probably move there too — type and methods are split across files. Pure relocation, no API change.

urvisavla (approved):
Nice work! left a couple of minor comments for consideration.

tamirms:
**`packfile.go:55`** — Good catch on the commit-message drift — the `HasContentHashExtract` reference is stale (earlier draft added the flag, reverted before the final push, commit message didn't get updated). Apologies for the noise.

I don't think a flag bit is the right addition, because `Format` is already the encapsulation point. The package docs at codec.go:14 say readers dispatch on Format to pick the matching decoder; extract, record layout, and item interpretation work the same way — the caller picks them based on Format. So Format is the version of the encoding: change what's hashed, change what's in a record, change the codec → bump Format. A flag bit for extract would be a half-measure for one axis of a more general rule (any schema change → new Format) that the current design already covers cleanly.

Worth making this more explicit in the docstring though — I'll add a line to the Format godoc spelling out the bump-on-any-encoding-change rule, since right now it's implicit.

**`reader.go:102`** — Agreed, pure relocation. Will move `Trailer` from reader.go into packfile.go alongside `marshal` / `unmarshalTrailer`.

PR #733, opened 2026-05-07
Add Layer-1 pkg/rocksdb wrapper (#685)

### LINKED ISSUE #685: Backfill 4: Layer 1 pkg/rocksdb wrapper (CF-aware, WAL, iteration, flock, per-store config struct)
## TL;DR

Build the **generic Layer-1 RocksDB wrapper** that becomes the **shared foundation** for every Layer-2 facade in the project — meta store today, hot ledger store + hot txhash store + hot events store later. CF-aware (zero / one / many column families per store), WAL-on default, flock-protected, idempotent `Open` with auto-mkdir, per-store config struct hardcoded by callers (never operator-facing). Each Layer-2 facade has its own RocksDB directory, its own config, its own flock, and its own block cache — but they all run through this same wrapper.

This slice also absorbs the Work-0 metastore relocation: as its first commit, `git mv` the existing `cmd/stellar-rpc/internal/backfill/meta_store.go` into the new `cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/` and create the `cmd/stellar-rpc/internal/fullhistory/{backfill, streaming, pkg}/` package tree.

## Why this is the foundation, and what derives from it

This wrapper is generic bytes-in / bytes-out. **Every** RocksDB-backed store in the project is a Layer-2 facade that builds on this wrapper, with its own typed API + key schema + hardcoded tuning. The wrapper supports them all transparently — particularly the divergent CF schemas — without per-facade special-casing in the wrapper itself.

| Layer-2 facade | RocksDB directory | CF schema | Status |
|---|---|---|---|
| Meta store (#689) | `{DEFAULT_DATA_DIR}/meta/rocksdb` | Default CF only | tracked in this epic |
| Hot ledger store (#584 reframed) | `{DEFAULT_DATA_DIR}/active/ledgers` | Default CF only | streaming-side |
| Hot txhash store | `{DEFAULT_DATA_DIR}/active/txhash` | **16 CFs**, routed by `txhash[0] >> 4` | streaming-side; tracked under the txhash store epic |
| Hot events store (Urvi's #665, sub-issue TBD) | `{DEFAULT_DATA_DIR}/active/events` | **Multiple CFs** (e.g., basic events + ledger→events-offsets in separate CFs) | events-side |

Each store has its **own RocksDB directory** + its **own flock** + its **own block cache** + its **own per-store `Config`**. The Layer-1 wrapper opens, manages, and closes them in isolation — there is no cross-store sharing of cache, flock, or config.

Reference commit `19a917907e836366d90f189dd0ac0190b45eb8e7` contains prior-art for the txhash 16-CF schema (look at `pkg/cf` and the RecSplit-related files); use that as a reference for the CF-routing helper that the txhash facade will build on top of this wrapper.

### First commit (mechanical relocation)

- Create the new package tree at `cmd/stellar-rpc/internal/fullhistory/{backfill, streaming, pkg}/` with placeholder `doc.go` files in each leaf package.
- `git mv cmd/stellar-rpc/internal/backfill/meta_store.go → cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/meta_store.go`; rename package `backfill` → `rocksdb`. The old `cmd/stellar-rpc/internal/backfill/` directory disappears.

### The wrapper itself

Grow the relocated seed file into the full Layer-1 wrapper at `cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/`.

#### `Config` struct

Per-store hardcoded by the calling Layer-2 facade. Never read from operator TOML.

```go
type Config struct {
    Path                    string   // on-disk directory; created on Open if it doesn't exist
    ColumnFamilies          []string // empty or ["default"] = default-only; ["cf-0", ..., "cf-f"] = 16 CFs; etc.
    EnableWAL               bool     // default true; cannot be re-derived for stores that require durability
    WriteBufferSizeMB       int
    MaxWriteBufferNumber    int
    TargetFileSizeMB        int
    MaxBytesForLevelBaseMB  int
    MaxTotalWALSizeMB       int
    BloomFilterBitsPerKey   int      // 0 to disable
    BlockCacheSizeMB        int
    // ... extend as facades demand
}
```

#### API — every method gets Go doc explaining the use case

```go
package rocksdb

// Open opens (or creates) a RocksDB store at cfg.Path with the per-store
// configuration in cfg. This is the entry point every Layer-2 facade calls.
//
// Idempotency: a same-process second call against the same canonical absolute
// Path returns the existing handle, not a fresh one. The wrapper maintains a
// process-level registry keyed by canonical Path. This lets multiple facades
// (or test setups) that point at the same directory share one handle without
// fighting for the flock.
//
// Auto-mkdir: if cfg.Path doesn't exist, Open creates it (and all parent
// directories) with permission 0700 by default. Stores that need different
// permissions can pass them through cfg.
//
// Flock: grocksdb's native flock under cfg.Path. A second Open from a
// DIFFERENT process against the same directory fails immediately with a
// clear error. A same-process second call follows the idempotency rule above.
//
// CF configuration: cfg.ColumnFamilies is fully specified at open time and
// fixed for the store's lifetime. The wrapper handles three concrete shapes:
//   - Empty or ["default"]: default CF only. Used by the meta store (#689)
//     and the hot ledger store (#584 reframed).
//   - 16 CFs ["cf-0", ..., "cf-f"]: nibble-routed by txhash[0] >> 4. Used by
//     the hot txhash store (#729).
//   - N named CFs: arbitrary multi-CF schemas. Used by the events store
//     (Urvi's #665, sub-issue TBD), where the schema is still being designed.
//
// On open, the store logs WAL size, L0 file count, SST count, total data
// size, active memtable size — essential for diagnosing slow restarts.
//
// Returns ErrInvalidConfig if cfg.Path is empty or other zero-valued required
// fields are missing.
func Open(cfg Config) (Store, error)

// Store is the per-RocksDB-instance handle that Layer-2 facades hold.
//
// Each Layer-2 facade (meta, hot ledger, hot txhash, hot events) owns its
// own Store with its own RocksDB directory + flock + block cache + Config —
// nothing is shared across facades.
//
// Concurrency: Store methods are safe for concurrent use by multiple
// goroutines. Internal locking is grocksdb's responsibility.
type Store interface {
    // Put writes a single key/value pair to the named CF.
    //
    // For atomic multi-write commits across one or more CFs (e.g., per-ledger
    // ingestion writing N transactions across 16 CFs in the txhash store, or
    // multi-CF events writes), use Batch instead — Put is for single-key,
    // single-CF writes.
    //
    // cf: the CF name. Pass "default" or "" (normalized to "default") for
    // single-CF stores. Pass the configured CF name for multi-CF stores.
    // Returns ErrCFNotFound if cf was not configured at Open.
    Put(cf string, key, value []byte) error

    // Get retrieves the value for key from the named CF.
    //
    // Returns (value, true, nil) if found; (nil, false, nil) if not found
    // (no error — absence is a normal outcome, not a failure). Returns
    // (nil, false, err) for I/O or store-level errors.
    Get(cf string, key []byte) (value []byte, found bool, err error)

    // Delete removes key from the named CF.
    //
    // Idempotent at the wrapper level: no error if the key didn't exist.
    // For atomic multi-key / multi-CF deletes (e.g., retention pruning
    // across all 16 txhash CFs in one go), use Batch.
    Delete(cf string, key []byte) error

    // Iterate returns an iterator over keys in the named CF that share prefix.
    //
    // Keys are returned in sorted byte order within the CF. Caller MUST call
    // Close() on the iterator when done. The iterator is a point-in-time
    // snapshot; concurrent writes are not reflected.
    //
    // Common consumers: ProgressSummary scans (meta store), range queries
    // (hot ledger store), retention pruning iteration (hot stores).
    Iterate(cf string, prefix []byte) Iter

    // Batch accumulates writes (Put / Delete) across one or more CFs in fn,
    // then commits them atomically.
    //
    // This is the primitive for per-ledger live-ingestion commits where one
    // ledger produces writes to MULTIPLE CFs:
    //   - Hot txhash store (#729): up to 16 CFs touched per ledger as
    //     transactions route by nibble (txhash[0] >> 4).
    //   - Hot events store (Urvi's #665): schema-dependent, but multi-CF
    //     writes are likely (basic events + ledger->events-offsets in
    //     separate CFs).
    //
    // All writes commit together or none commit. Use it whenever a single
    // logical unit (e.g., one ledger) produces multiple physical writes that
    // must be atomically observable.
    //
    // Semantics:
    //   - fn returning a non-nil error: NO writes are applied. The store is
    //     unchanged. The error from fn is returned to the caller.
    //   - fn returning nil but commit failing: NO writes are applied. The
    //     commit error is returned.
    //   - fn returning nil and commit succeeding: ALL writes are visible
    //     atomically.
    //   - Empty batch (fn calls no Put / Delete): no-op, returns nil.
    //   - Mixing Put + Delete on the same key within one batch: the final
    //     state reflects the LAST operation queued.
    //
    // Concurrency: multiple goroutines can run Batch concurrently. Each
    // batch's writes commit atomically as a unit; cross-batch ordering is
    // RocksDB's serialization (writes from goroutine A's batch and
    // goroutine B's batch do NOT interleave within either batch).
    //
    // The BatchWriter handed to fn must NOT be retained after fn returns —
    // its lifetime is scoped to the call. The wrapper owns all batch
    // resources; no manual Close is required by the caller.
    Batch(ctx context.Context, fn func(b BatchWriter) error) error

    // Close cleanly shuts down the underlying RocksDB and releases the flock.
    //
    // Idempotent: a second Close is a no-op. After Close, all other Store
    // methods return ErrStoreClosed.
    Close() error
}

// BatchWriter is the accumulator handed to Store.Batch's callback.
//
// All Put / Delete calls accumulate in memory; nothing is written to disk
// until the callback returns nil and the surrounding Batch commits.
//
// The BatchWriter's lifetime is scoped to the callback. Do not retain it
// after the callback returns; do not pass it across goroutine boundaries.
type BatchWriter interface {
    // Put queues a write of key=value into the named CF.
    //
    // No I/O occurs here; the write commits when the surrounding Batch's
    // callback returns nil. Returns ErrCFNotFound if cf was not configured
    // at Open (the surrounding Batch will return this error).
    Put(cf string, key, value []byte)

    // Delete queues a delete of key from the named CF.
    //
    // No I/O occurs here; the delete commits when the surrounding Batch's
    // callback returns nil. Mixing Put and Delete on the same key within
    // one batch is allowed; the final state reflects the LAST operation
    // queued.
    Delete(cf string, key []byte)
}

// Iter is the prefix-scan iterator returned by Store.Iterate.
type Iter interface {
    // Next advances to the next key. Returns false when no more keys match
    // the prefix or on iterator error. Call Err() afterward to disambiguate.
    Next() bool

    // Key returns the current key. Valid only after Next() returns true.
    // The returned slice is owned by the iterator; copy it if you need to
    // retain it past the next Next() call.
    Key() []byte

    // Value returns the current value. Same lifetime caveat as Key.
    Value() []byte

    // Err returns any error encountered during iteration. Should be checked
    // after Next() returns false to distinguish "no more keys" from "error".
    Err() error

    // Close releases iterator resources. MUST be called when done — the
    // underlying RocksDB iterator is not GC-managed.
    Close() error
}
```

#### What this wrapper does NOT do

- Does **not** own key schemas — those live in Layer-2 facades.
- Does **not** call `Flush()` or `SyncWAL()` from production code paths — relies entirely on RocksDB background threads. Tests may use `Flush` for assertions.
- Does **not** manage **cross-store** coordination, transactions, or replication. `Batch` is intra-store atomic only — atomicity across two different `Store` handles is not provided and is not a goal.
- Does **not** read TOML — `Config` is constructed in code by the facade.

## Acceptance criteria

- [ ] **All tests run against a real RocksDB instance via `t.TempDir()`.** The Layer-1 wrapper IS the real implementation — there is NO mock for this layer; faking RocksDB at this level would test nothing useful. The temp dir is created per-test, mutated by the test, and torn down by `t.TempDir()`'s cleanup. Mock implementations for higher-layer testing live in the Layer-2 facade slices (#689, #584, #729), not here.
- [ ] Existing `stellar-rpc` binary still builds; existing tests still pass.
- [ ] `cmd/stellar-rpc/internal/backfill/` no longer exists; `meta_store.go` (and its test) live under `cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/`.
- [ ] The `cmd/stellar-rpc/internal/fullhistory/{backfill, streaming, pkg}/` tree compiles with placeholder `doc.go` files.
- [ ] `Open` accepts the per-store `Config`; rejects zero-valued required fields (e.g., empty `Path`).
- [ ] `Open` is idempotent: same-process second call against the same `Path` returns the existing store handle, not a fresh one.
- [ ] `Open` creates the on-disk directory if it doesn't exist, with sensible default permissions.
- [ ] CF configuration: tested against three concrete shapes — default CF only (matches the meta-store and hot-ledger-store use cases), 16 CFs (matches the hot-txhash-store use case), and "multiple named CFs that aren't simply N-CF nibble-routed" (matches the hot-events-store use case).
- [ ] CF isolation: writes to one CF do not appear in another.
- [ ] `Iterate` returns keys in sorted byte order for prefix scans within a CF.
- [ ] flock: two `Open` calls from different processes against the same directory — second fails with a clear error.
- [ ] On-open log line includes WAL size, L0 count, total data size, memtable size.
- [ ] No manual `Flush()` calls in production code paths (background threads only). Tests may call `Flush` for assertions.
- [ ] **Batch — single-CF**: 100 puts queued in a batch against one CF commit atomically; all 100 visible after.
- [ ] **Batch — multi-CF**: 100 puts spread across 16 CFs (~6–7 per CF) commit atomically; every CF reflects exactly its writes; no cross-CF leakage.
- [ ] **Batch — mid-callback error**: callback returns a non-nil error after queuing N puts; assert ZERO of those puts are visible (true rollback, not partial commit). The error from the callback is returned to the caller.
- [ ] **Batch — empty**: callback queues nothing and returns nil → no-op, returns nil.
- [ ] **Batch — Put + Delete same key**: queue Put then Delete on the same key within one batch → final state shows the deletion.
- [ ] **Batch — concurrent**: goroutine A queues writes against `cf-0` while goroutine B queues against `cf-f`, both commit successfully, no interference (verified by reading both CFs after).
- [ ] **Batch lifetime**: `BatchWriter` handed to the callback is not retained after the callback returns (the wrapper owns all batch resources; verify no leak under stress).
- [ ] **Every public API method has Go doc** explaining its use case + the consumer pattern it serves (per the API block above).

## Logging convention

This slice does NOT build any `pkg/logging` package, scope helper, or logrus bootstrap. Components take a logger reference via constructor arg (e.g., `*logrus.Logger` or a narrow interface); the caller hands one in. Tests use a discard or capture logger. The on-open state-log just calls `logger.WithField("scope", "...").Info(...)` directly — whatever scoping shape the slice author picks, it's local to this slice; no shared helper.

Service-wide logger plumbing — global, concurrent-safe, structured writes to the single log file specified in `[LOGGING]` — is owned by the streaming epic (#704) and lands when the unified service entry point is built.

## Helper port — `pkg/format` + `pkg/memory` (demand-driven; first consumer = this slice)

This slice IS the first consumer of two cross-cutting helpers, so per the project's demand-driven port pattern, they get ported here:

- **`pkg/format`** — the canonical home for ALL byte-encoding helpers (`Uint32ToBytes` / `BytesToUint32` / `Uint64ToBytes` / `BytesToUint64` over `binary.BigEndian`) AND user-facing numeric formatters (`FormatBytes` / `FormatDuration` / `FormatRate` / `FormatNumber` / `FormatPercent`). **Project rule, zero exceptions in `fullhistory/`**: NO inline `fmt.Sprintf`, NO direct `binary.BigEndian` / `binary.LittleEndian` — every numeric / byte-encoding call routes through `pkg/format`. Port from `19a9179` with `*_test.go`.
- **`pkg/memory`** — memory-accounting helpers used for state logging + progress reporting. Port from `19a9179` with `*_test.go`.

Once ported here, every downstream slice (`#686` paths, `#689` metastore facade, `#690` writers, `#691` tasks, `#584` hot ledger store, `#729` hot txhash store, etc.) imports from this location. No other slice ports `pkg/format` or `pkg/memory` again.

> Reference commit `19a917907e836366d90f189dd0ac0190b45eb8e7` contains exploratory prior art under `full-history/backfill/`. Use `git show 19a9179:full-history/backfill/<file>.go` to read; do not check them out. Make explicit port / adapt / rewrite / discard calls per file. The 16-CF naming convention shown in `pkg/cf/cf.go` is informational reference for this slice's CF-isolation tests; `pkg/cf` itself is NOT ported here — that lands in #696 (real RecSplit) where it has its real consumer.
> Current backfill design doc on the active design-doc branch (`full-history/design-docs/01-backfill-workflow.md`) governs the meta-store's flag-after-fsync invariant; the Layer-1 wrapper is intentionally schema-agnostic, but its semantics (WAL-on, flush cadence, on-open logging) need to support that invariant.
> Aggressively port `*_test.go` companions when behavior still matches.
> One test → one implementation → repeat. No bulk file porting without a test driving each behavior.

karthikiyer56:
## Summary

- New `cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/` Layer-1 wrapper. The CI-only `meta_store.go` seed under `cmd/stellar-rpc/internal/backfill/` is replaced by it.
- Two-phase Store lifecycle: `New(cfg) (*Store, error)` validates Config, then `(*Store).Open() error` does the actual grocksdb open. Open is idempotent via `sync.Once` on the Store; Close is idempotent via `atomic.Bool`. Auto-mkdir on Open (mode 0700). `Flush()` is available for clean shutdown; `WriteOptions.DisableWAL(false)` is pinned in `doOpen` so the always-on WAL setting is visible at the call site.
- CF-aware: default-only (meta store, hot ledger store), 16 CFs nibble-routed (hot txhash store), arbitrary multi-named (events store). Same `Put`/`Get`/`Delete`/`Iterate`/`Batch` path internally for every shape.
- `Store.Get` uses grocksdb's zero-copy `GetPinnedCFV2`: the returned handle points directly into the pinned block-cache page, dropping the separate C-side buffer that the older `GetCF` path allocated. One memcpy on the read path (cache → Go slice) instead of two.
- `Store.Batch(fn)` is intra-store atomic via `grocksdb.WriteBatch`. Mid-callback error rolls back zero-visibility; empty batch is a no-op; the `*BatchWriter` goes inert after the callback returns; same-key Put + Delete in one batch — deletion wins. Signature carries no `context.Context` because grocksdb's `db.Write` is a blocking CGO call with no cancellation hook; we can add ctx back in a useful form (passed through to fn for callback-internal cancellation) when a real consumer needs it.
- Lifecycle `sync.RWMutex` on Store: every operation (Put/Get/Delete/Iterate/Batch/Flush) holds `RLock` for the duration of its underlying C call; Close takes the exclusive `Lock` after flipping the closed flag, waiting for in-flight RLocks to release before tearing down the C++ DB. Prevents the segfault that would otherwise occur when Close runs concurrently with an in-flight op. Reads don't block writes and writes don't block reads — only Close serializes. Cost is one atomic CAS per op (single CAS spans an entire Batch); reasoning and trade-off explicit in the `mu` field doc on Store.
- `Iterate` uses Go 1.23+ range-over-func: returns `iter.Seq2[Entry, error]`. Callers `for e, err := range store.Iterate(cf, prefix) { ... }`; the producer closure owns `defer it.Close()` and yields per-iteration errors, so callers can't forget cleanup and there is no separate `Err()` method. Inside the closure, `KeySlice()` / `ValueSlice()` return `OptimizedSlice` (value type) instead of `*Slice` (heap-allocated) — eliminates ~30k transient allocations per 10k-key scan.
- On-open log line includes elapsed open time (`time.Duration.Round(time.Microsecond).String()`), WAL size, L0 file count, total data size, active memtable size — surfaces slow restarts to the operator. Logger is `*supportlog.Entry` directly, matching the existing `daemon.go` / `event_test.go` pattern.
- Endianness for RocksDB-stored integers lives in `pkg/rocksdb/encoding.go` — unexported `byteOrder` (BigEndian) is the single source of truth, with `EncodeUint32` / `DecodeUint32` / `EncodeUint64` / `DecodeUint64` as the Layer-2 API. The rationale (BE required so byte-lex iterator order matches numeric range-scan order — LE would silently return scrambled ranges) lives inline in the `byteOrder` doc comment.
- ~34 tests, all real-RocksDB via `t.TempDir()`. Covers Open lifecycle (idempotency, two-stores-same-path collide, auto-mkdir, persistence across reopen, cross-process flock), Put/Get/Delete across default + 16-CF + multi-CF shapes, sorted Iterate, Batch atomicity, error paths (ops before Open, ops after Close, Iterate corners, Batch error paths), and the lifecycle lock (concurrent ops + Close validated under `-race`; Close blocks until in-flight iterate releases its RLock).

Closes #685


cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,452 @@
...
+// flock. Idempotent: a second Close is a no-op. Calling Close on a
+// Store that was never successfully Opened is also a no-op.
+func (s *Store) Close() error {
+	if !s.closed.CompareAndSwap(false, true) {
+		return nil
+	}
+	if s.db == nil {
+		// New'd but Open never succeeded — nothing to clean up.
+		return nil
+	}
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 16c09bd. Close now calls `s.openOnce.Do(func() {})` after the closed CAS — that either waits for an in-flight Open to settle (so we read a stable s.db) or claims openOnce so a later Open returns ErrStoreClosed without touching disk. Open also checks `s.closed.Load()` after openOnce.Do and returns ErrStoreClosed if Close raced ahead. Real openErr (e.g., grocksdb disk error) takes precedence. Added TestStore_ConcurrentOpenAndClose; CI runs with -race.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,452 @@
...
+		total += info.Size()
+	}
+	return total
+}
+
+// Iter is the prefix-scan iterator returned by Store.Iterate.
+//
+// Lifetime: Close MUST be called when iteration is done. Key / Value
+// byte slices are valid only between Next-returning-true and the next
+// Next call (RocksDB owns the buffer); copy them to retain past that.
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 16c09bd. Switched prefixIter.Key() / Value() to zero-copy — they now return `i.it.Key().Data()` directly, matching what the doc already said. Faster for Layer-2 facades that scan many keys; callers that need to retain a slice past Next() / Close() copy via `string(...)` or `append([]byte{}, ...)` as the doc requires.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb_test.go
```diff
@@ -0,0 +1,363 @@
...
+// Big-endian encoding sorts lexicographically the same way it sorts
+// numerically — what makes range queries O(window-size).
+func TestStore_Iterate_SortedPrefixScan(t *testing.T) {
+	s := openTestStore(t, nil)
+
+	inserts := map[string]string{
+		"chunk:00000000:lfs":    "1",
+		"chunk:00000001:lfs":    "1",
+		"chunk:00000002:lfs":    "1",
+		"chunk:00000005:txhash": "1", // not under our scan prefix
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 16c09bd. The comment now correctly says `chunk:00000005:txhash` matches the prefix despite its different suffix; only `index:00000000:txhash` is the excluded case.

cmd/stellar-rpc/internal/fullhistory/pkg/format/format.go
```diff
@@ -0,0 +1,165 @@
...
+	default:
+		return fmt.Sprintf("%dy %dmo %dd", years, months, days)
+	}
+}
+
+// Number formats an int64 with thousands separators ("1,234,567").
+// Negative numbers carry a leading "-".
+func Number(n int64) string {
+	if n < 0 {
+		return "-" + Number(-n)
```
leighmcculloch: Two's complement asymmetry means int64 ranges from -9223372036854775808 to 9223372036854775807 with the negative side reaching one further than the positive side.

So that when `n` is `math.MinInt64`, `-n` overflows back to `math.MinInt64` and the function calls itself until stack overflow.

For example: https://go.dev/play/p/k3QihF9qlMo

It may be worth using https://github.com/dustin/go-humanize rather than rolling this ourselves.

leighmcculloch (commented):
I was taking a peak at what this PR was doing out of curiosity, and noticed a stack overflow. I haven't gone deep because I'm lacking a lot of context, but wanted to mention the overflow briefly.

cmd/stellar-rpc/internal/fullhistory/pkg/format/format.go
karthikiyer56: I have started using humanize now. 
I still need a helper function to print the duration in human readable format across time ranges - us, ms, s, m, h. There isn't one available in the humanize pkg, so I wrote my own

cmd/stellar-rpc/internal/fullhistory/pkg/format/format.go
tamirms, line 1: What would you think about dropping `pkg/format` entirely and using `elapsed.Round(time.Millisecond).String()`? That gives `"123ms"` / `"4.5s"` / `"2m3.5s"` / `"72h0m0s"`.

The tradeoff is that the largest unit caps at hours: a hypothetical 50-hour open would log as `"50h0m0s"` rather than `"2d 2h"`. I think that tradeoff is fine for eliminating the code.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/experiment_endianness_test.go
tamirms, line 1: This test effectively never runs (`//go:build experiment` means it's not invoked in CI), and there aren't any assertions on the iterator output — `polluters` and `returned` are collected and printed but never compared against expected values. Both tests would pass if grocksdb returned a different answer tomorrow.

I think we should either delete it or replace it with a real test that drops the build tag and asserts `require.Equal(t, []uint32{100, 101, 150, 199, 200}, returned)` — turning it into a regression test that an iterator range scan visits BE-encoded keys in numeric order.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,636 @@
...
+// The returned value is a fresh copy owned by the caller.
+func (s *Store) Get(cf string, key []byte) ([]byte, bool, error) {
+	if err := s.checkOpen(); err != nil {
+		return nil, false, err
+	}
+	cfh, err := s.resolveCF(cf)
+	if err != nil {
+		return nil, false, err
+	}
+	slice, err := s.db.GetCF(s.ro, cfh, key)
```
tamirms: `Store.Get` does two value copies today: `GetCF` allocates a fresh C buffer and memcpys the value into it (copy 1, inside `rocksdb_get_cf`), then we `make([]byte, ...) + copy(out, slice.Data())` to produce a Go-owned slice (copy 2). Switching to `GetPinnedCFV2` drops copy 1 — the returned handle is a zero-copy reference into the block cache, so only the cache→Go-buffer copy required by the API contract remains.

```go
ps, err := s.db.GetPinnedCFV2(s.ro, cfh, key)
if err != nil { return nil, false, err }
defer ps.Destroy()
if !ps.Exists() { return nil, false, nil }
data := ps.Data()
out := append([]byte(nil), data...)
return out, true, nil
```

`GetPinnedCFV2` is the variant grocksdb's own docs flag as the recommended migration target.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,636 @@
...
+	}
+	return s.db.DeleteCF(s.wo, cfh, key)
+}
+
+// Iterate returns an iterator over keys in the named CF that share
+// prefix.
+// An empty prefix iterates every key in the CF.
+// Keys come back in sorted byte order.
+// Caller MUST Close the iterator.
+func (s *Store) Iterate(cf string, prefix []byte) Iter {
```
tamirms: What about `iter.Seq2[Entry, error]` instead of the explicit `Iter` interface?

```go
type Entry struct{ Key, Value []byte }

func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error]
```

Caller:

```go
for e, err := range store.Iterate(cf, prefix) {
    if err != nil { return err }
    // use e.Key, e.Value
}
```

Two real wins over the current `Iter`:

- Caller can't forget `Close()` — `defer it.Close()` lives in the producer closure and fires on any exit (normal completion, `break`, panic).
- Iterator error rides along per-iteration; error handling is one line in the loop body. No separate `Err()` getter, no `errIter` shim.

Lets you delete `Iter`, `errIter`, and the public `prefixIter` surface. The lifetime rule on `e.Key` / `e.Value` (valid only until next iteration) carries over unchanged.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,636 @@
...
+// succeeded).
+// This avoids the race where Close fires while Open is still
+// constructing the grocksdb DB, returns early on s.db == nil, and
+// leaves the just-opened DB (and its flock) leaked.
+//
+// Close does NOT auto-Flush.
+// Data is durable either way (WAL replay handles it on next Open),
+// but callers wanting a clean shutdown with no WAL replay should call
+// Flush first.
+func (s *Store) Close() error {
```
tamirms, line 477: The Close doc covers concurrent `Open` carefully but doesn't say anything about concurrent `Get` / `Put` / `Delete` / `Iterate` / `Batch`. The current implementation only checks `closed` at method entry — once a `Get` / `Put` / `Iterate` / etc. passes that check, nothing prevents `Close` from racing ahead and tearing down the DB underneath the in-flight C call.

Either document the contract explicitly ("callers must drain all operations before `Close`") or add a `sync.RWMutex`: every operation `RLock`s for its duration; `Close` `Lock`s.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/batch.go
```diff
@@ -0,0 +1,145 @@
...
+//     writes are visible to subsequent reads at the same instant.
+//   - fn made no Put/Delete calls → no-op, returns nil.
+//   - fn called Put then Delete (or vice versa) on the same key →
+//     the final state reflects whichever was queued LAST.
+//
+// Atomicity is INTRA-store only — every operation in fn lands in this
+// one Store, all-or-nothing.
+// Coordinating atomic writes across two different Store handles is
+// not a goal of this wrapper.
+func (s *Store) Batch(_ context.Context, fn func(*BatchWriter) error) error {
```
tamirms: The `_ context.Context` parameter is accepted but never used.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -0,0 +1,636 @@
...
+	defer k.Free()
+	return hasPrefix(k.Data(), i.prefix)
+}
+
+// Key returns the current key as a zero-copy view into the iterator's
+// internal buffer.
+// The slice is valid only until the next Next() / Close() call —
+// callers that need to retain it must copy (e.g., via string(...) or
+// append([]byte{}, k...)).
+func (i *prefixIter) Key() []byte {
```
tamirms: `i.it.Key()` returns a `*grocksdb.Slice` (heap-allocated struct) on every call — same for `i.it.Value()`, plus another in `Next()`. For a 10k-key scan that's ~700KB of GC pressure across ~30k transient `*Slice` allocations.

grocksdb v1.10 exposes `iter.KeySlice()` / `iter.ValueSlice()` returning `OptimizedSlice` (value type, no heap alloc). Switching `Key()` / `Value()` and the prefix-check in `Next()` to use those would eliminate the per-iteration allocations.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/experiment_endianness_test.go
karthikiyer56: the inclusion of this test and the deliberate tagging as experimental was to only show why the byte order when inserting **numeric keys** in rocksdb needs to be BE-oriented.
this test serves no other purpose.
I will delete it now

cmd/stellar-rpc/internal/fullhistory/pkg/format/format.go
karthikiyer56: done. fixed in `a4c4a21`. dropped `pkg/format` entirely; the open-log uses `elapsed.Round(time.Microsecond).String()` directly. microsecond rounding so a sub-ms open does not render as `"0s"`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/experiment_endianness_test.go
karthikiyer56: done. fixed in `a4c4a21`. test deleted; the rationale (concrete LE pollution example) moved inline into the `byteOrder` doc comment in `encoding.go`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done. fixed in `97862dc`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done. fixed in `a4c4a21`. `Iterate` returns `iter.Seq2[Entry, error]`; `Iter`/`errIter`/`prefixIter` deleted. range-over-func is now the iteration shape across fullhistory

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done. fixed in `a4c4a21`. lifecycle `sync.RWMutex` on Store: every op holds RLock for its C call; Close takes WLock after flipping the closed flag. reads do not block writes; only Close serializes. design tradeoff over doc-only drain contract spelled out in the `mu` field comment. two tests added: `TestStore_ConcurrentOpsAndCloseRaceFree` (chaos under `-race`) and `TestStore_CloseWaitsForInflightIterate` (deterministic).

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/batch.go
karthikiyer56: done. fixed in `0bf012e`. removed `_ context.Context` from `Store.Batch`. grocksdb `db.Write` is a blocking CGO call with no cancellation hook, so threading ctx through bought nothing. add it back when a real consumer needs callback-internal cancellation.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done. fixed in `a4c4a21`. inside the new `iter.Seq2` producer closure, `KeySlice()`/`ValueSlice()` (`OptimizedSlice` value type) replaced `Key()`/`Value()` (`*Slice` heap-allocated). zero heap allocs per iteration.

tamirms (approved):

PR #734, opened 2026-05-07
daemon: tune HTTP transport for captive-core clients

tamirms:
### What

Tune the `http.Client` Transport that stellar-rpc uses to call its embedded captive-core. `createStellarCoreClient` (admin HTTP port) and `createHighperfStellarCoreClient` (high-performance query port) both currently leave `Transport` nil, falling back to Go's `http.DefaultTransport` — which has `MaxIdleConnsPerHost = 2` and no explicit `IdleConnTimeout`.

Both clients now use a shared `captiveCoreHTTPTransport()` returning an `http.Transport` with:
- `MaxIdleConns: 500`
- `MaxIdleConnsPerHost: 500`
- `MaxConnsPerHost: 500`
- `IdleConnTimeout: 60s`

### Why

Under sustained `sendTransaction` + `getTransaction` load (both methods proxy to captive-core via these clients), the default 2-connection idle pool starves. New goroutines beyond that count must open fresh TCP connections to localhost for every call, serialising on TCP setup. The HTTP server queues the work, handler-side JSON-RPC durations stay under 100ms, but the outer HTTP request duration balloons to 8+ seconds and the server starts logging:

```
"Request processing for / exceed warning threshold of 5s"
"finished request" duration=8.4s status=200
```

After ~2-3 minutes of sustained load, vegeta-style clients start timing out their 30s requests, which in load-test scenarios cascades into account-lease starvation and the test reports `lease ... context deadline exceeded` on the targeter side — even though stellar-rpc itself returns 200s eventually.

#### Empirical impact

A/B against `stellar-rpc:latest` (with this patch vs without), 5-minute bench at 60 RPS Soroban + 100 RPS classic, sac-transfer mode, all other variables identical:

| Variant | sac success | simple-payment success | sac httpErr | simple-payment httpErr |
|---|---:|---:|---:|---:|
| **with this patch** | 98.9% | 97.1% | 0 | 0 |
| without this patch | 86.9% | 87.0% | 2,349 | 3,898 |

Net: **−12.0pp / −10.1pp inclusion success rate**, attributable solely to the connection-pool tune. The 6,247 `httpErr` events in the unpatched run are EOFs / "server closed idle connection" against this very pool.

The cliff onset is reliably at ~3 minutes into the run, which is roughly the time it takes for the working set of poll workers + sendTransaction handlers to exhaust the keepalive pool churn budget.

### When this matters

This change has **no effect** on:
- Unit and integration tests (small workloads, no keepalive churn).
- Light development workflows (e.g. `stellar contract invoke` or local CI).
- Anyone running stellar-rpc behind a reverse proxy that already absorbs connection churn (some quickstart setups do; bare deployments do not).

It **does** matter for:
- Sustained load testing (this is what surfaced the bug).
- Production-style deployments serving high request rates without an upstream proxy doing connection pooling.

The fix is a defaults-only change in two struct literals; no API surface, no config flag, no behavior change at low load.

### Known limitations

- 500 connections is empirically generous; a more disciplined value could be derived from `runtime.GOMAXPROCS()` × some factor, but 500 is well below any realistic process FD limit and well above the contended state we observed.
- `IdleConnTimeout: 60s` is set just under nginx's typical 75s `keepalive_timeout` in case anyone fronts captive-core's HTTP port with a proxy. Default Go behavior (no explicit timeout) inherits OS-level keepalive only; this gives us deterministic recycling.
- Both client functions could plausibly share a *single* Transport rather than calling `captiveCoreHTTPTransport()` twice; kept separate to make the construction path symmetric and allow per-client tuning if needed later.

### DIFF
--- a/cmd/stellar-rpc/internal/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/daemon/daemon.go
@@ -308,20 +308,18 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
 // against the same localhost endpoint and stalling requests in the HTTP queue.
 //
 // Cloned from http.DefaultTransport so we keep its other defaults
-// (Proxy=ProxyFromEnvironment, DialContext with timeouts, TLSHandshakeTimeout,
-// etc.) and only override the connection-pool fields. Fall back to a fresh
-// Transport in the unlikely event that http.DefaultTransport has been
-// replaced with a different concrete type.
+// (Proxy=ProxyFromEnvironment, DialContext with sane timeouts/keepalive,
+// TLSHandshakeTimeout, ForceAttemptHTTP2, ExpectContinueTimeout) and only
+// override the idle-connection-pool fields. http.DefaultTransport is
+// documented to be a *http.Transport, so the type assertion is safe; we
+// suppress the forcetypeassert linter accordingly.
 func captiveCoreHTTPTransport() *http.Transport {
-	var t *http.Transport
-	if base, ok := http.DefaultTransport.(*http.Transport); ok {
-		t = base.Clone()
-	} else {
-		t = &http.Transport{}
-	}
+	t := http.DefaultTransport.(*http.Transport).Clone() //nolint:forcetypeassert // http.DefaultTransport is documented as *http.Transport
+	// Increase only the idle-pool limits; deliberately leave MaxConnsPerHost
+	// at the default (0 = unlimited) so we don't introduce a hard cap on
+	// total in-flight connections.
 	t.MaxIdleConns = 500
 	t.MaxIdleConnsPerHost = 500
-	t.MaxConnsPerHost = 500
 	t.IdleConnTimeout = 60 * time.Second
 	return t
 }
--- a/cmd/stellar-rpc/internal/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/daemon/daemon.go
@@ -309,9 +309,16 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
 //
 // Cloned from http.DefaultTransport so we keep its other defaults
 // (Proxy=ProxyFromEnvironment, DialContext with timeouts, TLSHandshakeTimeout,
-// etc.) and only override the connection-pool fields.
+// etc.) and only override the connection-pool fields. Fall back to a fresh
+// Transport in the unlikely event that http.DefaultTransport has been
+// replaced with a different concrete type.
 func captiveCoreHTTPTransport() *http.Transport {
-	t := http.DefaultTransport.(*http.Transport).Clone()
+	var t *http.Transport
+	if base, ok := http.DefaultTransport.(*http.Transport); ok {
+		t = base.Clone()
+	} else {
+		t = &http.Transport{}
+	}
 	t.MaxIdleConns = 500
 	t.MaxIdleConnsPerHost = 500
 	t.MaxConnsPerHost = 500
--- a/cmd/stellar-rpc/internal/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/daemon/daemon.go
@@ -309,18 +309,24 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
 //
 // Cloned from http.DefaultTransport so we keep its other defaults
 // (Proxy=ProxyFromEnvironment, DialContext with sane timeouts/keepalive,
-// TLSHandshakeTimeout, ForceAttemptHTTP2, ExpectContinueTimeout) and only
-// override the idle-connection-pool fields. http.DefaultTransport is
-// documented to be a *http.Transport, so the type assertion is safe; we
-// suppress the forcetypeassert linter accordingly.
+// IdleConnTimeout=90s, TLSHandshakeTimeout, ForceAttemptHTTP2,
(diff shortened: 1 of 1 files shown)


cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -301,17 +301,48 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
...
+// (Proxy=ProxyFromEnvironment, DialContext with timeouts, TLSHandshakeTimeout,
+// etc.) and only override the connection-pool fields. Fall back to a fresh
+// Transport in the unlikely event that http.DefaultTransport has been
+// replaced with a different concrete type.
+func captiveCoreHTTPTransport() *http.Transport {
+	var t *http.Transport
+	if base, ok := http.DefaultTransport.(*http.Transport); ok {
+		t = base.Clone()
+	} else {
+		t = &http.Transport{}
```
leighmcculloch: Go's default transport has other defaults. Do they need to be set here?

For example, in the comments above it sounds like the intent is for the Proxy to be `ProxyFromEnvironment`, but if `Proxy` isn't set in the transport, which it isn't here, no proxy will be used.

`http.Transport`'s default behaviour is not the same as `http.DefaultTransport`'s configuration, so if the goal is to restore that default configuration we need to set all the same values here.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -301,17 +301,48 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
...
+	}
+	t.MaxIdleConns = 500
+	t.MaxIdleConnsPerHost = 500
+	t.MaxConnsPerHost = 500
```
leighmcculloch: Setting the `MaxConnsPerHost` sets a hard limit on new conns that means the transport will start blocking when it hits that limit.

The default value for `MaxConnsPerHost` is `0` which means unlimited. Ref: https://go.dev/src/net/http/transport.go

Given the PR description sounds like the intent is increase the number of allowed idle conns, it seems unnecessary to limit the max conns per host too.

Is it changing for some related reason?

leighmcculloch (commented):
Couple concerns concerns inline.

leighmcculloch (approved):

tamirms:
Closing this PR. After running follow-up diagnostics, I cannot substantiate the empirical claim in the original commit message, and the proposed mechanism doesn't survive direct measurement.

**Mechanism check:** captive-core's HTTP server (`stellar-core/lib/http/connection.cpp`, `do_write`) unconditionally calls `socket_.shutdown(both)` followed by `socket_.close()` after every response. There's no HTTP/1.1 keep-alive on this link. As a result, connections from stellar-rpc to captive-core never enter Go's idle connection pool — every request opens a new TCP connection and closes it after the response. Setting `MaxIdleConnsPerHost` to 2, 100, or 500 has no observable effect: the pool stays empty either way.

**Direct TCP measurement** (sampling `netstat` every 0.5s during a sustained 60+100 RPS bench, then a stress 300+200 RPS bench):
- Peak `ESTABLISHED` count on the regular HTTP port to captive-core: **5 at moderate load, 17 at stress**.
- `TIME_WAIT` count grows 1:1 with submitted-request count, plateauing at ~5500 (= ~90 closures/sec × 60s `TIME_WAIT` timeout).
- Pool of 500 idle connections never has anywhere near 500 to hold; `ESTABLISHED` never approaches the proposed `MaxConnsPerHost=500` cap.

**A/B vs unpatched at original conditions** (60+100 RPS, 5 min, sac-transfer): patched 17351/17407 = 99.7%, unpatched 17240/17397 = 99.1%. Within run-to-run noise. (OZ and soroswap collapsed due to a separate, self-induced surge-pricing issue at the bench's offered load and aren't a clean comparison; sac-transfer is well within ledger capacity and is the valid signal.)

The original 12pp / "cliff at 3 minutes" observation appears to have been environmental (network state variance, port-pool / `TIME_WAIT` inheritance between sequential A/B/C/D runs, run-order effects). I can't reproduce it under controlled conditions.

The pool-size knobs in this PR would only become relevant if captive-core's HTTP server is changed to honor HTTP/1.1 keep-alive (separate change against `stellar-core`). Until then there's no mechanism for them to apply, and on this RPC ↔ captive-core link as deployed they're a no-op.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -301,17 +301,56 @@ func mustOpenDatabase(cfg *config.Config, logger *supportlog.Entry, metricsRegis
...
+// is a hard cap on total in-flight conns, not a pool-sizing knob, and
+// adding one would limit throughput under concurrent load.
+func captiveCoreHTTPTransport() *http.Transport {
+	// http.DefaultTransport is documented as *http.Transport, so the
+	// assertion is safe in practice; nolint:forcetypeassert acknowledges
+	// the linter rule.
+	//nolint:forcetypeassert
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.MaxIdleConns = 500
+	t.MaxIdleConnsPerHost = 500
```
leighmcculloch, line 333: I think these settings may interact with file-descriptors, and given the default on many systems is 1024, that could be a lot of idle connections consuming file descriptors unnecessarily.

leighmcculloch (commented):
One more thought inline.

PR #737, opened 2026-05-13
Update packfile design doc to match merged writer and reader implementations

tamirms:
## Summary

The packfile design doc has been out of sync with the implementation. Most of the divergence came in with the writer (#702) and was never reflected in the doc; the reader (#712) added a few more deltas on top. This PR brings the doc end-to-end into alignment with the merged code.

**No code changes** — doc-only.

### From #702 (writer) — doc was never updated

- **Format model.** `WriterOptions.Format` is now a caller-assigned `Format uint32` identifier, not the old `RecordFormat` enum (`Compressed` / `Uncompressed` / `Raw`). The built-in formats are gone — codec is plug-in via `NewRecordEncoder`.
- **New WriterOptions fields.** `NewRecordEncoder func() RecordEncoder` (per-worker factory), `ContentHashExtract` for per-item transformation before hashing.
- **Trailer is 76 bytes (was 64).** New `format uint32` field at offset 8; only `flagContentHash` remains (`flagNoCompression` / `flagNoCRC` are gone). Magic constant corrected to `0x48434C53`. Worked-example file-size math updated.
- **Records.** Caller-encoded payload + library-managed item-size-index CRC. Per-record payload integrity is encoder-supplied (zstd's xxHash64, or caller's choice, or none with passthrough).
- **Writer single-goroutine constraint** documented (parallel to the Reader's concurrency disclosure).
- **AppendItem semantics**: no-args is a no-op, `[]byte{}` records an empty item, parts are copied, oversized items error.
- **Sync nit**: doc said `fdatasync`; the code calls `os.File.Sync()` which is `fsync`.
- **Concurrency sizing advice**: values above `runtime.NumCPU()` tend to hurt throughput.
- **ErrWriterClosed** added to the Errors section.

### From #712 (reader)

- **ReaderOptions is a struct** (was variadic `WithConcurrency`). Default `Concurrency` is 1 (serial), not 8.
- **RecordDecoder is a single concurrent-safe instance**, caller-owned, not closed by `Reader.Close`.
- **Error semantics.** `ReadItems` / `ReadRange` return errors instead of panicking. New `ErrPositionsUnsorted` sentinel.
- **ReadRange lifetimes.** Yielded slice is invalidated after the for-range loop exits (break or natural completion); iterator itself is not safe for concurrent iteration; `count == 0` is valid.
- **ReadItems cancellation / atomicity.** Context is checked at batch boundaries, not between items inside a batch; `ReadItems` is not atomic on error (fn may have run for some positions and not others). Empty positions slice is valid. fn's error is returned verbatim.
- **Close blocks** on the background open if it is still in flight.
- **Implementation notes.** `ReadItems` uses `errgroup.WithContext` (no more atomic-flag cancellation); single concurrent-safe decoder + package-level workspace pool replaces the old per-Reader decoder pool.

### New section

- **Codec Contract.** Documents `Format uint32`, the `RecordEncoder` / `RecordDecoder` interfaces, and the in-tree `zstd` codec (`*zstd.Compressor` / `*zstd.Decompressor`).

### Polish

- Disambiguated "format version" (the on-disk `version` byte vs the caller-assigned `Format` field) at the index-encoding section.
- Used KiB/MiB binary prefixes where referring to specific code constants (1 MiB pooled buffer, 256 KiB speculative read).
- Parallel-pipeline gate description corrected (triggered by encoder/hash, not by `Concurrency`).

## Verification

Doc was fact-checked end-to-end against the current implementation in `cmd/stellar-rpc/internal/packfile/` via five review passes: three parallel agents on API surface, on-disk format, and behavior/implementation; followed by deep audits dedicated to the writer side and the reader side. All trailer offsets, struct fields, method signatures, error names, default values, lifetime contracts, and implementation claims verified against code.

## Test plan

- [x] No code changes; no tests to run
- [x] Doc renders cleanly in GitHub preview
- [x] All anchors / cross-references resolve

### DIFF
--- a/design-docs/packfile-library.md
+++ b/design-docs/packfile-library.md
@@ -21,7 +21,9 @@ Non-goals:
 
 On disk, items are grouped into **records**. `ItemsPerRecord` controls how many items go into each record.
 
-Records can be stored in one of three formats: **Compressed** (zstd with built-in integrity, the default), **Uncompressed** (raw bytes with CRC32C integrity check), or **Raw** (raw bytes, no integrity check).
+Record bytes are transformed on the way to and from disk by a **caller-supplied codec** — a pair of `RecordEncoder` / `RecordDecoder` implementations the caller plugs into `WriterOptions` and `ReaderOptions`. With a nil codec, records are passthrough (written and read verbatim). The package ships no built-in codec; the `zstd` subpackage provides `*zstd.Compressor` and `*zstd.Decompressor` that satisfy the interfaces directly, and callers are free to provide their own (e.g. raw bytes with a trailing CRC32C).
+
+The trailer carries a caller-assigned `Format uint32` field. Readers dispatch on `Format` to pick the matching decoder (and any other decode-side choice such as the content-hash extract — see [Codec Contract](#codec-contract)). The library does not interpret `Format` values.
 
 The file ends with a compact **offset index** that maps each record to its byte position on disk. When a file is opened, the offset index is loaded into memory. After that, looking up any item is a single read to fetch the record containing it — no further index I/O. This is how packfile achieves the "one I/O on open, one I/O per lookup" goal.
 
@@ -44,9 +46,13 @@ Error handling omitted for clarity. All functions return errors.
 
 ```go
 w, _ := packfile.Create("output.pack", packfile.WriterOptions{
-    ItemsPerRecord: 128,  // items per record (default 128)
-    Concurrency:    4,    // parallel compression goroutines
-    ContentHash:    true, // compute SHA-256 content hash
+    ItemsPerRecord:   128,                              // items per record (default 128)
+    Format:           packfile.Format(1),               // caller-assigned codec identifier
+    NewRecordEncoder: func() packfile.RecordEncoder {   // per-worker encoder
+        return zstd.NewCompressor()
+    },
+    Concurrency:      4,                                // parallel encode/hash goroutines
+    ContentHash:      true,                             // compute SHA-256 content hash
 })
 defer w.Close() // removes file if Finish was not called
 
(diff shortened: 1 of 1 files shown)


design-docs/packfile-library.md
```diff
@@ -186,50 +259,85 @@ func (w *Writer) Close() error
...
+// Open returns a Reader immediately. File I/O and option validation run in
+// a background goroutine; both kinds of failure are deferred to the first
+// method that needs the result. Open itself has no error return.
 // Close must always be called.
-func Open(path string, opts ...ReaderOption) *Reader
-
-// WithConcurrency sets the max parallel goroutines for ReadItems.
-// Values less than 1 are clamped to 1. Default 8.
-func WithConcurrency(n int) ReaderOption
+func Open(path string, opts ReaderOptions) *Reader
```
karthikiyer56, line 293: Option validation isn't actually in the background. [`cmd/stellar-rpc/internal/packfile/reader.go#L179-L183`](https://github.com/stellar/stellar-rpc/blob/225231e915acc67a77ca9aae8dff8028dedc5d89/cmd/stellar-rpc/internal/packfile/reader.go#L179-L183) checks `Concurrency < 0` synchronously before `Open` returns; the code comment there even says *"Reject invalid options synchronously"*. The background goroutine only runs `doOpen(path)`, which takes `path` as its only argument and never touches `opts`.

The other two option fields (`RecordDecoder`, `ContentHashExtract`) aren't validated at all — they're just stored. So nothing of `opts` ever flows through the goroutine.

The pre-PR wording was already accurate.

Suggested rewording — split the two phases explicitly:

```go
// Open returns a Reader immediately. Option validation runs synchronously;
// file I/O runs in a background goroutine. Errors from either are deferred
// to the first method that needs the result, since Open itself has no error
// return.
```

karthikiyer56 (approved):

PR #738, opened 2026-05-14
Layer-2 storage facades: MetaStore + TxHashStore + LedgerStore

### LINKED ISSUE #584: Implement RocksDB-backed LedgerStore "Reader" for Full History Ledger Retrieval

### LINKED ISSUE #689: Backfill 5: Layer 2 metastore facade (typed schema on pkg/rocksdb, conscious defaults)
## Parent PRD

#678

## What to build

- Layer 2 meta-store facade at `cmd/stellar-rpc/internal/fullhistory/backfill/metastore/` on top of `pkg/rocksdb`.
- Owns the design-doc key schema: `chunk:{C:08d}:lfs`, `chunk:{C:08d}:txhash`, `chunk:{C:08d}:events`, `index:{N:08d}:txhash`, `config:chunks_per_txhash_index`. Values are `"1"` (or the stored uint32 for config).
- Typed methods: `HasChunkLFS(chunkID) bool`, `MarkChunkLFS(chunkID) error` — plus the txhash / events / index equivalents. `DeleteChunkTxHash(chunkID) error` for cleanup. `GetConfigChunksPerTxHashIndex() (uint32, bool, error)` / `SetConfigChunksPerTxHashIndex(uint32) error`. `ProgressSummary() (ProgressSummary, error)` that scans meta store for flag counts per task type (for getStatus).
- Hardcoded meta-store RocksDB config (per #678 "RocksDB configuration (per-store conscious defaults)"): WAL enabled, `WriteBufferSizeMB ≈ 4–16`, `MaxWriteBufferNumber = 2`, `TargetFileSizeMB ≈ 16`, `MaxBytesForLevelBaseMB ≈ 16`, `MaxTotalWALSizeMB ≈ 32`, `BloomFilterBitsPerKey = 10`. Defaults are refined from actual workload during this slice.
- Wire the `CHUNKS_PER_TXHASH_INDEX` immutability contract: the facade is what the config-validate slice stubs against.

## Acceptance criteria

- [ ] Typed methods round-trip against real RocksDB (via `t.TempDir()`).
- [ ] Key format tests: construction of each key matches the design doc example exactly.
- [ ] `ProgressSummary` returns correct `completed/pending/in_progress` counts per task type for populated meta-store states (table-driven).
- [ ] `SetConfigChunksPerTxHashIndex` is write-once: second call with the same value is OK; second call with a different value returns an error (or the metastore-facade exposes a separate "immutability check" method — design choice).
- [ ] Meta-store opens with hardcoded config (not grocksdb defaults); on-open log prints the config values chosen.
- [ ] No TOML knobs for RocksDB tuning — operator-facing config has only `[META_STORE].PATH`.
- [ ] No manual `Flush` calls; background flush handles durability.

## Blocked by

- Blocked by #685

## User stories addressed (from #678)

- 9, 13, 17, 25, 27, 28

> Commit `19a917907e836366d90f189dd0ac0190b45eb8e7` on this repo contains a working-but-exploratory reference implementation of the backfill pipeline under `full-history/backfill/`. Before writing any code for this slice:
> 1. Identify which files in the reference commit correspond to this slice. Use `git show 19a9179:full-history/backfill/<file>.go` to read them (do NOT check them out).
> 2. Read the matching section of `full-history/design-docs/03-backfill-workflow.md` on `feature/full-history`.
> 3. For each reference file, make an explicit port / adapt / rewrite / discard call:
>    - **Port** if it is a pure, self-contained module and matches the current design doc.
>    - **Adapt** if the shape is right but field names, package paths, or config keys need to change for the new home under `cmd/`.
>    - **Rewrite** if the file touches an unresolved open question or predates a design-doc change.
>    - **Discard** intermediate planning artifacts and anything under `full-history/design-docs-og/`.
> 4. Aggressively port `*_test.go` files when the behavior under test still matches the current design — tests encode validated behavior and re-deriving them wastes hours.
> 5. Follow TDD: for each ported or rewritten behavior, one test → one implementation → repeat. Do not bulk-port files without a test driving each behavior.

### LINKED ISSUE #729: RocksDB-backed hot txhash store (read + write paths, 16 CFs)
## TL;DR

The **RocksDB-backed hot active txhash store** for the unified `stellar-rpc` v2 service. A single Layer-2 typed facade on top of the Layer-1 `pkg/rocksdb` wrapper (merged on `feature/full-history`), with **16 column families** routed by `txhash[0] >> 4`. Covers **both** the write path (live ingestion writes here) and the read path (`getTransaction(byHash)` query handlers read here, then federate with the cold-side RecSplit reader).

Hot-side counterpart to the cold RecSplit-backed writer (#696) and reader (txhash cold reader sub-issue).

## Where it sits

- **One Layer-2 facade**, both read and write methods on the same struct.
- Built on the merged Layer-1 wrapper; same architecture as the meta store (#689) and the hot ledger store (#584 reframed) — just with a different CF schema.
- **CF schema**: 16 column families named to match the nibble convention (`cf-0`, `cf-1`, …, `cf-f`). Routing: `cfIndex = txhash[0] >> 4`.
- **On-disk location**: `{DEFAULT_DATA_DIR}/active/txhash`; operator-overridable via `[ACTIVE_STORAGE.TXHASH].PATH`.
- **Schema per CF**: key = `txhash` (32 bytes); value = `ledgerSeq` (uint32, encoded via `rocksdb.EncodeUint32` — big-endian, single source of truth).
- **Owned by**: streaming epic (#704); built on the Layer-1 wrapper.

## Patterns inherited from the merged Layer-1 wrapper

The Layer-1 wrapper landed with these patterns; this facade reuses them verbatim and does not re-invent:

- **Two-phase lifecycle.** `New(cfg) (*HotTxHashStore, error)` validates Config; `(*HotTxHashStore).Open() error` does the actual open, idempotent via `sync.Once`. `Close()` is idempotent via `atomic.Bool`. The facade internally holds a `*rocksdb.Store` and delegates.
- **No `context.Context` in operation signatures.** The underlying C call has no cancellation hook. `Batch` takes only the callback: `Batch(fn func(*TxHashBatchWriter) error) error`.
- **Lifecycle race safety.** The Layer-1 wrapper protects against teardown-during-op races via an internal `sync.RWMutex`. The Layer-2 facade inherits that for free.
- **Byte encoding for the value (ledgerSeq).** `rocksdb.EncodeUint32` / `DecodeUint32`. NEVER direct `binary.BigEndian`.
- **Logger.** `*supportlog.Entry` passed via the facade's `Config` struct. NO custom Logger interface, NO `NopLogger`. Tests use `supportlog.New() + SetOutput(&bytes.Buffer{})`; no fixture types.
- **Operator-facing formatting.** `humanize.Bytes(uint64)`, `humanize.Comma(int64)`, `elapsed.Round(time.Microsecond).String()`.

## Configuration

```toml
[ACTIVE_STORAGE.TXHASH]
PATH = "/data/stellar-rpc/active/txhash"   # optional; default {DEFAULT_DATA_DIR}/active/txhash
```

RocksDB tuning knobs are **hardcoded inside the Layer-2 facade**. Operator-facing TOML has only the on-disk path.

## Nibble routing

The CF for a given txhash is derived inline by the facade:

```go
// cfNameFor returns the column family name for a txhash, derived from the
// top nibble of byte 0. Inline helper inside the facade — no separate
// pkg/cf package until a second consumer exists.
func cfNameFor(txhash [32]byte) string {
    return fmt.Sprintf("cf-%x", txhash[0] >> 4)
}
```

## Layer-2 facade interface

This facade is a pure storage primitive. It exposes typed read / write / batch operations and nothing else. Domain-level orchestration verbs (live-ingestion sequencing, retention / aging-out) live in separate slices in the streaming epic.

```go
package txhashstore

// HotTxHashStore is the typed Layer-2 facade for the RocksDB-backed hot active
// txhash store, built on the Layer-1 pkg/rocksdb wrapper with 16 CFs named
// cf-0 … cf-f, routed by txhash[0] >> 4.
//
// The facade hides nibble routing and value encoding from callers. The
// streaming-side ingestion orchestrator (separate slice) calls into this
// facade's typed primitives — Put for single-tx writes, Batch for atomic
// multi-tx commits per ledger.
//
// Two-phase lifecycle: New() validates Config (cheap), Open() establishes
// the underlying store (idempotent via sync.Once); Close() tears down
// (idempotent via atomic.Bool). Concurrency: safe for many concurrent
// readers + concurrent writers / batches; lifecycle protection (no
// teardown-during-op) is inherited from the Layer-1 wrapper.
type HotTxHashStore struct { /* unexported fields */ }

// New validates cfg and returns a HotTxHashStore ready to be Opened.
func New(cfg Config) (*HotTxHashStore, error)

// Open opens the underlying RocksDB store and prepares for reads + writes.
// Idempotent. Creates the on-disk directory if missing.
func (s *HotTxHashStore) Open() error

// Close cleanly shuts down. Idempotent. Waits for in-flight operations to
// drain before tearing down the underlying store.
func (s *HotTxHashStore) Close() error

// Put writes a single (txhash → ledgerSeq) entry, routing internally by
// nibble (txhash[0] >> 4) to the correct CF. Caller does not see the CF
// names or the value encoding.
//
// For atomic multi-write commits where one logical unit (e.g., one
// ledger's worth of transactions) produces multiple writes that must
// become visible together, use Batch.
func (s *HotTxHashStore) Put(txhash [32]byte, ledgerSeq uint32) error

// Batch executes fn, accumulating typed (txhash → ledgerSeq) writes
// across however many of the 16 CFs the txhashes' nibbles cover, then
// commits them atomically. Either every queued write is visible, or none.
//
// NO context.Context — the underlying RocksDB Write call has no
// cancellation hook so threading ctx through would be misleading.
func (s *HotTxHashStore) Batch(fn func(*TxHashBatchWriter) error) error

// Lookup returns the ledgerSeq for the given txhash, routing internally
// by nibble. Returns rocksdb.ErrNotFound if not present in the hot store —
// caller falls back to the cold RecSplit reader for full-chain lookup.
func (s *HotTxHashStore) Lookup(txhash [32]byte) (ledgerSeq uint32, err error)

// Health is a quick open-status + per-CF sanity check.
func (s *HotTxHashStore) Health() error
```

### TxHashBatchWriter (concrete struct, not an interface)

```go
// TxHashBatchWriter is the typed accumulator handed to HotTxHashStore.Batch's
// callback. Lifetime is scoped to the callback — calls against a captured
// pointer after the callback returns are silently inert (the underlying
// Layer-1 *BatchWriter invalidates itself; this struct inherits that
// protection).
//
// Concrete struct, not an interface — there is exactly one implementation
// and nothing in the codebase mocks this boundary.
type TxHashBatchWriter struct { /* unexported fields */ }

// Put queues a (txhash → ledgerSeq) write. Nibble routing + value
// encoding are internal; caller hands typed args. The write commits
// atomically with the rest of the batch when the callback returns nil.
func (b *TxHashBatchWriter) Put(txhash [32]byte, ledgerSeq uint32)
```

## Data lifecycle (NOT part of this interface)

This store owns its read / write / batch surface and nothing else. Retention / aging / Phase 3 freeze are owned by a separate slice in the streaming epic. There is no `PruneLedgersBelow` method.

## Where the per-ledger atomic commit lives

The "for one ledger, atomically write all its transactions" sequencing is **NOT** in this slice. It belongs to the streaming-side ingestion orchestrator (separate slice, lives under the streaming epic). That orchestrator calls into this facade's `Batch` primitive:

```go
// In the streaming-side ingestion orchestrator (NOT in this slice).
err := txhashStore.Batch(func(b *txhashstore.TxHashBatchWriter) error {
    for _, tx := range ledger.Transactions {
        b.Put(tx.Hash, ledgerSeq)
    }
    return nil
})
```

Inside this slice, the implementation of `Batch` is a thin typed adapter over the Layer-1 wrapper's batch:

```go
// In #729 — typed adapter; hides CF routing + encoding.
func (s *HotTxHashStore) Batch(fn func(*TxHashBatchWriter) error) error {
    return s.layer1.Batch(func(b1 *rocksdb.BatchWriter) error {
        return fn(&TxHashBatchWriter{b1: b1})
    })
}

func (b *TxHashBatchWriter) Put(txhash [32]byte, ledgerSeq uint32) {
    b.b1.Put(cfNameFor(txhash), txhash[:], rocksdb.EncodeUint32(ledgerSeq))
}
```

### Task 1: Layer-2 facade scaffold

- [ ] Define `HotTxHashStore` struct + `Config` + `TxHashBatchWriter` in `cmd/stellar-rpc/internal/fullhistory/streaming/txhash/store.go`.
- [ ] Wire to the Layer-1 wrapper — 16 CFs configured at open, named `cf-0` … `cf-f`.
- [ ] Two-phase lifecycle as detailed above.
- [ ] Hardcoded RocksDB config tuned for this workload: high write rate, mostly point reads, bloom filter ~10 bits/key, block cache sized for the workload, WAL on.

### Task 2: Nibble routing

- [ ] Inline helper `cfNameFor(txhash) string` inside the facade — no separate `pkg/cf` package until a second consumer needs it. Unit-tested across all 16 nibble values.

### Task 3: Write path

- [ ] `Put(txhash, ledgerSeq)`: route by nibble → encode value via `rocksdb.EncodeUint32` → `Store.Put`.
- [ ] `Batch(fn)`: thin typed adapter over `Store.Batch`. The `*TxHashBatchWriter` handed to fn applies nibble routing + value encoding.

### Task 4: Read path

- [ ] `Lookup(txhash)`: route by nibble → encode key (32-byte txhash) → `Store.Get` (uses the wrapper's zero-copy `GetPinnedCFV2` internally) → decode value via `rocksdb.DecodeUint32`.

### Task 5: Health + metrics

- [ ] `Health()`: open-status + per-CF sanity check.
- [ ] Prometheus metrics:
  - `hot_txhash_store_lookup_total` (counter, labels: status)
  - `hot_txhash_store_lookup_duration_seconds` (histogram)
  - `hot_txhash_store_write_total` (counter)
  - `hot_txhash_store_write_duration_seconds` (histogram)
  - RocksDB-derived per-CF stats.

### Task 6: Testing

- [ ] Unit tests against `t.TempDir()` RocksDB. NO Layer-1 mocking.
- [ ] Nibble-routing tests: synthesize txhashes targeting each of the 16 CFs; assert isolation (write to CF X never appears in CF Y).
- [ ] **Race-condition coverage** (mandatory): a chaos test spawns concurrent goroutines doing `Put` / `Lookup` / `Batch` across multiple nibbles while `Close` races; tested under `-race`; no panic, post-Close ops return the wrapper's closed-state error. Plus a deterministic test that parks a `Lookup` or `Batch` via channels and asserts `Close` blocks until it releases. Tests use `bytes.Buffer + (*supportlog.Entry).SetOutput` for log capture; no fixture types; under ~50 lines each.
- [ ] Open idempotency + mkdir-on-missing tests.

### Task 7: Federation hook

- [ ] `Lookup` combines hot-store result + (caller's) cold-store result, with a clear handoff convention so the federated `getTransaction(byHash)` handler routes correctly across the retention boundary.

## Acceptance

- [ ] **Two-phase lifecycle**: `New(cfg)` validates; `(*HotTxHashStore).Open()` opens; both `Open` and `Close` idempotent.
- [ ] **Primary tests run against a real RocksDB instance via `t.TempDir()`.** Every method on `HotTxHashStore` is tested by writing/reading through the real 16-CF Layer-1-backed store. NO Layer-1 mocking.
- [ ] **No `context.Context`** in any of the facade's method signatures.
- [ ] **Race-condition coverage** as detailed in Task 6.
- [ ] **Mock `HotTxHashStore` deliverable**: in-memory map-backed impl in a sibling test-package; keyed by full txhash (no per-CF sharding inside the mock — typed-method semantics are the contract; CF sharding is an implementation detail of the real version). For consumer slices (federated reader, streaming-side ingestion orchestrator) without spinning up RocksDB.
- [ ] **Mock parity tests**: a small table-driven suite asserts mock and real impl produce identical results for the same sequence of typed calls.
- [ ] Layer-2 facade compiles + opens against real RocksDB via `t.TempDir()`.
- [ ] All 16 CFs configured at open; nibble routing tested for representative txhashes covering all 16 values.
- [ ] **`Put` correctness**: writes land in the CF dictated by `txhash[0] >> 4`; verified across all 16 nibble values.
- [ ] **`Batch` atomicity**: when fn queues writes targeting multiple nibbles (so multiple CFs), either every write is visible after, or none. Tested via mid-callback error injection AND via concurrent reader observing only fully-committed sets.
- [ ] **`Batch` typing**: the `*TxHashBatchWriter` handed to fn never exposes CF names or byte-level keys; only typed `Put(txhash, ledgerSeq)`.
- [ ] **`Lookup` round-trip**: every txhash written via `Put` or `Batch` is retrievable via `Lookup` with the correct ledgerSeq.
- [ ] Write throughput meets SLA in benchmarks (#730).
- [ ] Lookup latency P50 / P90 / P95 / P99 meets SLA.
- [ ] `Open` is idempotent and creates the on-disk directory if missing.
- [ ] **Interface contains no orchestration verbs** (`IngestLedger`, `PruneLedgersBelow`, etc.). Storage primitives only.
- [ ] **Every public API method has Go doc** explaining its use case + the consumer pattern, in plain English readable by a fresh contributor opening the file cold.

## Dependencies

- Blocked by the merged Layer-1 wrapper (formerly tracked under #685).
- Federation slice is separate (lands when streaming design doc lands).
- Tracked under the txhash store epic (cross-cutting per-entity view).
- Native parent: streaming epic #704.

karthikiyer56:
## Summary

**Per-domain store packages**

- `pkg/rocksdb` — Layer-1 only: `Store`, `BatchWriter`, `Tuning`, `Config`, iterators (`Iterate` prefix, `IterateRange` bounded), encoding helpers, and the wrapper-level sentinels (`ErrInvalidConfig`, `ErrCFNotFound`, `ErrStoreClosed`).
- `pkg/stores/ledger` — hot ledger store (`HotStore`) + `Entry`. Cold counterpart lands alongside when written.
- `pkg/stores/txhash` — hot txhash store (`HotStore`) + `Entry` + 16-CF nibble routing. Cold counterpart lands alongside.
- `pkg/stores/metastore` — service-wide metadata `Store` + `Entry` + string-keyed `BatchWriter`. Single tier — no cold counterpart.
- `pkg/stores` keeps only the cross-cutting `ErrNotFound` sentinel; per-domain stores propagate `rocksdb.ErrStoreClosed` directly.

**MetaStore — generic string-KV**

- `Get` / `Put` / `Delete` / `Batch` / `PrefixScan`. Encoding is caller-driven (`fmt.Sprintf` / `strconv` at the call site); the store treats values as opaque strings.
- `Batch(fn func(*BatchWriter) error)` commits Put/Delete operations atomically with one fsync.
- Default-CF only; CF naming hidden behind the facade.

**txhash HotStore**

- 16 CFs `cf-0..cf-f`, nibble-routed; routing hidden behind the facade.
- Public surface: `AddEntries(entries)` and `Get(hash)`.

**ledger HotStore**

- Opaque bytes — no compression opinion (callers compress with zstd at the streaming write path; documented on `ledger.Entry`).
- `AddLedgers` / `GetLedgerRaw` / `IterateLedgers` (walks `[start, end]` inclusive).

**Layer-1 wrapper (`pkg/rocksdb`)**

- `New(cfg)` validates and opens in one phase; `Close` flushes and tears down idempotently.
- `Iterate(cf, prefix)` for prefix scans.
- `IterateRange(cf, start, end)` for byte-lex inclusive range scans.
- `Store.IsClosed()` accessor for facade paths that short-circuit.
- WAL on + per-write Sync on + `NoCompression` wrapper-pinned.
- `DecodeUint32` / `DecodeUint64` panic on bad length.

**Close / Flush ownership at Layer 1**

- `Store.Close` auto-flushes the active memtable before teardown.
- Best-effort; logged on failure; teardown proceeds.
- Graceful shutdown ⇒ zero WAL replay on next `New` against the same path.
- Every facade `Close` is a one-line delegate.

**Helper package**

- `pkg/geometry` — `LedgersPerChunk` constant.

**Tests**

- Real RocksDB via `t.TempDir()` across every facade.
- Exhaustive coverage of Get / Put / Delete / Batch / PrefixScan / IterateRange semantics.
- Race-chaos + Close-blocks-on-RLock under `-race`.
- Graceful close + reopen round-trips.
- Open-failure cache/filter cleanup verified via `constructAndOpen` direct invocation.
- XDR ledger round-trip through `ledger.HotStore`.

Closes #689
Closes #584
Closes #729


cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/metastore.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	chunksPerTxIndex := ledgersPerTxIndex / geometry.LedgersPerChunk
+	firstChunkID := txIndexID * chunksPerTxIndex
+	return translateError(m.store.Batch(func(b *BatchWriter) error {
+		b.Put(defaultCFName, fmt.Appendf(nil, keyIndexTxHashFormat, txIndexID), []byte{value})
+		for i := range chunksPerTxIndex {
+			b.Delete(defaultCFName, fmt.Appendf(nil, keyChunkTxHashRawFormat, firstChunkID+i))
+		}
+		return nil
+	}))
```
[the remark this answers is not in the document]
karthikiyer56: not needed — `ledgersPerTxIndex` is validated at the write site (`UpdateConfigLedgersPerTxIndex` callers). By the time `MarkTxHashIndexComplete` reads it back, the invariant already holds; duplicating the check here would split responsibility for the same rule.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/hot_store.go
```diff
@@ -0,0 +1,172 @@
...
+	names := make([]string, txHashNumCFs)
+	for i := range txHashNumCFs {
+		names[i] = fmt.Sprintf("cf-%x", i)
+	}
+	return names
+}
+
+func cfNameForTxHash(hash [32]byte) string {
+	return fmt.Sprintf("cf-%x", hash[0]>>4)
+}
```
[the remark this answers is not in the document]
karthikiyer56: done `01a6e90` — precomputed `[16]string` cf-0..cf-f table; `cfNameForTxHash` is now an array index, no per-call `fmt.Sprintf`.

cmd/stellar-rpc/internal/fullhistory/pkg/testutil/testutil.go
```diff
@@ -0,0 +1,106 @@
...
+					SourceAccount: xdr.MustMuxedAddress(keypair.MustRandom().Address()),
+					Operations:    []xdr.Operation{},
+					Fee:           xdr.Uint32(seqNum + i),
+					SeqNum:        xdr.SequenceNumber(seqNum + i),
+				},
+				Signatures: []xdr.DecoratedSignature{},
+			},
+		}
+
+		txHash, _ := network.HashTransactionInEnvelope(txEnv, networkPassphrase)
```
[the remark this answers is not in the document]
karthikiyer56: done `01a6e90` — panic on `HashTransactionInEnvelope` error. Test scaffolding; loud-fail is the right shape.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/metastore.go
```diff
@@ -0,0 +1,232 @@
...
+	return m.getU32(keyStreamingLastCommittedLedger)
+}
+
+// UpdateConfigLedgersPerTxIndex sets the immutability marker.
+// Write-once enforcement lives at the call site, not here.
+func (m *MetaStore) UpdateConfigLedgersPerTxIndex(n uint32) error {
+	return translateError(m.store.Put(defaultCFName, []byte(keyConfigLedgersPerTxIndex), EncodeUint32(n)))
+}
+
+func (m *MetaStore) GetConfigLedgersPerTxIndex() (uint32, error) {
```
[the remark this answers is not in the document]
karthikiyer56: by design. The upfront `IsClosed()` check is added only at methods that short-circuit before reaching the wrapper (empty-slice writes, invalid-range `IterateLedgers`, `MarkTxHashIndexComplete`'s pre-read). Methods that always delegate (`Get*`, `Update*`) inherit the wrapper's own `checkOpen()`. Rule is "gate where the wrapper can't catch it", not "gate everywhere".

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
```diff
@@ -0,0 +1,159 @@
...
+// between yielded entries.
+func (l *LedgerStore) IterateLedgers(start, end uint32) iter.Seq2[stores.LedgerEntry, error] {
+	return func(yield func(stores.LedgerEntry, error) bool) {
+		if l.store.IsClosed() {
+			yield(stores.LedgerEntry{}, stores.ErrStoreClosed)
+			return
+		}
+		if start > end {
+			return
+		}
```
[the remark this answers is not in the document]
karthikiyer56: done `01a6e90` — godoc now says `start > end` yields no entries and no error.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -596,26 +470,111 @@ func (s *Store) doOpen() error {
...
+	}
+	if t.Level0FileNumCompactionTrigger > 0 {
+		o.SetLevel0FileNumCompactionTrigger(t.Level0FileNumCompactionTrigger)
+	}
+	if t.Level0SlowdownWritesTrigger > 0 {
+		o.SetLevel0SlowdownWritesTrigger(t.Level0SlowdownWritesTrigger)
+	}
+	if t.Level0StopWritesTrigger > 0 {
+		o.SetLevel0StopWritesTrigger(t.Level0StopWritesTrigger)
+	}
```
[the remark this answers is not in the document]
karthikiyer56: benign. grocksdb's own default for auto-compactions is on (`false`), which is exactly what `metaStoreTuning` and `ledgerTuning` want when they leave the knob unset. Zero-value == default == desired behavior; no facade ever wants to explicitly re-enable from a non-default starting point.

tamirms:
**Drop the interfaces; reshape into per-type packages**

`pkg/stores` defines three interfaces (`MetaStore`, `LedgerStore`, `TxHashStore`) with zero production consumers — the only uses outside `pkg/rocksdb` itself are `var _ stores.X = (*X)(nil)` satisfaction assertions in test files. The `//nolint:interfacebloat` on `MetaStore` is a tell — that lint exists precisely to discourage publishing wide interfaces ahead of narrow consumers. Go's guidance is to define interfaces where they're consumed; with no consumers, the right move is to not define them and let callers use the concrete types.

The deeper issue is that the PR organizes by storage technology, not by domain. `pkg/rocksdb` mixes a generic Layer-1 wrapper with three Layer-2 facades, and each of those types is getting a packfile-backed cold counterpart. "Ledger access" is one concept; splitting it across packages by which storage engine backs which half loses cohesion.

Proposed layout:

```
pkg/rocksdb/      Layer-1 only: Store, BatchWriter, Tuning, iterators, encoding,
                  ErrStoreClosed, ErrNotFound. No facades.
pkg/ledgers/      HotStore (RocksDB-backed), Entry, tuning, tests with inlined
                  XDR helpers. Cold reader lands here when written.
pkg/txhash/       HotStore, Entry, nibble routing + cfNameByNibble, tuning, tests.
                  Cold reader lands here when written.
pkg/metastore/    Store (single tier — service-wide RocksDB state),
                  ChunkEntry/IndexEntry/ChunkKey/IndexKey, ChunkArtifactKind,
                  MarkTxHashIndexComplete, tests.
pkg/geometry/     Unchanged.
```

`HotStore` vs `ActiveStore` is a coin flip — pick whichever matches the team's vocabulary. The PR description uses "hot"; the TOML config keys use `ACTIVE_STORAGE` / `IMMUTABLE_STORAGE`. Either reads fine; just be consistent.

Removed:

- `pkg/stores/` — three interface declarations + parallel error sentinels with no callers.
- `pkg/testutil/` — generically-named util packages are widely considered an antipattern in the Go community: no clear scope, accumulate unrelated helpers over time. The Go convention is that test fixtures live in `_test.go` when one package uses them — where they can't be imported and can't accidentally grow a shared API. Today one test file (`ledger_test.go`) uses these fixtures; they belong inline in `pkg/ledgers`.
- `pkg/rocksdb/errors.go` (`translateError`) — Layer-1 owns the error sentinels; facades propagate them directly. `errors.Is(err, rocksdb.ErrStoreClosed)` works from any caller.
- The `//nolint:interfacebloat` directive.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -596,26 +470,111 @@ func (s *Store) doOpen() error {
...
+	applyDBTuning(opts, t)
+	s.applySharedTableOptions(cfOpts, t)
+}
+
+// applyPinnedCFOptions sets the per-CF values that hold for every
+// facade — applied unconditionally so a future facade can't drift
+// off these defaults.
+func applyPinnedCFOptions(o *grocksdb.Options) {
+	o.SetMinWriteBufferNumberToMerge(2)
```
tamirms: Why is this pinned to 2? The RocksDB default is 1 (each memtable flushes as its own L0 file). What's the reason for the deviation?

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
```diff
@@ -0,0 +1,160 @@
...
+		MaxWriteBufferNumber:           2,
+		Level0FileNumCompactionTrigger: 4,
+		Level0SlowdownWritesTrigger:    20,
+		Level0StopWritesTrigger:        36,
+		TargetFileSizeMB:               64,
+		MaxBytesForLevelBaseMB:         128,
+		MaxBackgroundJobs:              8,
+		MaxOpenFiles:                   10_000,
+		BlockCacheMB:                   256,
+		BloomFilterBitsPerKey:          12,
```
tamirms: Don't we already know exactly which ledgers are in the hot store? We'll never query for a ledger that isn't here. I don't think we need a bloom filter at all on the ledger store.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
```diff
@@ -0,0 +1,160 @@
...
+		MaxTotalWalSizeMB:              1024,
```
tamirms: Given the active ledger store is bounded at 10K ledgers at ~200KB each (zstd-compressed) = ~2GB total before it's dumped + discarded, these four values look sized for a much larger / longer-lived store than this actually is:

- `MaxBytesForLevelBaseMB=128` (default 256): with 2GB total lifetime data, the whole store fits in L1+L2 either way — the non-default doesn't change behavior at this scale.
- `MaxBackgroundJobs=8` (default 2): at one 200KB ledger per 5s (~40KB/s), compaction work is trivial. 2 would suffice.
- `MaxOpenFiles=10_000` (default unlimited): with 64MB SSTs and 2GB lifetime data, at most ~32 SSTs exist concurrently. 10k is 300× the working set.
- `MaxTotalWalSizeMB=1024` (default 0): WAL is bounded at ~128MB by `WriteBufferMB × MaxWriteBufferNumber`. The 1GB threshold never fires.

Were these calibrated for this scale or carried from a different (larger / longer-lived) workload?

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/metastore.go
```diff
@@ -0,0 +1,232 @@
...
+		Level0FileNumCompactionTrigger: 4,
+		Level0SlowdownWritesTrigger:    20,
+		Level0StopWritesTrigger:        36,
+		TargetFileSizeMB:               16,
+		MaxBytesForLevelBaseMB:         64,
+		MaxBackgroundJobs:              2,
+		MaxOpenFiles:                   1000,
+		BlockCacheMB:                   8,
+		BloomFilterBitsPerKey:          0,
+		MaxTotalWalSizeMB:              8,
```
tamirms: Why these values? At metastore's scale — ~6K chunks × 3 chunk-state entries = ~380KB live data plus singletons, well under 1MB total over Stellar's full history — most of these non-defaults are inert:

- `WriteBufferMB=4` (default 64): memtable never fills. Overwrites of `streaming:last_committed_ledger` are merged in-place; live data caps at ~1MB regardless of write count.
- `TargetFileSizeMB=16` (default 64): SSTs never approach 16MB.
- `MaxBytesForLevelBaseMB=64` (default 256): L1 budget irrelevant; whole store fits in one SST.
- `MaxOpenFiles=1000` (default unlimited): SST count stays in single digits.

The one knob actually doing work is `MaxTotalWalSizeMB=8`: since the memtable never fills, this is what triggers flushes (~50 bytes per `UpdateLastCommittedLedger` × ~160K writes = 8MB WAL, hits every ~9 days at today's 5s ledger cadence). Without it the WAL would grow unbounded between organic flushes.

Were the shrunk values calibrated against this scale, or do they reflect intent that hasn't materialized?

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
```diff
@@ -0,0 +1,160 @@
...
+				b.Put(defaultCFName, EncodeUint32(e.Seq), e.Bytes)
+			}
+			return nil
+		}))
+	}
+}
+
+// DeleteLedgers removes the given sequences. Missing sequences are
+// silently skipped. Same single-vs-batch dispatch as AddLedgers.
+func (l *LedgerStore) DeleteLedgers(seqs []uint32) error {
```
tamirms: Why does this exist? The active store lifecycle is fill-to-10K → dump-to-cold → discard the whole store wholesale. There's no per-key retention within the hot window. When would this ever be called?

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/txhash.go
```diff
@@ -0,0 +1,178 @@
...
+				b.Put(cfNameForTxHash(e.Hash), e.Hash[:], EncodeUint32(e.LedgerSeq))
+			}
+			return nil
+		}))
+	}
+}
+
+// RemoveEntries deletes by hash atomically. Missing hashes are
+// silently skipped.
+func (s *TxHashStore) RemoveEntries(hashes [][32]byte) error {
```
tamirms: Why does this exist? When would this ever be called?

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
```diff
@@ -0,0 +1,160 @@
...
+				return
+			}
+		}
+	}
+}
+
+// GetLedgerRange returns the smallest and largest sequences in the
+// store, or (0, 0, nil) when empty. O(1) — backed by RocksDB
+// SeekToFirst + SeekToLast on SST file metadata, no data I/O.
+func (l *LedgerStore) GetLedgerRange() (uint32, uint32, error) {
```
tamirms: Why is this needed? The active store's bounds are derivable upstream — chunk start is deterministic from chunk_id, and `streaming:last_committed_ledger` is in the metastore. When would this ever be called?

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger.go
```diff
@@ -0,0 +1,41 @@
+package stores
+
+import "iter"
+
+// LedgerEntry — one (sequence, opaque bytes) pair.
+// The store stores Bytes verbatim and has no opinion on encoding.
+type LedgerEntry struct {
+	Seq   uint32
```
tamirms: Aren't ledgers supposed to be zstd-compressed on write / decompressed on read? The store treats bytes as opaque and the wrapper pins `Compression=None` — so the caller has to handle compression. But that contract isn't documented anywhere.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -428,88 +245,134 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
...
-// Calling Flush on a closed or never-Opened Store returns the
-// corresponding error.
-//
-// Holds the lifecycle read-lock for the duration of the underlying C
-// call. See the mu field doc on Store for what that lock is and is
-// not for.
+// Gap handling: yields every key that exists in [startKey, end-of-CF).
+// Holes (e.g., 100, 102, 105) are silent — callers comparing
+// consecutive yielded keys decide whether a gap is fatal.
+func (s *Store) IterateFrom(cf string, startKey []byte) iter.Seq2[Entry, error] {
```
tamirms: `IterateFrom`'s only caller (`LedgerStore.IterateLedgers`) open-codes an upper-bound check (`if seq > end { return }`) — the actual usage pattern is a bounded range, not unbounded forward.

`Iterate` (prefix scan) is the pattern the streaming retention scan needs — walking `hot:*`, `chunk:*`, `index:*` keys.

Could we replace `IterateFrom` with `IterateRange(cf, start, end)` walking keys in `[start, end]` inclusive (nil start = first key in CF, nil end = walk to end of CF), absorbing the upper-bound check into the wrapper? And keep `Iterate(cf, prefix)` as-is. Two functions matching the two real usage patterns, no open-coded termination at the call site.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/metastore.go
```diff
@@ -0,0 +1,103 @@
...
+
+// MetaStore is the typed contract for the service-wide
+// key/value/state store. Multi-row entries (Chunk/Index) plus typed
+// singletons (last-committed-ledger, ledgers-per-tx-index) plus the
+// one atomic transition cleanup_txhash needs.
+// AddEntries / DeleteEntries / MarkTxHashIndexComplete each commit
+// as one transaction; one fsync per call regardless of size.
+//
+//nolint:interfacebloat
+type MetaStore interface {
```
tamirms: The current API — typed `ChunkEntry`/`IndexEntry`/`ChunkKey`/`IndexKey` wrappers, sealed marker interfaces `MetaStoreEntry`/`MetaStoreKey`, generic `AddEntries`/`DeleteEntries` dispatched through type switches, dedicated `MarkTxHashIndexComplete`, typed singleton accessors — doesn't match how the metastore is actually used in either the streaming or backfill design docs.

The design docs treat the metastore as a generic string-keyed KV store:

- `meta_store.get(key)` / `put(key, value)` / `has(key)` / `delete(key)`
- `batch = meta_store.new_batch(); batch.put(...); batch.delete(...); batch.commit()` for atomic multi-key transitions
- `meta_store.prefix_scan("hot:chunk:")`, `prefix_scan("hot:index:")` for lifecycle iteration
- `meta_store.iter_range("chunk:00000000:", f"chunk:{N:08d}:")` for retention-bounded scans

The exact shape of a type-safe metastore API depends on design decisions that haven't been finalized yet in the streaming design. A generic string KV API decouples the metastore from those still-pending decisions, letting the design iterate quickly at this layer without API churn blocking it:

```go
type Store struct{ ... }
func (s *Store) Open() error
func (s *Store) Close() error

func (s *Store) Get(key string) (value string, ok bool, err error)
func (s *Store) Put(key, value string) error
func (s *Store) Delete(key string) error
func (s *Store) Has(key string) (bool, error)

func (s *Store) Batch(fn func(*Batch) error) error

func (s *Store) PrefixScan(prefix string) iter.Seq2[Entry, error]
func (s *Store) Range(start, end string) iter.Seq2[Entry, error]
```

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -142,56 +74,26 @@ type Store struct {
...
+	// freeing the C++ DB. Without this, a Put already past checkOpen
+	// would segfault against a torn-down DB.
 	mu sync.RWMutex
 
 	closed atomic.Bool
 }
 
 // New validates cfg and returns a Store ready to be Opened.
-// Returns ErrInvalidConfig if cfg.Path or cfg.Logger is missing.
 func New(cfg Config) (*Store, error) {
```
tamirms, line 97: Why is there a separate `Open` method? Every caller invokes `New` then `Open` back-to-back; nothing happens between them, and `Open` doesn't distinguish create-vs-open (it sets `CreateIfMissing=true` either way).

The split adds `sync.Once` + `openErr` + the `ErrStoreNotOpened` sentinel, and makes `Open` non-retryable on transient failure (the failed result is cached in `openErr` until you discard the Store and call `New` again).

Could `New` just do the open work? Single-phase `New(cfg) (*Store, error)` returning a ready-to-use Store removes all three.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -569,11 +446,8 @@ func (s *Store) doOpen() error {
 		cfOpts[i] = grocksdb.NewDefaultOptions()
 	}
 
-	// Time the grocksdb open call.
-	// A store with a populated WAL to replay or many L0 files to
-	// reconcile can take seconds to come up; the elapsed time goes
-	// straight into the on-open log line so a slow restart is visible
-	// in operator logs without further instrumentation.
+	s.applyTuning(opts, cfOpts)
```
tamirms, line 406: Cache/filter leak on Open failure. `applyTuning` at line 449 creates `s.cache` (via `NewLRUCache`) and `s.filter` (via `NewBloomFilter`) as side effects. If `OpenDbColumnFamilies` then fails, this cleanup destroys `opts` and `cfOpts` but never `s.cache`/`s.filter`. A later `Close()` won't free them either — Close short-circuits when `s.db == nil`. With `txHashTuning` that's ~512MB of LRU cache leaked per failed Open. Fix is to also destroy `s.cache`/`s.filter` (and nil them) here.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/encoding.go
```diff
@@ -2,78 +2,40 @@ package rocksdb
...
 }
 
-// DecodeUint32 reads 4 bytes encoded by EncodeUint32 and returns the
-// uint32.
-// Returns 0 if b is not exactly 4 bytes — callers that need to tell
-// "decode failed" from "the encoded value really was zero" must
-// validate len(b) == 4 themselves first.
+// DecodeUint32 returns the uint32 from a 4-byte big-endian buffer,
+// or 0 if len(b) != 4.
 func DecodeUint32(b []byte) uint32 {
```
tamirms, line 27: `DecodeUint32(b)` returns 0 silently when `len(b) != 4`. Iterator-key callers (`LedgerStore.IterateLedgers` does `DecodeUint32(e.Key)` without pre-checking length) get silent zeros instead of an error if data ever drifts from the expected encoding. `MetaStore.getU32` checks length defensively before calling; the helper itself doesn't. Suggest returning `(uint32, error)` or panicking on bad length — silent-zero is the worst option.

karthikiyer56:
> **Drop the interfaces; reshape into per-type packages**
> 
> `pkg/stores` defines three interfaces (`MetaStore`, `LedgerStore`, `TxHashStore`) with zero production consumers — the only uses outside `pkg/rocksdb` itself are `var _ stores.X = (*X)(nil)` satisfaction assertions in test files. The `//nolint:interfacebloat` on `MetaStore` is a tell — that lint exists precisely to discourage publishing wide interfaces ahead of narrow consumers. Go's guidance is to define interfaces where they're consumed; with no consumers, the right move is to not define them and let callers use the concrete types.
> 
> The deeper issue is that the PR organizes by storage technology, not by domain. `pkg/rocksdb` mixes a generic Layer-1 wrapper with three Layer-2 facades, and each of those types is getting a packfile-backed cold counterpart. "Ledger access" is one concept; splitting it across packages by which storage engine backs which half loses cohesion.
> 
> Proposed layout:
> 
> ```
> pkg/rocksdb/      Layer-1 only: Store, BatchWriter, Tuning, iterators, encoding,
>                   ErrStoreClosed, ErrNotFound. No facades.
> pkg/ledgers/      HotStore (RocksDB-backed), Entry, tuning, tests with inlined
>                   XDR helpers. Cold reader lands here when written.
> pkg/txhash/       HotStore, Entry, nibble routing + cfNameByNibble, tuning, tests.
>                   Cold reader lands here when written.
> pkg/metastore/    Store (single tier — service-wide RocksDB state),
>                   ChunkEntry/IndexEntry/ChunkKey/IndexKey, ChunkArtifactKind,
>                   MarkTxHashIndexComplete, tests.
> pkg/geometry/     Unchanged.
> ```
> 
> `HotStore` vs `ActiveStore` is a coin flip — pick whichever matches the team's vocabulary. The PR description uses "hot"; the TOML config keys use `ACTIVE_STORAGE` / `IMMUTABLE_STORAGE`. Either reads fine; just be consistent.
> 
> Removed:
> 
> * `pkg/stores/` — three interface declarations + parallel error sentinels with no callers.
> * `pkg/testutil/` — generically-named util packages are widely considered an antipattern in the Go community: no clear scope, accumulate unrelated helpers over time. The Go convention is that test fixtures live in `_test.go` when one package uses them — where they can't be imported and can't accidentally grow a shared API. Today one test file (`ledger_test.go`) uses these fixtures; they belong inline in `pkg/ledgers`.
> * `pkg/rocksdb/errors.go` (`translateError`) — Layer-1 owns the error sentinels; facades propagate them directly. `errors.Is(err, rocksdb.ErrStoreClosed)` works from any caller.
> * The `//nolint:interfacebloat` directive.


-----


 Yeah, I was torn between this. Standard guidance is don't introduce an interface until there's more than one concrete impl — and right now we only have RocksDB. 
The thinking was that the DAG / backfill work could be developed against  mocks of these interfaces, pre-defining them now to save the extract-later step. 
That said, the real backfill work's unit tests can also be done against real rocksdb tempdirs, much like how most of the UTS happen today.

I will do the rework in conjunection with addressing your other comments

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: A lot of these fine tuning parameters are a throwback to when we were operating under different assumpitsion

1/
we were still positing that rockdb would serve as the YEAR worth of ledger store.
As in, it would store 5millionish elements for ledger, and 2 billionish txhashes (without the 16 cf column split at that time).
when fine tuning them during my PoC - so as to complete ingestion faster whilst also having better query performance, I landed on these settings.

2/
we would still be having big machines. 128gb minimum for full history and so on.


that approach has since been abandoned ofcourse, these defaults have not.
not all at any rate.

A lot of these fine tuning parameters are probably not needed for the ledger/event rocsdkb hot store.

I have an ticket item in the full PRD to reevaluate all these config options and come with sane defaults.
That ticket  will serve as the eval grounds for what defaults we want to land on.

i am going to mark this particular one - `SetMinWriteBufferNumberToMerge` back to 1 in the meanwhile

The only thing that is proper, for the most part, is the txhashstore settings, since those are coming from a time when I did the PoC using 16 cfs around February

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: Same ratiionale as my response [here](https://github.com/stellar/stellar-rpc/pull/738/files/01a6e90ce581addd9a04a308cc4032cb84e75444#r3242214524)
yes, they are a throwback to when they were done with 1 year worth of ledgerstoer for backfill and streaming.
not applicable anymore.

My intent with this PR was to push out something to get the work unblocked for PoC next week.
I havent revisited these tunings before.
I was about to leave a comment to that effect before review started.
That being said, the defaults here make sense and can be saefly done for the ledger Store and I see no issue in retaning them.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: yup, we dont need it.
removing

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/metastore.go
karthikiyer56: > Were the shrunk values calibrated against this scale, or do they reflect intent that hasn't materialized?

I unilaterally reduced numbers across the board here significantly for metastore.
And no, these numbers have not been calibrated.
Yes. I realize that the data to be written is less than 1Mb and not even a single memtables is going to be filled at this scale.,. 4Mb fo the WriteBuffer just felt simpler.

That said, I am not sure what you are suggesting here in the comment to do.
Are you asking me to size down WriteBufferMb to 1mb?
TargetFileSize to Y?

Or simply eliminate all keys and just retain defaults for the meta store? 
as in, get rid of this entire tuning section for metastore?

tamirms: Just use defaults — drop the entire `metaStoreTuning()` function (pass an empty `Tuning{}` or remove the `Tuning` arg from `NewMetaStore` altogether).

At <1MB live data, none of these knobs do meaningful work:
- `WriteBufferSize` at default 64MB: memtable never fills regardless
- `MaxTotalWalSize` at default 0 (dynamically ~512MB given the other defaults): WAL never approaches that either
- Memtable + WAL just accumulate slightly until graceful `Close`, which auto-flushes

If ungraceful shutdown is a concern, the WAL replay is a few KB at most — trivial.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: This and the deletes in the txstore was just an attempt to provide it as a function even though there is no use case for it just yet.
that said, I have already revised my stance and am removing this myself in the next push

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/txhash.go
karthikiyer56: no reason anymore.
will be removed

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: will be removed

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/encoding.go
karthikiyer56: cant believe I missed this.
Claude. despite all my prompts/durectives/tenets for "check correctness/corner cases ad nauseum over cleverness", didnt catch this either.
panicking is better.
idt returning an error here is warranted

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: > Could New just do the open work? 

It could.
We dont need an `open` surface per se.
That removes a lot of dont open after close kinda things altogether. since there is no `open`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: good catch.
will fix it

cmd/stellar-rpc/internal/fullhistory/pkg/stores/metastore.go
karthikiyer56: The pseudo-code in the design docs does use `(key:str, value:str)` — but that's pseudo-code, and the string-string shape isn't load-bearing in either direction. Key obviously has to be a string; value doesn't have to be.

  The reason for the typed shape was to subsume value encoding inside the metastore rather than leak it to callers:

  - Chunk/index state values are uint8 — they can carry a typed enum (`ChunkLFSInProgress = 0`, `ChunkLFSDone = 1`) so the state machine reads in code at the call site, not in `strconv.Itoa` / `[]byte(EncodeUint32(...))` side conventions.
  - Singletons like `last_committed_ledger` are uint32 — typed accessor `UpdateLastCommittedLedger(seq uint32)` hides the EncodeUint32 / DecodeUint32 from callers.
  - Generic KV would push that encoding (or worse, ad-hoc `strconv` parsing) onto every caller, and every caller has to agree on the encoding. With a typed shape, the metastore owns it.

  The atomic-batch property is preserved either way — generic `Batch` and typed `AddEntries` are both one-fsync-per-call.

  That said, I'm not strongly opinionated. If you'd rather have the metastore be a string-KV with callers wrapping typed accessors at their own layer, I can rework.

karthikiyer56:
done `e6e8ce6` — restructured into `pkg/rocksdb` (Layer-1 only) plus per-domain packages `pkg/stores/ledger`, `pkg/stores/txhash`, `pkg/stores/metastore`. The new shape leaves room for cold-storage readers (packfile-backed) to land in `pkg/stores/ledger` and `pkg/stores/txhash` alongside their hot counterparts when they're written — same package, same domain. `pkg/stores` itself keeps only the shared `ErrNotFound` sentinel; `pkg/testutil` removed (XDR fixtures inlined into the one test that used them); `pkg/rocksdb/errors.go` (translateError) removed — facades propagate `rocksdb.ErrStoreClosed` directly.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done `0aa501d` — set back to 1.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: done `0aa501d` — bloom dropped; `ledgerTuning` now collapses to defaults.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: done `0aa501d` — collapsed entirely to `Tuning{}`; rationale inlined into the constructor's docstring.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/metastore.go
karthikiyer56: done `0aa501d` — `metaStoreTuning()` dropped; the metastore constructor uses a default `rocksdb.Config`. No explicit cache, filter, or WAL cap.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: done `0aa501d` — `DeleteLedgers` removed; the post-restructure type is `ledger.HotStore` in `pkg/stores/ledger`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/txhash.go
karthikiyer56: done `0aa501d` — `RemoveEntries` removed; the post-restructure type is `txhash.HotStore` in `pkg/stores/txhash`.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/ledger.go
karthikiyer56: done `0aa501d` — `GetLedgerRange` removed from `ledger.HotStore`; `rocksdb.Store.FirstLastKey` (its only consumer) dropped along with it.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger.go
karthikiyer56: done `0aa501d` — `ledger.Entry`'s docstring now documents the caller-side zstd convention: the store treats `Bytes` verbatim; the streaming write path compresses with zstd before `AddLedgers` and readers decompress after `GetLedgerRaw` / `IterateLedgers`. Not enforced at the store boundary.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done `0aa501d` — `IterateFrom` replaced with `IterateRange(cf, start, end)`. nil start = SeekToFirst, nil end = walk to end of CF. `ledger.HotStore.IterateLedgers` drops the open-coded upper-bound check; the wrapper handles it now.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/metastore.go
karthikiyer56: done `0aa501d` — went with the generic string-KV shape. `metastore.Store` exposes `Get` / `Put` / `Delete` / `Batch` / `PrefixScan`. Encoding is caller-driven (`fmt.Sprintf` / `strconv` at the call site). `Range` deliberately not added — hierarchical keys (`chunk:NNNNNN:kind`) make lex bounds awkward (e.g., `chunk:00000200:` sorts before `chunk:00000200:events`); `PrefixScan` + caller-side filter is cleaner. If a future caller needs a tight numeric range, it's a thin wrapper over the Layer-1 `rocksdb.Store.IterateRange` we already have.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done `0aa501d` — `rocksdb.New` does construction + open in one phase. Facade Open methods removed (each store's `New` returns a fully-open instance); `ErrStoreNotOpened` sentinel gone. A private `constructAndOpen` method behind `rocksdb.New` preserves the leak-fix regression test (test builds a `*rocksdb.Store` directly and inspects post-failure state on the half-built receiver).

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
karthikiyer56: done `0aa501d` — `rocksdb.OpenDbColumnFamilies` failure branch now destroys `s.cache` and `s.filter` (and nils them) alongside `opts` / `cfOpts`. The leak-regression test exercises `constructAndOpen` directly to inspect post-failure state on the half-built receiver.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/encoding.go
karthikiyer56: done `0aa501d` — `DecodeUint32` + `DecodeUint64` panic on bad length with a diagnostic message. Internal-invariant style — bad length means a writer somewhere wrote garbage, which is a code bug, not a runtime condition.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -1,28 +1,6 @@
...
-//   - Holding the cross-process LOCK file so two services can't
-//     stomp on the same directory.
-//   - Keeping the WAL on (configurable in code, but no facade flips
-//     it off).
-//   - Plain bytes-in / bytes-out Put / Get / Delete / Iterate /
-//     Batch / Flush / Close.
-//
-// Each facade gets its own dedicated RocksDB directory, flock, and
-// Config — nothing is shared between facades.
+// Package rocksdb wraps grocksdb: Layer-1 generic store +
```
tamirms, line 1: Nit: package doc references "Layer-2 typed facades" — those moved to `pkg/stores/{ledger,txhash,metastore}` in the restructure. This is now Layer-1 only.

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -428,88 +233,97 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
...
-// succeeded).
-// This avoids the race where Close fires while Open is still
-// constructing the grocksdb DB, returns early on s.db == nil, and
-// leaves the just-opened DB (and its flock) leaked.
-//
-// Close does NOT auto-Flush.
-// Data is durable either way (WAL replay handles it on next Open),
-// but callers wanting a clean shutdown with no WAL replay should call
-// Flush first.
+// IsClosed reports whether Close has been called. Used by Layer-2
```
tamirms, line 301: Nit: godoc references "Layer-2 facade methods" but those live in `pkg/stores/{ledger,txhash}` now. Could read "callers that short-circuit before reaching wrapper ops."

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -428,88 +233,97 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.db == nil {
-		// Either Open never ran, or it failed before setting s.db.
-		// Nothing to clean up.
 		return nil
 	}
+
+	if err := s.doFlush(); err != nil {
+		s.cfg.Logger.WithError(err).Warnf("rocksdb: graceful close Flush failed at %s; next Open will replay WAL", s.cfg.Path)
```
tamirms, line 324: Nit: `Open` is folded into `New` now. Log message should read "next `New` will replay WAL".

tamirms (approved):
Approved. Restructure addressed all substantive review items — interface drop, per-domain packages, tuning cleanup, leak fix, encoding panics, single-phase `New`, `IterateRange`, generic-KV metastore. Remaining items are doc/comment staleness filed as nits.

tamirms (approved):
Approved. Restructure addressed all substantive review items — interface drop, per-domain packages, tuning cleanup, leak fix, encoding panics, single-phase New, IterateRange, generic-KV metastore. Remaining items are doc/comment staleness filed as nits.

tamirms (approved):

PR #739, opened 2026-05-15
Cold ledger store: packfile-backed reader + writer (#695)

### LINKED ISSUE #695: Backfill 13: Land packfile dep + real ledger packfile writer (replaces stub)
## Parent PRD

#678

## What to build

- Land the external `packfile` library pin in `go.mod` (same PR as the real ledger writer impl).
- Replace the blank ledger writer stub at `backfill/writer/ledger/` with a real packfile-backed impl behind the existing `Writer` interface.
- `AppendLedger` buffers or streams (implementation's choice — interface-opaque) ledger bytes using the packfile API.
- `Finalize` flushes + fsyncs the `.pack` file at the path returned by `paths.LedgerPackPath`.
- Port relevant code from reference commit `19a9179`'s `lfs_writer.go` / `chunk_writer.go` where the shape still matches PR #633's format spec.
- No changes to `ProcessChunkTask` — it still calls the same interface.

## Acceptance criteria

- [ ] Integration test: build a chunk's pack file end-to-end, then read it back through the packfile library and assert every ledger round-trips.
- [ ] `.pack` file layout matches PR #633's format spec exactly.
- [ ] `go.mod` has the new packfile pin; `go.sum` updated.
- [ ] No changes to the `Writer` interface surface; `ProcessChunkTask` still passes its unit tests untouched.

## Blocked by

- Blocked by #691

## User stories addressed (from #678)

- 22

> Commit `19a917907e836366d90f189dd0ac0190b45eb8e7` on this repo contains a working-but-exploratory reference implementation of the backfill pipeline under `full-history/backfill/`. Before writing any code for this slice:
> 1. Identify which files in the reference commit correspond to this slice. Use `git show 19a9179:full-history/backfill/<file>.go` to read them (do NOT check them out).
> 2. Read the matching section of `full-history/design-docs/03-backfill-workflow.md` on `feature/full-history`.
> 3. For each reference file, make an explicit port / adapt / rewrite / discard call:
>    - **Port** if it is a pure, self-contained module and matches the current design doc.
>    - **Adapt** if the shape is right but field names, package paths, or config keys need to change for the new home under `cmd/`.
>    - **Rewrite** if the file touches an unresolved open question or predates a design-doc change.
>    - **Discard** intermediate planning artifacts and anything under `full-history/design-docs-og/`.
> 4. Aggressively port `*_test.go` files when the behavior under test still matches the current design — tests encode validated behavior and re-deriving them wastes hours.
> 5. Follow TDD: for each ported or rewritten behavior, one test → one implementation → repeat. Do not bulk-port files without a test driving each behavior.

karthikiyer56:
## Summary

- Adds `ColdStoreWriter` + `ColdStoreReader` at `cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_store.go`, siblings to `HotStore`. Two-phase write contract: `Commit` finalizes (trailer + fsync), `Close` cleans up a partial pack when `Commit` hasn't run.
- Each ledger is one packfile record (`ItemsPerRecord=1`). Zstd wired via packfile codec slots — `WriterOptions.NewRecordEncoder` returns a per-worker `*zstd.Compressor`, `ReaderOptions.RecordDecoder` takes a shared, concurrent-safe `*zstd.Decompressor`. The cold layer hands raw ledger bytes to `AppendLedger` and gets raw bytes back; compression is opaque.
- `Overwrite: true` on `packfile.Create` so a crashed mid-write at the same path is truncated and rewritten on retry — backfill doesn't need to clean up partial packs. `ContentHash` off; zstd's per-frame checksum already covers per-record integrity.
- `NewColdStoreReader` reads the trailer once, rejects packs whose `Format != formatLedgerCold`, requires `TotalItems > 0`, decodes the 4-byte big-endian `firstSeq` from `AppData`, derives `lastSeq = firstSeq + TotalItems - 1`.
- Writer enforces contiguous-append: a gap or out-of-order seq returns an error and does NOT advance internal state. `GetLedgerRaw` returns `stores.ErrNotFound` when seq is outside `[FirstSeq, LastSeq]`; otherwise it's a pass-through to `packfile.ReadItem`. `IterateLedgers` clamps to `[FirstSeq, LastSeq]`. Post-close calls return packfile's native errors verbatim — no boundary sentinel translation.
- Tests in `cold_store_test.go` cover round-trip table-driven × {1, 3, 7, 10} (byte-equality + XDR-parseability + iteration ordering), input validation, gap / out-of-order rejection, trailer + AppData inspection, truncate-on-retry (with documented intentional fd + goroutine leak), out-of-range → `ErrNotFound`, mid-walk break, bounds clamping, AppData-size + format-mismatch rejection, shared-decompressor across packs. Use-after-close and Close-idempotency tests pruned — packfile's own tests already cover those.

Closes #695


cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_writer.go
```diff
@@ -0,0 +1,104 @@
...
+	path string,
+	firstSeq uint32,
+	logger *supportlog.Entry,
+) (*ColdWriter, error) {
+	if path == "" {
+		return nil, rocksdb.ErrInvalidConfig
+	}
+	if logger == nil {
+		return nil, rocksdb.ErrInvalidConfig
+	}
```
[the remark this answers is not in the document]
karthikiyer56: done `011a4cf` — logger persisted, lifecycle events logged at Info/Warn.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_store.go
```diff
@@ -0,0 +1,152 @@
...
+) (*ColdStore, error) {
+	if path == "" {
+		return nil, rocksdb.ErrInvalidConfig
+	}
+	if decoder == nil {
+		return nil, rocksdb.ErrInvalidConfig
+	}
+	if logger == nil {
+		return nil, rocksdb.ErrInvalidConfig
+	}
```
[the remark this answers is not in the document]
karthikiyer56: done `011a4cf` — same fix on the reader.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_writer.go
```diff
@@ -0,0 +1,104 @@
...
+		return stores.ErrStoreClosed
+	}
+	var ad [appDataSize]byte
+	binary.BigEndian.PutUint32(ad[:], w.firstSeq)
+	if err := w.pw.Finish(ad[:]); err != nil {
+		return err
+	}
+	_ = w.enc.Close()
+	w.closed.Store(true)
+	return nil
```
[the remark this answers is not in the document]
karthikiyer56: done `011a4cf` — adopted the enc.Close error return. Skipped auto-close on Finish error: the writer's documented two-phase contract has callers using `defer Close()` to clean up after a failed Commit.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_writer_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+func TestNewColdWriter_TruncatesPreexistingFile(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "ledgers.pack")
+
+	crashed, err := NewColdWriter(path, 1, silentLogger())
+	require.NoError(t, err)
+	for i := range uint32(100) {
+		require.NoError(t, crashed.AppendLedger(1+i, []byte("stale-ledger-payload-padding-padding-padding")))
+	}
+	_ = crashed
```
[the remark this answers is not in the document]
karthikiyer56: done `011a4cf` — dropped the dead line, added a two-line comment naming the intentional fd leak.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_store.go
```diff
@@ -0,0 +1,269 @@
...
+	firstSeq uint32,
+	logger *supportlog.Entry,
+) (*ColdStoreWriter, error) {
+	if path == "" {
+		return nil, rocksdb.ErrInvalidConfig
+	}
+	if logger == nil {
+		return nil, rocksdb.ErrInvalidConfig
+	}
+	pw, err := packfile.Create(path, packfile.WriterOptions{
```
tamirms: Was there a specific downside in mind when this moved off `WriterOptions.NewRecordEncoder` / `ReaderOptions.RecordDecoder` and into the cold layer? The commit message on `3fe03cf5` says "delegated too much to packfile" but I couldn't tell from the diff what the concrete problem was — happy to be corrected if I'm missing it.

Reading `packfile/codec.go` and the `recordWorker` path, the abstraction looks like it was sized for exactly this case: the doc even names zstd as the typical implementation, `NewRecordEncoder`'s "per-worker, not concurrent-safe" contract matches `*zstd.Compressor`, and `RecordDecoder`'s "shared, concurrent-safe" contract matches `*zstd.Decompressor`.

tamirms: Both `atomic.Bool closed` fields can be deleted in favor of boundary sentinel translation. They're not doing the work they look like they're doing.

- Writer: with zstd inside `NewRecordEncoder`, the cold-store doesn't own `enc` anymore — `recordWorker` does via `defer enc.Close()` at `writer.go:335` — so there's no compressor lifecycle to gate. Post-close `AppendLedger`/`Commit` calls become `errors.Is(err, packfile.ErrWriterClosed) → stores.ErrStoreClosed` at the boundary.

- Reader: `packfile.Reader.Close` is already idempotent via `sync.Once` at `reader.go:750`. After Close, in-flight `ReadAt`s return `os.ErrClosed`-wrapped errors, and `errors.Is(err, os.ErrClosed) → stores.ErrStoreClosed` at `GetLedgerRaw`/`IterateLedgers` cleans them up. That's the whole job.

This also resolves the close-during-read race that `packfile.Reader.Close` warns about at `reader.go:135-137`: in-flight reads racing `Close` return `ErrStoreClosed` cleanly via the boundary translation. `TestColdStoreReader_ConcurrentReadsRaceFree` passes today by discarding errors; the same race yields the right sentinel under the proposed shape.

Net: drop both flags; boundary-translate sentinels in their place.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_store.go
```diff
@@ -0,0 +1,269 @@
...
+// AppendLedger rejects a gap or out-of-order seq without advancing
+// internal state.
+func (w *ColdStoreWriter) AppendLedger(seq uint32, ledgerBytes []byte) error {
+	if w.closed.Load() {
+		return stores.ErrStoreClosed
+	}
+	if seq != w.nextSeq {
+		return fmt.Errorf("cold: expected seq %d, got %d", w.nextSeq, seq)
+	}
+	compressed, err := w.enc.Encode(w.encBuf[:0], ledgerBytes)
```
tamirms: `HotStore.Entry`'s doc at `hot_store.go:15-19` says the caller compresses bytes with zstd before `AddLedgers` and decompresses after `GetLedgerRaw` / `IterateLedgers` — caller-side contract, store stores verbatim. `ColdStoreWriter.AppendLedger` and `ColdStoreReader.GetLedgerRaw` / `IterateLedgers` do the opposite: compression happens inside the store, callers see raw bytes both ways.

The write-side hazard:
```go
bytes, _ := hot.GetLedgerRaw(seq)   // zstd-compressed (per hot convention)
cold.AppendLedger(seq, bytes)       // zstd's it again
```
Doubly-zstd'd bytes in the cold pack. Silent on write; surfaces as garbage on a single-Decode read.

The read-side hazard is symmetric: a caller written against a generic `LedgerStore` interface can't interpret the returned bytes without knowing which backend served the call. The same `GetLedgerRaw` method name even means *different things* on the two stores today — decompressed on cold, compressed on hot — which compounds the confusion.

Best resolution is to make hot match cold: hot store internalizes compression so both stores' APIs take raw bytes on write and return raw bytes on read. *How* hot internalizes is a separate call — rocksdb-level via flipping the `NoCompression` pin at `rocksdb.go:477` (though block-level zstd at default 4–16KB block sizes won't recoup the ratio for 1.5MB-uncompressed / 200KB-compressed ledgers that span dozens of blocks each), or wrapper-level zstd inside `HotStore.AddLedgers` / `GetLedgerRaw` before/after the rocksdb call (per-value compression, cleaner for these value sizes). Either way, the `Entry` doc on `hot_store.go:15` and the cold writer's contract end up describing one model and the `GetLedgerRaw` semantic collision goes away.

karthikiyer56: yeah, so this is something that I have been meaning to hash out in comments here.
the dichotomy -  that hot store might be pushing the compression responsibility at call site VS cold store that handles it internally via zstd,  became evident as I was coding this up.

I have a revised opinion on this now.
I am leaning towards not doing any compression in rocksdb ledger hot store at all - not internally in hot_store.go nor at any call site during streaming mode where it might be used.

Each uncompressed ledger is around 1.2-1.5 MB (after compression, its around 120kb)
for our chunk worth of ledger in a hot store usecase, this translates to arounf `10k * 1.5 MB uncompressed lcm = 15GB` vs `1.2GB` in the compressed shape.
In this particular instance, ~12Gb of additional storage might be worth it - as in, we dont have to decompress on query path. (Perhaps we shave off ~1ms on decompression?)


One thing I dont want to do is to leverage rocksdb native compression via the compression related config knobs. II cant remember the exact numbers for the rocksdb natively compressed datastore size nor the query  time of the top of my head. I do remember that compression was not that good as 120Kb though.

At any rate, I am leaning towards this in decreasing order of preference:
1/ no compression anywhere whatsoever. key = ledgerSeq, val = 1.5mb of raw bytes
2/ zstd compression before inserting in rocksdb(no more call side)

wydt

cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_store.go
karthikiyer56: > Was there a specific downside in mind when this moved off... The commit message on 3fe03cf5 says "delegated too much to packfile" but I couldn't tell from the diff what the concrete problem was — happy to be corrected if I'm missing it.

Honestly...
this was a Claude Miss and one that I didnt catch either in my haste to put this PR out to close out the loop on "pushing out building blocks code".
i will rework this and push it out

karthikiyer56: done `7dac92e` — zstd moved onto `WriterOptions.NewRecordEncoder` / `ReaderOptions.RecordDecoder`.

PR #741, opened 2026-05-15
Add RPC ingestion load test driven by synthetic apply-load ledger bundles

### LINKED ISSUE #711: Release eval: Add repeatable Core apply-load integration test
Parent: #706 

## Summary

Add a repeatable integration test, modeled after Horizon's apply-load ledger generation test, that uses Stellar Core `apply-load` with one checked-in mixed classic and Soroban parameter set to generate synthetic ledger metadata for RPC ingestion validation.

This first version should prove the harness and artifact flow with a single stable profile. Broader parameter variation should be left for follow-up work once the baseline test is reliable.

## Scope

- Add an integration test that prepares an isolated temporary Core work directory, copies a checked-in apply-load config, and runs `stellar-core new-db`, `stellar-core new-hist`, and `stellar-core apply-load`.
- Use a single initial profile that mixes classic payment transactions with Soroban transactions, likely using Core `APPLY_LOAD_MODE="ledger-limits"`, non-zero `APPLY_LOAD_CLASSIC_TXS_PER_LEDGER`, and a small Soroban transaction/resource count suitable for GitHub-hosted runners.
- Parse the pre-benchmark checkpoint emitted by Core and keep only the benchmark ledgers for assertions or downstream ingestion fixtures.
- Verify the generated metadata is usable by RPC's ingestion path or by the synthetic-ingestion harness, including ledger count and basic evidence that both classic and Soroban activity are present.
- Passing synthetic ledgers from Core is already done work in the [`load-testing`](https://github.com/stellar/stellar-rpc/tree/load-testing) branch, that should be ported as part of this.

## Acceptance criteria

- The test uses one checked-in apply-load config with deterministic mixed classic and Soroban parameters.
- All Core state is under a temporary directory and leaves no dependency on developer-local paths.
- The generated benchmark ledger range is isolated from Core setup ledgers by parsing the final pre-benchmark checkpoint.
- The test asserts the expected benchmark ledger count and fails on Core command errors, missing metadata, or malformed ledger output.
- The generated output can be passed to RPC ingestion or a clearly named follow-up step without requiring a live Core stream.
- Future parameter matrices are explicitly out of scope for this issue and tracked as follow-up work.

## Notes

- Horizon's [`generate_ledgers_test.go`](https://github.com/stellar/stellar-horizon/blob/master/internal/integration/generate_ledgers_test.go) integration test is the reference pattern: initialize Core state, run `apply-load`, parse the pre-benchmark checkpoint, then stream benchmark ledgers and optional fixtures.
- Keep the initial profile small and stable enough for GitHub-hosted runners; larger or varied profiles should be added after this repeatable baseline exists.
- The test should pin or record the Stellar Core version/image used so generated ledger metadata can be compared across releases.

cjonas9:
### What

This is a PR implementing a repeatable CI ingestion load test on a full database of 7 days of ledgers. The approximate design is here:
Screenshot 2026-05-14 at 12 59 14 PM

This GHA workflow for this test, currently, is triggered on pushes to this branch (`apply-load`), but will later be modified to trigger on any release or on PR comments stating "run load test".

The workflow benchmarks RPC ingestion end-to-end on an ephemeral c5.2xlarge: it launches the box, pulls a mainnet-scale golden DB (~307GB, 1-week retention window), a BUILD_TESTS stellar-core, and three apply-load ledger bundles from S3 (sha-verified). After the box downloads and decompresses this data, its gp3 volume is throttled to 125 MiB/s, ingests the bundles, and posts a per-profile results table to the run summary / PR.

Main Pieces:
- `integrationtest/ingest_loadtest_test.go::TestIngestSyntheticLedgers`: byte-concatenates N bundles into one continuous stream (the backend rebases ledger seqs per ledger, so per-bundle seq resets are harmless), ingests onto the golden DB with retention trimming live, verifies exact classic/soroban op counts via parallel `getTransactions` walkers, and reports per-profile wall-clock/ledgers-sec/ms-ledger/latency quantiles.
-  `loadtest/testdata/apply-load-v27-*-cfg`: config files specifying three O3 target tx profiles, 1,000 ledgers each: sac (1,000 soroban TPL), oz (900), soroswap (250). All generate these + 1,000 classic payments/ledger to create ledger bundles (for local usage or S3) offline by `stellar-core apply-load`.
- `.github/workflows/load-test.yml`: push-triggered orchestrator. OIDC-assumes into AWS, launches an ephemeral `c5.2xlarge` (Ubuntu 22.04, 500GB gp3) with the runner script as user-data (shipped verbatim, TARGET_SHA/RUN_ID passed via a two-line env preamble), waits for SSM registration, delegates polling to the script, writes the results table to the step summary (and PR comment when one exists), fails the job on a fail verdict or timeout, and **always** terminates the instance.
- `run-load-test.sh`: both halves of the run in one self-contained script, coordinated by a /tmp marker protocol.
    - instance (user-data on the box): installs the toolchain, streams the golden DB + BUILD_TESTS core + all bundles from S3, builds the repo at the target SHA, throttles root volume MiB/s, runs the test, and writes results.md plus an ok/fail verdict. 
    - `orchestrate` (on the GHA runner): polls the box over SSM, drives the gp3 downshift handshake (500 -> 125 MiB/s after downloads complete, so fetches are fast but the benchmark runs on throttled I/O), and relays verdict + results as step outputs.

### Why

CI testing of RPC ingestion performance; benchmarking. This also serves as an automated regression testing framework, though future work should expand this to report some metric that allows one to compare a run's results to historical results.

### Known limitations

This is purely intended as a test of RPC's ingestion pipeline and seeks to see how it handles load in isolation (i.e. without captive core running). Future work should also seek to automatically refresh the S3 DB + ledger bundles on some pre-determined cadence.


cmd/stellar-rpc/internal/integrationtest/generate_ledgers_test.go
```diff
@@ -0,0 +1,594 @@
...
+
+	// Poll getHealth at 25ms granularity, recording the first time each
+	// sequence is observed for per-ledger latency stats.
+	arrivals := make(map[uint32]time.Time, cfg.NumSynthetic+1)
+	arrivals[startSeq-1] = startedAt // synthetic ingestion "began" at startedAt
+	ingestDeadline := 30 * time.Minute
+	deadline := time.After(ingestDeadline)
+	tick := time.NewTicker(25 * time.Millisecond)
+	defer tick.Stop()
+	latestSeen := startSeq - 1
```
leighmcculloch: **Misleading sentinel: `latestSeen` initialises to `0` for a fresh DB, so a deadline-timeout message reads "ingested through ledger 0" rather than "no ledgers seen from the expected range".**

With `startSeq = 1` (empty DB), `latestSeen = 0`. If the 30-minute deadline fires before any synthetic ledger lands:

```
t.Fatalf("RPC only ingested through ledger 0; wanted 1000 within 30m0s")
```

"Ledger 0" is not a real sequence; a developer debugging this CI timeout would look for ledger 0 in the DB and find nothing, which is confusing.

Suggestion — distinguish "nothing seen" from "partial ingestion":

```go
if latestSeen == startSeq-1 {
    t.Fatalf("RPC ingested no synthetic ledgers; expected range [%d, %d] within %s", startSeq, endSeq, ingestDeadline)
} else {
    t.Fatalf("RPC stalled at ledger %d; wanted %d within %s", latestSeen, endSeq, ingestDeadline)
}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
```diff
@@ -0,0 +1,344 @@
...
+  rm -rf "$WORK_DIR/stellar-rpc"
+
+  log "installing deps"
+  export DEBIAN_FRONTEND=noninteractive
+  apt-get update -qq
+  apt-get install -y -qq --no-install-recommends \
+    zstd awscli jq curl git build-essential ca-certificates \
+    libpq5 libsodium23 libunwind8 libc++1-14
+
+  GO_VERSION=1.22.7
```
marwen-abid: ❓ Shouldn't this be `1.25` ?

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,777 @@
...
+	"github.com/stellar/go-stellar-sdk/xdr"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/config"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/db"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/integrationtest/infrastructure"
+)
+
+const (
+	defaultLedgerBundlePath    = "./infrastructure/testdata/load-test-ledgers-v25.xdr.zstd"
+	defaultApplyLoadConfigPath = "./infrastructure/load-test/testdata/apply-load.cfg"
```
marwen-abid: ❓ Both defaults above don't seem to point to real files. Should we switch them to one of the v27 files or delete the defaults and require the associated env vars ?

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,777 @@
...
+				continue
+			}
+			// Arrivals are recorded contiguously, so exactly the sequences
+			// past latestSeen are new.
+			seen := min(health.LatestLedger, endSeq)
+			for seq := latestSeen + 1; seq <= seen; seq++ {
+				arrivals[seq] = now
+			}
+			latestSeen = max(latestSeen, seen)
+			if health.LatestLedger >= endSeq {
```
marwen-abid: ⛏️ it looks like `waitForIngest` can hang for the full deadline (150m) if the cfg's ledger count drifts from the bundle. 

`endSeq` is derived from cfg, but loadtests reachable latest is derived from bundle. Given that `waitForIngest` only exists on `LatestLedger >= endSeq`, this daemon may plateau and hang (the way I read it). It may be worth bailing on a stall if this is possible.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,777 @@
...
+		startSeq:           startSeq,
+		segments:           prof.segments,
+		captiveCoreVersion: versionInfo.CaptiveCoreVersion,
+	})
+}
+
+// waitForIngest polls getHealth at 25ms granularity until endSeq has been
+// ingested (or LOADTEST_INGEST_DEADLINE, default 30m, passes), recording the
+// first time each sequence is observed for per-ledger latency stats.
+func waitForIngest(t *testing.T, client *rpcclient.Client, startSeq, endSeq uint32) map[uint32]time.Time {
```
marwen-abid: ❓ We already are collecting some good prometheus metrics on ingestion (cmd/stellar-rpc/internal/ingest/service.go:51). Is there a reason why we're not using them ? 

I can see that they don't cover the full spectrum of what we want to measure, but it may be worth considering leveraging them for some parts, or augmenting them for more accurate measurements. 

When I asked Claude if this is possible, it said: 

>  Add an End-of-run Gather() to to get *mean* ms/ledger + throughput (_sum/_count), and recent p50/p90/p99 (quantiles = last 10 min only). 

```go
func ingestTotals(t *testing.T, d *daemon.Daemon) (sumSec float64, count uint64, q map[float64]float64) {
	mfs, err := d.MetricsRegistry().Gather()
	require.NoError(t, err)
	for _, mf := range mfs {
		if mf.GetName() != "soroban_rpc_ingest_ledger_ingestion_duration_seconds" {
			continue
		}
		for _, m := range mf.GetMetric() {
			if !hasLabel(m, "type", "total") {
				continue
			}
			s := m.GetSummary()
			sumSec, count = s.GetSampleSum(), s.GetSampleCount()
			q = map[float64]float64{}
			for _, qq := range s.GetQuantile() {
				q[qq.GetQuantile()] = qq.GetValue()
			}
		}
	}
	return // mean ms = sumSec/float64(count)*1000 ; ledgers/sec = float64(count)/sumSec
}
```

marwen-abid (commented):
This looks great! I love the design of this solution and learned a lot from reading this PR. 

I have a couple of ❓ questions and a ⛏️  nit (probably all due to my lack of understanding of how RPC works).

.github/workflows/e2e.yml
```diff
@@ -11,9 +11,13 @@ concurrency:
...
 jobs:
   run-system-test:
+    # Temporary branch used to iterate on the EC2-backed load-test workflow.
+    # The reusable system-test workflow is currently rejected for this PR path,
+    # so skip e2e here instead of failing every push to apply-load.
+    if: github.event_name != 'pull_request' || github.head_ref != 'apply-load'
     # set the git ref of stellar/system-test after the '@'
     # to specify which version of the workflow to call
-    uses: stellar/system-test/.github/workflows/test-workflow.yml@master
+    uses: stellar/system-test/.github/workflows/test-workflow.yml@789121c0914150a02122581f32ee62c4e42e1c84
```
tamirms: what's the motivation for this change?

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,234 @@
+name: Load test (ephemeral)
+# Launches a c5.2xlarge in Horizon (203618453975), polls it via SSM, posts
```
tamirms: what is Horizon (203618453975) ? what is SSM ?

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
```diff
@@ -0,0 +1,344 @@
...
+    ssm_touch "/tmp/volume-throttle-$1" && THROTTLE_SIGNALLED=1
+  }
+  reconcile_throttle() {
+    [ "$THROTTLE_SIGNALLED" -eq 1 ] && return 0
+    if [ -z "$ROOT_VOLUME_ID" ]; then
+      signal_throttle failed
+      return 0
+    fi
+    if [ "$THROTTLE_ATTEMPTED" -eq 0 ]; then
+      if aws ec2 modify-volume --region "$AWS_REGION" --volume-id "$ROOT_VOLUME_ID" \
```
tamirms: are we sure that modifying volume throughput is instantaneous ? how do we know there isn't some initial period where the throughput is not matching what we requested?

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,777 @@
...
+	require.NoError(t, sdb.Close())
+	require.NoError(t, err)
+
+	i := infrastructure.NewTest(t, &infrastructure.TestConfig{
+		NetworkPassphrase:      prof.networkPassphrase,
+		SQLitePath:             sqlitePath,
+		HistoryRetentionWindow: initialCount,
+		LoadTest: config.LoadTestConfig{
+			File:      ledgerPath,
+			Frequency: 1 * time.Millisecond, // frequency with which we emit ledgers for ingestion
```
tamirms: I think the frequency and the number of ledgers we ingest should be configurable. I understand you're setting it to 1 ms to see the fastest possible close time rpc could sustain. another use case is to see what the ingestion duration is if the ledger close time is at 5 seconds or 2 seconds or some other configurable amount.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,777 @@
...
+			}
+		}
+	}
+}
+
+// countIngestedOps verifies the ingested range by paginating every transaction
+// through getTransactions, returning the total classic Payment and Soroban
+// InvokeHostFunction op counts. At millions of txs and 200 txs/page a single
+// cursor chain takes hours, so the range is split across parallel walkers.
+func countIngestedOps(t *testing.T, client *rpcclient.Client, startSeq, endSeq uint32) (int, int) {
```
tamirms: I wonder if it is necessary to verify every single ledger was ingested in this manner? could we instead check a random sample of transactions to see if they were ingested in the db? you can also do a sql query against the sqlite db to see that there are no gaps in the ledgers table and that the number of ingested ledgers matches our expectations

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
```diff
@@ -0,0 +1,344 @@
+#!/usr/bin/env bash
```
tamirms, line 1: we require go to be installed to run rpc and the ingestion load test. so I wonder if most of the logic in this bash script could live in a go file. I think that it would be easier to understand and maintain a go script than a large bash script.

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
cjonas9: Yes, good catch!

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
cjonas9: Yes, definitely. There's definitely some required shell, but cramming it all into one shell script is super excessive and messy (though I did like that it kept everything for the instance held in one place as user data). I'll see about refactoring some of this out

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
cjonas9: Great question, this really goes hand-in-hand with your comment about readability of this script: modifying volume throughput is indeed asynchronous. But this function runs in a loop around line 96 until it either times out (set to 900s, which is generous but not absurd) or properly downshifts

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
cjonas9: Great catch, moved the default config/bundle to SAC

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
cjonas9: will do, good point!

.github/workflows/e2e.yml
cjonas9: I'm glad you caught this and I'm going to leave this thread open so I remember to fix it, but if I remember correctly, this had to do with some weirdness in trying to get the GHA to trigger without anything being merged to main. I'm near-certain this won't end up being here after all PR review changes are made and this is verified as working

.github/workflows/load-test.yml
cjonas9: These are some AWS details, the former is the account ID for profile Horizon, the latter is AWS Systems Manager. I figured that info could be helpful to someone maintaining this (and for their agent that's scanning this and needs to know that)

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/run-load-test.sh
tamirms: but how do we know that the downshift is fully in effect? Is it possible that that the request to the AWS backend succeeds but it takes some while for the change to be applied to the drive?

cjonas9: Wow excellent point, I just learned something new: my initial answer was that we check DescribeVolumes.Throughput to ensure it's in effect, but as it turns out, that's insufficient. Beyond checking that, we **also** need to key on `DescribeVolumesModifications` -> `ModificationState == "completed"` (and `Progress == 100`). That's the robust check and I should've caught that sooner, thanks for the thoughtfulness on that!!

Also, to be clear on this, even with the speed as high as it is (at 500 MiB/s), this is still a pretty big bottleneck and the default write speeds would make it 3x worse. The download isn't very bad, but the memory write speed is slow for a 300Gb DB, which is why I would like to see this work. If you're very convinced it's not worth it, I am happy to listen nonetheless

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/refresh/refresh-tool
tamirms, line 1: we shouldn't need to include binaries in the git repo

cjonas9: Sorry, yes, these are some untracked files I've been keeping around for a future PR on the monthly refresh workflow, they just got accidentally added to the last commit

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,485 @@
...
+	i := infrastructure.NewTest(t, &infrastructure.TestConfig{
+		NetworkPassphrase:      cfg.NetworkPassphrase,
+		SQLitePath:             sqlitePath,
+		HistoryRetentionWindow: initialCount,
+		LoadTest: config.LoadTestConfig{
+			File:      ledgerPath,
+			Frequency: 1 * time.Millisecond, // frequency with which we emit ledgers for ingestion
+		},
+	})
+	startedAt := time.Now().UTC()
```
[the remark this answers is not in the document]
cjonas9: Note that this value is captured in the final JSON but isn't rendered or used to derive anything about latency/throughput, it's intended to capture literally when the test started.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
cjonas9: Really good question! I actually wasn't fully up on the metrics that we exposed, but having looked into them, I'm not entirely sure I think that this is totally worth it. TLDR upfront: we could, but this would involve handling these metrics as well as the custom ones our tests gathers for an ambiguous payoff.

As your Claude alluded to, the percentiles reported here reflect the last 10 minutes of data points in a sliding window. This would give us an approximation, but the whole ingesting section takes about 40 minutes. Secondly, these report p50/90/99, and generally, I think p95 is worth including over p90 (I'm not sure of how adding it to the metrics handler would affect the upstream).

From what I can tell, also, this does give us something we don't already have: pure/isolated DB-write time. No doubt it has its uses (as it would let us break our wall clock measurements into an exact overhead vs. write time disambiguation), but for the effort + considering this still would be a 10-minute windowed measurement, I feel ok leaving it out unless it's decided it should be included.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
cjonas9: Great catch! added in a 3m stall timeout, considering whether or not it's worth it to see about bypassing this being derived from the config altogether.
Update: having thought about this a little more, I'm leaning towards leaving it as-is. It could rederive this straight from the bundles, but realistically, I think that for now, it serves as a smoke-check that one's config matches the bundles it's supposed to correspond to.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,595 @@
...
+
+	ingestDeadline := 30 * time.Minute
+	if v := os.Getenv("LOADTEST_INGEST_DEADLINE"); v != "" {
+		var err error
+		ingestDeadline, err = time.ParseDuration(v)
+		require.NoError(t, err, "invalid LOADTEST_INGEST_DEADLINE")
+	}
+	arrivals := make(map[uint32]time.Time, endSeq-startSeq+1)
+	deadline := time.After(ingestDeadline)
+	tick := time.NewTicker(25 * time.Millisecond)
```
tamirms: The per-ledger timing here is reconstructed by polling `getHealth` at 25ms granularity, but the daemon already measures this exactly when it records `ledger_ingestion_duration_seconds`. Since the load test runs the daemon in-process, we could capture the ground truth directly instead of inferring it: add an optional `OnLedgerIngested(seq uint32, d time.Duration)` hook to `ingest.Config` (nil in production, same shape as `OnIngestionRetry`), fire it where the duration is observed, and have the harness record each `(seq, duration)`. That gives exact per-ledger durations with no 25ms quantization, any quantile, and a clean per-segment split — with no change to the production metric. It would also drop the startup back-stamping bias (the first observed ledgers currently all share the first poll's timestamp) and make the quantiles exact rather than nearest-rank. Happy to sketch the wiring.

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/runner/orchestrate.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	in := &ssm.GetCommandInvocationInput{CommandId: &id, InstanceId: &r.instanceID}
+	_ = ssm.NewCommandExecutedWaiter(r.client).Wait(ctx, in, commandWaitTimeout)
+	inv, err := r.client.GetCommandInvocation(ctx, in)
+	if err != nil {
+		// Unreadable result is "not ready", not a dispatch failure.
+		return "", nil //nolint:nilerr
+	}
+	return aws.ToString(inv.StandardOutputContent), nil
```
tamirms, line 184: Results come back inline via `cat /tmp/done /tmp/results.md` over SSM, and `GetCommandInvocation.StandardOutputContent` truncates (~2,500 chars). The current table is already in that ballpark with three profiles plus the metrics table, so a slightly larger body or the debug tail could be silently cut and corrupt the PR comment. Since the box already has S3 access, one option is to have it `PutObject` a structured `result.json` (verdict + raw bench JSON + rendered markdown) under a `runs/<run-id>/` prefix and have `orchestrate` poll that key, instead of shelling over SSM and parsing stdout sentinels. That removes the size limit, the sentinel parsing, and the `/tmp` marker protocol. Relatedly, `/tmp/download-complete` is now vestigial — nothing acts on it since the throttle handshake was removed, and its comment still reads "benchmark running" — so it's worth removing either way. Would need `PutObject` on the instance profile.

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/runner/instantiate.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// runStreaming runs name in dir (with extra env appended), streaming combined
+// output to our log; on failure the error carries the last tailN lines so the
+// verdict explains what broke.
+func runStreaming(ctx context.Context, dir string, env []string, tailN int, name string, args ...string) error {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(), env...)
+	var buf strings.Builder
```
tamirms: `runStreaming` buffers the entire child output in this `strings.Builder` only to print the last `tailN` lines on failure. For the `go test -v` run, which can stream verbose output for up to ~170m on a 16GB box that's also running the ingesting daemon, this is an unbounded allocation. Since the bootstrap already tees everything to `/var/log/user-data.log`, the simplest fix is to skip the in-memory buffer and read the tail of that file on failure (a bounded ring buffer would also work if you'd rather not depend on the file).

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,213 @@
...
+
+      - name: Launch EC2 instance
+        id: launch
+        run: |
+          COMMON_TAGS="{Key=$TEST_TAG_KEY,Value=$TEST_TAG_VAL},
+            {Key=pr,Value=${{ steps.target.outputs.pr_tag_value }}},
+            {Key=ref,Value=${{ github.ref_name }}},
+            {Key=sha,Value=${{ github.sha }}},
+            {Key=run-id,Value=${{ github.run_id }}}"
+          RUN_INSTANCES_JSON=$(aws ec2 run-instances \
```
tamirms, line 104: One safety net worth adding: the job terminates the instance with `if: always()`, but a force-cancelled or crashed runner skips that step and would leave a `c5.2xlarge` + 500GB volume billing until someone notices. Setting `--instance-initiated-shutdown-behavior terminate` on this `run-instances` call plus a `shutdown -P +<ttl>` at the top of `run-load-test.sh` gives a hard self-terminate ceiling independent of the runner, with no extra IAM.

go.mod
```diff
@@ -98,7 +100,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/jmoiron/sqlx v1.3.5 // indirect
-	github.com/klauspost/compress v1.18.1 // indirect
+	github.com/klauspost/compress v1.18.1
```
tamirms, line 103: `klauspost/compress` is imported directly now (in `instantiate.go`) but it's still in the indirect block with just the `// indirect` comment stripped. A `go mod tidy` would move it to the direct-require block — worth running so the tidy check stays green.

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
```diff
@@ -0,0 +1,595 @@
...
+	sdb, err := db.OpenSQLiteDB(sqlitePath)
+	require.NoError(t, err)
+	preTestLast, initialCount, err := getLedgerBounds(t.Context(), sdb)
+	require.NoError(t, sdb.Close())
+	require.NoError(t, err)
+
+	i := infrastructure.NewTest(t, &infrastructure.TestConfig{
+		NetworkPassphrase:      prof.networkPassphrase,
+		SQLitePath:             sqlitePath,
+		HistoryRetentionWindow: initialCount,
```
tamirms, line 100: Setting the retention window to `initialCount` keeps the DB the same size through the run, which is the right call for the large golden DB. One edge case: if a seed DB ever has `0 < initialCount < totalLedgers`, trimming drops ledgers inside `[startSeq, endSeq]` and `verifyLedgerRange` then fails in a way that's hard to trace. A short guard or comment noting the assumption would save a future debugging session.

.github/workflows/load-test.yml
cjonas9: Amazing catch, I'd tried to be wary of orphaned boxes, but didn't know about this option as a final backstop.

cmd/stellar-rpc/internal/integrationtest/infrastructure/load-test/runner/instantiate.go
cjonas9: Good catch, there was never a good justification for buffering all of this other than the debugging I was doing earlier on. With that in mind, having wrestled with this a bit, I do think I still prefer flushing from some in-memory buffer as it doesn't deal with the same races that reading from the tail of a file we're teeing to. I've turned this into in a small ring buffer to avoid that, but if you truly prefer reading from `/var/log/user-data.log`, let me know (I could really see this going either way)

tamirms: > I do think I still prefer flushing from some in-memory buffer 

sure, that approach is fine too!

cmd/stellar-rpc/internal/integrationtest/ingest_loadtest_test.go
cjonas9: I've had to iron out some implementation details for this, but since it's a great idea and major simplification, I've attempted to work this out.

There's two notable kinks that showed up here:
- The OnLedgerIngested(seq uint32, d time.Duration) hook needed to be visible to the daemon in MustNew(...) because that both creates the daemon and starts ingestion. There were a few choices here:
    - add variadic argument to MustNew(...)
    - add MustNewWithoutIngest(...) that MustNew(...) aliases for then starts ingestion
    - carry the hook in the ingest load test config
Of these, I went with the latter. While there's definitely a good argument to be made that you shouldn't carry a function in a config (especially given that it's not configurable), the other two touched the daemon code in a way that felt somewhat invasive for what we gain out of this. Between the other two, I'd likely have preferred some function that lets one create the daemon without starting ingest as that could have utility elsewhere, but that's something for another day.
- To my understanding, there is a difference between the metric displaying when a ledger becomes visible to a poller and the delta between the time the ledger is received/committed. We now have the latter number, through this implementation, but since we're not polling anymore, we're not measuring the former at all (which is probably best for an ingestion-pipeline load test).

Let me know what you end up thinking of all of this!

tamirms (approved):

PR #742, opened 2026-05-18
fix(simulate): Use core ledger close time

jia57b:
### Summary

Use the latest ingested core ledger close time for preflight simulation instead of the local wall-clock time.

This change aligns simulateTransaction with the actual ledger timestamp observed by stellar-core, which is especially important for tests and workflows that manipulate ledger time
(for example via forceclosetime). Previously, simulation used time.Now(), which could drift from the real chain time and produce inconsistent results between simulation and on-chain execution.

### What changed

- Added GetLatestLedgerTime(ctx) to the LedgerReader interface.
- Implemented latest-ledger-close-time retrieval using the existing DB cache / ledger range data.
- Threaded the latest ledger close time through the preflight worker pool and preflight parameter structs.
- Updated preflight ledger info construction to use the ingested core ledger close time instead of the host system clock.
- Updated mocks and unit tests to support the new ledger-time lookup path.

### Why

simulateTransaction should execute against the same time context as the latest ledger known to RPC, not the machine's current time. Using wall-clock time can make simulation:

- disagree with core when ledger time has been forced or drifted
- produce timestamps that do not match submitted transaction behavior
- cause flaky integration tests around ledger-time-sensitive contracts

With this change, simulation uses the same ledger-close-time source that RPC ingests from core, making simulation results deterministic and consistent with chain state.

### Testing

- Updated unit tests 
- Run unit tests: go test ./cmd/stellar-rpc/internal/methods/... ./cmd/stellar-rpc/internal/preflight/... ./cmd/stellar-rpc/internal/db/...


Shaptic (approved):
This is great, thank you! Would you mind just adding a line under a new `### Fixed` header in the `Unreleased` section of the changelog as a quick release note?

leighmcculloch:
In addition to the ask about by @Shaptic the lints appear to be failing and would need resolving before merging.

cmd/stellar-rpc/internal/preflight/preflight_test.go
```diff
@@ -249,12 +249,31 @@ func getPreflightParameters(t testing.TB, protocolVersion uint32) Parameters {
...
+		ProtocolVersion:   25,
+		BucketListSize:    200,
+	}
+
+	ledgerInfo := getLedgerInfo(params)
+	defer freeLedgerInfo(ledgerInfo)
+
+	require.Equal(t, uint64(params.LedgerTime), uint64(ledgerInfo.timestamp))
+	require.Equal(t, uint32(params.LedgerSeq), uint32(ledgerInfo.sequence_number))
+	require.Equal(t, uint32(params.ProtocolVersion), uint32(ledgerInfo.protocol_version))
```
Shaptic: Just gonna fix the CI errors inline

```suggestion
	require.EqualValues(t, params.LedgerTime, ledgerInfo.timestamp)
	require.EqualValues(t, params.LedgerSeq, ledgerInfo.sequence_number)
	require.EqualValues(t, params.ProtocolVersion, ledgerInfo.protocol_version)
```

cmd/stellar-rpc/internal/methods/simulate_transaction_test.go
```diff
@@ -189,6 +206,367 @@ func TestSimulateTransactionFeeBumpMissingSorobanData(t *testing.T) {
...
+	}
+
+	return xdr.TransactionEnvelope{
+		Type: xdr.EnvelopeTypeEnvelopeTypeTx,
+		V1: &xdr.TransactionV1Envelope{
+			Tx: tx,
+		},
+	}
+}
+
```
Shaptic: This is a dupe so we can drop it I think?
```suggestion
```

cmd/stellar-rpc/internal/preflight/preflight_test.go
```diff
@@ -249,12 +249,31 @@ func getPreflightParameters(t testing.TB, protocolVersion uint32) Parameters {
...
+		BucketListSize:    200,
+	}
+
+	ledgerInfo := getLedgerInfo(params)
+	defer freeLedgerInfo(ledgerInfo)
+
+	require.EqualValues(t, params.LedgerTime, ledgerInfo.timestamp)
+	require.EqualValues(t, params.LedgerSeq, ledgerInfo.sequence_number)
+	require.EqualValues(t, params.ProtocolVersion, ledgerInfo.protocol_version)
+	require.Equal(t, uint64(params.BucketListSize), uint64(ledgerInfo.bucket_list_size))
```
Shaptic:
```suggestion
	require.EqualValues(t, params.BucketListSize, ledgerInfo.bucket_list_size)
```

PR #755, opened 2026-05-28
ledger: split cold store into reader + writer; rework hot store

tamirms:
## Summary

**Stacked on #739** (base is `karthik/backfill-impl/ledger-cold-store`; GitHub will retarget to `feature/full-history` once #739 merges). This shows only the incremental evolution of the cold ledger store since #739.

Part of the series extracting reviewable components from the `rpc-hack` branch. This PR also folds in the shared `stores` error sentinels (no separate sentinels PR), since the ledger store is their first consumer.

Beyond moving `cold_store.go` into `cold_reader.go` + `cold_writer.go`, the store APIs and behavior change:

### Cold writer (`ColdStoreWriter` -> `ColdWriter`)
- `NewColdWriter` takes a new `ColdWriterOptions{Concurrency, BytesPerSync}` — enables **parallel zstd encoding** across packfile worker goroutines for large backfills, plus fsync-cadence control. The #739 constructor was serial-only.
- Dropped the `logger` parameter.

### Cold reader (`ColdStoreReader` -> `ColdReader`)
- **Lazy open**: `OpenColdReader` does no I/O; the trailer read + format check are deferred to the first method call (was eager in #739).
- **API change**: `FirstSeq()`/`LastSeq()` now return `(uint32, error)` instead of `uint32`, since the lazy header load can fail at first access.
- Dropped the caller-injected `*zstd.Decompressor`; uses a package-level shared `coldPackDecoder`.

### Hot store
- **zstd compression internalized**: `Entry` now holds uncompressed bytes; the store compresses on write / decompresses on read via a per-store `sync.Pool` of compressors (so concurrent `AddLedgers` don't contend). Previously the caller compressed before `AddLedgers` and decompressed after `GetLedgerRaw`.
- `NewHotStore` -> `OpenHotStore`; `AddLedgers([]Entry)` -> variadic `AddLedgers(...Entry)`; documented the `Close` concurrency contract.

### Shared
- New `stores.ErrStoreClosed` / `ErrInvalidConfig` / `ErrCorrupt`; both stores translate backing-primitive errors into these at their public-method boundaries.

## Reviewer notes
- The hot-store zstd move and the lazy reader's `FirstSeq`/`LastSeq` signature change are the two things worth the closest look.

## Test plan

- [x] `go build ./cmd/stellar-rpc/internal/fullhistory/pkg/stores/...`
- [x] `go test ./cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/`
- [x] Resulting tree is byte-identical to the corresponding paths on `rpc-hack`


cmd/stellar-rpc/internal/fullhistory/pkg/stores/ledger/cold_reader.go
```diff
@@ -0,0 +1,185 @@
...
+		// count calc below.
+		if start > end || end < h.firstSeq || start > h.lastSeq {
+			return
+		}
+		if start < h.firstSeq {
+			start = h.firstSeq
+		}
+		if end > h.lastSeq {
+			end = h.lastSeq
+		}
```
chowbao, line 157: Just wanted to double check. Is this range logic is what we expect/want? 

I thought in RPC v1 if a startLedger is < what is available it will error out.

I think as a user it'd be a little odd to get a returned ledger range that was a subset of what I requested without any notification/warning/error

Also it seems like [GetLedgerRaw](https://github.com/stellar/stellar-rpc/pull/755/changes#diff-dcc79772b283c2b1a04f0d490b88cba0c49e43f2cb1a760522de6600480438fdR116) returns an error if the seq is outside the cold header so maybe it's more consistent for both to error out if any part of the seq, start, end are outside the coldheader

tamirms: good catch, I just pushed a fix to validate the ledger range

chowbao (approved):

PR #760, opened 2026-06-04
bench(fullhistory): make apply-load synthetic LCM usable by the read benches at target TPS

chowbao:
## 📊 Results report (addresses #762)

Benchmark report committed at **`cmd/stellar-rpc/scripts/bench-fullhistory/results/2026-06-09-synthetic-apply-load.md`** — the three synthetic datasets (sac 10k / token 9k / soroswap 2.5k TPS @ 600 ms) run through the full read + ingest suite on c6id.8xlarge, with a comparison to the pubnet chunk-5860 baseline.

Datasets + configs + CSVs + machine-readable `RESULTS.md` are in GCS at `gs://rpc-full-history/synthetic-ledgers/2026-06-04-apply-load-20k/`.

---

## Summary

Addresses **#762** — a *controllable* synthetic-ledger dataset whose transaction profile we set deliberately, feeding the existing `cold-*` / `hot-*` ingest + query benches. Builds on the apply-load driver from the parent branch (`a8c82958`).

It delivers the issue's acceptance criteria via the **`--source=lcm`** path: `apply-load-gen.sh` runs stellar-core `apply-load` to emit a framed `LedgerCloseMeta` stream, and `cold-ingest --source=lcm` turns it into the cold packfiles (+ hot store) the query benches read — unchanged. This is used instead of the proposed `--source=synthetic` + `ingest/loadtest.ApplyLoad` wiring because the pinned `go-stellar-sdk` doesn't yet include `ingest/loadtest` (stellar/go-stellar-sdk#5940), so `--source=lcm` reaches the same goal with **no dependency bump**. Swapping in `loadtest.ApplyLoad` later is a drop-in producer change; the ingest/query path is unaffected.

## apply-load config

`apply-load-gen.sh` writes the upstream `docs/apply-load-benchmark-*.cfg` shape **plus meta output**, parameterized per profile:

- `APPLY_LOAD_MODE="benchmark"`, `APPLY_LOAD_MODEL_TX=sac|custom_token|soroswap`
- `APPLY_LOAD_MAX_SOROBAN_TX_COUNT` = txs-per-ledger (the density knob)
- `APPLY_LOAD_LEDGER_MAX_DEPENDENT_TX_CLUSTERS` = `CLUSTERS` (default **8**) — parallel apply threads, a generation-speed knob only; capped at 8 (known multi-threaded-apply perf issues above that)
- `APPLY_LOAD_NUM_LEDGERS` = ledgers to close after setup (`NUM_LEDGERS`)
- `METADATA_OUTPUT_STREAM` + `DISABLE_TX_META_FOR_TESTING=false`, BL pre-gen disabled, `GENESIS_TEST_ACCOUNT_COUNT` ≥ 2× TPL, quorum/history boilerplate.

## Target load shapes (10k SAC / 9k OZ / 2.5k Soroswap @ 600 ms blocks)

TPS is taken at the network's **600 ms** block time (`CLOSE_TIME_MS`), so per-ledger tx count = `TPS × 0.6`. (The ledger header `closeTime` is whole seconds in XDR, so the sub-second cadence is modeled by density, not timestamps.)

| profile | model tx | target | txs/ledger @600ms |
|---|---|---|---|
| sac | `sac` | 10 000 TPS | **6 000** |
| token/OZ | `custom_token` | 9 000 TPS | **5 400** |
| soroswap | `soroswap` | 2 500 TPS | **1 500** |

A small `NUM_LEDGERS` is enough — TPS is set by density, not ledger count.

## What it took

1. **`apply-load-gen.sh` path + pubnet passphrase** — absolutize `OUT_ROOT`; default `NETWORK_PASSPHRASE` to pubnet (the bench binary hardcodes it).
2. **Make the synthetic LCM consumable (`lcm_fixup.go`)** — apply-load's streamed meta holds the same txs in the tx-set and `TxProcessing` but in different order, and the stored result hash matches no envelope under any passphrase, so the SDK ingest reader (pairs by hash) rejected it with `unknown tx hash in LedgerCloseMeta`, breaking roundtrip `cold-txpage`/`cold-txhash`. We repair it: each result's unique fee-charged account identifies its envelope, so we stamp the real hash (a correct pairing). Default-on (`--lcm-fix-tx-hashes`). Plus partial final chunk (`--lcm-allow-partial`) + `cold-ledgers`/`cold-txhash` cursor clamping, so runs below a full 10k-ledger chunk work; `NUM_LEDGERS` knob.
3. **Single-contract events corpus (`corpus.go`)** — relaxed the ≥3-contract floor to "enough filterable terms (anchors + topic values ≥ max K)"; a single SAC contract's `transfer` events over many accounts give the needed 15 terms. Pubnet behaviour unchanged.
4. **`sac` BATCH_SAC=1** — `BATCH_SAC>1` folds transfers into one tx and only that tx is streamed, so the pack carried ~1 tx/ledger; `=1` makes each transfer its own tx so density equals the TPS target.
5. **600 ms block model (`CLOSE_TIME_MS`, default 600)** and **`CLUSTERS` default 8** for generation speed.

## Verification — generated all 3 profiles (100 ledgers each, CLUSTERS=8) and decoded the packs

| profile | tx/ledger (decoded) | TPS @600ms | tx-hash fixup | cold-ledgers | cold-txpage | cold-txhash | cold-events |
|---|---|---|---|---|---|---|---|
| sac | 6 000 | 10 000 | 600022 / 600022 | ✅ | ✅ | ✅ miss-rate 0 | ✅ |
| token/OZ | 5 400 | 9 000 | 540028 / 540028 | ✅ | ✅ | ✅ miss-rate 0 | n/a |
| soroswap | 1 500 | 2 500 | 164445 / 164445 | ✅ | ✅ | ✅ miss-rate 0 | ✅ |

Fixup paired 100% (0 skipped) on all three; read benches run 0-errors / miss-rate 0.

## Known limitation

`cold-events` is **not** supported for `token`/`custom_token` — its events are not 4-topic (a workload property). Use `sac` or `soroswap` for event benches.

---
**Base note:** targeted at `rpc-hack`, which doesn't yet contain `a8c82958` nor the 2026-06-03 bench-report commits, so the diff also carries those preceding `bench(fullhistory)` commits + a merge of `rpc-hack` (conflicts were stale report docs + a txpage helper, all resolved to rpc-hack; the apply-load files merged cleanly). The substantive new work is the apply-load commits above.


cmd/stellar-rpc/scripts/bench-fullhistory/lcm_fixup.go
```diff
@@ -0,0 +1,168 @@
+package main
```
tamirms, line 1: this is no longer necessary now that https://github.com/stellar/stellar-core/pull/5319 has been merged. Even without that fix in stellar-core, the better workaround is to just configure the network passphrase in the benchmark commands to be "Apply Load"

cmd/stellar-rpc/scripts/bench-fullhistory/sources.go
```diff
@@ -26,8 +29,38 @@ import (
 const (
 	sourcePack = "pack"
 	sourceBSB  = "bsb"
+	sourceLCM  = "lcm"
```
tamirms, line 32: instead of introducing a new source, I think it would be better to make have a post processing step after apply load which takes the generated ledgers and then converts them into the pack format. then we can run the benchmarks using the source as pack pointing to the synthetic ledgers packfile

PR #767, opened 2026-06-06
ledger,rocksdb: hot-store FirstSeq/LastSeq range accessors

tamirms:
## Summary

Adds `HotStore.FirstSeq()` / `LastSeq()` to the hot ledger store so a caller can learn the store's covered ledger range without an external chunk hint — the cold reader (`ColdReader.FirstSeq`/`LastSeq`) already exposes the equivalents, so this restores symmetry.

They delegate to two new `rocksdb.Store` primitives:

- `Store.FirstKey(cf)` / `LastKey(cf)` — the smallest / largest key in a CF (or `ok=false` if the CF is empty), via a single `SeekToFirst`/`SeekToLast` boundary seek under the store's `RLock`. No scan.

`edgeKey` copies the boundary key out before the iterator closes (matching the existing `Get`/`Iterate` ownership pattern); `edgeSeq` decodes it with `DecodeUint32`.

Ported from `rpc-hack`, with the documentary named returns dropped and a `//nolint:funcorder` on the unexported helpers (kept adjacent to their callers) to satisfy this branch's lint.

A second commit modernizes the six C-buffer copy sites in `rocksdb.go` (`Get`, `BatchMultiGet`, `Iterate`, `IterateRange` ×2, and the new `FirstKey`/`LastKey`) from `append([]byte(nil), x...)` to `bytes.Clone` so the new accessors match their siblings — equivalent, since `bytes` is already imported and the nil-vs-empty difference for a zero-length source is unobservable here.

## Test plan

- `TestStore_FirstLastKey`: empty CF (`ok=false`); out-of-order inserts → min/max (big-endian key order == numeric); non-default CF resolution; unknown CF → `ErrCFNotFound`.
- `TestStore_OpsAfterCloseFailWithErrStoreClosed`: adds `FirstKey`/`LastKey` (post-close → `ErrStoreClosed`).
- `TestHotStore_FirstLastSeq`: empty store (`ok=false`); out-of-order `AddLedgers` → min/max.
- `go build` / `go vet` / `golangci-lint` clean on the `rocksdb` and `ledger` packages.

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
+++ b/cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
@@ -292,6 +292,46 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
 	}
 }
 
+// FirstKey returns the smallest key in cf, or ok=false if the CF is empty.
+// Cheap: a single boundary seek (no scan).
+func (s *Store) FirstKey(cf string) ([]byte, bool, error) {
+	return s.edgeKey(cf, false)
+}
+
+// LastKey returns the largest key in cf, or ok=false if the CF is empty.
+// Cheap: a single boundary seek (no scan).
+func (s *Store) LastKey(cf string) ([]byte, bool, error) {
+	return s.edgeKey(cf, true)
+}
+
+//nolint:funcorder // helper grouped with FirstKey/LastKey for readability
+func (s *Store) edgeKey(cf string, last bool) ([]byte, bool, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if err := s.checkOpen(); err != nil {
+		return nil, false, err
+	}
+	cfh, err := s.resolveCF(cf)
+	if err != nil {
+		return nil, false, err
+	}
+
+	it := s.db.NewIteratorCF(s.ro, cfh)
+	defer it.Close()
+	if last {
+		it.SeekToLast()
+	} else {
+		it.SeekToFirst()
+	}
+	if !it.Valid() {
+		// Empty CF (it.Err() is nil) or a mid-seek RocksDB error.
+		return nil, false, it.Err()
+	}
+	// Copy: the KeySlice is freed when the iterator closes.
+	return append([]byte(nil), it.KeySlice().Data()...), true, it.Err()
+}
+
 // IterateRange yields (key, value) for keys in [start, end] byte-lex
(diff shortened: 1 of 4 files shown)


cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -292,6 +292,46 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
 	}
 }
 
+// FirstKey returns the smallest key in cf, or ok=false if the CF is empty.
```
chowbao: nit: comment clarification. CF is empty on `ok=false` and nil error

Sounds like empty CF will be handled elsewhere

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
```diff
@@ -292,6 +292,46 @@ func (s *Store) Iterate(cf string, prefix []byte) iter.Seq2[Entry, error] {
...
+// Cheap: a single boundary seek (no scan).
+func (s *Store) FirstKey(cf string) ([]byte, bool, error) {
+	return s.edgeKey(cf, false)
+}
+
+// LastKey returns the largest key in cf, or ok=false if the CF is empty.
```
chowbao: nit: comment clarification. CF is empty on ok=false and nil error

Sounds like empty CF will be handled elsewhere

chowbao (approved):

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
tamirms: Good call — reworded in b4b496b9 to spell out that a keyless CF returns `(nil, false, nil)` (emptiness via `ok`, not an error), and to disambiguate the `cf` argument (`""` selects the default CF; an unregistered name returns `ErrCFNotFound`). Thanks!

cmd/stellar-rpc/internal/fullhistory/pkg/rocksdb/rocksdb.go
tamirms: Good call — reworded in b4b496b9 to spell out that a keyless CF returns `(nil, false, nil)` (emptiness via `ok`, not an error), and to disambiguate the `cf` argument (`""` selects the default CF; an unregistered name returns `ErrCFNotFound`). Thanks!

Issue #765, opened 2026-06-05
Ingestion: all data types for both cold and hot stores


tamirms:
## TL;DR

Implement the data-ingestion path for RPC v2: for each ledger close meta, write the relevant data into the configured storage tiers. Three data types (ledger / events / txhash) × two tiers (hot RocksDB / cold immutable files) = six per-tier ingesters, plus a top-level service per tier that fans an LCM out across the enabled ingesters with hot-tier parallelism, configurable data-type selection, and Prometheus metrics. A working prototype of all of this exists in the `rpc-hack` bench harness and can serve as inspiration (see "Reference / prior work" below).

## What to build

Code lands in `cmd/stellar-rpc/internal/fullhistory/ingest/`.

### Per-ingester contract

Two interfaces, one per tier:

```go
type HotIngester interface {
    Ingest(ctx context.Context, lcm xdr.LedgerCloseMetaView) error
}

type ColdIngester interface {
    Ingest(ctx context.Context, lcm xdr.LedgerCloseMetaView) error
    Finalize(ctx context.Context) error
    Close() error
}
```

**Ownership model.** Hot ingesters receive a long-lived store via dependency injection (the daemon owns the store's lifecycle). Cold ingesters open their own per-chunk writer in the constructor and own it for the chunk's duration — `Finalize` commits on success, `Close` drops a partial output file on failure.

Six implementations:

| | Hot (RocksDB) | Cold (immutable files) |
|---|---|---|
| **Ledger** | `ledger.HotStore.AddLedgers` | `ledger.ColdWriter.AppendLedger` + `Commit` |
| **Events** | `eventstore.HotStore.IngestLedgerEvents` | `eventstore.ColdWriter` + `WriteColdIndex` |
| **TxHash** | `txhash.HotStore.AddEntries` | append `(txHash, ledgerSeq)` per ledger → sorted dump per chunk on `Finalize` |

The global txhash MPHF build over the per-chunk sorted dumps is a separate downstream step, not in scope here.

### Metrics

Metrics are emitted through a small sink interface so the same ingester code can report to **Prometheus** (production) or **CSV** (the bench-command migration, follow-up task) interchangeably. The Prometheus implementation ships with this task and registers via the existing daemon convention (`internal/daemon/metrics.go`). Exact sink shape is up to the implementor.

Required signals:

- **Per-ingester** — hot: per-ledger ingest latency, throughput, item count, error count. Cold: per-chunk ingest latency, item count, error count.
[...]

Issue #764, opened 2026-06-05
XDR view extractors: events, tx-hashes, tx-details, tx-pages from LedgerCloseMetaView


tamirms:
## TL;DR

Implement four extractor functions that derive the data shapes RPC v2 needs from an `xdr.LedgerCloseMetaView`, using zero-copy aliasing into the source LCM. The perf benefits of zero-copy XDR views were validated on `rpc-hack`, so v2 will adopt them. Most of this code exists on `rpc-hack` already; the task is largely promoting it from bench-quality into a production package.

## What to build

Four extractors over `xdr.LedgerCloseMetaView`:

- **events** — yields events-index payloads.
- **tx-hashes** — yields the `txHash` for each transaction in the LCM.
- **tx-details by hash** — given an LCM + target txhash, returns the transaction's envelope / result / meta.
- **tx-pages** — yields transactions in order with cursor support for the paginated `getTransactions` API.

All outputs zero-copy, aliasing into the source LCM buffer.

The stellar-rpc-side code lives in `cmd/stellar-rpc/internal/fullhistory/views/` (or similar).

## Determine the SDK vs stellar-rpc split

Part of this task: decide which pieces belong in `go-stellar-sdk` (alongside the view types) versus `stellar-rpc`. Rule of thumb — low-level XDR navigation primitives go to the SDK; RPC-specific data shapes (e.g., `events.Payload`, pagination cursors) stay here.

## Source material on rpc-hack

Most of this code exists on `rpc-hack` already and can be cleaned up rather than written from scratch:

- **events** — `cmd/stellar-rpc/internal/events/ingest_view.go` (production-ready; originally in PR #756, will be scoped out of that PR and land via this task instead).
- **tx-hashes / tx-details by hash** — `cmd/stellar-rpc/scripts/bench-fullhistory/tx_hash_helpers.go`, `extract_views.go`, `bench_cold_txhash.go` (bench-quality; needs promotion).
- **tx-pages** — `bench_cold_txpage.go` / `bench_hot_txpage.go` for reference; reusability TBD.

## Tests + benchmarks

- Per-extractor unit tests.
- Per-extractor `go test -bench` benchmarks so perf regressions are catchable.
- Differential correctness against the full-decode (`UnmarshalBinary` → derive) path on a checked-in fixture set covering recent ledgers, protocol-transition ledgers, and known edge cases (empty / sponsorship / large tx / V0/V1/V2 variants).
[...]

PR #778, opened 2026-06-08
views: zero-copy XDR view extractors — events, tx-hashes, tx-details, tx-pages (#764)

chowbao:
Implements #764 — the four zero-copy XDR view extractors, promoted from the `rpc-hack` bench harness into a production package `cmd/stellar-rpc/internal/fullhistory/views/`. Each turns a raw `LedgerCloseMeta` (wrapped as `xdr.LedgerCloseMetaView`) into the shape RPC v2 needs, **without** `UnmarshalBinary` — outputs alias the view buffer (callers copy what they retain).

> **Stacked.** Based on a throwaway integration base (`fh-ingest-base` = `feature/full-history` + #756 merged) so it compiles against the events index. Will retarget to `feature/full-history` once #756 lands. The ingestion PR (#765) stacks on this branch.

## The four extractors (all of #764)

| #764 requirement | Entrypoint | File |
|---|---|---|
| **events** — events-index payloads | `ExtractEvents(lcm) ([]events.Payload, error)` | `events.go` |
| **tx-hashes** — txHash per tx | `ExtractTxHashes(lcm) ([]txhash.Entry, error)` | `txhash.go` |
| **tx-details by hash** — envelope/result/meta for a target hash (`getTransaction`) | `ExtractTxDetailsByHash(lcm, hash, passphrase) (Transaction, bool, error)` | `txdetails.go` |
| **tx-pages** — transactions in order with cursor (`getTransactions`) | `ExtractTransactions(lcm, startIdx, limit, passphrase) ([]Transaction, error)` | `txdetails.go` |

- **events / tx-hashes** read only `TxProcessing` (which carries the hash + meta), so they take no passphrase; `ExtractEvents` returns `ErrV0Unsupported` for V0 (V0 has no contract events).
- **tx-details / tx-pages** also need each tx's **envelope** (in the `TxSet`, which is in agreed-set/hash-sorted order, not apply order), so they pair envelope→tx **by hash** (`network.HashTransactionInEnvelope`) and take the network passphrase. `startIdx`/`limit` are the `getTransactions` cursor. They return a package-local `views.Transaction`.
- Supporting: `envelopes.go` (version/phase-aware envelope enumeration + hashing, used by the two read-path extractors).

## Tests
- **Differential** vs the parsed path: `ExtractEvents`/`ExtractTxHashes` vs the struct `LCMToPayloads`/`lcm.TransactionHash`; `ExtractTxDetailsByHash`/`ExtractTransactions` vs `db.ParseTransaction` (driven by the SDK `LedgerTransactionReader`) — field-by-field, wire-identical bytes, across LCM **V0/V1/V2** and meta **V1–V4** (incl. V0Components + ParallelTxs phases, staged top-level + per-op events, diagnostic events).
- `pairing_test.go` — reversed-order regression (TxSet order ≠ apply order) guarding the by-hash pairing.
- Negative paths (unsupported meta version, missing-hash, `startIdx<0`/`limit<0`), a zero-copy aliasing assertion, and per-extractor `go test -bench` benchmarks.

## SDK vs stellar-rpc split (issue #764 acceptance item)

**Rule:** low-level XDR navigation → `go-stellar-sdk`; RPC-specific data shapes → stellar-rpc.

**Determination: nothing in this task belongs in the SDK** — the boundary already sat where the rule wants it. The SDK ships all the low-level navigation, so this package is a pure *consumer* that composes those primitives into RPC shapes.

**SDK-side (consumed, nothing added)** — production imports:
- `xdr.LedgerCloseMetaView` + the generated `*View` accessors (`.V()/.V0..V2()`, `.Iter()`, `.At()`, `.Count()`, `.Raw()`, `.Value()`) — zero-copy, schema-generated XDR navigation.
- `network.HashTransactionInEnvelope` — canonical protocol-level tx-hash (envelope↔tx pairing).
- `toid` — protocol-level transaction/operation-ID (apply order / cursor).

**stellar-rpc-side (this package)** — the four extractors plus the RPC-specific shapes (`events.Payload`, `txhash.Entry`, the package-local `views.Transaction`) and semantics (event-ID/term ordering, `startIdx/limit` paging, envelope-by-hash pairing). Documented in `views/doc.go`.

The differential tests pull in the SDK's `ingest.LedgerTransactionReader` only as the *reference oracle* to validate the view path against — it is deliberately **not** a production dependency.

**Related layering choice (within stellar-rpc):** the read-path extractors return a package-local `views.Transaction` rather than `internal/db.Transaction` (even though `db` doesn't import `views`, so there'd be no cycle) — keeping `views` a clean leaf and avoiding an inversion of the `db → views` layering.

## Differences from `rpc-hack`

A faithful promotion of rpc-hack's view-extraction logic (the V1–V4 event walk and XDR-view navigation are essentially line-for-line), with the bench-only scaffolding dropped (roundtrip-comparison harness, CLI/CSV/cache) and a few deliberate changes:

- **One correctness fix (the notable divergence):** rpc-hack paired each transaction's envelope to its TxProcessing entry **by array position**, which is only correct when TxSet order == apply order — it isn't in general. That's a latent bug still present on rpc-hack; its differential tests missed it because their fixtures built the TxSet in apply order. Here, envelopes are paired **by hash** (`envelopesByHash`, mirroring the SDK's `LedgerTransactionReader`), with `pairing_test.go` as a reversed-order regression.
- **Packaging:** moved from `package events` + the bench `package main` into one leaf `package views`; read-path extractors return `views.Transaction` (flat `LedgerSequence`/`LedgerCloseTime`, raw `[32]byte` hash) rather than `internal/db.Transaction`.
- **Events:** term keys are no longer precomputed inline onto the payload; they're derived downstream via `events.TermsForBytes`. The (previously ignored) passphrase param was dropped from `ExtractEvents`; it's now a real, used input on the read-path extractors.
- **Hardening not in rpc-hack:** V3 contract-event gating on `IsSorobanTx` (matches `db.ParseTransaction`; rpc-hack over-emitted for a classic-envelope V3 tx), `startIdx<0`/`limit<0` rejection, an explicit missing-hash error, and differential tests vs `db.ParseTransaction` across LCM V0/V1/V2.

## Follow-up (per #764 acceptance, not in this PR)
- Full-history sweep harness that walks every ledger and asserts view-extraction == full-decode (separate tracking issue).

## Test plan
- [x] `go test ./cmd/stellar-rpc/internal/fullhistory/views/` (+ `-bench`), `go vet`, `gofmt` — green.

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
+++ b/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
@@ -22,9 +22,9 @@ type envPart struct {
 }
 
 // envPartFromView decodes the envelope view into an envPart: the RETURNED raw
-// bytes stay the zero-copy .Raw() slice; the binary decode is transient and
-// used only to compute the transaction hash (the map key the TxProcessing
-// entries reference). Mirrors ingest.LedgerTransactionReader.storeTransactions,
+// bytes stay the zero-copy .Raw() slice; the binary decode feeds the
+// transaction hash (the map key the TxProcessing entries reference), the
+// envelope type, and the soroban flag. Mirrors ingest.LedgerTransactionReader.storeTransactions,
 // which hashes every TxSet envelope (via network.HashTransactionInEnvelope,
 // which handles the fee-bump case) because the TxSet is in agreed-set /
 // hash-sorted order, NOT TxProcessing apply order.
@@ -33,10 +33,13 @@ func envPartFromView(env xdr.TransactionEnvelopeView, passphrase string) (envPar
 	if err != nil {
 		return envPart{}, fmt.Errorf("views: envelope.Raw: %w", err)
 	}
-	// Transient decode purely to compute the hash; the returned raw slice is
-	// the original zero-copy .Raw() view buffer, not this decoded value. The
-	// type and soroban flag are read off this decode too (we have it in hand),
-	// rather than via a separate view traversal.
+	// Decode once and read everything we need off it: the transaction hash
+	// (network.HashTransactionInEnvelope needs a decoded envelope — it builds
+	// the tx-hash preimage, not a hash of the raw bytes), the envelope type,
+	// and the soroban flag (which must inspect Tx.Ext). The decode is required
+	// regardless for isSoroban, so hashing piggybacks on it rather than being
+	// extra work. The RETURNED raw slice stays the zero-copy .Raw() view
+	// buffer, not this decoded value.
 	var decoded xdr.TransactionEnvelope
 	if uerr := decoded.UnmarshalBinary(raw); uerr != nil {
 		return envPart{}, fmt.Errorf("views: envelope decode for hashing: %w", uerr)
--- a/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
+++ b/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
@@ -33,16 +33,10 @@ func envPartFromView(env xdr.TransactionEnvelopeView, passphrase string) (envPar
 	if err != nil {
 		return envPart{}, fmt.Errorf("views: envelope.Raw: %w", err)
 	}
-	tv, err := env.Type()
-	if err != nil {
-		return envPart{}, fmt.Errorf("views: envelope.Type: %w", err)
-	}
-	t, err := tv.Value()
-	if err != nil {
-		return envPart{}, fmt.Errorf("views: envelope.Type value: %w", err)
-	}
 	// Transient decode purely to compute the hash; the returned raw slice is
-	// the original zero-copy .Raw() view buffer, not this decoded value.
+	// the original zero-copy .Raw() view buffer, not this decoded value. The
+	// type and soroban flag are read off this decode too (we have it in hand),
+	// rather than via a separate view traversal.
 	var decoded xdr.TransactionEnvelope
 	if uerr := decoded.UnmarshalBinary(raw); uerr != nil {
 		return envPart{}, fmt.Errorf("views: envelope decode for hashing: %w", uerr)
@@ -51,7 +45,7 @@ func envPartFromView(env xdr.TransactionEnvelopeView, passphrase string) (envPar
 	if err != nil {
 		return envPart{}, fmt.Errorf("views: hash envelope: %w", err)
 	}
-	return envPart{raw: raw, typ: t, hash: hash, isSoroban: envelopeIsSoroban(decoded)}, nil
+	return envPart{raw: raw, typ: decoded.Type, hash: hash, isSoroban: envelopeIsSoroban(decoded)}, nil
 }
 
 // envelopeIsSoroban mirrors ingest.LedgerTransaction.IsSorobanTx /
--- a/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
+++ b/cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
@@ -1,7 +1,12 @@
 package views
 
 import (
+	"crypto/sha256"
+	"encoding/binary"
+	"errors"
 	"fmt"
+	"hash"
+	"strings"
 
 	"github.com/stellar/go-stellar-sdk/network"
 	"github.com/stellar/go-stellar-sdk/xdr"
@@ -9,11 +14,11 @@ import (
 
 // envPart is one envelope (raw bytes + type) gathered while enumerating a
 // TxSet. raw aliases the view buffer (zero-copy). hash is the transaction
-// hash computed from the envelope under the network passphrase — the key the
-// TxProcessing entries reference. isSoroban mirrors
-// ingest.LedgerTransaction.IsSorobanTx (computed from the SAME transient decode
-// used for hashing), used by the read path to gate V3 contract events the way
-// the struct path does.
+// hash computed from the envelope wire bytes under the network passphrase —
+// the key the TxProcessing entries reference. isSoroban mirrors
+// ingest.LedgerTransaction.IsSorobanTx (read from the inner Tx.Ext view
+// discriminant), used by the read path to gate V3 contract events the way
+// the struct path does (see the trusted-input invariant in doc.go).
 type envPart struct {
 	raw       []byte
 	typ       xdr.EnvelopeType
@@ -21,69 +26,118 @@ type envPart struct {
 	isSoroban bool
 }
 
-// envPartFromView decodes the envelope view into an envPart: the RETURNED raw
-// bytes stay the zero-copy .Raw() slice; the binary decode feeds the
-// transaction hash (the map key the TxProcessing entries reference), the
-// envelope type, and the soroban flag. Mirrors ingest.LedgerTransactionReader.storeTransactions,
-// which hashes every TxSet envelope (via network.HashTransactionInEnvelope,
-// which handles the fee-bump case) because the TxSet is in agreed-set /
-// hash-sorted order, NOT TxProcessing apply order.
-func envPartFromView(env xdr.TransactionEnvelopeView, passphrase string) (envPart, error) {
-	raw, err := env.Raw()
-	if err != nil {
-		return envPart{}, fmt.Errorf("views: envelope.Raw: %w", err)
-	}
-	// Decode once and read everything we need off it: the transaction hash
-	// (network.HashTransactionInEnvelope needs a decoded envelope — it builds
-	// the tx-hash preimage, not a hash of the raw bytes), the envelope type,
-	// and the soroban flag (which must inspect Tx.Ext). The decode is required
-	// regardless for isSoroban, so hashing piggybacks on it rather than being
-	// extra work. The RETURNED raw slice stays the zero-copy .Raw() view
-	// buffer, not this decoded value.
-	var decoded xdr.TransactionEnvelope
-	if uerr := decoded.UnmarshalBinary(raw); uerr != nil {
-		return envPart{}, fmt.Errorf("views: envelope decode for hashing: %w", uerr)
-	}
-	hash, err := network.HashTransactionInEnvelope(decoded, passphrase)
-	if err != nil {
-		return envPart{}, fmt.Errorf("views: hash envelope: %w", err)
-	}
-	return envPart{raw: raw, typ: decoded.Type, hash: hash, isSoroban: envelopeIsSoroban(decoded)}, nil
+// txHasher computes transaction hashes straight from envelope view bytes.
+// The tx-hash preimage is
+//
+//	SHA256(networkID ‖ envelope-type tag (4 bytes) ‖ Transaction XDR)
+//
+// — exactly what network.HashTransactionInEnvelope produces — and every
+// piece is available as a wire-byte slice off the view, so no envelope
+// decode is needed. The network ID is computed once per extraction and the
+// SHA-256 state is reused, so hashing allocates nothing per envelope.
+type txHasher struct {
+	networkID [32]byte
+	h         hash.Hash
 }
 
-// envelopeIsSoroban mirrors ingest.LedgerTransaction.IsSorobanTx /
-// GetSorobanData: a tx is Soroban iff its (inner, for fee-bump) tx envelope
-// carries SorobanTransactionData in its Ext. Only the v1 Tx and fee-bump inner
-// v1 Tx shapes can carry Soroban data; every other envelope type is classic.
-//
-// NB: the Tx -> V1.Tx.Ext.GetSorobanData (and fee-bump -> inner) access path
-// here mirrors the SDK's ingest.LedgerTransaction.IsSorobanTx/GetSorobanData;
-// if the SDK ever relocates where soroban-data lives, this is a known drift
-// point to update in lockstep.
-func envelopeIsSoroban(env xdr.TransactionEnvelope) bool {
-	switch env.Type {
-	case xdr.EnvelopeTypeEnvelopeTypeTx:
-		_, ok := env.V1.Tx.Ext.GetSorobanData()
-		return ok
-	case xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
-		_, ok := env.FeeBump.Tx.InnerTx.V1.Tx.Ext.GetSorobanData()
-		return ok
-	default:
-		return false
+// newTxHasher derives the network ID from passphrase. The empty-passphrase
+// guard matches network.hashTx's, surfaced once per extraction instead of
+// once per envelope.
+func newTxHasher(passphrase string) (*txHasher, error) {
+	if strings.TrimSpace(passphrase) == "" {
+		return nil, errors.New("views: empty network passphrase")
 	}
+	return &txHasher{networkID: network.ID(passphrase), h: sha256.New()}, nil
+}
+
+// sum computes SHA256(networkID ‖ tag ‖ parts...).
+func (th *txHasher) sum(tag xdr.EnvelopeType, parts ...[]byte) [32]byte {
+	th.h.Reset()
+	th.h.Write(th.networkID[:])
+	var tagBuf [4]byte
+	binary.BigEndian.PutUint32(tagBuf[:], uint32(tag)) //nolint:gosec // enum tag is small and non-negative
+	th.h.Write(tagBuf[:])
+	for _, p := range parts {
+		th.h.Write(p)
+	}
+	var out [32]byte
+	th.h.Sum(out[:0])
+	return out
+}
+
+// ed25519KeyTypePrefix is the 4-byte XDR discriminant
+// CryptoKeyTypeKeyTypeEd25519 (0) that converting a TransactionV0 to a
+// Transaction prepends to the source account key on the wire.
+var ed25519KeyTypePrefix = [4]byte{0, 0, 0, 0}
+
+// envPartFromView reads one envelope view into an envPart without decoding
+// the envelope: the type comes from the union discriminant view, the
+// transaction hash is computed by th straight from the inner transaction's
+// wire bytes, and the soroban flag is read from the inner Tx.Ext view. The
+// returned raw slice stays the zero-copy .Raw() view buffer.
+//
+// A TX_V0 envelope hashes as its V1 conversion (network.HashTransactionV0):
+// on the wire that conversion is exactly the 4-byte ed25519 key-type prefix
+// followed by the unchanged TransactionV0 bytes — the V0 optional-TimeBounds
+// and V1 Preconditions encodings are identical (absent ⇒ 0 ⇒ PRECOND_NONE,
+// present ⇒ 1 + TimeBounds ⇒ PRECOND_TIME), as are all fields after them
+// (both Ext arms are void).
+func envPartFromView(env xdr.TransactionEnvelopeView, th *txHasher) (envPart, error) {
+	typ, err := xdr.Try(func() xdr.EnvelopeType { return env.MustType().MustValue() })
+	if err != nil {
+		return envPart{}, fmt.Errorf("views: envelope type: %w", err)
+	}
+	switch typ {
+	case xdr.EnvelopeTypeEnvelopeTypeTx,
+		xdr.EnvelopeTypeEnvelopeTypeTxV0,
+		xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
+	default:
+		return envPart{}, fmt.Errorf("views: invalid transaction envelope type %v", typ)
+	}
+	p := envPart{typ: typ}
+	err = xdr.TryVoid(func() {
+		p.raw = env.MustRaw()
+		switch typ {
+		case xdr.EnvelopeTypeEnvelopeTypeTx:
+			tx := env.MustV1().MustTx()
+			p.hash = th.sum(typ, tx.MustRaw())
+			p.isSoroban = txExtIsSoroban(tx)
+		case xdr.EnvelopeTypeEnvelopeTypeTxV0:
+			p.hash = th.sum(xdr.EnvelopeTypeEnvelopeTypeTx,
+				ed25519KeyTypePrefix[:], env.MustV0().MustTx().MustRaw())
+		case xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
+			fbTx := env.MustFeeBump().MustTx()
+			p.hash = th.sum(typ, fbTx.MustRaw())
+			// IsSorobanTx inspects the fee-bump INNER transaction.
+			p.isSoroban = txExtIsSoroban(fbTx.MustInnerTx().MustV1().MustTx())
+		}
+	})
+	if err != nil {
+		return envPart{}, fmt.Errorf("views: envelope (%v): %w", typ, err)
+	}
+	return p, nil
+}
+
+// txExtIsSoroban reads Tx.Ext's union discriminant (1 ⟺
+// SorobanTransactionData present), mirroring
+// ingest.LedgerTransaction.IsSorobanTx / GetSorobanData. Must-style: panics
+// with *xdr.ViewError on malformed input, recovered by the caller's TryVoid.
+func txExtIsSoroban(tx xdr.TransactionView) bool {
+	return tx.MustExt().MustV().MustValue() == 1
 }
 
 // envYield is the per-envelope sink the TxSet enumerators drive. It is
 // called once per envelope (in agreed-set order, NOT TxProcessing apply
 // order); returning stop==true ends enumeration early (used by the
-// by-hash lookup once it finds its target), returning an error aborts it.
+// by-hash lookups once every wanted hash is resolved), returning an error
+// aborts it.
 type envYield func(envPart) (stop bool, err error)
 
 // enumerateEnvelopesFromV0TxSet walks a V0 TransactionSet, invoking yield for
 // every envelope. The TxSet is in agreed-set order (NOT TxProcessing apply
 // order); callers pair envelopes to transactions by hash, so enumeration order
 // is irrelevant here. Returns stopped==true if yield asked to stop early.
-func enumerateEnvelopesFromV0TxSet(txSet xdr.TransactionSetView, passphrase string, yield envYield) (stopped bool, err error) {
+func enumerateEnvelopesFromV0TxSet(txSet xdr.TransactionSetView, th *txHasher, yield envYield) (stopped bool, err error) {
 	txs, err := txSet.Txs()
 	if err != nil {
 		return false, fmt.Errorf("views: V0 TxSet.Txs: %w", err)
@@ -93,7 +147,7 @@ func enumerateEnvelopesFromV0TxSet(txSet xdr.TransactionSetView, passphrase stri
 		if eerr != nil {
 			return false, fmt.Errorf("views: V0 envelope at %d: %w", i, eerr)
 		}
-		p, perr := envPartFromView(env, passphrase)
+		p, perr := envPartFromView(env, th)
 		if perr != nil {
 			return false, perr
 		}
@@ -114,7 +168,7 @@ func enumerateEnvelopesFromV0TxSet(txSet xdr.TransactionSetView, passphrase stri
 // Like the V0 walker, the order is the agreed-set order (NOT TxProcessing apply
 // order); pairing is by hash so order does not matter. Returns stopped==true if
 // yield asked to stop early.
-func enumerateEnvelopesFromGeneralized(txSet xdr.GeneralizedTransactionSetView, passphrase string, yield envYield) (stopped bool, err error) {
+func enumerateEnvelopesFromGeneralized(txSet xdr.GeneralizedTransactionSetView, th *txHasher, yield envYield) (stopped bool, err error) {
 	v1Set, err := txSet.V1TxSet()
 	if err != nil {
 		return false, fmt.Errorf("views: V1TxSet: %w", err)
@@ -141,7 +195,7 @@ func enumerateEnvelopesFromGeneralized(txSet xdr.GeneralizedTransactionSetView, 
 			if err != nil {
 				return false, fmt.Errorf("views: V0Components: %w", err)
 			}
-			stop, err := enumerateV0Components(comps, passphrase, yield)
+			stop, err := enumerateV0Components(comps, th, yield)
 			if err != nil {
 				return false, err
 			}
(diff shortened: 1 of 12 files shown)


cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
```diff
@@ -0,0 +1,588 @@
...
+// dependency, keeping this package a clean leaf.)
+type Transaction struct {
+	Hash              [32]byte
+	ApplicationOrder  int32 // 1-based apply order within the ledger
+	FeeBump           bool
+	Successful        bool
+	Envelope          []byte     // raw xdr.TransactionEnvelope
+	Result            []byte     // raw xdr.TransactionResult
+	Meta              []byte     // raw xdr.TransactionMeta
+	Events            [][]byte   // raw xdr.DiagnosticEvent (V3/V4 diagnostic)
```
chowbao, line 22: Are diagnostic events part of full history rpc v2?

The ledgers we backfill with from GCS/S3 do contain diagnostic events but that doesn't necessarily mean the transaction view needs it as well

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return envPart{}, fmt.Errorf("views: envelope.Type value: %w", err)
+	}
+	// Transient decode purely to compute the hash; the returned raw slice is
+	// the original zero-copy .Raw() view buffer, not this decoded value.
+	var decoded xdr.TransactionEnvelope
+	if uerr := decoded.UnmarshalBinary(raw); uerr != nil {
+		return envPart{}, fmt.Errorf("views: envelope decode for hashing: %w", uerr)
+	}
+	hash, err := network.HashTransactionInEnvelope(decoded, passphrase)
```
Shaptic: Doesn't this more-or-less just re-marshal it anyway? It's a little more involved than that and obviously the code would be really gross but the round trip here does feelsbadman a bit.

Alternatively, since you're already unmarshaling the envelope, you can remove the code above that fetches type via views and just do it the "normal" way for brevity.

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
```diff
@@ -0,0 +1,244 @@
...
+// v1 Tx shapes can carry Soroban data; every other envelope type is classic.
+//
+// NB: the Tx -> V1.Tx.Ext.GetSorobanData (and fee-bump -> inner) access path
+// here mirrors the SDK's ingest.LedgerTransaction.IsSorobanTx/GetSorobanData;
+// if the SDK ever relocates where soroban-data lives, this is a known drift
+// point to update in lockstep.
+func envelopeIsSoroban(env xdr.TransactionEnvelope) bool {
+	switch env.Type {
+	case xdr.EnvelopeTypeEnvelopeTypeTx:
+		_, ok := env.V1.Tx.Ext.GetSorobanData()
```
Shaptic: I will say that I do not like this method of "is soroban"ing but it is what it is

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,460 @@
...
+//
+// Ordering matches LCMToPayloads exactly: per-tx in apply order; within
+// each tx, top-level TransactionEvents first (V4 only, dispatched on
+// Stage with ledger-wide before/after counters), then per-operation
+// events in (op, event) order.
+//
+// Supported shapes:
+//
+//   - LCM V1, V2 (apply order = TxProcessing array order)
+//   - TransactionMeta V1, V2 (no events, skipped)
```
Shaptic: There's also TransactionMeta"V0" which is just the `Operations *[]OperationMeta` field being filled out which doesn't appear to be considered here

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+	ctView, err := scp.CloseTime()
+	if err != nil {
+		return 0, 0, fmt.Errorf("views: CloseTime: %w", err)
+	}
+	ctVal, err := ctView.Value()
+	if err != nil {
+		return 0, 0, fmt.Errorf("views: CloseTime value: %w", err)
+	}
+	return seqVal, int64(ctVal), nil //nolint:gosec // TimePoint is uint64
```
Shaptic: There's gooooootta be a better way to do this than per-property error inspection.... It'd be nice if you could do:

```go
ctVal := header.HeaderW().LedgerSeqW().ValueW().ScpValueW().CloseTimeW().ValueW()
if ctVal.hasError {
  // ... has the first error that occurred during unwrapping
}
```

where the `*W()` variant returns a "wrapped" version that tracks whether or not there's an error, e.g.

```
struct LedgerHeaderW {
  LedgerHeader
  hasError error
}

func (*hdr LedgerHeaderHistoryEntryView) HeaderW() LedgerHeaderW {
	inner, err := header.Header()
	return LedgerHeaderW{LedgerHeader: inner, hasError: err}
}

// and so on
```

where each child method just propagates the error until the end if it's present. I could be making up Go semantics here and maybe this isn't possible, but it'd be a _way_ nicer api ux if we could chain it this way. @tamirms is this even tenable in Go or am I making up syntax?

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,460 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("views: meta.V: %w", err)
+	}
+	v, err := vView.Value()
+	if err != nil {
+		return nil, fmt.Errorf("views: meta.V value: %w", err)
+	}
+
+	switch v {
+	case 1, 2:
```
Shaptic: yeah so here there'd need to be a case 0 as well

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,460 @@
...
+// payloadsFromV4Meta extracts V4 events:
+//
+//  1. Top-level TransactionEvents first, dispatched on Stage:
+//     BeforeAllTxs -> uses ledger-wide state.beforeIndex
+//     AfterAllTxs  -> uses ledger-wide state.afterIndex
+//     AfterTx      -> uses per-tx counter (reset here)
+//  2. Then per-operation contract events in (op, event) order.
+//
+// Order matches LCMToPayloads / db.InsertEvents and the (TxIdx, OpIdx,
+// EventIdx) sentinels match the struct path's encoding.
```
Shaptic, line 215: Maybe I'm misunderstanding here: [this order](https://github.com/stellar/stellar-rpc/blob/99a61f337b66635ba6f9d70d2403ee5faed1d7c1/cmd/stellar-rpc/internal/db/event.go#L97-L119) from current RPC operates on a per-ledger basis. Are you suggesting that this creates a list of orders that are then correctly interleaved later?

cmd/stellar-rpc/internal/fullhistory/views/pairing_test.go
```diff
@@ -0,0 +1,240 @@
...
+	header := xdr.LedgerHeaderHistoryEntry{
+		Header: xdr.LedgerHeader{
+			ScpValue:  xdr.StellarValue{CloseTime: xdr.TimePoint(closeTimestamp)},
+			LedgerSeq: xdr.Uint32(ledgerSeq),
+		},
+	}
+	txSet := xdr.GeneralizedTransactionSet{V: 1, V1TxSet: &xdr.TransactionSetV1{Phases: phases}}
+
+	switch version {
+	case 1:
```
Shaptic, line 138: what about v0?

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
Shaptic: If we want parity with the [current API](https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getTransaction) then we do need diagnostic events, yeah. But not in `getEvents`.

Shaptic (commented):
First pass

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: I had that question too but it seems like current RPC doesn't support V0

```
- db.ParseTransaction → ingestTx.GetTransactionEvents() → SDK ledger_transaction.go:308 has a default:
  that returns "unsupported TransactionMeta version: 0".
```

Hmm but I guess so with this implementation of rpc v2 we should?

chowbao: Oh I see. I remember one other thing. Full history rpc v2 would realistically never have V0 txmeta because it would ingest with the unified events and backfill flags set to true cause otherwise there wouldn't be any events for classic

Anyhoo I'm forcing claude to add it in

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: ~It doesn't interleave later. I forgot about the event ordering~

Edit: It passes off the interleaving to the ingest system downstream. This is just the extraction of all events with the correct IDs to be used to order later

It follows the same part of getting the payload out as the code you linked where [ops happen after all tx events](https://github.com/stellar/stellar-rpc/blob/main/cmd/stellar-rpc/internal/db/event.go#L195).

And to get the correctly ordered events you'd just iterate through the event indexes

cmd/stellar-rpc/internal/fullhistory/views/events.go
Shaptic: Yeah current RPC doesn't need to support it because it'd never come up, but with full history it's possible. You're 100% right about the flags, though - but better to be defensive!

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
chowbao: Done below.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Fixed

cmd/stellar-rpc/internal/fullhistory/views/pairing_test.go
chowbao: Fixed

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: It seems possible but that wrapped types would have to go into the go-stellar-sdk and can't be done here.

I'm not sure that's worth it at this time just to make the error handling a little nicer

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// envPartFromView decodes the envelope view into an envPart: the RETURNED raw
+// bytes stay the zero-copy .Raw() slice; the binary decode feeds the
+// transaction hash (the map key the TxProcessing entries reference), the
+// envelope type, and the soroban flag. Mirrors ingest.LedgerTransactionReader.storeTransactions,
+// which hashes every TxSet envelope (via network.HashTransactionInEnvelope,
+// which handles the fee-bump case) because the TxSet is in agreed-set /
+// hash-sorted order, NOT TxProcessing apply order.
+func envPartFromView(env xdr.TransactionEnvelopeView, passphrase string) (envPart, error) {
```
tamirms: **should-fix:** The decode can be removed entirely. The transaction hash can be computed directly from the wire bytes as `SHA256(networkID ‖ envType-tag ‖ innerTx.Raw())`. If the network ID is computed once per call and the SHA256 state is reused, hashing allocates nothing per envelope. The TX_V0 case hashes as its V1 conversion, which on the wire only adds the 4-byte muxed key-type prefix before the source key; the V0 optional-TimeBounds encoding and the V1 Preconditions encoding are identical, as are all the fields after them. The `isSoroban` flag can be read from `V1().Tx().Ext()`, or the fee-bump inner equivalent. Regarding aad67769: once the envelope type and the soroban flag are derived from the view, nothing needs the decode anymore. Separately, it would be worth adding TX_V0-envelope fixtures — no current fixture exercises one.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
```diff
@@ -0,0 +1,594 @@
...
+	enumerateEnvs func(passphrase string, yield envYield) (stopped bool, err error)
+}
+
+// dispatchLCM opens lcm, reads its discriminator, and returns the
+// version-specific handles for the read path (V0/V1/V2). Shared by
+// ExtractTxDetailsByHash and ExtractTransactions. The header + TxProcessing
+// opening is delegated to openHeaderAndTxProc; only the version-specific
+// TxSet open (TransactionSetView vs GeneralizedTransactionSetView) and its
+// envelope enumerator are handled here.
+func dispatchLCM(lcm xdr.LedgerCloseMetaView) (lcmDispatch, error) {
```
tamirms: **should-fix:** This switches on the same discriminant twice: `dispatchLCM` calls `openHeaderAndTxProc` and then re-opens `lcm.V0()/V1()/V2()` to get the TxSet. The comment saying the TxSet open "cannot share this path" doesn't quite hold, because the `enumerateEnvs` closure already absorbs the type differences. Could this be a single dispatch that returns the header, the TxProcessing iterable, and the envelope enumerator, with the events path simply ignoring the enumerator?

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
```diff
@@ -0,0 +1,594 @@
...
+	metaIsV3 bool
+}
+
+// envelopesByHash builds the txhash -> envelope map mirroring
+// ingest.LedgerTransactionReader.storeTransactions: every TxSet envelope is
+// enumerated and hashed (the TxSet is in agreed-set order, NOT TxProcessing
+// apply order), so a TxProcessing entry's TransactionHash can locate its OWN
+// envelope rather than the one that happens to sit at the same array index.
+// The returned envPart raw bytes remain zero-copy aliases of the view buffer.
+func envelopesByHash(d lcmDispatch, passphrase string) (map[[32]byte]envPart, error) {
```
tamirms: **suggestion:** This hashes every TxSet envelope even when `limit` is 1. Since the page's hashes are collected before the envelopes are enumerated, the enumeration could stop as soon as every page hash has been resolved. For small pages that is most of the cost of the call.

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,476 @@
...
+// LedgerSequence/LedgerCloseTime helpers use, but staying view-side.
+// viewChain threads a single error through a sequence of zero-copy view
+// accessors. Once any step fails, later steps short-circuit and the first
+// (labeled) error is retained — collapsing the repeated
+// `v, err := f(); if err != nil { return wrap(err) }` ladder that view
+// navigation otherwise requires into a declarative sequence plus one terminal
+// check. This is the package-local stand-in for the fluent `.HeaderW().…W()`
+// chaining wished for in review; a true fluent API would need SDK-generated
+// wrapper types.
+type viewChain struct{ err error }
```
tamirms: **should-fix:** Two things here. First, the doc block attached to `viewChain` opens with `readLedgerHeader`'s documentation, so the wrong declaration is documented and `readLedgerHeader` has none. Second, the comment says a fluent API "would need SDK-generated wrapper types", but the Go SDK already generates them: the `Must*` accessors chain fluently and [`Try`](https://github.com/stellar/go-stellar-sdk/blob/ff1e140adca5/xdr/xdr_views_generated.go#L102) recovers their failures as errors. With that style this helper disappears entirely.

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,476 @@
...
+// 0..N-1. V3 has no top-level TransactionEvents, so the ledger-wide
+// before/after counters in state are not touched.
+//
+// Note: this view path emits SorobanMeta.Events whenever SorobanMeta is
+// present and relies on the core invariant "SorobanMeta present ⟺
+// soroban transaction" — it does NOT re-check the envelope via
+// IsSorobanTx the way the struct path does. This is why the differential
+// test's V3 fixtures, which always pair a present SorobanMeta with a
+// soroban envelope, agree on both paths.
+func payloadsFromV3SorobanMeta(metaView xdr.TransactionMetaView, state *txWalkState, dst []events.Payload) ([]events.Payload, error) {
```
tamirms, line 160: **question:** `ExtractEvents` emits V3 events whenever SorobanMeta is present, relying on the core invariant, while the read path additionally gates them on the envelope (`gateV3ContractEvents`). On an inconsistent LCM, the events index and `getTransaction` would therefore disagree about the same transaction. That is fine under the trusted-input assumption, but since this is a package-level contract, would you move the invariant statement into `doc.go` so that both halves reference one place?

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,476 @@
...
+	applyIdx       uint32 // 1-based, matches ingest reader's tx.Index
+	beforeIndex    uint32 // ledger-wide BeforeAllTxs event counter
+	afterIndex     uint32 // ledger-wide AfterAllTxs event counter
+}
+
+// txProcessingIterable abstracts over the SDK TxProcessing array view
+// types (V0/V1/V2 LCM use distinct generated views) so the dispatch
+// switches above can share the iteration body. All yield a
+// txResultMetaView with the methods we need.
+type txProcessingIterable interface {
```
tamirms: **nit:** This could be a single generic function, for example `func widen[E txResultMetaView](seq iter.Seq2[E, error]) iter.Seq2[txResultMetaView, error]`, which would drop both the interface and the adapter struct.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
```diff
@@ -0,0 +1,594 @@
...
+	idx := 0
+	for txView, iterErr := range d.tp.iter() {
+		if iterErr != nil {
+			return Transaction{}, false, fmt.Errorf("views: TxProcessing iter: %w", iterErr)
+		}
+		h, herr := readTxHash(txView)
+		if herr != nil {
+			return Transaction{}, false, herr
+		}
+		if bytes.Equal(h[:], hash[:]) {
```
tamirms: **nit:** Both sides are `[32]byte`, so `h == xdr.Hash(hash)` works instead of `bytes.Equal`.

cmd/stellar-rpc/internal/fullhistory/views/events.go
```diff
@@ -0,0 +1,476 @@
...
+func readTxHash(tx txResultMetaView) (xdr.Hash, error) {
+	var c viewChain
+	rp := step(&c, "Result", tx.Result)
+	hv := step(&c, "TransactionHash", rp.TransactionHash)
+	hb := step(&c, "TransactionHash value", hv.Value)
+	if c.err != nil {
+		return xdr.Hash{}, c.err
+	}
+	if len(hb) != 32 {
+		return xdr.Hash{}, fmt.Errorf("views: tx hash length %d != 32", len(hb))
```
tamirms: **nit:** This check can't fire: `HashView` is a fixed opaque[32], so `Value()` always returns exactly 32 bytes on success.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
```diff
@@ -0,0 +1,594 @@
...
+//	carry per-op contract Events. Mirrors
+//	ingest.LedgerTransaction.GetTransactionEvents()/GetDiagnosticEvents().
+//
+// Empty (not nil) slices match db.ParseTransaction's behavior — it always
+// allocates via make([][]byte, 0, len(...)) regardless of whether the
+// source slice was nil, so returning nil here would diverge from the
+// struct path purely on the nil-vs-empty axis.
+//
+//nolint:cyclop,funlen // linear dispatch on meta version
+func extractEventRawsFromMeta(mv xdr.TransactionMetaView) (metaEvents, error) {
```
tamirms, line 396: **nit:** gofmt has reflowed the V0/V3/V4 list in this doc comment into stray indented blocks, which will render as code in godoc. It needs reformatting as a proper list.

cmd/stellar-rpc/internal/fullhistory/views/txdetails_test.go
```diff
@@ -0,0 +1,538 @@
...
+	raw[off] ^= 0xFF
+	assert.Equal(t, raw[off], field[0], "field did not track mutation of raw — not aliased")
+	assert.NotEqual(t, before, field[0])
+}
+
+// indexOfSubslice returns the offset at which sub begins within buf such
+// that buf[off:off+len(sub)] shares backing storage with sub, or -1. Since
+// the view fields are sub-slices of buf, a content match at a unique offset
+// confirms aliasing for our fixtures.
+func indexOfSubslice(buf, sub []byte) int {
```
tamirms: **nit:** A content match would also pass for a copy that happens to match at a unique offset. `events_test.go` proves aliasing correctly with `unsafe.SliceData` pointer containment; could we reuse that helper here?

cmd/stellar-rpc/internal/fullhistory/views/pairing_test.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// TestExtractTransactions_HashPairing_OrderMismatch is the P0 regression
+// guard: the TxSet (agreed-set order) is built REVERSED relative to
+// TxProcessing (apply order). Positional pairing (envs[k] -> parts[k]) would
+// hand each returned Transaction another tx's Envelope/FeeBump; hash-keyed
+// pairing returns each tx's OWN envelope. We assert each returned Envelope
+// equals the envelope wire-bytes for that tx's OWN hash.
+//
+// This test FAILS on the old positional code and passes after the
```
tamirms: **nit:** Several comments narrate review and benchmarking history: "wished for in review", "previously-0%-coverage", "FAILS on the old positional code", and the rpc-hack provenance notes. The PR description already preserves that history. Could the code comments keep just the forward-looking contracts?

tamirms (commented):
I have two main suggestions; the details are in the inline comments.

**1. Error handling.** The package handles errors with an explicit check after every accessor call, plus a hand-rolled `viewChain`/`step` helper. The Go SDK already provides a better option: every view has generated `Must*` accessors that chain fluently, and `xdr.Try`/`TryVoid` convert their failures back into errors (see [the "Try / TryVoid" section of `xdrgen/views_api.md`](https://github.com/stellar/go-stellar-sdk/blob/ff1e140adca5/xdrgen/views_api.md#try--tryvoid)). The `viewChain` doc comment says that a fluent API would need SDK-generated wrapper types, but those already exist. If each exported extractor wraps its body in a single recover boundary, the package shrinks to roughly half its size and the `//nolint:cyclop,funlen` suppressions become unnecessary. Error quality is preserved because `ViewError` already carries the kind, byte offset, and detail. The views documentation says Must methods are intended for trusted input, which is what we have here.

**2. Envelope pairing should not need to decode.** `envPartFromView` calls `UnmarshalBinary` and [`network.HashTransactionInEnvelope`](https://github.com/stellar/go-stellar-sdk/blob/ff1e140adca5/network/main.go#L57) on every envelope in the TxSet, and this dominates the cost of the read path. The decode is avoidable, because all three of its outputs can be derived from the view directly: the envelope type comes from the discriminant, `isSoroban` comes from the Ext discriminant, and the hash can be computed from the wire bytes. Details in the inline comment.

On the SDK-split question from #764, I would revisit the conclusion that nothing belongs in the SDK. View-side envelope hashing, `IsSorobanTx`, and the version-agnostic LCM helpers that this package hand-rolls all mirror existing struct-side SDK APIs: `readLedgerHeader` mirrors [`lcm.LedgerSequence()` and `lcm.LedgerCloseTime()`](https://github.com/stellar/go-stellar-sdk/blob/ff1e140adca5/xdr/ledger_close_meta.go#L21-L29), and the TxSet walkers mirror [`lcm.TransactionEnvelopes()`](https://github.com/stellar/go-stellar-sdk/blob/ff1e140adca5/xdr/ledger_close_meta.go#L62). Keeping copies of that logic here means we have to track SDK behavior changes by hand. The `envelopeIsSoroban` comment already describes itself as "a known drift point to update in lockstep".

cmd/stellar-rpc/internal/fullhistory/views/events.go
Shaptic: Oh yeah 100% this is just a side comment from the actual PR here

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
chowbao: Done in da9ee156. `envPartFromView` no longer decodes the envelope: the hash is computed straight from the wire as `SHA256(networkID ‖ envType tag ‖ innerTx.Raw())`, with the network ID derived once per extraction and a reused SHA-256 state, so per-envelope hashing allocates nothing. TX_V0 hashes as its V1 conversion (4-byte ed25519 key-type prefix + the unchanged V0 bytes — the optional-TimeBounds/Preconditions equivalence is documented on the function), and `isSoroban` now reads the inner `Tx.Ext` view discriminant. Added TX_V0 fixtures with TimeBounds present and absent, paired against the reference path.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
chowbao: Done — `dispatchLCM` (now in dispatch.go) is the single switch, returning the header, the TxProcessing sequence, and a lazily-opened envelope enumerator. `openHeaderAndTxProc` is gone; `ExtractEvents` and `ExtractTxHashes` use the same dispatch and just never touch the enumerator.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
chowbao: Done — `ExtractTransactions` collects the page's hashes first and `envelopesForHashes` stops the TxSet enumeration as soon as every wanted hash resolves.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Done — `viewChain`/`step` are deleted; `readLedgerHeader` and `readTxHash` use the generated `Must*` chains under `xdr.Try`, and the doc comment is back on the right declaration.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Moved into doc.go as the package's trusted-input invariant; `payloadsFromV3SorobanMeta` and `gateV3ContractEvents` both reference it now.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Done — a generic `widen[E]` replaces the interface + adapter struct.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
chowbao: Done.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Removed — `readTxHash` converts with `xdr.Hash(hb)` directly.

cmd/stellar-rpc/internal/fullhistory/views/txdetails.go
chowbao: Reformatted as a proper bullet list.

cmd/stellar-rpc/internal/fullhistory/views/txdetails_test.go
chowbao: Done — `assertAliasesRaw` now proves aliasing by `unsafe.SliceData` pointer containment; the content scanner is gone.

cmd/stellar-rpc/internal/fullhistory/views/pairing_test.go
chowbao: Swept — all of them are gone (the "wished for in review" one went with `viewChain` itself).

cmd/stellar-rpc/internal/fullhistory/views/envelopes.go
chowbao: Closing the loop: the decode is gone entirely now — the hash preimage is built from the wire bytes directly (da9ee156), so neither the decode nor the round-trip exists anymore.

cmd/stellar-rpc/internal/fullhistory/views/events.go
chowbao: Turns out the SDK already generates this: `Must*` accessors panic with `*ViewError` and `xdr.Try` recovers them. da9ee156 switches to that style and deletes the local helper.

Shaptic: Bruh I keep forgetting that that exists

chowbao:
Closing in favor of **stellar/go-stellar-sdk#5949** + **#779**.

The conclusion of this PR ("nothing in this task belongs in the SDK") was reversed: the zero-copy XDR view extractors are general-purpose XDR navigation — the view twins of the parsed `ingest.LedgerTransaction` / `LedgerTransactionReader`, `network.HashTransactionInEnvelope`, and `TransactionResult.Successful` — so they were **promoted into the go-stellar-sdk** (stellar/go-stellar-sdk#5949):

- `ingest`: `DispatchLedgerCloseMetaView`, `TransactionEventsView` + `TransactionEventsFromMeta` / `DiagnosticEventsFromMeta`, `TransactionView` + `TransactionViewByHash` / `TransactionViewRange`
- `network`: `TransactionViewHasher`
- `xdr`: `LedgerCloseMetaView.LedgerCloseTime`, `TransactionResultView.Successful`

validated wire-identical against the parsed path across LCM V0/V1/V2 and `TransactionMeta` V0–V4.

The remaining RPC-specific glue (the thin `views` adapters — `ExtractEvents` / `ExtractTxHashes` and the `events.Payload` / `txhash.Entry` shapes, plus the `EventIdx` removal) is folded into **#779**, which has been retargeted onto `fh-ingest-base` and now consumes the SDK directly. Nothing here is lost — it lives in #5949 and #779.

Issue #728, opened 2026-05-05
TxHash cold store — streamhash index (build + read)


karthikiyer56:
## TL;DR
The cold (immutable-file) half of the txhash store: **build** a streamhash MPHF index per group of chunks, and **read** `getTransaction(byHash)` point lookups against it. Build input comes from the cold txhash ingester in **#765**. Counterpart to the cold ledger reader (#725) and events cold-segment reader (#657).

## Index format
Cold tx hashes are grouped into indexes of a configurable number of chunks each (default 1,000) — one streamhash MPHF per group, in its own file, over the `(txhash, ledgerSeq)` pairs in that range. Each tx hash maps to a fixed ledger seq stored directly in streamhash's per-key payload, so the index file is self-contained. A small per-key fingerprint rejects most MPHF false positives, with downstream txhash-in-LCM verification catching the rest.

## Build
For each index, merge the per-chunk sorted tx hash files produced by #765's cold txhash ingester (over the index's chunk range) into one streamhash index via the sorted-mode builder — keys arrive pre-sorted, so the build is single-pass.

## Read
Resolve a txhash to its transaction: look up the ledger seq in the streamhash index, read that ledger's LCM (cold ledger reader, #725), and pull the transaction's envelope / result / meta via the tx-details-by-hash view (#764). An index miss returns not-found; the tx-details view also rejects MPHF false positives — a candidate ledger that doesn't actually contain the txhash. rpc-hack `cold_format.go` / `cold_reader.go` are the starting point for the index lookup.

## Acceptance
- [x] Build produces a per-index streamhash index from #765's per-chunk tx hash files.
- [x] Lookups return the correct `ledgerSeq` for indexed tx hashes; clean not-found for misses.
- [x] MPHF false positives are caught (fingerprint + downstream verification).

## Dependencies
- Build input: cold txhash ingester (#765).
- Read path: tx-details-by-hash view (#764) to extract the transaction from the looked-up ledger; cold ledger reader (#725) to fetch that ledger's LCM.

Issue #725, opened 2026-05-05
Ledger query path over packfiles (immutable-file reader)


tamirms:
## TL;DR

Build the ledger-flavored query layer that turns packfile reads (decoder primitive #721) into `getLedger(seq)` / `getLedgers(start, end)` responses over the immutable-file half of the ledger store.

Symmetric counterpart to the RocksDB-backed reader in #584, but for the on-disk frozen artifacts produced by the ledger packfile writer in #695. The events stack already has this shape via "[events] Cold segment reader" (https://github.com/stellar/stellar-rpc/issues/657); this is the ledger version.

## What to build

A ledger reader package that:
- Opens N pack files (one per chunk, as produced by #695).
- Maps `ledgerSeq` → `(packfile, position)` using the file naming or an index sidecar.
- Decompresses and decodes via the packfile decoder primitive (#721).
- Returns `xdr.LedgerCloseMeta` with the same interface shape used by #584 so a coordinator can fall through from hot to cold without per-backend logic at the call site.

## Acceptance

- [ ] `GetLedger(ctx, seq)` over packfiles.
- [ ] `GetLedgers(ctx, start, end)` with cross-packfile spans.
- [ ] Round-trip tests writer (#695) → reader.
- [ ] Interface compatible with #584 so call sites don't branch on backend.

## Dependencies

- Blocked by: #721 (packfile decoder primitive), #695 (ledger packfile writer — for round-trip tests).

## Part of

Full-History Ledger Store epic #724.

tamirms:
Closed by #755. Cold reader query path lives in `ColdReader.GetLedgerRaw` / `ColdReader.IterateLedgers`.

PR #780, opened 2026-06-08
txhash: cold-store streamhash index — build + read (#728)

tamirms:
Part of #728 — implements the buildable slice: the cold (immutable-file) half of the txhash store.

> **Base:** stacked on `events-eventstore-core` (#756) for `github.com/stellar/streamhash`. Retarget to `feature/full-history` once #756 lands.

## What this adds
A per-index streamhash MPHF over `(txhash, ledgerSeq)` and a `ColdReader` that resolves a tx hash to the ledger it was committed in.

- **`cold_format.go`** — per-index on-disk format: 3-byte ledger-seq payload (offset from `MinLedger`), 1-byte fingerprint, and an 8-byte `[MinLedger, MaxLedger]` coverage anchor in the index metadata. `ColdReader.Get(hash) -> ledgerSeq`, `ErrNotFound` on miss, idempotent mmap `Close`; `MinLedger()`/`MaxLedger()` report the index's ledger span so a reader can pick an index without probing it. Rejects a payload width ≠ `ColdPayloadSize` at open.
- **`cold_index.go` + `cold_merge.go`** — `BuildColdIndex(min, max)` merges per-chunk sorted `.bin` files into one streamhash index via a parallel **O_DIRECT fan-in merge** (ported from streamhash `cmd/bench`), fed single-pass into the sorted-mode builder so I/O, merge CPU, and the MPHF build overlap. Validates that the coverage span fits the payload and **every entry falls in `[min, max]`** — so the advertised coverage is a real invariant, not just metadata. Header/size guards, `ctx` cancellation, first-error-wins pipeline.
- **`odirect_{linux,other}.go`** — O_DIRECT page-cache bypass on Linux (cached-open fallback on tmpfs/EINVAL; no-op elsewhere).

## Tuning (measured on warm macOS + cold Linux NVMe over 382M real keys)
- streamhash block-build workers default to **NumCPU/2** (the end-to-end gate; ~2.7× over serial).
- merge leaves capped at **NumCPU/2**: NumCPU/2 leaves + NumCPU/2 builders fill the cores without oversubscription — ~+18% e2e on cold NVMe vs NumCPU, neutral on warm.
- k1 merge tiebreak **kept** for byte-reproducible output regardless of input order (k0 alone already satisfies streamhash's block routing).
- Sorted k-way merge kept over `NewUnsortedBuilder` (measured 1.7–6× faster for the pre-sorted inputs).

Peak build memory is bounded by **file count, not entry count** — ~145 MiB for a full 1000-chunk index, ~constant from 5M to 50M+ keys (the index file is `pwrite`-streamed, not held in RAM).

## Out of scope (blocked on unbuilt deps)
- Production build wiring to the cold txhash ingester (**#765**) — the `.bin` input format is the documented seam for it.
- The `getTransaction` read assembly over the tx-details-by-hash view (**#764**) + cold ledger reader (**#725**). `ColdReader` stops at `ledgerSeq`.

## Tests
Build/query round-trip, miss, concurrent reads, fan-in tree (sorted + complete), large-file refill, ledger coverage (metadata + per-entry bounds), and error/format guards. `go test -race` + `golangci-lint` clean. No benchmarks ship — the tuning sweeps that produced the figures above were removed once the decisions were settled.


cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_index.go
```diff
@@ -0,0 +1,173 @@
...
+// streamhash MPHF. The merge is in cold_merge.go.
+//
+// .bin input format, one file per chunk named <chunkID:08d>.bin:
+//
+//	 8 bytes  entry count  (uint64 LE)
+//	then, per 20-byte entry:
+//	16 bytes  txhash[:16]  (the key; streamhash routes on this)
+//	 4 bytes  ledger seq   (uint32 LE, absolute)
+//
+// Entries within a file must be sorted ascending by the big-endian uint64
```
chowbao, line 15: Is there a reason we use both little and big endian?

I wonder if everything can just be big-endian just for consistency

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_index.go
```diff
@@ -0,0 +1,173 @@
...
+			return 0, err
+		}
+		total += count
+	}
+	return total, nil
+}
+
+// scanBinHeader opens path, reads its declared entry count, and verifies
+// the file is exactly binHeaderSize + count*binEntrySize bytes.
+func scanBinHeader(path string) (uint64, error) {
```
chowbao, line 151: Is it worth doing a distinct on top of the count == total comparison?

I guess it's a tradeoff of absolute correctness (no duplicate txhash) vs most likely correctness (count == total and assume there are no duplicates). Realistically the only time a problem would happen is if there is an equal number of duplicate txhash and skipped txhash which would be extremely unlikely

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_reader.go
```diff
@@ -0,0 +1,88 @@
...
+			return 0, stores.ErrNotFound
+		}
+		return 0, fmt.Errorf("txhash: cold index query: %w", err)
+	}
+	return r.minLedger + uint32(payload), nil //nolint:gosec // 24-bit payload (width enforced at open)
+}
+
+// CoveredRange returns the inclusive [minLedger, maxLedger] the index covers.
+// Immutable; safe after Close.
+func (r *ColdReader) CoveredRange() (uint32, uint32) {
```
chowbao: nit: Hmm in other code we normally have a min and max function right? For completeness it might be useful to make func for minLedger() and maxLedger()

chowbao (approved):

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_reader.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, fmt.Errorf("txhash: open cold index %s: %w", path, err)
+	}
+	return &ColdReader{idx: idx, minLedger: minLedger, maxLedger: maxLedger}, nil
+}
+
+// Lookup returns the ledgerSeq the hash was committed in, stores.ErrNotFound
+// if it wasn't in the build set (residual fingerprint false positives are
+// caught downstream), or ErrClosed after Close. streamhash keys on the first
+// 16 bytes, so the full 32-byte hash and hash[:16] are equivalent.
+func (r *ColdReader) Lookup(hash [32]byte) (uint32, error) {
```
chowbao: nit: Does it make sense to name this `Get()`? It seems like the hot path calls this Get()

https://github.com/stellar/stellar-rpc/blob/feature/full-history/cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/hot_store.go#L157

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_index.go
tamirms: Good thing to check. streamhash already gives us this — its sorted builder rejects out-of-order keys (`ErrUnsortedInput`) and the MPHF solver rejects equal keys (`ErrDuplicateKey`), so a duplicate txhash across the chunk files fails the build rather than silently corrupting the index; `feedMergedKeys` surfaces it as an error.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_index.go
tamirms: Fair question — I went back and forth on it and decided to keep the on-disk integers little-endian. The `.bin` files are an internal scratch format written and read on the same node, and every target we run on (amd64 + arm64) is little-endian native, so LE is a zero-cost load while BE adds a byte-swap; it's also the usual convention for internal binary formats (RocksDB, Arrow), and streamhash itself reads keys little-endian. The big-endian you see in the merge is only how we interpret the raw key *bytes* for lexicographic ordering (to match streamhash's block routing) — a comparison detail, not a stored integer. So the stored format (count, seq, and the `[min,max]` metadata) is uniformly little-endian. Happy to revisit if these files ever need to be portable across architectures.

PR #783, opened 2026-06-10
Simulate tx auth v2 flag

Ryang-21:
### What

  Add an optional `useUpgradedAuth` boolean to the `simulateTransaction` request. When set, authorization entries recorded during simulation use Soroban `AddressV2` ("v2") credentials instead of `Address` ("v1"). The flag is threaded from the request through the preflight worker pool and across the CGO bridge into the Rust simulation path. Also bumps the current-protocol `soroban-env-host` / `soroban-simulation` dependencies to `27.0.0`.

### Why   

  [rs-soroban-env#1689](https://github.com/stellar/rs-soroban-env/pull/1689) added a recording-auth parameter (`use_address_v2`) that selects v1/v2 `Address` credential recording independently of the ledger protocol version. As stated in that PR, the intent is to "allow RPC to select the preferred mode when calling simulation … without a hard protocol-based switch" — so v2 credentials can be exercised through RPC for testing *before* the protocol that requires them activates. This change surfaces that capability as a request-level opt-in.

  The flag only has an effect in the recording auth modes (`record` / `record_allow_nonroot`); in `enforce` mode the caller supplies their own auth entries, so the credential format is already fixed.
  
### Known limitations

  - Requires the new `SimulateTransactionRequest.UseUpgradedAuth` field in `go-stellar-sdk` (this [pr](https://github.com/stellar/go-stellar-sdk/pull/5948)). Until that lands in a release and `go.mod` is bumped, the field is supplied via a local `replace` directive that is intentionally not committed — so this branch does not build in CI on its own yet.
  - v2 credentials are only emitted by the current soroban-env host. On protocols served by the previous host (which predates v2), `useUpgradedAuth` is silently ignored and v1 credentials are produced. This is by design, not an error.
  - The end-to-end integration test (`TestSimulateInvokeContractTransactionuUseUpgradedAuth`) requires a captive-core running protocol ≥ 27 and is skipped otherwise; the unit/handler tests run without it.


Shaptic:
This is gonna fail CI until you actually reference the correct version of the SDK - `go get -u github.com/stellar/go-stellar-sdk@main` after https://github.com/stellar/go-stellar-sdk/pull/5948 merges

Shaptic (approved):

PR #786, opened 2026-06-14
Cache preflight snapshot source lookups

mihaieremia:
Discussion: https://github.com/stellar/stellar-protocol/discussions/1958

### What

Adds a per-preflight cache around `SnapshotSourceGet`, including cached misses, with focused unit coverage for cache hits and misses.

### Why

External read-only storage can make one simulated transaction read the same target market, governance, reserve, oracle, or contract-instance keys repeatedly. In lending-style flows, a controller may touch several supplied and borrowed assets while repeatedly checking the same market configuration or oracle state.

Preflight already routes Rust host ledger-entry requests through `SnapshotSourceGet`. Caching those lookups for the duration of one preflight avoids repeated Go/Rust boundary work and repeated ledger-entry getter calls for the same keys. This is especially useful for external instance reads because the host records the target contract instance entry as the footprint key.

This does not change production transaction semantics, ledger state, auth, footprint contents, or result formatting. It only avoids duplicate snapshot fetch work during simulation/preflight.

### Known limitations

- Cache lifetime is one preflight only.
- This PR does not add the external-storage ABI to RPC by itself; it is a simulation/preflight companion optimization for the env and SDK PRs.
- Full external-storage simulation assertions should be added once RPC consumes an env/preflight dependency that includes the new protocol-28 ABI.

### Validation

- `go test ./cmd/stellar-rpc/internal/preflight -run 'TestSnapshotSourceGetCaches|TestGetPreflight'`
- `gofmt -l cmd/stellar-rpc/internal/preflight`
- `git diff --check`


cmd/stellar-rpc/internal/preflight/preflight.go
```diff
@@ -50,40 +99,53 @@ func SnapshotSourceGet(handle C.uintptr_t, cLedgerKey C.xdr_t) C.ledger_entry_an
...
+		return cachedLedgerEntryAndTTL{}
 	}
 	entries, _, err := h.ledgerEntryGetter.GetLedgerEntries(h.ctx, []xdr.LedgerKey{ledgerKey})
 	if err != nil {
 		h.logger.WithError(err).Error("SnapshotSourceGet(): GetLedgerEntries() failed")
-		return C.ledger_entry_and_ttl_t{}
+		return cachedLedgerEntryAndTTL{}
 	}
 	if len(entries) > 1 {
 		h.logger.WithError(err).Error("SnapshotSourceGet(): GetLedgerEntries() returned more than one entry")
```
leighmcculloch: The "GetLedgerEntries() returned more than one entry" log line reuses the stale `nil` `err` from the prior (successful) `GetLedgerEntries` call instead of describing the actual anomaly (there's no new error to report here — the condition is `len(entries) > 1`, not an error return). If a `LedgerKey` ever legitimately maps to >1 entry, the emitted log always shows a nil error object, which is useless for diagnosing why it happened. Pre-existing behavior, but this PR rewrites the whole function so it's a good opportunity to fix it (e.g. log the entries themselves, or drop the `err` field).

---
_Generated by [Claude Code](https://claude.ai/code)_

cmd/stellar-rpc/internal/preflight/preflight.go
```diff
@@ -50,40 +99,53 @@ func SnapshotSourceGet(handle C.uintptr_t, cLedgerKey C.xdr_t) C.ledger_entry_an
 		panic("invalid handle type: expected snapshotSourceHandle")
 	}
 	ledgerKeyXDR := GoXDR(cLedgerKey)
+	return ledgerEntryAndTTLToC(h.getLedgerEntryAndTTL(ledgerKeyXDR))
+}
+
+func (h snapshotSourceHandle) getLedgerEntryAndTTL(ledgerKeyXDR []byte) cachedLedgerEntryAndTTL {
+	cacheKey := string(ledgerKeyXDR)
+	if h.cache != nil {
```
leighmcculloch: This `h.cache != nil` guard (also repeated around lines 127 and 145) looks unreachable in production: both call sites (`getFootprintTTLPreflight`, `getInvokeHostFunctionPreflight`) always build the handle via `newSnapshotSourceHandle`, which unconditionally sets a non-nil cache. Not a functional bug, but it's a dead branch that no test exercises (nothing constructs a `snapshotSourceHandle` with `cache == nil`), and it could mislead a future refactor into thinking a nil cache is a supported state. Consider dropping the nil checks or making `cache` non-optional in the type.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #787, opened 2026-06-15
docs(full-history): split getTransaction design into its own doc; reorganize design-docs

tamirms:
## What this delivers

Net of this branch against `feature/full-history`: three new design docs under
`design-docs/`, plus removal of the stale numbered-doc README — written in a
plain, concept-first voice that leans on the pseudocode rather than
re-narrating it in prose.

- **`full-history-streaming-workflow.md`** — the streaming daemon design:
  geometry, the catalog and the one write protocol, backfill
  (postcondition-driven planning → primitives → execution model), hot-DB
  ingestion, the lifecycle run (freeze → rebuild → discard → prune), the
  reader contract, and the correctness invariants (INV-1…4) with their audits.
- **`gettransaction-full-history-design.md`** — the transaction-by-hash
  subsystem end to end: the hot `txhash` CF, the `.bin`/`.idx` streamhash
  formats, the rolling window-index rebuild, the `getTransaction` query path,
  and capacity numbers. Scoped to storage / rebuild / query — the crash-safety
  machinery lives in the streaming doc. The streaming doc references it for
  tx-index depth, mirroring how the events subsystem already has its own doc.
- **`full-history-design-explorer.html`** — a self-contained interactive
  companion to both docs (no external dependencies; open in a browser), with a
  widget per concept: geometry, the artifact lifecycles and one write protocol,
  a crash simulator, the derived last-committed-ledger, the rolling index, a
  chunk boundary end to end, the resolver and `executePlan`, the startup
  backfill loop, the reader contract, and the invariants.

The full-history design docs are consolidated into the top-level `design-docs/`
alongside the existing events and packfile docs, with the numeric prefixes
dropped. `full-history/design-docs/03-backfill-workflow.md` is unchanged vs.
this base (retained as the historical standalone-backfill design; the streaming
doc records that it is subsumed).

Docs only — no code changes.

## Notes for review
- Easiest to review by reading the rendered markdown / opening the HTML
  explorer rather than the raw diff.


chowbao (approved):
LGTM

I don't think these need to go in the design but I had a few comments/questions

* Verification process (will probably defer to a separate issue/PR)
  * The design assumes that the data backfilled/ingested is complete and correct but there aren't any verification steps in place checking if the LCM contents are correctly ingested or not. For example, checking if for a given LCM the number of transactions ingested == the number of transactions marked as within the LCM
* There are ledger within the Stellar network that have zero transactions (very early ledgers around genesis and when the network was live but not very active). Zero transaction ledgers would also exist now/in the future if core keeps running but maybe there is no way for people to submit transactions to the network. We should check to make sure when backfilling/frontfilling that the system knows the difference between missing and empty data like txhash rolling index is okay with empty .bin files
* One clarification with backfill and frontfill on first startup

> Backfill has a contract — given a range, ensure every artifact derived from every ledger in it is durable and servable — and resolves what's missing before scheduling anything

> skipping the tip chunk that captive core is actively ingesting. Covers first-ever start, downtime gaps, and retention widening.

What happens in the case that while backfilling the live chunk moves to the next chunk. Is it possible to have multiple hot chunks ready to freeze?
`and resolves what's missing before scheduling anything` does this mean RPC won't serve/run until the backfill is complete? In that case doesn't it make sense to just start captive core after the backfill range is complete instead of skipping the tip/live chunk?

tamirms:
@chowbao 

I agree with you on the verification process. I think we'll need a separate design doc (can be more lightweight like a github issue) to scope out the requirements there.

I also looked into the 0 transaction case. If there is an entire 10 million consecutive ledger range where there are 0 transactions (could be the case in a new stand alone network), this would create a problem because streamhash does not support indexes where there are 0 keys. But, we can fix this easily in streamhash by allowing a 0 key index which would return not found on every query.


> What happens in the case that while backfilling the live chunk moves to the next chunk

The backfill operation is run in a loop which should handle the case of the tip advancing to the next chunk during the backfill operation:

```go
	earliest := cat.EarliestLedger()
	lastCommitted := lastCommittedLedger(cat)

	// Step 1: backfill from the floor up to the last complete chunk at the tip,
	// leaving the partial tip chunk to ingestion. Re-pass while the tip moves.
	backfilledThrough := int64(-1)
	for {
		tip, err := networkTip(cfg)
		if err != nil {
			if lastCommitted < earliest {
				fatalf("network tip unavailable and no local history to serve: %v", err)
			}
			tip = lastCommitted // backend down, but local data exists: serve it
		}
		anchor := max(tip, lastCommitted)
		rangeEnd := lastCompleteChunkAt(anchor)
		rangeStart := retentionFloorChunk(rangeEnd, cfg.RetentionChunks, earliest)
		midChunk := lastCommitted != chunkLastLedger(chunkID(lastCommitted))
		nearTip := int64(tip)-int64(lastCommitted) < LedgersPerChunk
		if nearTip && midChunk {
			rangeEnd = chunkID(lastCommitted) - 1 // leave the partial resume chunk to ingestion
		}
		if rangeEnd < rangeStart || rangeEnd <= backfilledThrough {
			break
		}
		if err := executePlan(ctx, cfg, resolve(cfg, rangeStart, rangeEnd)); err != nil {
			return err
		}
		lastCommitted = max(lastCommitted, chunkLastLedger(rangeEnd))
		backfilledThrough = rangeEnd
	}
	resumeLedger := lastCommitted + 1
```

> Is it possible to have multiple hot chunks ready to freeze?

During the live ingestion phase (after we finished backfilling), it is possible to have hot chunks which are ready to freeze but it is very unlikely. Each freeze would probably complete in a matter of minutes and it takes 14 hours for a hot chunk to reach its capacity. For there to be backlog of hot chunks it means the system is in a bad / degraded state and would need investigation.

> does this mean RPC won't serve/run until the backfill is complete? In that case doesn't it make sense to just start captive core after the backfill range is complete instead of skipping the tip/live chunk?

yes, RPC won't serve any requests until the backfill loop is complete. But, I don't understand the second question. After the backfill step is complete, we do actually start up captive core to do the live ingestion phase.

chowbao:
> > does this mean RPC won't serve/run until the backfill is complete? In that case doesn't it make sense to just start captive core after the backfill range is complete instead of skipping the tip/live chunk?
> 
> yes, RPC won't serve any requests until the backfill loop is complete. But, I don't understand the second question. After the backfill step is complete, we do actually start up captive core to do the live ingestion phase.

I should worded this better. If I'm backfilling from genesis to now I would assume that the actual current ledger would be at least a chunk or two ahead of the "now" (`the tip chunk that captive core is actively ingesting`). I was just suggesting that we can backfill as close to current as possible instead of spinning up captive core for the old current ledger tip.

In other words
* I want to spin up a new rpc v2 from genesis to now. "now" in this example can be like 1,000,000
* Backfilling from 2 to 1,000,000 takes some arbitrary amount of time backfilling with BSB
* By the time backfill BSB gets to 1,000,000 the current ledger might actually be 1,050,000
* The way the design is worded right now makes me think that captive core would have been started at 1,000,000 which means that captive core is ingesting an extra 5 chunks slowly instead of just letting BSB do those 5 chunks

This is a small optimization. This probably doesn't matter in the long run because rpc v2 wouldn't start serving until the backfill is complete and caught up anyways. But I just wanted to call it out

Edit: Oh actually the code in your comment might have handled this scenario cause that code would run until it is in the latest chunk which at that point we would start up captive core

urvisavla (approved):

design-docs/gettransaction-full-history-design.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+The two tiers hand off with no gap. A chunk's hot table is dropped only *after* the cold index covers that chunk. So a freshly frozen chunk keeps being answered from its hot table until the index can answer for it, and only then does the hot table go away. Every transaction is findable in exactly one tier at all times.
+
+## 4. Geometry
+
+Two units organize the map. Every structure below is named by them:
+
+- **Chunk** — 10,000 ledgers (hardcoded). The unit of the hot DB and of the sorted runs.
+- **Window** — 1,000 chunks = 10,000,000 ledgers (hardcoded). The unit of the cold index.
```
karthikiyer56, line 59: The window or (`chunksPerTxIndex` as it was previously named) is decidedly fixed to be 10million ledgers? instead of being configurable during the first ever run and then fixed forever?

Wouldnt this have implications for pruning or for  initial startup for users who dont want a big retention window? Wont they have a rather long startup time?
(I am yet to read the streaming doc in this case the answer to this question is answered there)

design-docs/gettransaction-full-history-design.md
```diff
@@ -0,0 +1,253 @@
...
+
+Storing the full hash makes the hot tier **exact**: a lookup either finds the hash or it doesn't. There are no false positives to screen out and nothing to verify. The table is tuned for point lookups — bloom filters on, ordering off.
+
+### 5.2 Write path
+
+Writing is straightforward. As each ledger is ingested, one `(hash, seq)` entry is added for every transaction in it, in the same atomic write that stores the rest of the ledger. So a ledger's hashes are written all-or-nothing, together with the rest of the ledger.
+
+### 5.3 Lifetime
+
+A chunk's hot table lives from the moment the chunk starts ingesting until the cold index covers it. Coverage can lag the chunk's freeze by a while; until it lands, the chunk is simply answered from its hot table.
```
karthikiyer56, line 92: > A chunk's hot table lives.....

this is slightly misleading.
the hot table/store for the txHash lookup is at the cadence of the window(TxIndex).

Also.
I am not sure what `Coverage can lag the chunk's freeze by a while; until it lands, the chunk is simply answered from its hot table.` means here.

When I am reading this, it seems like chunk is front and center cadence wise when it comes to freezing a txIndex - which is not the case. 


Perhaps this entire section can be replaced with
```
A window/txIndex's hot table lives from the moment the first ledger in the first chunk in the window starts ingesting till the last ledger in the last chunk in the window is done ingesting ANd then the cold index for the entite window is frozen
```

design-docs/gettransaction-full-history-design.md
tamirms: 1,000 chunks is more of an upper bound. The coverage of a tx hash index never extends below the retention window and if your retention is smaller than a window, the coverage gets clamped to it.

So let's say your retention is 2 chunks (20,000 ledgers) and the latest ledger in the network is 65,000,000, which falls in chunk 6499. At 1,000 chunks per index, chunk 6499 belongs to window 6, so the nominal coverage would be [6000, 6999]. But 6000 is way below the retention floor, so the coverage gets clamped to [6498, 6499]. When the tip advances into chunk 6500, the index is rebuilt with coverage [6499, 6500] and chunk 6498 is pruned away. This also means we always rebuild the tx hash index after completing each chunk.

The formula for the coverage is [max(retention_floor, tip - 1,000), tip].

So to your questions: pruning just falls out of the per-chunk rebuild — each rebuild drops the chunk that aged past the floor. And startup isn't a problem either, since a node with 2 chunks of retention only backfills 20,000 ledgers, not 10 million. The 1,000 chunk number just caps how many chunks a single index can merge

design-docs/gettransaction-full-history-design.md
tamirms: one of the changes compared to the previous design is that all the hot data is now scoped by chunk and lives in a single rocks db instance per chunk — events, tx hash data, and ledgers each get their own CF in that same rocks db. so the tx hash store no longer holds an entire window's worth of data, it only holds a single chunk's worth.

separately there's the tx hash index, which is rebuilt after every chunk completes and covers all the transactions in the current window except the one hot chunk. so the hot store handles the chunk that's still being written, and the index handles everything behind it in the window.

karthikiyer56 (approved):

PR #791, opened 2026-06-17
Add GHA coordinator for performance evaluation task scatter/gather

### LINKED ISSUE #707: Release Eval: Add the coordinator GitHub Action
Parent: #706 

## Summary

Add an overarching GitHub Action that runs the release evaluation suite for `release/**` branches and coordinates benchmark, ingestion, request-load, and reporting jobs.

## Scope

- Create `.github/workflows/release-evaluation.yml`.
- Trigger on `push` to `release/**` and `workflow_dispatch` with inputs for release ref, baseline ref, profile, and threshold mode.
- Run on GitHub-hosted `ubuntu-24.04` runners.
- Reuse existing repo setup patterns: checkout, `.github/actions/setup-go`, `rustup update`, `stellar/actions/rust-cache`, and `make build-libs` before Go builds/tests.
- Build the `stellar-rpc` binary once per job as needed and pass generated artifacts between jobs.
- Coordinate jobs for endpoint benchmarks, benchmark comparison, mainnet backfill ingestion, Core `apply-load` synthetic ingestion, `stellar-rpc-blaster` request load, and final reporting.
- Add workflow-level concurrency keyed by workflow and ref, matching existing workflow behavior.

## Acceptance criteria

- The workflow can be started manually for any ref and automatically starts for `release/**` pushes.
- Threshold values are centralized in workflow inputs or checked-in config with TBD defaults where the spec does not yet define values.
- Each child job uploads raw logs/metrics and exposes a concise status to the final report job.
- A failing release gate fails the workflow; missing TBD thresholds produce an explicit neutral/warning state until configured.

## Notes

- Existing workflows already use `release/**` for Go, Stellar RPC, and e2e checks; this workflow should follow the same branch prefix.
- GitHub-hosted runners are noisy for performance measurement, so the workflow should favor repeatable setup, explicit runner metadata, and conservative interpretation. We will use a dedicated runner in the future.

cjonas9:
### What

Adds a GitHub Actions coordinator (`load-test-coordinator.yml`) that, on a push to a `release/**` branch, launches the release performance-evaluation leg(s) as callable workflows and posts their consolidated results as a sticky comment on the release PR. The apply-load ingest load test (#741) is adapted into a callable leg that reports back (outputs + an S3 result object) instead of commenting itself:
image

Decomposes a great deal of the existing infrastructure into more modular components for future coordinator-driven load test. Of the diff, **+870/-721** of it is **purely refactoring existing code** into a new `harness` package, so review of it does not necessarily have to be as thorough as coordinator-heavy sections. Here is the breakdown of that:
- Generic code lifted out of `runner/orchestrate.go` and `runner/instantiate.go` into an importable `harness` package, mechanical changes only
    - package consists of `gather.go`, `s3.go`, `exec.go`, `harness.go`, and `harness_test.go`
- Generic code lifted out of `run-load-test.sh` and placed into `bootstrap-common.sh` mostly as helper functions
    - `run-load-test.sh` is now extremely slim and only serves to actually run the ingestion load test
- Generic code lifted out of `load-test.yml` and put in `ec2-leg.yml`, which handles the EC2 lifecycle with no test-specific hardcoding

Results of the load tests are commented on a sticky comment on the `release` PR. If >1 push is made/multiple runs are requested, previous run's results are folded into dropdown menus like this:

---------------
🧪 Performance Evaluation Test #N...
`[recent results for run N]`
<details>
<summary>Performance Evaluation Test #N-1</summary>
`[results for run N-1]`
</details>

<details>
<summary>Performance Evaluation Test #...</summary>
`[results for run ...]`
</details>

<details>
<summary>Performance Evaluation Test # 1</summary>
`[results for run 1]`
</details>

---------------

Or, you can just look in the comments below to see how this looks!

The high-level flow of the coordinator is below:
```mermaid
flowchart TB
  trigger["push to release/** "] --> plan["perf-eval-coordinator.yml<br/>- plan<br/>- resolve ref + PR"]
  plan --> lt["load-test.yml<br/>runs synthetic ledger ingestion test"]
  plan -.-> future["future legs"]
  lt --> report["report<br/>aggregate → sticky PR comment"]
  future -.-> report
```

### Why

This is so that the remaining perf-eval tests slot in cleanly: each new test becomes another callable leg the coordinator launches and folds into the same report, with no per-leg reporting or permissions plumbing to duplicate.

### Known limitations

N/A. Remaining work for the epic this task is a part of mainly includes finishing off the other perf eval tasks (only the ingest load test is complete at this point).


.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+    needs: plan
+    uses: ./.github/workflows/load-test.yml
+    secrets: inherit
+    with:
+      target_ref: ${{ needs.plan.outputs.target_sha }}
+
+  report:
+    name: Aggregate + report
+    needs: [plan, load-test]
+    if: always()
```
tamirms: **[should-fix]** The coordinator's `cancel-in-progress: true` combined with `if: always()` here has a surprising interaction: jobs guarded with `always()` still run after a workflow run is cancelled (that's documented behavior — the force-cancel API exists precisely because a normal cancel doesn't stop `always()` jobs; see https://github.blog/changelog/2023-09-21-github-actions-force-cancel-workflows/ and https://github.com/orgs/community/discussions/26303). So when push B supersedes push A's in-flight run, A's report job still executes, sees the cancelled leg's empty outputs, and edits the sticky comment with a "Performance Evaluation Test #N — ❌ No result object published" entry. B's run then folds that entry into history permanently. A couple of quick consecutive pushes to a release branch would accrete false-failure entries (and burn test numbers).

Would `if: ${{ !cancelled() }}` work here instead? It still runs the report when a leg genuinely fails, but skips it when the run was superseded.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/coordinator-runner.go
```diff
@@ -0,0 +1,215 @@
...
+		b.WriteString("\n" + olderStr + "\n")
+	}
+	return b.String()
+}
+
+// splitAtFirstDetails partitions lines at the first "<details>" line: the lines
+// before it (the prior run's own results) and the rest (older runs' drop-downs).
+func splitAtFirstDetails(lines []string) ([]string, []string) {
+	for i, line := range lines {
+		if line == "<details>" {
```
tamirms: **[should-fix]** Splitting the previous body at the first literal `<details>` line works today because `renderPerfMarkdown` only emits tables, but it silently corrupts the comment the moment any leg's markdown contains its own drop-down — which feels like a natural formatting choice for the longer output future legs will produce. I reproduced it with:

```go
legMD := "table\n\n<details>\n<summary>per-profile breakdown</summary>\n\ninner\n</details>"
run1 := renderComment(commentInput{legs: okLeg(legMD), ...})
run2 := renderComment(commentInput{legs: okLeg(legMD), prev: run1, ...})
```

In `run2`, the "Test #1" drop-down is truncated at the inner `<details>` line and the leg's own block escapes as a fake top-level sibling entry after it; the corruption compounds on every subsequent run, and nothing would flag it except a garbled comment.

I think the underlying issue is that the renderer reverse-engineers its own previous output — the marker regexes, the heading regex, and this split are all parsing presentation to recover state. The direction I'd push for: store the history as data and make the comment a pure render of it. Either a base64 JSON blob embedded in an HTML comment inside the sticky comment, or a small S3 state object keyed by PR (`prs/<pr>/history.json`) with the comment derived from it. That eliminates the line-oriented parsing entirely (this whole class of bug goes away), makes run numbering `len(history)+1` instead of regex extraction, and makes capping history a slice operation. Since no production comments exist yet, there's no legacy format to stay compatible with — this is as cheap as it will ever be.

If that feels like too much for this PR, the minimal fix is an explicit sentinel line (e.g. `<!-- perf-eval-history -->`) between the current run and the folded history, split on that instead of `<details>`.

Related either way: the fold currently keeps every prior run, and GitHub caps issue comments at 65,536 characters. Once a long-lived release PR crosses that, the PATCH starts failing with only a `::warning` and the comment silently freezes. Capping the folded history at the last N runs (10 seems generous) would avoid that cliff — with the data-backed history it's a one-line slice.

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,301 @@
...
+
+# Only the intrinsic infra permissions necessary
+permissions:
+  id-token: write   # OIDC AssumeRole into the GHA role
+  contents: read    # checkout
+
+# Non-cancelling group keyed on the tested commit + leg. An instance can run until
+# the box’s self-terminate ceiling (bootstrap-common.sh), so new runs queue.
+concurrency:
+  group: ec2-leg-${{ inputs.target_ref }}-${{ inputs.run_label }}
```
tamirms: **[should-fix]** I have two doubts about this block. First, the semantics of workflow-level `concurrency` inside a *called* workflow are murky — the reusable-workflows docs only describe job-level concurrency interactions between caller and called workflows (https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations). Second, there are standing reports that `${{ inputs.* }}` doesn't resolve in workflow-level concurrency groups (https://github.com/orgs/community/discussions/45734). If the inputs come out empty here, the group collapses to the literal `ec2-leg--` shared by every leg across all refs and labels — and with `cancel-in-progress: false` that would serialize all perf-eval legs repo-wide into one queue at ~3.5 hours per slot.

Since the coordinator's own per-ref group already provides the supersede behavior, would it make sense to move this to `jobs.ec2-leg.concurrency`, where inputs are documented to work? Alternatively, if you've verified the resolved group name on one of the real runs, a comment noting that would settle it.

.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+        run: |
+          set -uo pipefail
+          MARKER='<!-- load-test-results -->'
+          PERF_EVAL=./cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval
+
+          # Read the existing sticky comment (id + body), if any.
+          CID=""; PREV=""
+          if [ -n "$PR" ]; then
+            CID=$(gh api "repos/$REPO/issues/$PR/comments" --paginate \
+              --jq "map(select(.body | startswith(\"$MARKER\"))) | .[0].id // empty" || true)
```
tamirms: **[hardening]** The lookup selects the first comment whose body starts with the marker, regardless of author. Anyone can post a marker-prefixed comment on a public release PR, and the workflow would then adopt it: fold its content into the official results history and PATCH *their* comment in place under the bot's identity. It's content spoofing rather than anything worse, but the fix is one extra filter — also requiring `.user.login == "github-actions[bot]"` (or `.user.type == "Bot"`).

.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+            [ -n "$CID" ] && PREV=$(gh api "repos/$REPO/issues/comments/$CID" --jq '.body' || true)
+          fi
+
+          # Render the new body (fetches each leg's S3 result, folds prior runs).
+          printf '%s' "$PREV" | go run "$PERF_EVAL" > /tmp/comment.md
+          cat /tmp/comment.md >> "$GITHUB_STEP_SUMMARY"
+
+          # Post / edit the sticky comment.
+          if [ -n "$PR" ]; then
+            jq -Rs '{body: .}' /tmp/comment.md > /tmp/payload.json
```
tamirms, line 139: **[should-fix]** The step deliberately runs without `set -e` (per the comment above it, so posting hiccups can't pre-empt the gate — makes sense). A side effect is that if `go run "$PERF_EVAL"` fails, `/tmp/comment.md` is empty and the script proceeds to PATCH a blank body over the existing comment. GitHub rejects blank comment bodies, so in practice the damage is just a warning, but that's incidental protection. A `[ -s /tmp/comment.md ] || { echo "::warning::render failed, skipping post"; }` style guard before the posting block would make the intent explicit.

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,301 @@
...
+            if [ -n "$EXTRA_ENV" ]; then printf '%s\n' "$EXTRA_ENV"; fi
+            cat "$PERF_EVAL_DIR/bootstrap-common.sh"
+            cat "$LEG_SCRIPT"
+          } > /tmp/user-data.sh
+
+      - name: Launch EC2 instance
+        id: launch
+        run: |
+          COMMON_TAGS="{Key=$TEST_TAG_KEY,Value=$TEST_TAG_VAL},
+            {Key=ref,Value=${{ inputs.target_ref }}},
```
tamirms: **[hardening]** This interpolates the input directly into the shell script. Today it's safe — the coordinator passes a `git rev-parse`d SHA and `run_label` is hardcoded by the wrapper — but the input is documented as "ref or SHA", and a ref containing `$(…)` would expand during shell evaluation. Since the rest of the file already routes parameters through `env:`, doing the same for the two inputs used in `run:` blocks would close this off cheaply.

.gitignore
```diff
@@ -3,3 +3,11 @@ captive-core/
...
 !test.toml
 *.sqlite*
+
+# Generated load-test ledger corpora (hundreds of MB; canonical copies live in
+# s3://stellar-rpc-ci-load-test/ledgers/).
+cmd/stellar-rpc/internal/integrationtest/infrastructure/testdata/*.xdr.zstd
+cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/testdata/*.xdr.zstd
+
+# Compiled refresh tool (build artifact; rebuild with `go build` in refresh/).
+cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/refresh/refresh-tool
```
tamirms: **[question]** I couldn't find `perf-eval/ingest-load-test/refresh/` on this branch or on main. Is the refresh tool landing in a follow-up PR? If so, this entry (and its comment) probably belongs there — as-is it points at nothing and is a little confusing to trace.

.github/workflows/e2e.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   run-system-test:
     # set the git ref of stellar/system-test after the '@'
     # to specify which version of the workflow to call
-    uses: stellar/system-test/.github/workflows/test-workflow.yml@master
+    uses: stellar/system-test/.github/workflows/test-workflow.yml@789121c0914150a02122581f32ee62c4e42e1c84
```
tamirms: **[question]** Pinning `stellar/system-test` to a bare SHA instead of `@master` seems orthogonal to the coordinator work — presumably master broke? A short comment saying why it's pinned and when it can be unpinned (or a line in the PR description) would help whoever finds this pin six months from now.

.github/workflows/load-test.yml
```diff
@@ -1,216 +1,35 @@
-name: Load test (ephemeral)
-# Launches a c5.2xlarge in Horizon (203618453975), polls S3 for the result object
-# the box publishes, posts results to the PR, terminates. Box bootstrap lives in
-# run-load-test.sh; runner-side polling in runner/orchestrate.go.
+name: Apply-load ingestion (callable)
 
 on:
   workflow_call:
+    inputs:
+      target_ref:
```
tamirms: **[question]** Nothing in this repo called the old `load-test.yml`, and its interface just changed in two ways: it now has a required `target_ref` input, and it no longer posts PR comments itself. If any external workflow calls it — the release pipeline in stellar/stellar-rpc-internal comes to mind — it would now fail validation. Can you confirm nothing outside this repo references it?

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,301 @@
...
+        description: 'Root gp3 throughput (MiB/s)'
+        required: false
+        type: number
+        default: 125
+      ssm_registration_timeout:
+        description: 'Seconds to wait for the SSM agent to register'
+        required: false
+        type: number
+        default: 240
+      results_timeout:
```
tamirms: **[suggestion]** The defaults maintain the necessary ordering (results 210m < job 225m < self-terminate 240m ≤ role 240m) and the descriptions hint at it, but these are now four independent per-leg knobs, and a future leg overriding one of them can silently break the chain — a box that outlives its role's credentials, or a job that dies before its own poll deadline. Would it be worth either deriving the buffers from a single per-leg duration input, or asserting the ordering in a cheap validation step so a misconfigured leg fails in seconds instead of hours?

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh
```diff
@@ -0,0 +1,12 @@
+# Apply-load ingestion leg. Concatenated after bootstrap-common.sh in the
```
tamirms, line 1: **[nit]** The leg script has the executable bit but no shebang, since it only ever runs concatenated into the user-data. Making it 644 like `bootstrap-common.sh` would signal that it's a fragment rather than a runnable script.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/exec.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd.Stdout, cmd.Stderr = w, w
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("%w\n%s", err, lastLines(tail.String(), tailN))
+	}
+	return nil
+}
+
+// BailInstance writes the failure body (titled per leg) and exits non-zero for
+// the bash wrapper.
+func BailInstance(resultsFile, title, runID, targetSHA, msg string) error {
```
tamirms, line 32: **[nit]** This was fine as a private helper, but as an exported harness API the `os.Exit(1)` behind an `error` return type is a small landmine for future leg authors (deferred cleanup in their `instantiate` silently won't run). Even just a doc-comment sentence saying it never returns would help; returning the error and letting each runner's `main` own the exit would be the tidier shape if you're touching it anyway.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/harness.go
```diff
@@ -0,0 +1,91 @@
...
+// leg's runner uses one for its own messages.
+func NewLogger() *supportlog.Entry {
+	l := supportlog.New()
+	l.SetLevel(supportlog.InfoLevel)
+	return l
+}
+
+var logger = NewLogger()
+
+func Run(instantiate func(context.Context) error) {
```
tamirms: **[suggestion]** `Gather` uses nothing leg-specific, yet it's only reachable through each leg's runner binary via the argv dispatch — which is the sole reason `ec2-leg.yml` needs the `runner_dir` input at all. If gather lived as its own `package main` (say `perf-eval/gather`), the workflow would always run the same package, the `runner_dir` input disappears, and leg runners become instantiate-only programs whose deployment location is obvious from the package layout. Cheap to do now; more annoying after a few more legs copy the pattern.

.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Render + post sticky report
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR: ${{ needs.plan.outputs.pr_number }}
+          REPO: ${{ github.repository }}
+          AWS_REGION: us-east-1
+          TARGET_SHA: ${{ needs.plan.outputs.target_sha }}
+          TARGET_REF: ${{ inputs.target_ref || github.ref_name }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          LEGS: '[{"label":"Apply-load ingestion","bucket":"${{ needs.load-test.outputs.bucket }}","key":"${{ needs.load-test.outputs.result_key }}","passed":"${{ needs.load-test.outputs.passed }}","verdict":"${{ needs.load-test.outputs.verdict }}"}]'
```
tamirms: **[suggestion]** The box→runner half of each leg already reports through S3, but the runner→report half switches to GHA job outputs — and that switch is what drives most of the per-leg plumbing: `Gather` growing `verdict`/`result_key`/`bucket` outputs, `ec2-leg.yml` re-exporting them, `load-test.yml` existing almost entirely to re-export them again, this hand-assembled JSON, and the per-leg check in the gate. Adding a leg currently touches four places.

An alternative worth weighing before more legs land: have the plan job emit a JSON list of leg tuples, fan out with one matrix job calling `ec2-leg.yml`, and have the report job list `s3://<bucket>/runs/<run_id>-<attempt>/` and read every `result.json` it finds — cross-checking the discovered set against the planned set so a leg that never published fails loudly. The gate reduces to the leg job's own result (it already exits 1 on failure or timeout, so the current output-reading gate duplicates logic the leg enforces itself). After that, adding a leg is one JSON entry plus its box script, and the usual matrix-outputs-collide objection is moot precisely because the results live in S3. Non-blocking — but if the remaining perf-eval legs are close behind, this is cheaper now than after the wrapper pattern calcifies.

.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+  report:
+    name: Aggregate + report
+    needs: [plan, load-test]
+    if: always()
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.plan.outputs.target_sha }}
```
tamirms: **[suggestion]** The tradeoff comment in `ec2-leg.yml` about harness-and-code traveling together makes sense for the legs — but the renderer is different: its contract is with the *previous comment bodies*, cross-run state written by whatever version ran last, not with the code being benchmarked. Checking out `target_sha` here means a `workflow_dispatch` against an older ref runs an old parser against a newer comment format. Checking out `github.sha` for the report job would pin the renderer to the workflow's own version while the legs keep benchmarking `target_sha`. (Adopting a data-backed history per the fold comment would mostly dissolve this, but it's a one-line change regardless.)

.gitignore
cjonas9: Ah, super sorry, this wasn't supposed to make it into the PR and I overlooked it. It happened because of some offline experimentation I was doing with a workflow to reset the apply-load load test assets automagically, will remove!

.github/workflows/load-test.yml
cjonas9: I added this yml in the apply-load load test PR, so we don't have to worry about this.

.github/workflows/e2e.yml
cjonas9: This is going to be re-pinned to master right before merge, so thank you for flagging it! I'll add a comment in the meanwhile.

.github/workflows/ec2-leg.yml
cjonas9: Great callout -- I think it's a bit messy to have these presented as four different knobs one can turn, and I don't really see a huge use case for keeping that behavior. In favor of deriving the buffers from a timeout with simple semantics (e.g. "total e2e runtime budget"), so I'll make that change

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,301 @@
...
+        run: |
+          AMI=$(aws ec2 describe-images \
+            --owners 099720109477 \
+            --filters \
+              "Name=name,Values=ubuntu/images/hvm-ssd*/ubuntu-jammy-22.04-amd64-server-*" \
+              "Name=architecture,Values=x86_64" \
+              "Name=state,Values=available" \
+            --query 'sort_by(Images, &CreationDate)[-1].ImageId' \
+            --output text)
+          echo "ami=$AMI" >> "$GITHUB_OUTPUT"
```
Shaptic: This seems.... complex. A default seems fine.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/bootstrap-common.sh
```diff
@@ -0,0 +1,140 @@
...
+RESULTS_FILE="${RESULTS_FILE:-/tmp/results.md}"
+BUCKET="${BUCKET:-stellar-rpc-ci-load-test}"
+RESULT_KEY="${RESULT_KEY:-}"
+DEFAULT_BRANCH=main
+
+# Install the AWS CLI + jq early so any later failure can still publish a result to S3.
+log "installing aws cli + jq"
+export DEBIAN_FRONTEND=noninteractive
+apt-get update -qq
+apt-get install -y -qq --no-install-recommends awscli jq curl ca-certificates
```
Shaptic, line 37: I think `apt-get` is deprecated
```suggestion
apt update -qq
apt install -y -qq --no-install-recommends awscli jq curl ca-certificates
```

.github/workflows/ec2-leg.yml
cjonas9: So, I think two things:
- it is simply more idiomatic/intuitive to move this to `jobs.ec2-leg`, so I'm happy to implement that for the sake of resolving the ambiguity
- we actually did have two runs with identically-labeled legs that should have been serialized under this failure mode if the group collapsed to `ec2-leg--`, but thankfully, the issue verifiably didn't occur.

.github/workflows/load-test-coordinator.yml
```diff
@@ -0,0 +1,144 @@
...
+            echo "pr_number=$PR"
+          } >> "$GITHUB_OUTPUT"
+
+  # The fan-out. Each leg is its own callable workflow (a thin wrapper over the
+  # shared ec2-leg.yml). Further legs slot in as sibling jobs the report
+  # aggregates.
+  load-test:
+    name: Apply-load ingestion
+    needs: plan
+    uses: ./.github/workflows/load-test.yml
```
Shaptic: Instead of using this file which is just a bunch of re-exports of outputs and another workflow call, can we just call the leg directly here?

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/coordinator-runner.go
```diff
@@ -0,0 +1,215 @@
...
+			n, _ := strconv.Atoi(m[1])
+			return n
+		}
+	}
+	return 0
+}
+
+// fetchMarkdown returns the leg result object's rendered markdown, or "" when the
+// object is missing or unreadable (a timeout or pre-publish failure).
+func fetchMarkdown(ctx context.Context, client *s3.Client, bucket, key string) string {
```
Shaptic: Could we simplify this by fetching the entire `Result`? It seems like we could reuse that code instead of dupeing it

.github/workflows/load-test-coordinator.yml
cjonas9: I'd originally thought this could take some weight out of the coordinator and make it trivial in the future to trigger runs of legs individually should that need ever be desired, but looking into this now, I think those reasons aren't airtight. Good catch!!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/bootstrap-common.sh
cjonas9: I'm not so sure it's deprecated, I think you use it in scripts because apt isn't designed to be script-friendly?
From the [man page](https://manpages.ubuntu.com/manpages/jammy/man8/apt.8.html) for it:
> All features of [apt(8)](https://manpages.ubuntu.com/manpages/jammy/man8/apt.8.html) are available in dedicated APT tools like [apt-get(8)](https://manpages.ubuntu.com/manpages/jammy/man8/apt-get.8.html) and [apt-cache(8)](https://manpages.ubuntu.com/manpages/jammy/man8/apt-cache.8.html) as well. [apt(8)](https://manpages.ubuntu.com/manpages/jammy/man8/apt.8.html) just changes the default value of some options (see [apt.conf(5)](https://manpages.ubuntu.com/manpages/jammy/man5/apt.conf.5.html) and specifically the Binary scope). So you should prefer using these commands (potentially with some additional options enabled) in your scripts as they keep backward compatibility as much as possible.

Thoughts?

.github/workflows/load-test-coordinator.yml
cjonas9: I was going back and forth on this/whether it should land here or in a follow-up, but it's a great catch and synthesizes well with some of the other edits I've made. Good call!!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/coordinator-runner.go
cjonas9: Under your suggestion here and one of Tamir's more structural comments above, this whole function ended up being dead/extremely inlineable code. Good catch!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/coordinator-runner.go
cjonas9: This is a moot point because I do think that this was a genuinely better design than my first pass here, but in the first point about nested drop-downs: I'm not sure that I see much of a use case for this, and that it's a bit early to expect the follow-up tests to warrant their own nested drop-downs. But, with that said, this is generally more robust as a design for the renderer. Added the main change + your suggestion asserting comment size + run history cap!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/bootstrap-common.sh
Shaptic: Oh hm, yeah that seems legit - idk where that bit came from in my mind, I think I just switched my brain to `apt` and that grew into an assumption of deprecation :shrug: should've checked my work! All good then

Shaptic (dismissed):
Awesome! Final result is way more concise and easier to read, glad we iterated on this 👍

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,260 @@
...
+          echo "::error::SSM agent never registered for $INSTANCE_ID — verify AmazonSSMManagedInstanceCore is attached to the stellar-rpc-ci-load-test role"
+          exit 1
+
+      # Polls S3 for the result object and relays verdict/passed + the result
+      # location (bucket/key) as step outputs the coordinator reads.
+      - name: Poll for results
+        id: results
+        env:
+          INSTANCE_ID: ${{ steps.launch.outputs.instance_id }}
+        run: go run "./$PERF_EVAL_DIR/gather"
```
tamirms, line 222: **[should-fix]** This step points at `perf-eval/gather`, but that directory never made it onto the branch, so every leg run since `2bdcc22d` dies here (or would, once the launch step is fixed). The run on `2bdcc22d` shows it directly: "Poll for results" failed with `stat .../perf-eval/gather: directory not found`. I suspect the new package got lost around the `0d93c008` merge and never `git add`ed. `harness.Gather` is also unreferenced right now, so nothing on the branch can poll for results. It might be worth a quick local `go run ./cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/gather` sanity check after committing it, since the sticky comment updating on every push makes the branch look healthier than the runs actually are.

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,260 @@
...
+        env:
+          TARGET_REF: ${{ inputs.target_ref }}
+          RUN_LABEL: ${{ inputs.run_label }}
+        run: |
+          COMMON_TAGS="{Key=$TEST_TAG_KEY,Value=$TEST_TAG_VAL},
+            {Key=ref,Value=$TARGET_REF},
+            {Key=sha,Value=${{ steps.resolve.outputs.sha }}},
+            {Key=run-id,Value=${{ github.run_id }}}"
+          RUN_INSTANCES_JSON=$(aws ec2 run-instances \
+            --image-id resolve:ssm:/aws/service/canonical/ubuntu/server/22.04/stable/current/amd64/hvm/ebs-gp2/ami-id \
```
tamirms: **[should-fix]** The latest run (29047822518) fails at "Launch EC2 instance" with:

```
An error occurred (SsmAccessDenied) when calling the RunInstances operation: User: arn:aws:sts::203618453975:assumed-role/gha-rpc-ci/GitHubActions is not authorized to perform: ssm:GetParameters on resource: arn:aws:ssm:us-east-1::parameter/aws/service/canonical/ubuntu/server/22.04/stable/current/amd64/hvm/ebs-gp2/ami-id
```

`resolve:ssm` makes RunInstances fetch the parameter with the caller's credentials, so the `gha-rpc-ci` role needs `ssm:GetParameters` on `arn:aws:ssm:us-east-1::parameter/aws/service/canonical/*` (note the empty account field: public parameters live in an account-less ARN). Two ways out: add that grant to the role policy, or make the input default a pinned AMI id, which needs no IAM change and is arguably the simpler default Shaptic was suggesting. If you keep `resolve:ssm`, maybe note the required grant in a comment so the next person who clones this setup into another account knows why launches fail.

.github/workflows/ec2-leg.yml
```diff
@@ -0,0 +1,260 @@
...
+      TEST_TAG_KEY: test
+      TEST_TAG_VAL: stellar-rpc-ci-load-test
+      SSM_REGISTRATION_TIMEOUT: ${{ inputs.ssm_registration_timeout }}
+      POLL_INTERVAL: ${{ inputs.poll_interval }}
+      DEBUG_LOG_LINES: 40
+      DEBUG_LOG_EVERY_POLLS: 5
+      PERF_EVAL_DIR: cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval
+      LEG_SCRIPT: ${{ inputs.leg_script }}
+      BUCKET: stellar-rpc-ci-load-test
+      RESULT_KEY: runs/${{ github.run_id }}/${{ inputs.run_label }}/result.json # deterministic per run + leg
```
tamirms, line 101: **[should-fix]** Dropping `run_attempt` from the key is what lets the report job derive fetch locations deterministically, and it also makes report-only re-runs work, so I think the key itself is right. But it opens a hole on leg re-runs: if attempt 1 published a fail object (verdict fail, as opposed to timing out without publishing), then "Re-run failed jobs" launches a fresh box, and gather finds attempt 1's object at the same key immediately, before the new box even boots. The leg fails within seconds every time, so such a re-run can never succeed.

The box preamble still exports `RUN_ID=<run_id>-<run_attempt>`, so the result object already records which attempt wrote it. If the poll step passed the expected `RUN_ID` through to gather and gather treated a `runId` mismatch as not-ready (same as absent), stale objects from prior attempts would be ignored and the new box's publish would win. The report job stays attempt-agnostic and reads whatever the final object is, which is what you want there.

.github/workflows/ec2-leg.yml
cjonas9: shoot, totally true, I accidentally missed adding gather/main.go. _d'oh_ sorry!

tamirms (approved):

Issue #770, opened 2026-06-08
Design: concurrent query serving across all data types


tamirms:
## TL;DR

Produce a reviewed design for serving queries across every data type's hot + cold stores while ingestion and lifecycle routines concurrently mutate the store set. **The deliverable is a design, not code** — once it's approved, we carve implementation issues from it. Query-side counterpart to #765 (ingestion).

## The problem the design must solve

Three kinds of thread touch the store set at the same time:

- **1 ingestion thread** — appends new ledgers to the hot store.
- **1 lifecycle thread** — freeze (hot → cold), prune cold stores past the retention window, delete a hot store once its freeze completes.
- **Many query threads.**

A query in flight must never read a store the lifecycle thread is removing, and must see a consistent view of which stores cover its range.

## Deliverable

A design doc covering:
- How a query routes across the live set of hot + cold stores per data type — lookup by ledger range, run the per-chunk engine against each, merge to the requested order/limit.
- The reader-safety contract — how queries stay correct while the ingestion and lifecycle operations above mutate the store set concurrently.
- Design review and sign-off before any implementation.

## Out of scope

- Implementation — broken into separate issues once the design is approved.

tamirms:
this task was completed in https://github.com/stellar/stellar-rpc/pull/843

PR #794, opened 2026-06-18
txhash: look up a transaction by hash across the hot and cold stores (#728)

tamirms:
Implements the read side of #728: given a transaction hash, find and return the transaction.

A by-hash lookup has no ledger to narrow the search on, so it fans out across every index that might hold the transaction — the **hot indexes** (recent, un-frozen chunks) and the **cold indexes** (older, frozen history) — and confirms each candidate against the actual ledger before returning it.

## How it works
- **`HashIndex`** — a one-method interface (`Get(hash) (uint32, error)`) that resolves a tx hash to its ledger sequence. The hot store satisfies it exactly (a hit is the truth); a cold index satisfies it with a fingerprint (a hit is only a candidate). Both `*HotStore` and `*ColdReader` implement it through their existing `Get`.
- **`TxReader.GetTransaction(hash)`** — scans the exact (hot) tier first, then the fingerprinted (cold) tier; for each candidate it reads the ledger through an injected `LedgerSource` and pulls the transaction out with the SDK's tx-details-by-hash view. That view confirms the hash is actually in the ledger, which is what rejects the cold indexes' fingerprint false positives. An exact hit short-circuits before any cold index is touched.
- **Resilience & consistency** — a transient index error falls through to the remaining indexes and is surfaced only if nothing else resolves, so a hot-store blip never masks a cold-resident transaction. If an *exact* index maps a hash to a ledger that doesn't contain it, the index and the ledger store disagree (corruption), so it surfaces as `ErrInconsistent` rather than a silent miss.

## Not in this PR (owned by query serving, #770)
The concrete hot+cold `LedgerSource`, discovering/opening the cold index files and managing their lifecycle, and wiring this into the `getTransaction` JSON-RPC handler. Cold-index build orchestration is deferred per #728.

## Testing
Tests against real cold streamhash indexes and real ledger fixtures: hit resolves the transaction, clean miss, false-positive rejection, hot+cold federation (real RocksDB + index), exact-source inconsistency, transient-error fall-through, fan-out across cold indexes, and constructor validation. `go test -race`, `go vet`, and `golangci-lint` all pass.

> Based on `feature/full-history`.


cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/read_assembly.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+	return r
+}
+
+// GetTransaction returns the transaction for hash. found is false (nil error)
+// on a genuine miss; an exact source that fails to verify yields
+// ErrInconsistent. The returned view aliases the LedgerSource bytes.
+func (r *TxReader) GetTransaction(hash [32]byte) (ingest.LedgerTransactionView, bool, error) {
+	if txv, found, err := r.scan(hash, r.exact, true); found || err != nil {
```
marwen-abid: ❓ It seems to me like a transient hot-store blip (e.g. connection drops) aborts the entire lookup before cold is ever scanned. Even for a tx that lives only in cold store.

Why not collect the error but fall through to inexact, then only surface the error when cold also misses ?

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/read_assembly.go
```diff
@@ -0,0 +1,93 @@
...
+			r.inexact = append(r.inexact, s)
+		}
+	}
+	return r
+}
+
+// GetTransaction returns the transaction for hash. found is false (nil error)
+// on a genuine miss; an exact source that fails to verify yields
+// ErrInconsistent. The returned view aliases the LedgerSource bytes.
+func (r *TxReader) GetTransaction(hash [32]byte) (ingest.LedgerTransactionView, bool, error) {
```
marwen-abid, line 50: ⛏️ No `context.Context` passed here, is that on purpose ?

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/read_assembly.go
```diff
@@ -0,0 +1,93 @@
...
+// extracted view aliases them; a seq it cannot serve must return
+// stores.ErrNotFound or stores.ErrOutOfRange.
+type LedgerSource interface {
+	GetLedgerRaw(seq uint32) ([]byte, error)
+}
+
+// TxReader resolves a tx hash to its transaction by federating candidate
+// sources and verifying each against the real ledger. Exact sources are
+// consulted before inexact ones, regardless of construction order.
+type TxReader struct {
```
marwen-abid, line 30: ⛏️ I'm probably not seeing the full picture, but chatting about this with Claude, it seems like we can simplify the abstraction by getting rid of both `candidate_source.go` and `cold_reader_set.go` entirely. We can replace them with a one-method interface and let TxReader hold the two tiers of stores:

```go
// HashIndex resolves a tx hash to the single ledger sequence it was committed
// in, or stores.ErrNotFound on a miss. Both the hot store (exact: a hit is the
// truth) and a cold index (fingerprinted: a hit is only a candidate) satisfy it
// with their existing Get — no adapter.
type HashIndex interface {
      Get(hash [32]byte) (uint32, error)
}

type TxReader struct {
      hot        HashIndex   
      cold       []HashIndex // inexact tier, scanned in order
      ledgers    LedgerSource
      passphrase string
}

func NewTxReader(hot HashIndex, cold []HashIndex, ledgers LedgerSource, passphrase string) (*TxReader, error) {
        if ledgers == nil {
                return nil, fmt.Errorf("txhash: nil ledger source: %w", stores.ErrInvalidConfig)
        }
        if passphrase == "" {
                return nil, fmt.Errorf("txhash: empty passphrase: %w", stores.ErrInvalidConfig)
        }
        return &TxReader{hot: hot, cold: cold, ledgers: ledgers, passphrase: passphrase}, nil
  }
```

marwen-abid (approved):
LGTM! All ⛏️ nits and ❓questions. 

Re-tagged the team for review because I definitely missed something.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/read_assembly.go
tamirms: thanks, I implemented this suggestion!

cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/read_assembly.go
tamirms: I think we should do this in a separate PR once the query design is nailed down. This code is currently simplified because it assumes a static list of hot and cold tx hash indexes. The final iteration should handle the fact that indexes can be added and removed along with the handling context cancellation

PR #779, opened 2026-06-08
ingest: full-history ingestion for cold + hot stores (#765)

chowbao:
Implements #765 — the full-history data-ingestion path (cold + hot stores) for each `LedgerCloseMeta`. Code at `cmd/stellar-rpc/internal/fullhistory/ingest/`.

> **Rebased onto `fh-ingest-base` and absorbed the view layer from #778, which is now closed.** The zero-copy XDR view extractors that #778 introduced have been **promoted into the go-stellar-sdk** (stellar/go-stellar-sdk#5949); this PR consumes them and keeps only the thin RPC-specific glue. Both retarget to `feature/full-history` once #756 lands.

## SDK dependency (stellar/go-stellar-sdk#5949)

The view extractors now live in the SDK as the zero-copy twins of the parsed path. Per the #5949 review, the SDK exports only the complete extractors (the navigation scaffolding is unexported):
- `ingest.ExtractTxHashes` — per-ledger transaction hashes in apply order.
- `ingest.ExtractLedgerEvents` — per-transaction contract events + hash from one `TxProcessing` walk.
- `ingest.LedgerTransactionViewByHash` / `LedgerTransactionViewRange` (+ `ingest.LedgerTransactionView`) — view parallel of `LedgerTransaction` / `LedgerTransactionReader`.
- `network.TransactionViewHasher` — view twin of `network.HashTransactionInEnvelope` (returns just the hash).

go.mod pins the #5949 branch pseudo-version (`f92b870f`); **bump to the merged version once #5949 lands.**

## RPC-side view adapters (`internal/fullhistory/views/`)

The local `views` package collapses to thin adapters that add only RPC-specific shapes/policy:
- `ExtractEvents` composes `ingest.ExtractLedgerEvents` (hash + events from one walk) with the `events.Payload` shape and the `Stage→(TxIdx, OpIdx)` cursor sentinels (`events.StageSentinels`).
- `ExtractTxHashes` wraps `ingest.ExtractTxHashes` into `txhash.Entry`.
- `ExtractTxDetailsByHash` / `ExtractTransactions` delegate to the SDK read path; `views.Transaction` aliases `ingest.LedgerTransactionView`.
- `dispatch.go` + `envelopes.go` deleted (moved to the SDK).

## EventIdx removed from the events payload

[...]

Issue #663, opened 2026-04-02
[events] Segment coordination and query routing


urvisavla:
Implement the component that coordinates all segments:
- Track current hot segment, freezing segment (if any), cold segment registry
- Segment lookup by ledger range
- Query routing: identify overlapping segments, query sequentially (ascending or descending), stop at result limit
- Handle queries spanning hot + cold segments

Add tests for: single-segment queries, cross-segment queries, ascending/descending order, limit handling.

PR #796, opened 2026-06-19
events,eventstore: single-chunk query engine (#663)

urvisavla:
Implements #663 — the per-chunk query coordinator. Built on the `Reader` interface (PR #756) so it serves `HotStore` and `ColdReader` without tier branching. Rebased onto `feature/full-history` now that #779 has landed; consumes that PR's `EventIdx` payload field and the SDK view-extractor API.

## Scope coverage

Each bullet from the issue:

- **Filter semantics** — AND within a filter (contract ID + topics 0..3), union across filters, restricted to an optional window. ✅
- **Cap at `MaxEvents`, ascending or descending** — `Order` enum; descending is added here (rpc-hack was asc-only). ✅
- **Post-filter to drop term-hash collision false positives** — view-path navigation over `ContractEventBytes`, lazy ContractId + topic resolution, zero per-event allocation. ✅

## Notable deltas vs the rpc-hack reference

The issue says the rpc-hack `query.go` is "a reference, not a spec." Two substantive changes from that reference:

1. **Window input is `EventIDRange`, not `LedgerRange`** — chunk-relative event-ID bounds. Ledger-bounded engines force the adapter to post-skip events at-or-before the cursor within the first ledger, which stalls when a single ledger holds more events than `MaxEvents` (realistic for Soroban traffic — busy DEXes, token launches). Event-ID windowing lets the adapter express "strictly after event X" directly, eliminating the stall. Adapters that think in ledgers call `EventIDRangeForLedgers(ofs, L1, L2)` to translate. The rationale is documented on top of `query.go`.

2. **Latent bug fixed in the match-all path** — rpc-hack's `fetchAllInRange` appended `events.Payload` values whose `ContractEventBytes` aliased the reader's iteration buffer. Once the iterator advanced, every retained slice pointed at stale memory. Fix: `bytes.Clone(p.ContractEventBytes)` before append.

The struct post-filter path is gone — this branch's `events.Payload` carries only `ContractEventBytes`, so the rpc-hack view/struct dispatch collapses to the view path only.

## What's out of scope

Per the issue: routing a query across multiple chunks. That coordinator is the immediate follow-up; it owns the `protocol.EventFilter` translation, cursor mapping, and `payload → EventInfo` shaping.

## Tests

43 tests, `-race` clean, `go vet` / `gofmt` clean.

- Filter combinations (single contract, single topic, AND, OR across filters, match-all, mixed-success, unknown terms).
- Ordering and limit in both directions (asc/desc match-all, asc/desc filtered, both with `MaxEvents` cap).
- Range clipping (in-window, end-beyond-chunk, start-past-chunk, range-empties-union).
- Range × filter × cap × order combinations.
- Post-filter collision rejection (mirror-poisoned via `ConcurrentBitmaps.AddTo`).
- Validation (negative `MaxEvents`, short ContractID, cancelled ctx).
- Empty chunk, chunk-with-empty-ledgers paths.
- `EventIDRangeForLedgers` helper translation + out-of-range error.
- **HotStore/ColdReader parity** — 8 scenarios covering the major code paths through `Query` against a `ColdReader` built by freezing a hot fixture. Validates the Reader-abstraction claim.

## Test plan

- [x] \`go test -race ./cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/\`
- [x] \`go vet\`, \`gofmt\`
- [x] \`go build ./...\`

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
+++ b/cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
@@ -1,0 +1,738 @@
+package eventstore
+
+// query.go is the events-side coordinator that turns a {filters,
+// eventIDRange, maxEvents, order} spec into matching events for one
+// Chunk. It is built on the Reader interface, so it works against
+// HotStore and ColdReader without branching.
+//
+// Semantics:
+//
+//   - A Filter is a conjunction of equality constraints on the
+//     indexed fields (contract ID + topics 0..3). A field left
+//     nil/empty is a wildcard.
+//   - The query result is the union of the per-filter intersections,
+//     intersected with QueryOptions.Range, then truncated to
+//     QueryOptions.MaxEvents (when non-zero). Ascending order keeps
+//     the lowest event IDs; descending keeps the highest.
+//   - An empty []Filter is treated as one wildcard filter
+//     (match-all in the chunk), matching the getEvents convention.
+//
+// The engine's input window is a CHUNK-RELATIVE event-ID range
+// [Start, End), not a ledger range. This is deliberate: ledger-bounded
+// queries can't express "everything strictly after event X" without
+// the caller doing fragile post-skip work, which breaks down when a
+// single ledger holds more events than MaxEvents. Adapters that speak
+// ledgers (the eventual multi-chunk coordinator / RPC handler)
+// translate ledger bounds to event-ID bounds via the chunk's
+// LedgerOffsets before calling Query.
+//
+// Optimization shape: the caller guarantees ≤15 filters and ≤15
+// unique terms total across all filters. The implementation
+// dedupes terms across filters and issues a SINGLE Reader.LookupKeys
+// call for all unique terms, then assembles per-filter
+// intersections by re-using the returned bitmaps. On the cold path
+// this collapses what would otherwise be one MPHF+index.pack round
+// trip per filter into a single coalesced read.
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"slices"
+
+	"github.com/RoaringBitmap/roaring/v2"
+
+	protocol "github.com/stellar/go-stellar-sdk/protocols/rpc"
+	"github.com/stellar/go-stellar-sdk/xdr"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/events"
+)
+
+// Filter is one item in the union of an events Query. Within a
+// single filter, every non-empty field is an equality constraint
+// AND-ed together against the corresponding indexed field of an
+// event. nil or empty slices are wildcards.
+//
+// Topics[i] constrains topic position i. Topic positions beyond
+// protocol.MaxTopicCount are not indexed (see events.TermsFor) and
+// must be left empty here.
+type Filter struct {
+	ContractID []byte
+	Topics     [protocol.MaxTopicCount][]byte
+}
+
+// Order selects the direction matches are returned (and, when
+// MaxEvents caps the result, which end of the range is kept).
+// The zero value is ascending, matching the getEvents v1 default.
+type Order int
+
+const (
+	// OrderAscending returns matches in ascending event-ID order
+	// (ledger-ascending, since event IDs are dense in [0, EventCount)).
+	// When MaxEvents caps, the lowest IDs win.
+	OrderAscending Order = iota
+	// OrderDescending returns matches in descending event-ID order.
+	// When MaxEvents caps, the highest IDs win.
+	OrderDescending
+)
+
+// EventIDRange restricts a Query to a half-open chunk-relative
+// event-ID window [Start, End). Both bounds are chunk-relative
+// event IDs in [0, EventCount).
+//
+// Zero values are interpreted as the chunk's endpoints:
+//
+//	Start == 0 → 0 (first event in chunk; this is also the literal value)
+//	End   == 0 → EventCount (last event +1; sentinel)
+//
+// An EventIDRange{} (the zero value) therefore means "every event
+// the chunk has ingested so far." End == 0 is treated as a sentinel
+// because a literal [0, 0) range would be trivially empty and there's
+// no productive use for that exact value as a query bound.
+//
+// Out-of-range End values are clipped to EventCount silently.
+// A clipped-empty range (post-clip Start >= End), or an empty chunk
+// (EventCount == 0), makes Query return (nil, nil) with no error.
+type EventIDRange struct {
+	Start, End uint32
+}
+
+// resolve returns the post-clip half-open bounds for this range
+// against the chunk's total event count. Returns (start, end, ok)
+// where ok is false when the chunk is empty or the clipped range
+// is empty.
+func (r EventIDRange) resolve(eventCount uint32) (uint32, uint32, bool) {
+	if eventCount == 0 {
+		return 0, 0, false
+	}
+	start, end := r.Start, r.End
+	if end == 0 || end > eventCount {
+		end = eventCount
+	}
+	return start, end, start < end
+}
+
+// EventIDRangeForLedgers translates an inclusive ledger window
+// [startLedger, endLedger] into the half-open EventIDRange covering
+// those ledgers' events, using ofs's per-ledger event-count snapshot.
+// Both bounds must lie inside ofs's [StartLedger, EndLedger) range;
+// out-of-range bounds surface a wrapped error from
+// LedgerOffsets.EventIDs.
+//
+// The engine works in event-ID space only — this is the canonical
+// adapter-side translation for callers that think in ledger ranges
+// (the multi-chunk coordinator / RPC handler converting a getEvents
+// request). Callers that already have chunk-relative event IDs (e.g.
+// cursor continuation past a specific event) build an EventIDRange
+// directly.
+func EventIDRangeForLedgers(ofs *events.LedgerOffsets, startLedger, endLedger uint32) (EventIDRange, error) {
+	firstID, _, err := ofs.EventIDs(startLedger)
+	if err != nil {
+		return EventIDRange{}, fmt.Errorf("events: range start ledger %d: %w", startLedger, err)
+	}
+	_, lastID, err := ofs.EventIDs(endLedger)
+	if err != nil {
+		return EventIDRange{}, fmt.Errorf("events: range end ledger %d: %w", endLedger, err)
+	}
+	return EventIDRange{Start: firstID, End: lastID}, nil
+}
+
+// QueryOptions configures a Query call.
+//
+// MaxEvents caps the result size (0 = unlimited). Must be
+// non-negative; Query rejects a negative value up-front. Which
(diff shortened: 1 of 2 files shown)


cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,738 @@
...
+// resolve returns the post-clip half-open bounds for this range
+// against the chunk's total event count. Returns (start, end, ok)
+// where ok is false when the chunk is empty or the clipped range
+// is empty.
+func (r EventIDRange) resolve(eventCount uint32) (uint32, uint32, bool) {
+	if eventCount == 0 {
+		return 0, 0, false
+	}
+	start, end := r.Start, r.End
+	if end == 0 || end > eventCount {
```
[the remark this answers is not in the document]
urvisavla: Fixed in 57635ea1. Range is now a mandatory value type — the caller pins End from a snapshot at request entry; the engine no longer invents an upper bound. The `End == 0` sentinel is gone entirely, so `EventIDRange{0, 0}` from an empty leading ledger stays empty. `TestQuery_EmptyLeadingLedgerRangeStaysEmpty` is the regression test.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,738 @@
...
+
+	// ───── 6. Fetch payloads ─────
+	//
+	// selectEventIDs drains the union into the ascending []uint32
+	// FetchEvents requires, keeping the lowest IDs (ascending) or
+	// highest IDs (descending) when MaxEvents caps. FetchEvents
+	// returns owned payloads in ascending order; the defensive
+	// post-filter preserves that order, and we reverse once at the
+	// end for descending output.
+	ids := selectEventIDs(union, opts.MaxEvents, descending)
```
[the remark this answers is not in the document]
urvisavla: Acknowledged. The cap-vs-post-filter gap is real but ~2^-64 per term with xxh3_128; cold path's 4-byte fingerprint (cold_reader.go) catches another ~2^-32 before the bitmap layer. Deferring the refill loop until a real workload exhibits it.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,670 @@
...
+// protocol.MaxTopicCount are not indexed (see events.TermsFor).
+type Filter struct {
+	ContractID []byte
+	Topics     [protocol.MaxTopicCount][]byte
+}
+
+// Order selects the direction matches are returned (and, when
+// MaxEvents caps the result, which end of the range is kept).
+// The zero value is ascending.
+type Order int
```
chowbao: nit: this probably could have been a bool named `OrderDecending` (or `OrderAscending`). But I see the benefit of also labeling OrderAscending and OrderDescending explicitly so feel free to ignore if you want

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,670 @@
...
+//
+//nolint:gocognit,cyclop,funlen
+func Query(ctx context.Context, r Reader, filters []Filter, opts QueryOptions) ([]events.Payload, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	if opts.MaxEvents < 0 {
+		return nil, fmt.Errorf("events: MaxEvents must be non-negative, got %d", opts.MaxEvents)
+	}
+	if opts.Range.End < opts.Range.Start {
```
chowbao: nit: Wouldn't we want to return early in the `start == end` case?

Doesn't seem to need to do the filter validation and everything else if the payload is gonna be a valid empty []events.Payload anyways

chowbao: I guess in theory if `start == end` then the eventCount would also be 0 which means the `Range.resolve(eventCount)` would be the early `nil, nil` return

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,670 @@
...
+	if hasMatchAllFilter(filters) {
+		return fetchAllInRange(ctx, r, start, end, opts.MaxEvents, descending)
+	}
+
+	// ───── 1. Dedupe terms across filters ─────
+	//
+	// filterPlans[i] holds the indices into uniqueKeys that filter i
+	// requires intersected. Empty plans were handled by the match-all
+	// short-circuit above.
+	filterPlans := make([][]int, len(filters))
```
chowbao, line 218: Was there supposed to be a maximum filter/term/index lookup limit?

I'm not sure if anything was ever agreed to and the [getEvent v2 discussion](https://github.com/orgs/stellar/discussions/1872#discussioncomment-15696991) doesn't seem to explicitly set a max term limit either

We can probably defer this to later but we should find out if we do want to have an explicit max term/max index lookup for getEvents()

We might just define this as a replacement for [MaxFiltersLimit](https://github.com/stellar/go-stellar-sdk/blob/main/protocols/rpc/get_events.go#L16)

chowbao (approved):
LGTM

I had a few nits/questions but I think realistically they can all be deferred til after MVP is done

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: Added the check.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: yes, It was verbally agreed that we'd limit it to 15 terms though the spec hasn't been updated to reflect that yet. I don't recall whether that limit was for total unique terms items or not so I left that out for now.

urvisavla: Also, tbh, the implementation will likely need to change a bit for multi-chunk query. We may not want to do filter validation on every chunk nor dedup terms per chunk. I wasn't sure whether to do it in this PR since it felt out of scope for the issue.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: Yeah, no reason it couldn't be a bool. This felt a bit more explicit to me.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,644 @@
...
+// End > EventCount is rejected as a caller bug (wrong chunk's
+// offsets or stale snapshot) — under the snapshot-isolation contract
+// a properly-pinned End never exceeds the chunk's current EventCount,
+// since chunks only grow.
+type EventIDRange struct {
+	Start, End uint32
+}
+
+// EventIDRangeForLedgers translates an inclusive ledger window
+// [startLedger, endLedger] into the half-open EventIDRange covering
```
tamirms: this comment mentions that the range is inclusive but the comment for EventIDRange says the interval is half-open: `[Start, End)`

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,644 @@
+package eventstore
+
+// query.go is the events-side coordinator that turns a {filters,
+// Range, MaxEvents, Order} spec into matching events for one Chunk.
```
tamirms: Small doc nit: now that the `Order` enum became the `Descending bool` in the final commit, this opening line still describes the spec as `{filters, Range, MaxEvents, Order}`.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
```diff
@@ -0,0 +1,644 @@
...
+// fetchAllInRange returns every event in the chunk-relative event-ID
+// window [start, end), truncated to maxEvents (0 = unlimited) and
+// ordered per descending. Used for the match-all short-circuit. Streams
+// via Reader.FetchRange — no []uint32{0..count} materialization, and
+// on the cold path one direct ReadRange call instead of the
+// per-position coalescing logic FetchEvents would run.
+//
+// Caller contract: start < end (the [start, end) window is non-empty)
+// and both bounds are valid chunk-relative event IDs. The resolve()
+// step in Query enforces both.
```
tamirms: This still points at `resolve()`, but that helper was removed in `0f1b1b6b`.

tamirms (commented):
A few optional readability ideas, all non-blocking:

- The cardinality comparator in step 3 could collapse to `cmp.Compare(a.GetCardinality(), b.GetCardinality())`.
- `indexOfOrAddTerm`'s scan could use `slices.Index`.
- The all-wildcard check and the per-filter term construction read naturally as methods on `Filter` (`isMatchAll()`, `termKeys()`), which would thin out both `hasMatchAllFilter` and `Query`'s step 1; similarly `EventIDRange` could grow `isEmpty()` / `check()` so `Query`'s preamble reads as intent.

One thing I'd explicitly *not* do is methodize the per-event matching path (`matchesAnyFilterView` / `collectTopicViewBytes`) — the lazy cache shared across filters is the whole point there, and a per-filter method would re-resolve per filter.

tamirms (approved):
looks great! left a few comments with some nits and minor suggestions

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: Fixed in d9214816.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: Clarified in d9214816 — now explicit: closed ledger window `[L1, L2]` → half-open EventIDRange `[firstID, lastID)`.

cmd/stellar-rpc/internal/fullhistory/pkg/stores/eventstore/query.go
urvisavla: Fixed in d9214816.

urvisavla:
Took all three in d9214816: `cmp.Compare` for the cardinality sort, `slices.Index` in `indexOfOrAddTerm`, plus `Filter.isMatchAll()` / `Filter.termKeys()` / `EventIDRange.isEmpty()` / `EventIDRange.check()`.

PR #801, opened 2026-06-23
Cache oldest ledger alongside latest ledger

Shaptic:
### What
Extends `dbCache` to cache oldest ledger alongside latest ledger.

### Why
Every endpoint call that returns the available ledger range (most of them) needs to decode the entire oldest LCM without caching.

### Known limitations
n/a


cmd/stellar-rpc/internal/db/db.go
```diff
@@ -339,6 +351,19 @@ func (w writeTx) Commit(ledgerCloseMeta xdr.LedgerCloseMeta, durationMetrics map
 			w.globalCache.latestLedgerSeq = ledgerSeq
 			w.globalCache.latestLedgerCloseTime = ledgerCloseTime
 		}
+		// Invalidate the cached oldest-ledger scalars when trimLedgers (run
+		// above with this same retention window) has removed the ledger they
+		// describe. cutoff mirrors trimLedgers: rows with sequence < cutoff are
+		// deleted. Only invalidate when retention is actually trimming and the
+		// cached oldest was at/below the cutoff, so the lazy recompute happens
+		// at most once per trim rather than on every read.
+		if w.historyRetentionWindow != 0 && ledgerSeq+1 > w.historyRetentionWindow {
```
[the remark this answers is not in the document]
Shaptic: Adding below the cached minimum is out of scope: backfill happens before serving, so the cache would be invalidated at the time.

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -201,11 +201,49 @@ func (r ledgerReader) GetLedgerRange(ctx context.Context) (ledgerbucketwindow.Le
...
+	firstLedgerSeqCache := r.db.cache.firstLedgerSeq
+	firstLedgerCloseTimeCache := r.db.cache.firstLedgerCloseTime
 	r.db.cache.RUnlock()
 
-	// Make use of the cached latest ledger seq and close time to query only the oldest ledger details.
+	// Fully cached: both ends known, no query at all. This is the hot path for
+	// read-heavy workloads (e.g. getTransaction polling), which previously
+	// decoded the entire oldest LedgerCloseMeta blob on every single call.
+	if latestLedgerSeqCache != 0 && firstLedgerSeqCache != 0 {
+		return ledgerbucketwindow.LedgerRange{
```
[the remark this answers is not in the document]
Shaptic: It's still accurate for all other endpoints

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -339,6 +351,19 @@ func (w writeTx) Commit(ledgerCloseMeta xdr.LedgerCloseMeta, durationMetrics map
...
+			cutoff := ledgerSeq + 1 - w.historyRetentionWindow
```
tamirms, line 361: do we need to worry about the case where `w.historyRetentionWindow` is larger than `ledgerSeq + 1` and  `cutoff` wraps around?

Shaptic: Hmmm very interesting edge case. This would only happen in the case you are on a standalone or recently-reset network and your retention window exceeds the current state, right? In this case, `ledgerSeq+1 > w.historyRetentionWindow` fails, right? So this block isn't hit.

tamirms: my bad, you're right! you can ignore that comment

tamirms (approved):

PR #817, opened 2026-06-24
Streaming daemon: Phase 1 layer 1 — foundations (#815)

chowbao:
**Epic [#808](https://github.com/stellar/stellar-rpc/issues/808) → Phase 1 [#815](https://github.com/stellar/stellar-rpc/issues/815) → this PR (layer 1 of 3).**

- Phase 1 ([#815](https://github.com/stellar/stellar-rpc/issues/815), Backfill) stands up the *whole* engine — all three data types plus the window index — under its first caller, the startup catch-up loop. It lands as a stacked, reviewable series.
- **This PR is the bottom layer:** the cold substrate every primitive and the orchestrator above it stand on.
  - No primitives or orchestration yet — only the durable-state machinery, its crash-safety guarantees, and the config/path/lock plumbing the daemon boots from.
- **Design:** [#787](https://github.com/stellar/stellar-rpc/pull/787) — `design-docs/full-history-streaming-workflow.md` + `gettransaction-full-history-design.md` (section links inline below).

---

## What this layer implements

New `cmd/stellar-rpc/internal/fullhistory/streaming` package, **cold-only**: the catalog and the immutable cold artifacts (`.pack` / events segment / `.bin` / `.idx`). The hot tier and live loop are Phase 2 ([#816](https://github.com/stellar/stellar-rpc/issues/816)).

- **Geometry & window arithmetic** — `window.go`
  - the chunk↔window↔index layer over the merged `pkg/chunk`; `Windows` bound to one immutable `chunks_per_txhash_index`, plus `MaxChunksPerTxhashIndex` and `IsTerminalCoverage`.
  - → [Geometry](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#geometry) · gettx [§4](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/gettransaction-full-history-design.md#4-geometry)
- **Key schema + key↔path bijection** — `keys.go`, `paths.go`
  - the `chunk:` / `index:` / `config:` key families, the `freezing|frozen|pruning` states, and `Layout` as the single source of the strict key→file mapping that lets cleanup walk keys instead of listing directories.
[...]

Issue #815, opened 2026-06-24
Streaming daemon: Phase 1 — Backfill (catch-up + all primitives)


chowbao:
_Part of #808 (epic)_

**Phase 1 of 2 — Backfill.** Brings up the entire type-agnostic daemon under its *first* caller: the startup catch-up loop. Backfill owns **every primitive the daemon has** — `processChunk` (all per-chunk artifacts), `buildTxhashIndex` (the window index), and `resolve`/`executePlan` — so this phase stands up the whole engine and **all three data types + the index** at once. The daemon boots from one TOML, catches up to the network tip from a `LedgerBackend` (BSB by default), and serves. The lifecycle (Phase 2) is just a *second* caller of these same primitives, so there is no separate "events" or "tx-hash" slice — each data type is only a `Kind`/CF/cold-writer parameter to `processChunk`, and the index is a backfill primitive.

### Context
The daemon keeps two tiers — **hot** (recent ledgers in a per-chunk RocksDB) and **cold** (older ledgers as immutable files). All durable state is tracked in a small **catalog** (meta-store RocksDB) via a strict **one-write protocol** (mark→fsync→flip), so cleanup walks catalog keys and never lists directories. This phase builds that substrate and the primitives that materialize cold artifacts, then drives them from a backend until on-disk coverage reaches the tip. Phase 1 reads from the cold tier only; the hot DB and the live loop arrive in Phase 2.

### Scope

**Foundations — catalog, protocol, geometry, config**
- **Geometry & layout** (on `pkg/chunk`):
  - chunk id ↔ first/last ledger; `LedgersPerChunk = 10_000`; genesis = ledger 2
  - bucket id (`chunk/1000`, `%05d`) for the on-disk directory layout
  - the signed-chunk sentinel `chunk −1` for the "nothing ingested yet" watermark
  - window arithmetic: `indexID`, `windowFirst/LastChunk`, `chunksInIndex`, `MaxChunksPerTxhashIndex`, coverage `[lo,hi]`
- **Catalog key schema**:
  - key families: `chunk:{c:08d}:{ledgers|events|txhash}`, `hot:chunk:{c:08d}`, `index:{w}:{lo}:{hi}`, `config:*` pins
  - a strict **key↔path bijection** — every key maps to exactly one file path
  - states: `freezing|frozen|pruning` (per-chunk artifacts) and `transient|ready` (hot DB)
[...]

Issue #777, opened 2026-06-08
RPC v2 — Roadmap


tamirms:
# RPC v2 — Roadmap

Living roadmap for shipping Stellar RPC v2. At a minimum, it's refreshed each sprint to set the next sprint's targets.

Status: ✅ done · 🟡 in progress · ⬜ not started.
Priority: **P0** must-ship for MVP · **P1** important, not MVP-blocking · **P2** deferrable · **P3** furthest out.

## 🎯 Sprint 75 — target
- **Query serving**
  - implement the v2 events API; support the v1 events API in rpc v2
    - P0 — #774 — @urvisavla
  - serve the API from the v2 stores: handlers shared with v1, rpc v1 and v2 built as separate binaries from the same repo/branch
    - P0 — #772 — @karthikiyer56
- **Benchmarks**
  - experiments and writeup to achieve the Phase 1–3 ingestion targets
    - P0 — #800 — @tamirms
  - query benchmark commands: reads across all data types
    - P1 — #856 — @marwen-abid
  - read-endpoint SLAs under ingestion load
    - P1 — #870 — @marwen-abid
  - PoC: speed up transaction-status reads (ledger hint on getTransaction, WebSocket status push)
    - P2 — #905 — @marwen-abid
  - PoC: remote captive Core service, streaming ledgers to RPC over the network
    - P2 — #906 — @marwen-abid

---

The work to ship an MVP release of RPC v2 falls under the categories below. Each sprint, we make progress by completing tasks across any number of them.

Not all work is broken into issues yet — some is design-gated. We settle the design first, then create the implementation issues from it.

## DB layer — ✅ · P0
*When complete: any ledger, transaction, or event from anywhere in history can be written and read back — recent data in a live, writable database, older data in immutable files written once and never changed. These are the storage read/write primitives; wiring them to the JSON-RPC endpoints comes later (Query serving).*

| Data type | Status | Issues |
|---|---|---|
| Ledger | ✅ | #695 / #725 / #583 / #584 |
| Events | ✅ | core #756 ✅, query engine #663 ✅ |
| TxHash | ✅ | #729, #728 |
| XDR view extractors | ✅ | #764 |
| Ingestion | ✅ | #765 |

## Unified ingestion workflow — ✅ · P0
*When complete: one daemon mode that catches up on startup (backfilling history as a subroutine), then ingests live ledgers from captive core — freezing completed chunks from the live database into immutable files, and pruning data once it ages out of the retention window.*
[...]

PR #818, opened 2026-06-24
Streaming daemon: Phase 1 layer 2 — primitives (#815)

chowbao:
**Part of #815** — Phase 1 (Backfill), layer 2 of 3. Stacked on [PR1A](https://github.com/stellar/stellar-rpc/pull/817) (`streaming-phase1-foundations`).

The **primitives** — single-pass cold materialization for all three data types:

- `processChunk` — materializes a chunk's ledgers (`.pack`), events (cold segment), and tx-hash (sorted `.bin`) artifacts for every `Kind`, with per-kind idempotency
- the cold `backfillSource` (frozen `.pack` → bulk backend)
- `buildTxhashIndex` — k-way merge of the in-window `.bin` runs into a sorted `.idx` (byte-format-identical to the single-index build)
- adds the `ingest.RunColdChunk` single-chunk cold primitive

**Verification:** `go build` + `go vet` + `go test -short` green on `./cmd/stellar-rpc/internal/fullhistory/...` (cgo RocksDB toolchain). Note: the full `cmd/stellar-rpc` binary link requires the Rust `libpreflight`/`libxdr2json` (CI `make build-libs`); `go vet ./cmd/stellar-rpc/` type-checks the entrypoint locally. `golangci-lint` runs in CI.

Stack: `streaming-phase1-primitives` → `streaming-phase1-foundations`


chowbao (commented):
_(summary removed — see inline notes)_

cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
```diff
@@ -331,3 +314,107 @@ func runOneChunkCold(
...
+// single-chunk, explicit-layout sibling of RunCold: same constructors,
+// ColdService, and drain loop (sequence/overrun validation + full-range check
+// before Finalize), differing only in producing one chunk and taking explicit
+// per-type roots (via buildColdIngestersIn).
+//
+// The cold ingesters overwrite any prior attempt's files at their canonical
+// paths (the package doc's artifact model), so RunColdChunk is the
+// re-materialization primitive the streaming freeze protocol drives: a partial
+// from a crashed attempt is inert scratch the next call overwrites.
+func RunColdChunk(
```
chowbao: Single-pass cold materialization of one chunk — [The primitives](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#the-primitives) under the [One write protocol](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#one-write-protocol). `ColdDirs` realizes the split [Directory layout](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#directory-layout) (`.bin` under `txhash/raw`).

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,281 @@
...
+//   - Mark-then-write: put each remaining kind "freezing" BEFORE any I/O, write
+//     the files (RunColdChunk, from the source backfillSource chose), fsync each
+//     file + its dirents (BarrierNewFile), then flip the keys to "frozen".
+//
+// It re-derives no extractor or writer — RunColdChunk is the same merged cold
+// ingester set RunCold uses; processChunk only chooses the source and drives the
+// protocol around the freeze.
+//
+//nolint:cyclop // sequential one-write protocol; splitting the ordered steps fragments the invariant
+func processChunk(ctx context.Context, chunkID chunk.ID, artifacts catalog.ArtifactSet, cfg ProcessConfig) error {
```
chowbao, line 64: Per-chunk freeze via the [One write protocol](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#one-write-protocol) (rule 1): mark-then-write → fsync barrier → flip frozen; a frozen kind self-skips ([Catalog keys](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#catalog-keys)).

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,281 @@
...
+
+// backfillSource implements the cold source-preference order for one chunk. It
+// returns the chosen ingest.ChunkSource, a closer (a no-op for the cold pack /
+// bulk branches), and an error.
+//
+// Preference order:
+//  1. The frozen local .pack via the ledger cold reader, when ledgers is NOT among
+//     the requested outputs (re-derivation without a download).
+//  2. The configured bulk backend, gated by a bounded WaitForCoverage.
+func backfillSource(
```
chowbao, line 140: Source-preference order from [The primitives](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#the-primitives); the bulk wait is [Scenario coverage §3](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#scenario-coverage). (Design's hot-DB branch is deferred to Phase 2 / #820.)

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
```diff
@@ -0,0 +1,238 @@
...
+//  4. Commit — CommitTxHashIndex's one atomic synced batch.
+//
+// buildTxhashIndex never deletes a file: removal is exclusively the sweeps' job
+// (buildThenSweep / the tick's prune scan). The ordering covers the crash matrix
+// (§7.6): a crash before step 4 leaves the predecessor frozen and the new
+// coverage as "freezing" debris; after, the new coverage is frozen and the
+// demoted keys are "pruning" sweep work.
+//
+//nolint:cyclop // sequential one-write protocol; splitting the ordered steps fragments the invariant
+func buildTxhashIndex(ctx context.Context, w geometry.TxHashIndexID, lo, hi chunk.ID, cfg BuildConfig) error {
```
chowbao, line 50: The rolling rebuild — [§7.2](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/gettransaction-full-history-design.md#72-the-rebuild); the atomic commit upholds [INV-2](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#invariants) (one frozen index per window). ⚠️ the inline `§7.6` refs have no match in #787 — crash-safety is in the streaming doc ([Convergence](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/full-history-streaming-workflow.md#convergence)).

cmd/stellar-rpc/internal/fullhistory/backfill/perf_test.go
```diff
@@ -0,0 +1,251 @@
...
+//     drives on every boundary (build.go); and
+//  2. a direct txhash.BuildColdIndex over the identical inputs — the merged
+//     cold path the bench harness on rpc-hack measures.
+//
+// The two .idx files must be byte-for-byte identical. That is what licenses
+// PERF.md to transfer the bench harness's measured ≈4.2 B/tx and ≈1-min
+// figures to the streaming daemon: the streaming rebuild is not a re-derivation
+// of the format, it is the same txhash.BuildColdIndex call wrapped in catalog
+// bookkeeping, and the bookkeeping does not touch the artifact.
+func TestStreamingRebuild_ByteIdenticalToColdPath(t *testing.T) {
```
chowbao, line 57: Pins the `.bin`/`.idx` formats to [§6.1](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/gettransaction-full-history-design.md#61-the-per-chunk-sorted-run-bin)/[§6.2](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/gettransaction-full-history-design.md#62-the-per-window-index-idx) and the byte-identity that lets [Part 4](https://github.com/stellar/stellar-rpc/blob/c586667a150777a724434596710b2a944c77f438/design-docs/gettransaction-full-history-design.md#part-4-capacity--performance) sizing transfer to the daemon.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/catalog"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/geometry"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/ingest"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/pkg/chunk"
+)
+
+// ErrBackendCoverageTimeout is returned when the bulk backend's tip never reaches
+// the chunk within the wait deadline.
+var ErrBackendCoverageTimeout = errors.New("streaming: backend never covered chunk within deadline")
```
chowbao: Error messages have `streaming:` in them. We removed the streaming package name but I don't know a better term to prefix on error messages. We could use `backfill` but that seems worse IMO

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
```diff
@@ -0,0 +1,207 @@
...
+	if newIndexDir {
+		if mkErr := os.MkdirAll(indexDir, 0o755); mkErr != nil {
+			return fmt.Errorf("streaming: buildTxhashIndex mkdir %s: %w", indexDir, mkErr)
+		}
+	}
+
+	minLedger := lo.FirstLedger()
+	maxLedger := hi.LastLedger()
+	if berr := txhash.BuildColdIndex(ctx, inputs, idxPath, minLedger, maxLedger, cfg.BuildOpts...); berr != nil {
+		return fmt.Errorf("streaming: buildTxhashIndex build window %s coverage [%s,%s]: %w", w, lo, hi, berr)
```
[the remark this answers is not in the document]
chowbao: This is already a known edge case. Choosing to skip for now as pubnet won't have empty .bin from no transactions

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,230 @@
...
+	if timeout <= 0 {
+		timeout = 5 * time.Minute
+	}
+	return &pollingBackendWaiter{Tip: tip, Interval: interval, Timeout: timeout}
+}
+
+func (w *pollingBackendWaiter) WaitForCoverage(ctx context.Context, chunkLastLedger uint32) error {
+	deadline := time.Now().Add(w.Timeout)
+	for {
+		tip, err := w.Tip(ctx)
```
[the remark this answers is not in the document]
chowbao: Addressed in `808eb622` — `WaitForCoverage` now bounds each `Tip` call by the overall deadline (`context.WithDeadline(ctx, deadline)`), so a hung tip query can't outlast `Timeout`. Covered by `TestPollingBackendWaiter_TipBoundedByDeadline`.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if cfg.Backend == nil {
+		return nil, noClose, fmt.Errorf(
+			"streaming: chunk %s has no local copy and no bulk backend is configured", chunkID)
+	}
+	if cfg.BackendWaiter != nil {
+		if werr := cfg.BackendWaiter.WaitForCoverage(ctx, chunkID.LastLedger()); werr != nil {
+			return nil, noClose, werr
+		}
+	}
```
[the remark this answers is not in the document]
chowbao: Addressed in `808eb622` — the bulk branch now *requires* `BackendWaiter` (matching the design's unconditional `waitForBackendCoverage` and the field doc), so a missing waiter is a loud config error instead of a silently-skipped coverage gate. Covered by `TestBackfillSource_BulkBackendRequiresWaiter`.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+			return fmt.Errorf("streaming: backend tip query: %w", err)
+		}
+		if tip >= chunkLastLedger {
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("%w: tip %d < needed %d after %s",
+				ErrBackendCoverageTimeout, tip, chunkLastLedger, w.Timeout)
+		}
+		timer := time.NewTimer(w.Interval)
```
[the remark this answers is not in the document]
chowbao: I'm not sure this is a problem

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+func (w *pollingBackendWaiter) WaitForCoverage(ctx context.Context, chunkLastLedger uint32) error {
+	deadline := time.Now().Add(w.Timeout)
+	for {
+		// Bound the tip query by the overall deadline — the parent ctx may carry no
+		// deadline, so a hung backend would otherwise outlast Timeout.
+		tipCtx, cancel := context.WithDeadline(ctx, deadline)
+		tip, err := w.Tip(tipCtx)
+		cancel()
+		if err != nil {
+			return fmt.Errorf("streaming: backend tip query: %w", err)
```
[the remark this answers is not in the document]
chowbao: I'm not sure this is a problem

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
tamirms: You don't need a prefix at all — that's the Go convention for *wrapped* errors. Every error here is wrapped with `%w`, so the chain already carries the context (`cold ingest chunk 5: open stream: connection refused`), and a package prefix just repeats at every wrap level (`fullhistory: cold ingest: fullhistory: open stream: …`). The stdlib uses a package prefix only on leaf/sentinel errors that aren't wrapped; for wrapped contextual errors the operation + the ids you already include are the useful part. So rather than choosing between `streaming`/`backfill`/`fullhistory`, I'd just drop it: `fmt.Errorf("cold ingest chunk %s: %w", …)`. If you ever want subsystem attribution, that's the logger's (or a sentinel error's) job, not a prefix on every message.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+	defer func() { _ = closeSource() }()
+
+	if err := cat.MarkChunkFreezing(chunkID, kinds...); err != nil {
+		return fmt.Errorf("streaming: mark freezing chunk %s %s: %w", chunkID, artifacts, err)
+	}
+
+	// Snapshot which bucket dirs exist before the write, so the barrier below
+	// fsyncs the grandparent dirent only for dirs THIS freeze created. (Buckets
+	// are created in arbitrary order, so a chunk-id heuristic can't tell.)
+	bucketExisted := make(map[string]bool)
```
tamirms: Could this snapshot and the `newParent` plumbing into `BarrierNewFile` just go away? It exists to skip the grandparent fsync when the bucket already existed — but fsyncing an unchanged directory is nearly free (no dirty metadata to flush). If `BarrierNewFile` always did file → parent → grandparent, you'd delete this loop, the `newParent` parameter, and the `newIndexDir` branch in `txindex.go`, for one fewer before-the-write step to get subtly wrong. (geometry's `DeepestExistingDir`/`FsyncNewDirs` already do "fsync a created dir chain" if you'd rather keep the optimization but not hand-roll it.)

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Choose the source before marking "freezing": a source error must not leave
+	// "freezing" debris for a chunk we then refuse to produce.
+	source, closeSource, err := backfillSource(ctx, chunkID, artifacts, cfg)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = closeSource() }()
+
+	if err := cat.MarkChunkFreezing(chunkID, kinds...); err != nil {
```
tamirms: This mark → write → barrier → flip skeleton is the crash-safety invariant, and it's spelled out twice (here and in `buildTxhashIndex`), with a third copy in the queued live-ingestion PR — `openHotTierForChunk` is `PutHotTransient` → mkdir/open → fsync → `FlipHotReady`. Three sites share this exact order. A small `oneWrite(mark, create, barrier, flip)` helper would keep the order in one place so none of the three can drift; the states and the create step differ per site, the order doesn't. Worth establishing here and adopting in the hot-tier open.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+	if werr := cfg.BackendWaiter.WaitForCoverage(ctx, chunkID.LastLedger()); werr != nil {
+		return nil, noClose, werr
+	}
+	cfg.Logger.Debugf("backfillSource: chunk %s from bulk backend", chunkID)
+	return cfg.Backend, noClose, nil
+}
+
+// pollingBackendWaiter polls Tip until it reaches chunkLastLedger, ctx is
+// canceled, or Timeout elapses.
+type pollingBackendWaiter struct {
```
tamirms: `github.com/cenkalti/backoff/v4` is already a direct dependency — this hand-rolled poll loop could use it and shed the timer/deadline/`select` bookkeeping. `backoff.Retry` with a `ConstantBackOff` (the interval), `WithContext` (cancellation), and `WithMaxElapsedTime` (the timeout) covers all of it; wrap the tip-query error in `backoff.Permanent` to keep its current fatal-on-error behavior. The executor's own retry logic (#819) could share the same primitive too.

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
```diff
@@ -0,0 +1,207 @@
...
+		})
+	if err != nil {
+		return nil, err
+	}
+	return refs, nil
+}
+
+// forEachChunkTxHashState calls fn with each chunk's txhash state over [lo, hi].
+// Centralizes the chunk.ID wraparound guard (the top can be the max id).
+func forEachChunkTxHashState(cat *catalog.Catalog, lo, hi chunk.ID, fn func(chunk.ID, geometry.State) error) error {
```
tamirms: go.mod is on 1.26, and the catalog already iterates this way (`for e, err := range c.store.PrefixScan(...)`), so this callback is a bit against the grain. As an `iter.Seq2[chunk.ID, geometry.State]` it'd read `for cid, state := range cat.TxHashStates(lo, hi) { … }`, and the two callers (`txhashBinInputs`, `demotedTxhashRefs`) could use a plain `for` with `break`/`continue` instead of returning sentinel errors from a closure.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+	if rerr != nil {
+		return fmt.Errorf("streaming: cold ingest chunk %s %s: %w", chunkID, artifacts, rerr)
+	}
+
+	// Durability barrier before the keys flip: fsync each file + its dirents
+	// (grandparent too for a bucket dir this freeze created).
+	for _, kind := range kinds {
+		for _, path := range layout.ArtifactPaths(chunkID, kind) {
+			newParent := !bucketExisted[filepath.Dir(path)]
+			if berr := geometry.BarrierNewFile(path, newParent); berr != nil {
```
tamirms: Side note, visible here but really a #824 / geometry-package thing: the fsync helpers `BarrierNewFile` & co. don't belong in `geometry`. `BarrierNewFile`, `FsyncDir`, `FsyncParentDirs`, `DeepestExistingDir`, `FsyncNewDirs`, `DeleteFileIfExists`, `RmdirIfEmpty` are generic `os`-level durability ops with **zero** geometry or domain references — they only landed in `geometry` because they shared `paths.go` with the path bijection in the monolith. `geometry` should be naming + math (identity and location, no I/O); crash-safe file ordering is a different concern that'd sit better in a small `durable`/`fsync` leaf package imported by `catalog`, `backfill`, and the hot-tier code. Worth folding into the #824 split.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Done in 4f85de7d — dropped the prefix entirely rather than picking a new one

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Done in `7a51cf81`. `BarrierNewFile` now always fsyncs file → parent → grandparent (the `newParent` bool is gone), so the `bucketExisted` snapshot loop in `processChunk` and the `newIndexDir` branch in `buildTxhashIndex` both disappear — and with the stat gone the index dir is just an idempotent `os.MkdirAll`. Always-fsync-grandparent is a strict superset of the old durability (an unchanged dir has no dirty metadata to flush, so it's nearly free), so it's never less safe.

I kept the hand-rolled two-level barrier rather than reaching for `DeepestExistingDir`/`FsyncNewDirs`: after the startup root creation, only a single bucket/index dir is ever created on demand, so file → parent → grandparent is exactly the depth needed — those helpers are for the multi-level root chain at startup.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Done in `7a51cf81`. Added `oneWrite(mark, create, barrier, flip func() error) error` and adopted it in both `processChunk` and `buildTxhashIndex`, so the mark → create → barrier → flip order is defined once and the sites only supply the per-step closures.

I scoped it to the `backfill` package for now since both current callers live here; when #820 adds the hot-tier `openHotTierForChunk` (`PutHotTransient` → mkdir/open → fsync → `FlipHotReady`) it can promote/relocate `oneWrite` and adopt it as the third caller. Folding the ordering out also dropped both functions under the cyclop threshold (13 < 15), so I removed their `//nolint:cyclop` directives.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Done in `7a51cf81`. `WaitForCoverage` now drives `backoff.Retry`: the tip-query error is wrapped in `backoff.Permanent` (stays fatal), and the low-tip case returns the `ErrBackendCoverageTimeout`-wrapped error — which `Retry` surfaces once `MaxElapsedTime` stops the loop, so callers still classify the timeout via `errors.Is`.

Two deviations from the literal recipe, both deliberate:
- `WithMaxElapsedTime` is an `ExponentialBackOffOpts`, not a wrapper around `ConstantBackOff`, so the constant poll is expressed as `NewExponentialBackOff` with `WithMultiplier(1)` + `WithRandomizationFactor(0)` (+ `WithInitialInterval`/`WithMaxInterval` = the interval).
- I kept a per-call `context.WithDeadline` inside the operation: the backoff bounds total *retry* time but not a single in-flight `Tip` call, so dropping it would regress the hung-query bound added in `808eb622` (and its `TestPollingBackendWaiter_TipBoundedByDeadline`).

Added `TestPollingBackendWaiter_TimeoutReturnsSentinel` to cover the timeout→sentinel path. Agreed the #819 executor retry can share this primitive.

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
chowbao: Done in `7a51cf81`. `forEachChunkTxHashState` is now `txHashStates(...) iter.Seq2[…, error]`, and `txhashBinInputs`/`demotedTxhashRefs` use a plain `for cs, err := range …` with normal error handling and `continue` instead of returning sentinel errors from a closure.

One adjustment from the sketch: since `catalog.State` can return an error, I followed the codebase's `iter.Seq2[T, error]` scan convention (the same shape as `PrefixScan`/`IterateLedgers`) rather than `iter.Seq2[chunk.ID, geometry.State]` — yielding a small `chunkTxHashState{Chunk, State}` value paired with the error. I left it as a `backfill` free function (not a `cat.TxHashStates` method) to avoid growing the catalog API for a backfill-local helper, but happy to move it to a method if you'd prefer that form.

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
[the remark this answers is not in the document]
tamirms: Agreed it's fine to keep out of #818 — but I'd capture it as a tracked follow-up rather than skip it in a comment, because it's more than a quiet-network optimization: it stalls the daemon. When a coverage has zero transactions, every `.bin` is empty, `BuildColdIndex` returns `ErrEmptyBuildSet` and writes no `.idx`, so `buildTxhashIndex` fails, the coverage stays `freezing`, and the next tick re-plans and fails again — the daemon can't advance past that window. And since the rolling index builds `[lo, current]`, a window that merely *starts* with a quiet chunk trips it on the first build, so it's routine on testnet/futurenet/standalone (and pubnet's early history during a full-history backfill), not only a fully-empty window.

Could you open a follow-up issue to track it? The cleaner fix is probably upstream — have the txhash cold format / streamhash treat a zero-key index as a valid empty result (reads resolve to not-found) instead of `ErrEmptyBuildSet`, so `buildTxhashIndex` needs no special-casing. A stopgap here (treat `ErrEmptyBuildSet` as a frozen-empty coverage) would unblock it if the streamhash change lands later.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Done now instead — promoted `oneWrite` to `catalog.OneWrite` in `catalog_protocol.go`, right under the protocol header it implements (the file already documents the same mark→create→barrier→flip steps). `processChunk` and `buildTxhashIndex` now call it across the package boundary.

It's a zero-dependency pure function and `catalog` never imports `backfill`, so there's no import cycle; #820's `openHotTierForChunk` adopts it as the third caller by import alone, with no later relocation. Pushed in `7e56b926`.

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
chowbao: Added to #824

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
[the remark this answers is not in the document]
chowbao: Opened #826 to track this. Added comment for this issue in https://github.com/stellar/stellar-rpc/issues/777 as well

cmd/stellar-rpc/internal/fullhistory/backfill/process.go
```diff
@@ -0,0 +1,238 @@
...
+	rerr := ingest.RunColdChunk(ctx, cfg.Logger, source, dirs, chunkID, cfg.Sink, ingestConfigFor(artifacts))
+	if rerr != nil {
+		return fmt.Errorf("cold ingest chunk %s %s: %w", chunkID, artifacts, rerr)
+	}
+
+	// Durability barrier before the keys flip: fsync each file + its dirents
+	// (grandparent too for a bucket dir this freeze created).
+	for _, kind := range kinds {
+		for _, path := range layout.ArtifactPaths(chunkID, kind) {
+			newParent := !bucketExisted[filepath.Dir(path)]
```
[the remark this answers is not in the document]
chowbao: Addressed in `7a51cf81`. `BarrierNewFile` now fsyncs the grandparent unconditionally, so the `bucketExisted` snapshot and the `newParent` skip are gone — along with the retry-after-failed-attempt hole this flagged. That is the "fsync the grandparent unconditionally" remedy: on a reused bucket dir left by a non-frozen attempt, the grandparent dirent is now persisted before `FlipChunkFrozen` regardless of which attempt created the dir.

cmd/stellar-rpc/internal/fullhistory/backfill/txindex.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	minLedger := lo.FirstLedger()
+	maxLedger := hi.LastLedger()
+	if berr := txhash.BuildColdIndex(ctx, inputs, idxPath, minLedger, maxLedger, cfg.BuildOpts...); berr != nil {
+		return fmt.Errorf("buildTxhashIndex build window %s coverage [%s,%s]: %w", w, lo, hi, berr)
+	}
+
+	if barErr := geometry.BarrierNewFile(idxPath, newIndexDir); barErr != nil {
```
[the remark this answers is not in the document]
chowbao: Addressed in `7a51cf81`. `BarrierNewFile` now fsyncs the grandparent unconditionally, and the index dir is a plain idempotent `os.MkdirAll` (the `newIndexDir` stat/branch is gone). So on a reused index dir left by a canceled/failed build, the grandparent dirent is persisted before `CommitTxHashIndex` regardless of which attempt ran the `MkdirAll` — closing the crash window this flagged.

tamirms:
**Two remaining items in the `backfill`/`ingest` area — both more structural than the last round, and both cheap to do now precisely because none of this is wired into the daemon yet (`ProcessConfig`/`processChunk`/`RunColdChunk` have no production constructor; the daemon still runs the legacy `ingest.BackfillMeta` path). Fixing them before the streaming layers stack on top is much cheaper than after.**

## 1. The ledger-source boundary between `backfill` and `ingest`

**The problem.** `ingest.ChunkSource` exposes only `OpenStream(chunkID) → LedgerStream`, which discards the one thing the bulk source can actually report: how far it currently reaches. To get that back, `backfill` bolts on a parallel apparatus — `BackendWaiter` + `pollingBackendWaiter` + an injected `Tip func` — to poll a "latest ledger" the source already knows. Separately, `ingest` still carries the dead multi-chunk batch path (`RunCold` → `runOneChunkCold` → `buildColdIngesters`) next to the live single-chunk `RunColdChunk`, which duplicates it almost line-for-line. These are two symptoms of one thing: the source abstraction is drawn in the wrong place.

**What the SDK already gives us.** `ledgerbackend.LedgerStream` (`RawLedgers(ctx, range) iter.Seq2[[]byte,error]`) is a universal streaming currency everything already speaks — the local pack (`packStream` already implements it) and the external backends (`NewBufferedStorageStream`, `NewCaptiveCoreStream`) all return it. So `ChunkSource`/`OpenStream` is a redundant wrapper over it.

The tip is a separate value — and one trap worth flagging: the SDK's `LedgerBackend.GetLatestLedgerSequence` is **not** the tip we want. It's prepared-relative — on `BufferedStorageBackend` it errors before `PrepareRange` and otherwise returns the buffer's latest, not the source's frontier. The frontier tip is an independent value per backend: `datastore.FindLatestLedgerSequence` for the lake, and the history archive's `GetRootHAS().CurrentLedger` for captive core (whose own `GetLatestLedgerSequence` likewise needs a running, prepared core, so it's no use here).

**Proposed shape.** The materializer becomes source-blind — it takes the raw ledger iterator and nothing else, and is faked in tests with a literal slice iterator:

```go
func WriteColdChunk(ctx, logger, chunkID chunk.ID,
	raw iter.Seq2[[]byte, error], dirs ColdDirs, sink MetricSink, cfg Config) error
```

`backfill` owns the external-source seam, one interface:

```go
// An external ledger source (BSB now, captive core later) we fetch from and may
// have to wait for. The local frozen .pack is NOT a Backend — it's complete, so
// it's a bare ledgerbackend.LedgerStream with nothing to wait on.
//   bsbSource     (now)   Tip = datastore.FindLatestLedgerSequence
//   captiveSource (later) Tip = historyarchive GetRootHAS().CurrentLedger
type Backend interface {
	ledgerbackend.LedgerStream
	Tip(ctx context.Context) (uint32, error)
}
type ProcessConfig struct { /* … */ Backend Backend }   // was: Backend ChunkSource + BackendWaiter + Tip func
```

This `Backend` is a **new** interface and deliberately *not* an evolution of today's `BackendWaiter` — that interface, its `pollingBackendWaiter` implementation, and `NewPollingBackendWaiter` are removed entirely. The coverage-polling they did becomes just the `Tip` method above plus the `waitForCoverage` function below; nothing in `backfill` implements `BackendWaiter` anymore.

The one impl wired now:

```go
type bsbSource struct {
	ledgerbackend.LedgerStream          // NewBufferedStorageStream
	ds datastore.DataStore              // long-lived; tip only; daemon-Closed
}
func (s *bsbSource) Tip(ctx context.Context) (uint32, error) { return datastore.FindLatestLedgerSequence(ctx, s.ds) }
```

Coverage-waiting collapses to one `waitForCoverage` function that polls `Backend.Tip`. This is what removes `BackendWaiter`, `pollingBackendWaiter`, and `NewPollingBackendWaiter`:

```go
func waitForCoverage(ctx, b Backend, target uint32, interval, timeout time.Duration) error // polls b.Tip
```

The local pack moves to its codec owner as a bare stream, constructed inside `backfillSource` (not injected) when a chunk's ledgers are already frozen locally and we're deriving the other artifacts from them:

```go
func NewPackStream(path string) ledgerbackend.LedgerStream   // the existing packStream, relocated to pkg/stores/ledger
```

so the call site is uniform — both arms are just a `LedgerStream`:

```go
src, err := backfillSource(ctx, chunkID, artifacts, cfg)   // pack OR cfg.Backend
raw := src.RawLedgers(ctx, ledgerbackend.BoundedRange(chunkID.FirstLedger(), chunkID.LastLedger()))
ingest.WriteColdChunk(ctx, cfg.Logger, chunkID, raw, dirs, cfg.Sink, ingestConfigFor(artifacts))
```

**Captive core is deferred, but the path stays open.** No need to implement it now — the `Backend` interface is the seam. It's one new struct (embed `NewCaptiveCoreStream` + a `Tip` from the history archive) plus one daemon factory case, with zero changes to `ProcessConfig`, `backfillSource`, `waitForCoverage`, or `WriteColdChunk`. Captive's tip comes from `historyarchive.NewArchivePool(cfg.HistoryArchiveURLs, …).GetRootHAS().CurrentLedger`. Listing the intended impls in the `Backend` doc comment keeps the intent visible without a stub.

**Net:** delete `ingest/source.go` whole (`ChunkSource`, `ChunkSourceFunc`, `packSource`, `dataStoreSource`, `NewDataStoreSource`); the dead `RunCold`/`runOneChunkCold`/`buildColdIngesters`; and `backfill`'s `BackendWaiter`/`pollingBackendWaiter`/`NewPollingBackendWaiter`/`Tip`. Rename `buildColdIngestersIn` → `buildColdIngesters` and `RunColdChunk` → `WriteColdChunk`. Roughly −7 types and −3 constructors for +1 interface.

Two things to keep honest in the code: the `ingest` *package* still imports `ledgerbackend` for `RunHot`, so only the materializer function is source-blind, not the package; and since source resolution (pack-stat, coverage wait) now runs in `backfill` before `WriteColdChunk`, a pack-missing or coverage-timeout failure is metered there rather than as an ingest `ColdChunkTotal` attempt — worth a line in the doc.

## 2. `catalog.OneWrite` — revert the two backfill sites to straight-line

First, apologies for the round-trip: I asked for the `oneWrite` wrapper last round, and having gone through both call sites properly since, straight-line reads better and I'd suggest reverting it. Sorry for sending you back and forth on it.

Why the wrapper doesn't pay off: every step wraps its error with per-site context (`chunkID`/`artifacts` in `processChunk`; `w`/`lo`/`hi`/`cov`/`idxPath` in `buildTxhashIndex`), so the closures can't collapse to one-liners — they stay multi-line and carry the closure wrapper on top. And `buildTxhashIndex` pays extra: because the flip needs the `cov` the mark returns, it hoists `var cov …; var idxPath string` above the call and does `var merr error; cov, merr = …` (no `:=`). Straight-line, those are ordinary downward locals:

```go
// one-write:mark — returns the coverage the commit (flip) needs.
cov, err := cat.MarkTxHashIndexFreezing(w, lo, hi)
if err != nil { return fmt.Errorf("buildTxhashIndex mark freezing %s: %w", geometry.TxHashIndexKey(w, lo, hi), err) }
idxPath := layout.TxHashIndexFilePath(cov)
// one-write:create  (os.MkdirAll + txhash.BuildColdIndex(ctx, inputs, idxPath, …))
// one-write:barrier (geometry.BarrierNewFile(idxPath))
// one-write:flip — CommitTxHashIndex(cov) promotes cov, demotes predecessor, demotes window .bin on terminal.
if err := cat.CommitTxHashIndex(cov); err != nil { return fmt.Errorf("buildTxhashIndex commit %s [%s,%s]: %w", w, lo, hi, err) }
return nil
```

Nothing is lost on safety: the wrapper only sequences four calls — it never checks their contents (a wrong fsync path, the wrong catalog method, a dropped step all pass it), so it isn't what guarantees crash-safety. That guarantee is `catalog/crashsafety_test.go`, which drives the protocol to each crash point and asserts the surviving state is recoverable. A `// one-write:mark|create|barrier|flip` label per step keeps the four beats visible and greppable. `processChunk` converts the same way.

So: delete `catalog.OneWrite` and inline both sites.

chowbao:
Both done in `096463ef`.

**1. Source boundary.**
- Deleted `ingest/source.go` (`ChunkSource`/`packSource`/`dataStoreSource`) and the dead `RunCold` path.
- Materializer is now source-blind: `WriteColdChunk(ctx, logger, chunkID, raw iter.Seq2[[]byte, error], dirs, sink, cfg)`.
- `buildColdIngestersIn` → `buildColdIngesters`; `packStream` → `ledger.NewPackStream(path)`.
- `backfill` owns the seam: `Backend{ ledgerbackend.LedgerStream; Tip }` (bsbSource now, captive later), `BackendWaiter`/polling collapsed into `waitForCoverage`, `ProcessConfig` down to one `Backend`, `backfillSource` returns a bare `LedgerStream`.

**2.** `catalog.OneWrite` reverted — inlined straight-line at both sites with `// one-write:` labels.

tamirms (approved):
Approving. The source-boundary redesign and the `OneWrite` revert both landed cleanly in `096463ef`, and the rest of the primitives layer matches the design — the resolver/executor and daemon wiring are correctly left to their later layers, and the empty-window (#826) and `geometry`/fsync (#824) items are tracked. Build and unit tests are green; the two red checks are the unrelated XDR Go/Rust rev mismatch.

A few optional cosmetic nits, none blocking:

- `process.go:93` — the one-write comment is garbled (a dangling "The" and a doubled `// //`); `txindex.go` has the clean version to copy.
- `geometry/paths.go:109` — the `TxHashRawRoot` doc still references `RunCold`, which this PR removed; the `TestRunCold_*` names in `ingest_test.go` are similarly stale now that they call `WriteColdChunk`.
- `ingest/service.go` — the `ColdService` struct comment says the timer starts "from the Finalize/Close call," but it actually starts at construction (the constructor comment is the accurate one).

Issue #772, opened 2026-06-08
Wire the v2 stores into the API handlers and support both v1 and v2


tamirms:
## TL;DR
Serve the RPC API from the v2 stores. v1 and v2 will coexist through a deprecation window, so the v1 code is not removed as originally planned: both binaries build from this repo and share the JSON-RPC handlers, with the data layer abstracted behind store interfaces so the same handler runs over the v1 SQLite stores or the v2 hot + cold stores.

## Architecture
1.  One repo, two binaries. The existing v1 binary keeps its SQLite ingestion and stores; the v2 binary runs the v2 ingestion, lifecycle, and stores. Separate binaries keep the v1 artifact's dependency footprint unchanged (the v2 stores bring in RocksDB/CGO, which the v1 binary should not carry).
2.  Shared handlers over store interfaces. The interfaces are designed around what the handlers need, not inherited from today's internal/db shapes; this is the opportunity to refactor the serving path into its cleanest form. Each interface gets two implementations: SQLite-backed, and v2-backed behind the query router from #865. The seam covers ingestion-fed serving state as well as store reads (getFeeStats serves from in-memory fee windows populated by ingestion, so both pipelines must feed them).
3.  Separate ingestion pipelines. Both binaries build from the same SDK, so a protocol release is absorbed once and reaches both at build time.
4.  Figure out the testing story for the shared handlers: unit tests on the handlers and integration tests, covering both the SQLite-backed and v2-backed configurations.

## Endpoints
Every endpoint reads the ledger range and the protocol version, so every endpoint needs at least that much backed by the v2 stores.

Cut over to the v2 stores behind shared handlers:
- getLedgers
- getTransaction / getTransactions
- getFeeStats (fee windows fed by v2 ingestion)
- getLatestLedger, getHealth, getNetwork, getVersionInfo (ledger range and protocol version only)

Ledger-entry state comes from captive core, not the stores; these read only the ledger range:
- getLedgerEntries, simulateTransaction, sendTransaction

Events:
- getEvents and the getTransaction `events.parsed` field are tracked in #774. Their API and query semantics are out of scope here, but the event store plugs into the same store interfaces, query router, and reader-safety contract this issue establishes, so the interface design must be aligned with #774.

## Outcome
- The `feature/full-history` branch is merged to main and retired. v1 and v2 are maintained together on main in this repo from then on.

## Depends on
- #770 (concurrent serving design) and #865 (query router, implemented in PR #894).

[...]

Issue #808, opened 2026-06-23
[Epic] Unified ingestion workflow: full-history streaming daemon


chowbao:
Tracking issue for implementing the **unified ingestion workflow** — the full-history streaming daemon. Design: **#722** (`design-docs/full-history-streaming-workflow.md` + `gettransaction-full-history-design.md`); part of the **#777** RPC v2 roadmap.

## What we're building
One daemon, one mode: it backfills old history and follows the live network. It keeps two tiers — **hot** (recent ledgers in a per-chunk RocksDB) and **cold** (older ledgers as immutable files) — and does three things:
- **Backfill** on startup: bring on-disk coverage in line with the retention window from a `LedgerBackend` (BSB by default).
- **Ingest** live ledgers from captive core into one multi-CF hot DB per chunk (one atomic synced `WriteBatch` per ledger).
- **Freeze → discard → prune**: migrate full chunks to immutable cold files and prune past-retention/superseded artifacts, in a background lifecycle goroutine.

Durable state lives in a small **catalog** (meta-store) written via a strict **one-write protocol** (mark→fsync→flip), so cleanup walks keys and never lists directories. Correctness rests on four invariants (INV-1 read correctness, INV-2 single canonical state, INV-3 disk↔catalog match, INV-4 retention bound).

## How the work is decomposed
**By phase, not by data type.** Backfill already owns *every* primitive the daemon has — `processChunk` (all per-chunk artifacts), `buildTxhashIndex` (the window index), and `resolve`/`executePlan` — and the lifecycle is just a *second caller* of them (its stage 1 *is* `resolve` + `executePlan`). So the work splits cleanly in two, mirroring the doc ("the startup backfill loop and the lifecycle run are [backfill's] two callers"). There is no separate "events" or "tx-hash" slice — each data type is only a `Kind`/CF/cold-writer parameter to `processChunk`, and the index is a backfill primitive, so all three data types come up together in Phase 1. **Tests ship in the same issue/PR as the code they cover.**

### Sub-issues

- [x] **#815 — Phase 1 · Backfill (catch-up + all primitives).** Boots from one TOML, catches up to the tip from a backend for **all three data types + the window index**, and serves.
  - catalog / one-write protocol / keys / config
  - every primitive: `processChunk` (ledgers + events + tx-hash), `backfillSource`, `buildTxhashIndex` + the `index:{w}` key family + `CommitIndex`
  - `resolve` / `executePlan` + the startup catch-up loop
[...]

PR #819, opened 2026-06-24
Streaming daemon: Phase 1 layer 3 — orchestration + daemon (closes #815)

### LINKED ISSUE #815: Streaming daemon: Phase 1 — Backfill (catch-up + all primitives)

chowbao:
**Closes #815** — Phase 1 (Backfill), layer 3 of 3. Stacked on #818 (`streaming-phase1-primitives`).

Orchestration + the daemon entrypoint, completing the cold-only backfill daemon.

### What this layer adds
- **Derived progress / watermark** (`progress.go`) — recomputed from durable catalog keys on every start, never stored; signed-`int64` sentinel arithmetic so the genesis / "nothing complete" markers can't wrap.
- **`resolve`** (`backfill/resolve.go`) — a pure catalog diff over `[rangeStart, rangeEnd]` → a `Plan`: per-chunk ledgers/events rule + per-window txhash-index rule (with the trailing-window cap). Recomputed every run, so a restart re-plans cleanly.
- **`executePlan`** (`backfill/execute.go`) — one bounded worker pool; index builds wait on their in-coverage chunk builds **before** acquiring a slot (no slot-starvation deadlock at `workers=1`); `cenkalti/backoff` retries.
- **`run` + the catch-up loop** (`startup.go`) — `networkTip`-bounded catch-up (`anchor = max(tip, watermark)`, mid-chunk resume exclusion, first-start-with-no-tip fatal) → `serveReads` handoff. No hot tier / live loop / lifecycle in Phase 1.
- **Daemon entrypoint** (`daemon.go`) — LoadConfig → lock roots → open catalog → build boundaries → `validateConfig` (pin/confirm `earliest_ledger`) → Prometheus registry → supervised run loop. Plus the `full-history` CLI subcommand (folds into the general RPC `start` command at the #772 cutover).
- **Production backend wired** — `buildProductionBoundaries` builds the BSB backend via `backfill.NewBSBBackendFromConfig`, releasing the datastore at shutdown; frontfill-only when no datastore is configured. `[backfill.datastore]` + `[backfill.bsb]` reuse the SDK config types verbatim (`datastore.DataStoreConfig` + `ledgerbackend.BufferedStorageBackendConfig`), so **any** SDK datastore — GCS, S3, Filesystem — works as the bulk source. The daemon drives the cold tier through `ingest.ColdService` + `NewPrometheusSink`.

This branch is a runnable backfill daemon: it boots from one TOML, catches up from the configured lake (or an injected backend), and serves — the read server itself is a #772 no-op for now (reads still come from the v1 daemon until the cutover).

### Reviewing
A per-file design map (each file → its #815 deliverable + #787 design section) and a suggested reading order are posted as review comments on this PR.

### Verification
`go build` + `go vet` + `go test -short` green on `./cmd/stellar-rpc/internal/fullhistory/...` (cgo RocksDB toolchain); `golangci-lint` runs in CI. The full `cmd/stellar-rpc` binary link needs the Rust `libpreflight`/`libxdr2json` (CI `make build-libs`); `go vet ./cmd/stellar-rpc/` type-checks the entrypoint locally.

Stack: `streaming-phase1-daemon` → `streaming-phase1-primitives` (#818)


chowbao:
**Note (non-blocking): a zero-txhash-key window → infinite restart loop on synthetic/standalone networks**

Surfaced while writing the daemon catch-up E2E (`TestRunDaemon_CatchUpMaterializesAllColdTypesAndIndex`, fa0083f9):

A *complete* chunk whose 10,000 ledgers are all transaction-free produces an **empty txhash `.bin`**. `buildTxhashIndex` then fails with `txhash: cannot build a cold index with zero keys`; that error propagates `runBackfill → startStreaming`, and `superviseStreaming` classifies it as **restartable** (not a fatal sentinel). So the daemon restarts (~5s backoff), re-derives the watermark, re-resolves the same still-unbuilt window, fails again → an **infinite restart loop** (~7s/cycle). It is *not* slow ingest — a successful catch-up of the same chunk runs in ~0.3s.

**Not a production concern:** mainnet/testnet never produce an all-zero-tx 10k-ledger chunk. But a **standalone/quickstart network, or a naive synthetic test backend** (e.g. an all-zero-tx fixture), hits it easily. The new E2E sidesteps it by giving the chunk a sparse few one-tx ledgers so the `.bin` has keys to index.

**Decision: no fix for now** — it can't happen in prod. If ever revisited: either build a degenerate/empty index for a zero-key window, or make the zero-key error a fatal sentinel so `superviseStreaming` surfaces it instead of retrying forever.

cmd/stellar-rpc/internal/fullhistory/daemon.go
```diff
@@ -0,0 +1,333 @@
+package fullhistory
```
chowbao, line 1: daemon files could stay at the root of fullhistory/

We can always move it later and have the file reorg/restructure deferred to #824

cmd/stellar-rpc/main.go
```diff
@@ -79,8 +83,33 @@ func main() {
 		},
 	}
 
+	// TODO(#772): standalone subcommand only for the Phase-1 cold-only daemon. Once
+	// Phase-2 ingestion lands, full-history backfill folds into the general RPC start
+	// command (no separate entrypoint), and this subcommand goes away.
+	var fullHistoryConfigPath string
+	fullHistoryCmd := &cobra.Command{
+		Use:   "full-history",
```
chowbao, line 91: **#815 → Daemon / CLI entrypoint.** Registers the `full-history` subcommand + the `--config` loader and wires SIGINT/SIGTERM → context-cancel into `RunDaemon` — the single seam from the binary into the whole subsystem. Design: [Daemon flow → Startup](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#startup). (TODO #772: this standalone subcommand folds into the general RPC `start` command at the read-path cutover.)

cmd/stellar-rpc/internal/fullhistory/daemon.go
```diff
@@ -0,0 +1,266 @@
...
```
chowbao, line 1: **#815 → Daemon / CLI entrypoint** (boot + supervisor). The ordered boot: `LoadConfig` → `LockRoots` (single-process flock) → open catalog → build boundaries → `validateConfig` → construct the Prometheus registry (`ingest.NewPrometheusSink`, cold tier via `ColdService`) → supervised run loop. `supervise` makes the design's *"startup is the recovery path"* literal — re-derive from durable keys and restart. Design: [Daemon flow → Startup](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#startup) · [Configuration](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#configuration).

cmd/stellar-rpc/internal/fullhistory/config_validate.go
```diff
@@ -0,0 +1,159 @@
+package fullhistory
```
chowbao, line 1: **#815 → Foundations: Config.** `validateConfig` — the structural form checks (`workers > 0`, `max_retries ≥ 0`, chunk-aligned ≥-genesis `earliest_ledger`) and the network-dependent earliest resolution (`now`/numeric require a reachable backend via `networkTip`), plus the first-start pin + restart immutability. Note: this layer adopts #817's *chunks_per_txhash_index is a constant* change, so the design's `cpi ∈ [1, Max]` structural check is dropped (it is now `geometry.ChunksPerTxhashIndex`). Design: [Configuration](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#configuration).

cmd/stellar-rpc/internal/fullhistory/progress.go
```diff
@@ -0,0 +1,135 @@
+package fullhistory
```
chowbao, line 1: **#815 → Orchestration: Derived progress.** `lastCommittedLedger`/watermark recomputed from durable catalog keys (the highest chunk with ledgers + events + (txhash \| covered-by-index) all `frozen`), clamped by `earliest − 1` with the sub-genesis sentinel — **never written** to the catalog. Signed-`int64` arithmetic throughout so the genesis / "nothing complete" sentinels can't wrap. Design: [Daemon flow → Startup](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#startup).

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -0,0 +1,273 @@
+package fullhistory
```
chowbao, line 1: **#815 → Orchestration: startStreaming + the catch-up loop** (named `run`/`catchUp` here). Derives the watermark, then loops: `networkTip` (bounded backoff + readiness reject + stalled-tip guard) → `anchor = max(tip, lastCommitted)` → mid-chunk resume exclusion → `RunBackfill` per pass. First-start-with-no-tip is fatal (`ErrFirstStartNoTip`); the serve handoff is the no-op `ServeReads` placeholder (#772). Phase-1 reads cold-only — the hot source / live loop is Phase 2. Design: [Daemon flow → Startup](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#startup).

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,146 @@
+package backfill
```
chowbao, line 1: **#815 → Orchestration: resolve.** The pure catalog diff over `[rangeStart, rangeEnd]` → a `Plan`: per-chunk `ChunkBuild`s (non-`frozen` kinds only) + the per-window `IndexBuild` rule, with the trailing-window clause (desired `Hi` capped by `rangeEnd`). A loggable/diffable value recomputed every run, so a steady-state restart plans nothing. Design: [Backfill → Postcondition-driven planning](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#postcondition-driven-planning).

cmd/stellar-rpc/internal/fullhistory/backfill/execute.go
```diff
@@ -0,0 +1,247 @@
+package backfill
```
chowbao, line 1: **#815 → Orchestration: executePlan.** One bounded worker semaphore (`Workers`) + `withRetries` (cenkalti/backoff). The index-build stratum waits on its in-coverage chunk builds' success-signalling done-channels **before** acquiring a slot — the no-slot-starvation-at-`workers=1` invariant from Acceptance. `RunBackfill` drives `resolve` + `executePlan`. Design: [Backfill → Execution model](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#execution-model) · [Concurrency model](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md#concurrency-model).

cmd/stellar-rpc/internal/fullhistory/observability/observability.go
```diff
@@ -0,0 +1,268 @@
+package observability
```
chowbao, line 1: **#815 → Daemon registry / operability** (the wiring half of [#808 pt 3](https://github.com/stellar/stellar-rpc/issues/808#issuecomment-4788731096)). The daemon's **control-plane** `Metrics` sink — catch-up / freeze / rebuild / prune phase timings + derived-progress gauges — distinct from the per-type `ingest.MetricSink` the registry wires for the cold tier. Phase-1 wires the metrics backfill actually exercises; `LastCommitted`/`ChunkBoundary` await the Phase-2 live loop.

chowbao (commented):
Annotating each file this layer authors with the Phase-1 deliverable (#815) and the design section (#787, the `streaming-design` docs) it implements, for reviewers.

This PR is **layer 3 of 3** of #815 — the **Orchestration** + **Daemon / CLI entrypoint** strata. The sibling strata land in the stacked parents:
- **Foundations** (geometry, catalog, one-write protocol, locks, retention floor) → #817
- **The primitives** (`processChunk`, `buildTxhashIndex`/`buildThenSweep`, `backfillSource`) → #818

So files like `retention.go` (#817) and `backfill/txindex.go` + `process.go` (#818) show up in the tree but are owned by the parent layers; the inline notes below cover the files this layer adds.

Design docs: [streaming workflow](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/full-history-streaming-workflow.md) · [getTransaction](https://github.com/stellar/stellar-rpc/blob/streaming-design/design-docs/gettransaction-full-history-design.md).

chowbao:
**Suggested reading order** — execution/dependency order, not alphabetical. Per-file design mapping is in the inline review.

`fullhistory` (boot & run loop)
1. `main.go` — CLI entry → `RunDaemon`
2. `daemon.go` — boot sequence + restart supervisor
3. `config_validate.go` — `earliest_ledger` gate (pin + restart immutability)
4. `progress.go` — derived watermark
5. `startup.go` — `run` + catch-up loop

`backfill` (engine)
6. `resolve.go` — catalog diff → `Plan`
7. `execute.go` — bounded worker pool + `RunBackfill`
8. `txindex.go` — one-write index build + sweep

`observability`
9. `observability.go` — control-plane metrics

Parent-layer context (in the tree, not this PR): `retention.go` → #817; `process.go` + most of `txindex.go` → #818.

cmd/stellar-rpc/internal/fullhistory/daemon.go
```diff
@@ -0,0 +1,266 @@
...
+) (Boundaries, error) {
+	b := Boundaries{
+		// TODO(#772): wire the full-history RPC read server; no-op until the cutover.
+		ServeReads: func(context.Context) error { return nil },
+	}
+
+	// Absent a configured backend this is frontfill-only: NetworkTip returns a
+	// not-configured error and Backend stays nil.
+	tip := &notConfiguredTip{}
+	b.NetworkTip = tip
```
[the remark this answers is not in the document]
chowbao: Fixed in 490943cf. `buildProductionBoundaries` now wires the BSB backend via `backfill.NewBSBBackend` when `[backfill.bsb].bucket_path` is set — re-adding the `backendTip` adapter (`NetworkTip` = `Backend.Tip`) and releasing the datastore handle through `Boundaries.Cleanup` at shutdown; with no `bucket_path` it stays frontfill-only. `[backfill.bsb]` also gains `type` (`"GCS"` default / `"S3"`) and `region` (required for S3) so the lake can be either object store, with the batch schema read from the lake manifest (`LoadSchema`).

cmd/stellar-rpc/internal/fullhistory/progress.go
```diff
@@ -0,0 +1,135 @@
...
+	highest := int64(-1)
+	for c, k := range frozen {
+		if !k.ledgers || !k.events {
+			continue
+		}
+		if !k.txhash && !covered(c) {
+			continue
+		}
+		if id := int64(c); id > highest {
+			highest = id
```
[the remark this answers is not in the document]
chowbao: Keeping the watermark as max-by-id — it matches the design's derivation (`max(frozen/hot maxima, earliest − 1)`), and in Phase 1 it isn't exploitable:

`catchUp` recomputes `rangeStart` from the retention **floor** every pass (not from `lastCommitted`), so a post-crash gap below a later-frozen chunk is re-planned by `resolve` on the next run. On the local-progress path (tip unreachable), `catchUp` still runs `RunBackfill` over the full `[floor, rangeEnd]` range — it does **not** skip repair. If the backend can't fill the hole, `RunBackfill` errors and the daemon restarts; `ServeReads` is never reached past an unfillable gap. And in Phase 1 `ServeReads` is the no-op (#772), so nothing serves off `lastCommitted` at all.

The legitimate kernel is Phase 2: once a live read path serves concurrently using `lastCommitted` as a floor, a transient pre-repair gap could be exposed. That's a Phase-2 (live loop / read path) concern, tracked there rather than changing the documented max-by-id derivation in this layer.

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		frozen, hasFrozen, err := cat.FrozenTxHashIndex(w)
+		if err != nil {
+			return Plan{}, err
+		}
+		if hasFrozen {
+			stored := coverageRange{Lo: frozen.Lo, Hi: frozen.Hi}
+			if stored.covers(desired) {
+				continue // steady-state restart, risen floor, or finalized window
```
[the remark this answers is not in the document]
chowbao: Fixed in 33d10cde. The `covers(desired)` branch no longer unconditionally `continue`s: it now checks `windowHasPruning(w)` (any demoted `pruning` coverage or `.bin` input in the window) and, if found, schedules a **sweep-only `IndexBuild` at the frozen coverage**. `buildTxhashIndex` skips the rebuild via its already-frozen skip-check, and `buildThenSweep` finishes the prune — so a crash between `CommitTxHashIndex` and the sweep is repaired on the next `RunBackfill`, including for a terminal window that never rebuilds again.

This is the all-or-nothing rule for the resolver: only `frozen` is trusted, so any `freezing`/`pruning` leftover is driven to completion on restart rather than stranded. Steady state (no leftover) still resolves to an empty plan. Added a resolver-path test (`TestResolve_FrozenCoverageWithPruningLeftoverSchedulesSweep`) since the prior crash-recovery tests only exercised `buildThenSweep` directly.

chowbao:
Addressed the codex P2 ("Sweep abandoned freezing coverages too", pullrequestreview-4594651771) in de43867 — the symmetric sibling of the pruning-leftover fix.

`buildThenSweep` now sweeps `freezing` debris alongside `pruning` coverages (the just-built coverage is `frozen` by then, so it's never swept; `SweepTxHashIndexKey` already handles a freezing partial), and `resolve`'s debris check (`windowHasDebris`) detects orphan `freezing` coverages so a fully-frozen window with leftover freezing debris still gets a sweep-only `IndexBuild`. A `freezing` *.bin* stays a per-chunk rebuild, not a sweep. New resolver-path test added.

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,204 @@
...
+	// Reuse the txHashStates scanner (centralizes the wraparound guard).
+	for cs, serr := range txHashStates(cat, desired.Lo, desired.Hi) {
+		if serr != nil {
+			return IndexBuild{}, false, serr
+		}
+		if cs.State != geometry.StateFrozen {
+			needs[cs.Chunk] = needs[cs.Chunk].Add(geometry.KindTxHash)
+		}
+	}
+	return IndexBuild{Index: w, Lo: desired.Lo, Hi: desired.Hi}, true, nil
```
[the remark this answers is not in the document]
chowbao: Confirmed — this is the known zero-transaction-window stall, deferred and tracked as #826.

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -0,0 +1,273 @@
...
+	if cfg.ServeReads == nil {
+		return errors.New("nil StartConfig.ServeReads")
+	}
+	return nil
+}
+
+// networkTip samples backend.NetworkTip, retrying a transient error on backoff
+// (bounded by maxAttempts) and rejecting a sub-genesis tip as "not ready" so an
+// unready backend never pins a garbage floor. ctx cancellation aborts the wait.
+func networkTip(
```
tamirms, line 243: This hand-rolls a ctx-aware retry-with-sleep loop, but the package already standardizes on `cenkalti/backoff` for its other retries (`execute.go`'s `withRetries`, and `backend.go`'s `waitForCoverage`). Could we converge `networkTip` onto the same primitive? `backoff.WithContext` already provides the ctx-cancellable sleep this loop builds by hand, and the "a sub-genesis tip is permanent, don't retry it" case maps cleanly onto `backoff.Permanent(err)`. That would leave one retry idiom in the package instead of two.

Minor follow-on: once this moves, `supervise` (in `daemon.go`) is the only remaining hand-rolled `time.NewTimer`/`select` sleep. Its three-way clean/fatal/restart classification genuinely doesn't fit `backoff.Retry`, so the loop itself should stay, but the sleep could share a small `sleepCtx(ctx, d) error` helper.

chowbao: Done in e631edc3.

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,208 @@
...
+	if rangeEnd < rangeStart {
+		return Plan{}, nil // no complete chunk exists yet
+	}
+	cat := cfg.Catalog
+	txLayout := cat.TxHashIndexLayout()
+
+	// Per-chunk work, unioned across kinds; one ChunkBuild per chunk.
+	needs := map[chunk.ID]catalog.ArtifactSet{}
+
+	for c := rangeStart; ; c++ {
```
tamirms: There are five chunk-id / window-id loops in the `fullhistory` tree written as an empty-condition `for c := lo; ; c++ { …; if c == hi { break } }`, each with a comment that it "guards chunk.ID wraparound" — this one, plus `resolve.go:201`, `execute.go:182` (`awaitCoverageBuilds`), `txindex.go:216` (`txHashStates`), and `catalog_protocol.go:142`.

The wraparound is real in the general case — `for c := lo; c <= hi; c++` does spin forever when `hi == MaxUint32`, since `c++` wraps past it — but it can't be reached here. `chunk.ID` (and `TxHashIndexID`) are `uint32`, and a chunk id is `(ledgerSeq - 2) / 10000` over a `uint32` ledger sequence, so the largest chunk id that can ever exist is about 429,496. For the loop to wrap, `hi` would have to be `MaxUint32` (~4.29 billion), roughly 10,000× beyond any value the type can produce. So the guard defends an input the protocol and the type jointly make unreachable, and the cost is moving the loop bound out of the header and carrying a comment that implies a hazard that doesn't exist.

Would you be open to replacing the five with the plain counted form `for c := lo; c <= hi; c++`? It's behaviorally identical for every reachable input and reads more directly. If you'd rather keep the shape, it'd at least be worth rewording the comments so they don't read as protecting against a reachable wraparound — a future reader is likely to copy the pattern on the belief that it's load-bearing. Either way it's probably best as all-five-or-none, since the only thing the current form has going for it is its uniformity.

chowbao: Yeah I can change it

chowbao: Done in 2dfc7bb6 — went with all-five → plain counted form.

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,202 @@
...
+		Lo: max(txLayout.FirstChunk(w), rangeStart),
+		Hi: min(txLayout.LastChunk(w), rangeEnd), // capped by range end
+	}
+
+	frozen, hasFrozen, err := cat.FrozenTxHashIndex(w)
+	if err != nil {
+		return IndexBuild{}, false, err
+	}
+	stored := coverageRange{Lo: frozen.Lo, Hi: frozen.Hi}
+	if hasFrozen && stored.covers(desired) {
```
tamirms, line 106: The design assigns debris cleanup to the lifecycle's prune stage (`eligiblePruneOps`, run on every lifecycle run including the startup seed) — that's always been the design. Its implementation lands in a later PR, but the responsibility is already homed there, so this handling in `resolve` duplicates it rather than filling a real gap.

The prune stage sweeps exactly this set: index keys in `freezing`/`pruning`, demoted `pruning` `.bin` inputs, and the redundant-`.bin`-in-a-finalized-window case — the same set `windowHasDebris` detects and the sweep-only `IndexBuild` (plus the `freezing` arm of `buildThenSweep`'s sweep) reclaims here. So this isn't only a Phase-1 gap-filler: once the lifecycle is implemented, one run will call `resolve` in stage 1 (hitting this path) *and* `eligiblePruneOps` in stage 3, sweeping the same leftovers twice — and it puts reclamation inside `resolve`, whose contract everywhere else is just "diff desired-vs-catalog → plan builds."

The codex thread that prompted this ("Re-run sweeps when coverage is already frozen") concluded the debris is "stranded forever," but that only holds when Phase 1 is viewed in isolation. Once the lifecycle exists, the next startup's first run sweeps it via the prune stage, so it's stranded at most until that PR, and in steady state never.

Proposal: drop `windowHasDebris`, the `stored.covers(desired)` debris branch in `resolveWindow`, and the `freezing` arm of `buildThenSweep`'s sweep (keep the `pruning` arm — that eager post-build sweep *is* design-specified), along with their tests, and let the prune stage own debris. That keeps `resolve` faithful to the design's plain `continue` on a covered window.

chowbao: Done in 5f045675 — dropped the resolve-side debris handling and let the lifecycle prune stage own it.

cmd/stellar-rpc/internal/fullhistory/progress.go
```diff
@@ -0,0 +1,132 @@
...
+// "Highest complete chunk" arithmetic runs in int64 (-1 = "nothing complete") to
+// avoid uint32 wraparound on the pre-genesis sentinel; completeThrough is the chokepoint.
+
+// preGenesisLedger is the watermark when nothing below the floor is complete
+// (FirstLedgerSeq-1, "ingest from genesis"); completeThrough returns it for the sentinel.
+const preGenesisLedger uint32 = chunk.FirstLedgerSeq - 1
+
+// completeThrough maps a signed chunk index to its "complete through" last ledger:
+// c < 0 ⇒ preGenesisLedger (no uint32 wraparound); c >= 0 ⇒ chunk.ID(c).LastLedger().
+func completeThrough(c int64) uint32 {
```
tamirms: `completeThrough` and `preGenesisLedger` don't appear in the design doc, and what they encode is already spelled out there. `completeThrough(c)` is just "chunk c's last ledger" with the `c < 0` sentinel folded in — and the doc's `lastCommittedLedger` handles that sentinel with an explicit branch instead of a helper:

```go
base := earliest - 1
cold := highestDurableChunk(cat)   // -1 if none
if cold >= 0 {
    return max(base, chunk.ID(cold).LastLedger()), nil   // == doc's chunkLastLedger(cold)
}
return base, nil
```

`watermarkMidChunk` can likewise use `chunk.ID(c).LastLedger()` directly (the ledger is ≥ genesis there, so `c ≥ 0`). That drops both helpers and makes the code greppable against the doc's `chunkLastLedger` / `lastCommittedLedger`.

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -0,0 +1,284 @@
...
+				return 0, fmt.Errorf("%w: %w", ErrFirstStartNoTip, err)
+			}
+			// Restart with local progress: serve what's below lastCommitted, skip catch-up.
+			tip = lastCommitted
+		}
+
+		// max() guards a lagging bulk tip: the tip alone could regress the floor below
+		// pruning or drop a complete watermark chunk.
+		anchor := max(tip, lastCommitted)
+		rangeStart := chunk.IDFromLedger(effectiveRetentionFloor(anchor, retentionChunks, earliest))
```
tamirms: The design doc names this `retentionFloorChunk` and returns a `ChunkID`; here it's `effectiveRetentionFloor` returning a ledger, so every chunk-side caller round-trips through `chunk.IDFromLedger(...)` to get back to a chunk. Suggest renaming to `retentionFloorChunk` and returning a `ChunkID` to match the doc — then this becomes `rangeStart := retentionFloorChunk(...)`, and only the `metrics.Watermark` call (which wants a ledger) does the `chunkFirstLedger` conversion. The definition lives in `retention.go` (#817), so it spans the stack.

cmd/stellar-rpc/internal/fullhistory/daemon.go
```diff
@@ -0,0 +1,314 @@
...
+	Metrics observability.Metrics
+
+	// IngestSink is the per-type cold-path ingest sink; nil ⇒ a *ingest.PrometheusSink.
+	IngestSink ingest.MetricSink
+}
+
+const defaultRestartBackoff = 5 * time.Second
+
+// Boundaries bundles the external boundaries run and validateConfig inject.
+type Boundaries struct {
```
tamirms: `Boundaries`/`BuildBoundaries` is heavier than the job needs. Three of the four fields are the same object — `NetworkTip` is `backendTip{backend}`, `Cleanup` is the backend's datastore teardown — so this is really "the backfill backend, viewed three ways" plus the unrelated `ServeReads`. It's also a second injection layer on top of `daemonOptions` (a factory, inside the options bag, returning another bag of seams), with two dead params (`paths`, `cat` are ignored in `buildProductionBoundaries`).

Suggest collapsing to the one thing that varies: an injectable `backfill.Backend` (nil ⇒ frontfill) plus `ServeReads`, with `runDaemonWith` deriving `NetworkTip` and `Cleanup`. That removes the struct, the factory, the dead params, and the vague name; tests already inject a `fakeBackend` with `Tip()`, so they lose nothing. Not in the design doc — pure impl-seam cleanup.

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -0,0 +1,284 @@
...
+	logger.WithField("last_committed", lastCommitted).
+		Info("read server returned — cold-only daemon shutting down cleanly")
+	return nil
+}
+
+// catchUp runs the catch-up loop, returning lastCommitted as backfill makes
+// progress. Each pass samples networkTip, anchors on max(tip, lastCommitted),
+// computes [rangeStart, rangeEnd] with the mid-chunk exclusion, and breaks on an
+// empty or non-advancing range.
+func catchUp(ctx context.Context, cfg StartConfig, lastCommitted, earliest uint32) (uint32, error) {
```
tamirms: The doc unified its terminology on "backfill" — it's used throughout (42×) and "catch-up" appears nowhere; it calls this very loop "Backfill" that *"re-runs if the tip advances mid-pass."* `catchUp` reintroduces "catch-up" as a parallel term, and it spreads into the `"catch-up pass"` log lines and the `CatchupPass`/`CatchupProgress` metrics. Since `RunBackfill` is already the single pass, suggest renaming the loop to something backfill-rooted that captures "re-pass until the tip" — e.g. `backfillToTip` — with the logs and metrics following (`BackfillPass`/`BackfillProgress`). Keeps the code greppable against the doc, same as the other naming items.

cmd/stellar-rpc/internal/fullhistory/progress.go
chowbao: Done in dcde887d. Dropped both — `lastCommittedLedger` now derives the watermark inline (signed `max` of the floor term `earliest-1` and the cold term `chunk.ID(cold).LastLedger()`, floored at `FirstLedgerSeq-1`), matching the doc's `lastCommittedLedger`/`chunkLastLedger`.

One adjustment: `watermarkMidChunk` keeps an explicit `c < 0 → false` guard rather than calling `chunk.ID(c).LastLedger()` unconditionally. The fresh-start watermark is the sub-genesis sentinel (`lastCommitted = FirstLedgerSeq-1`, `chunkIDOfLedger → -1`), and it reaches `watermarkMidChunk` both in the catch-up loop (line ~114, "Signed so genesis reads as a boundary") and the `{"genesis sentinel is a boundary", preGenesisLedger, false}` test — so `c` is not always `≥ 0` there. The guard mirrors the explicit-branch style you proposed for `lastCommittedLedger`. `preGenesisLedger` is now a test-only const (the name was only load-bearing in test expectations).

cmd/stellar-rpc/internal/fullhistory/startup.go
chowbao: Done in dcde887d.

cmd/stellar-rpc/internal/fullhistory/daemon.go
chowbao: Done in dcde887d.

cmd/stellar-rpc/internal/fullhistory/startup.go
chowbao: Done in 307e14d5.

cmd/stellar-rpc/internal/fullhistory/progress.go
chowbao: The adjustment seems weird. Let me see if there is way around it

chowbao: Correction — removed the guard in a51f19e1; the unconditional `chunk.ID(c).LastLedger()` you suggested is correct here, and I had it wrong above.

`chunkIDOfLedger` only ever returns `-1` (for any sub-genesis ledger) or a real chunk id `≥ 0`, and the only sub-genesis value `watermarkMidChunk` is ever passed is the fresh-start sentinel `preGenesisLedger` (= `FirstLedgerSeq-1`). For `c = -1`, `chunk.ID(-1).LastLedger()` *wraps* — `(MaxUint32+1)` overflows to `0` — back to exactly `preGenesisLedger`, so `lastCommitted != …` already yields `false` (a boundary) with no special case. (Same uint32 aliasing the old `TestCompleteThrough` documented.) So the guard was redundant, not load-bearing — my earlier "c is not always ≥ 0" reasoning was beside the point. `TestWatermarkMidChunk/genesis_sentinel_is_a_boundary` still passes. Left a comment noting the wrap so it isn't re-added.

cmd/stellar-rpc/internal/fullhistory/observability/observability.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/geometry"
+)
+
+// Metrics is the daemon's control-plane sink — times/counts the daemon's phases
+// (backfill, freeze/rebuild, prune) plus derived progress gauges; distinct from
+// per-data-type ingest.MetricSink. All methods must be safe for concurrent use.
+// LastCommitted and ChunkBoundary await the Phase-2 live-ingestion loop; the rest are wired.
+type Metrics interface {
```
tamirms, line 12: A lot of these are redundant or don't carry information you'd actually use. Suggested cuts:

**Three gauges hold the same number.** `watermark_ledger`, `catchup_backfilled_ledger`, and `last_committed_ledger` are all set from the same `lastCommitted` value. Keep one. I'd keep `last_committed_ledger` — the design doc says "last committed" everywhere (21 times) and "watermark" twice, so it's the name people will recognize.

**Tip, lag, and target aren't needed.** `catchup_target_ledger` is just the tip. `ingestion_lag_ledgers` is tip minus last-committed, which anyone can compute. And captive core already publishes the latest ledger it has, so the daemon doesn't need to report the tip at all — to get lag, subtract our `last_committed_ledger` from captive core's latest-ledger metric in the query. None of the three earns its place.

**Counters that don't add anything.** `catchup_passes_total` and `freeze_indexes_total` are just the `_count` of the `catchup_pass` and `rebuild` phase-duration histograms. `rebuilt_chunks_total` and `rebuild_chunks_per_index` both count chunks folded into index rebuilds, which isn't a number you'd act on. `freeze_chunks_total` counts chunks written, which just tracks `last_committed_ledger` moving. Drop all of these.

**`chunk_boundaries_total` has no code that sets it** — it's reserved for the Phase-2 ingestion loop. Like the #817/#818 reviews: add the metric when the code that fills it lands.

**`cold_tier_bytes` is computed the slow way.** It walks every cold directory and calls `stat` on every file — about 30,000 files at mainnet size, worse as history grows, and much worse if the files are on network storage. Disk usage is something the OS tracks per filesystem: one `statfs` call returns the used bytes, and it's more accurate (real disk blocks, not apparent file sizes). `node_exporter` already reports it per mount.

**Worth keeping:** `last_committed_ledger`, `retention_floor_ledger`, `pruned_ops_total`, and `phase_duration_seconds` by phase. Four instead of fifteen.

cmd/stellar-rpc/internal/fullhistory/daemon.go
```diff
@@ -0,0 +1,291 @@
...
+	return backendTip{backend}
+}
+
+// notConfiguredTip is the NetworkTipBackend placeholder when no bulk backend is
+// configured: every sample returns a clear not-configured error (until #772 wires
+// the real lake tip).
+type notConfiguredTip struct{}
+
+func (notConfiguredTip) NetworkTip(context.Context) (uint32, error) {
+	return 0, errors.New("no bulk backend configured ([backfill.bsb].bucket_path empty); " +
```
tamirms: Small one: this error points operators at `[backfill.bsb].bucket_path`, but that key doesn't exist — the bucket path lives in `[backfill.datastore.params]`, and the thing that actually triggers frontfill-only mode is an empty `[backfill.datastore].type` (the check just above at line 236). An operator who wants to configure a backend and reads this gets sent to the wrong place. Suggest pointing it at `[backfill.datastore].type`.

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -0,0 +1,288 @@
...
+func withinOneChunkOfTip(tip, lastCommitted uint32) bool {
+	return int64(tip)-int64(lastCommitted) < int64(chunk.LedgersPerChunk)
+}
+
+// watermarkMidChunk reports whether lastCommitted falls strictly inside a chunk.
+// The only sub-genesis value it sees is the fresh-start sentinel preGenesisLedger,
+// where chunkIDOfLedger yields -1 and chunk.ID(-1).LastLedger() wraps (MaxUint32+1
+// overflows to 0) back to exactly preGenesisLedger — so the comparison reports a
+// boundary (false) without a special case.
+func watermarkMidChunk(lastCommitted uint32) bool {
```
tamirms: Same naming point as the metrics one: this function (plus a few comments, the log "watermark derived", and the error "startup derive watermark") call the value "watermark," but the variable is `lastCommitted`, the function is `lastCommittedLedger`, and the log field is `last_committed`. The design doc uses "last committed" throughout and "watermark" only twice in passing. Could we settle on "last committed" everywhere — `watermarkMidChunk` -> `lastCommittedMidChunk`, and the stray "watermark" mentions to match? One name for one thing.

cmd/stellar-rpc/internal/fullhistory/startup_test.go
```diff
@@ -0,0 +1,455 @@
...
+	assert.Equal(t, chunk.ID(0), passes[0][0], "rangeStart is chunk 0 (genesis floor)")
+	assert.Equal(t, watermark, last, "watermark does not regress below where pruning advanced")
+}
+
+// ---------------------------------------------------------------------------
+// run — the backfill + serve flow.
+// ---------------------------------------------------------------------------
+
+// A young-network first start does no backfill then serves reads once.
+func TestRun_FirstStartCatchUpThenServe(t *testing.T) {
```
tamirms: Leftover from the catch-up -> backfill rename: this test (and `TestRunDaemon_CatchUpMaterializesAllColdTypesAndIndex` in daemon_test.go, plus the "catches up" prose around it) still says "CatchUp." Worth renaming to "Backfill" to match.

cmd/stellar-rpc/internal/fullhistory/observability/observability.go
chowbao: Done in ac64eacf — cut to the four you flagged (`last_committed_ledger`, `retention_floor_ledger`, `pruned_ops_total`, `phase_duration_seconds` by phase).

cmd/stellar-rpc/internal/fullhistory/daemon.go
chowbao: Fixed in 5d91d07f. The error now points at `[backfill.datastore].type` (the empty-type check just above that actually selects frontfill-only mode), instead of the non-existent `[backfill.bsb].bucket_path`.

cmd/stellar-rpc/internal/fullhistory/startup.go
chowbao: Done in 5d91d07f — settled on "last committed" everywhere.

cmd/stellar-rpc/internal/fullhistory/startup_test.go
chowbao: Done in 5d91d07f. Renamed `TestRun_FirstStartCatchUpThenServe` → `...BackfillThenServe` and `TestRunDaemon_CatchUpMaterializesAllColdTypesAndIndex` → `TestRunDaemon_BackfillMaterializesAllColdTypesAndIndex`, plus the "catches up" prose in those two tests' comments. (Left the generic English "catches up" in resolve_test.go / backend.go — those aren't the catch-up term-of-art.)

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
```diff
@@ -0,0 +1,165 @@
...
+			if state != geometry.StateFrozen {
+				needs[c] = needs[c].Add(kind)
+			}
+		}
+	}
+
+	// The txhash kind: one rule per overlapping window (resolveWindow runs the diff).
+	var builds []IndexBuild
+	for _, w := range indexesOverlapping(txLayout, rangeStart, rangeEnd) {
+		build, ok, err := resolveWindow(cat, txLayout, w, rangeStart, rangeEnd, needs)
```
tamirms: nit: this should probably be renamed to resolveTxHashIndex

tamirms (approved):
left one nit about renaming resolveWindow

but everything looks great! feel free to merge 🎉

cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
chowbao: Done in a20ebf5f — renamed `resolveWindow` → `resolveTxHashIndex` (def + call + comment).

Issue #706, opened 2026-04-29
O3.1.1 - Create an Automated Release Eval Path for RPCv1


Shaptic:
## Summary

Create a pre-release performance evaluation path for Stellar RPC that is connected by one GitHub Action runnable on `release/**` branches and by manual dispatch. The evaluation should catch version-to-version regressions across endpoint benchmarks, ingestion load, and realistic request load.

## Release decision signals

- Go benchmark results for each JSON-RPC endpoint complete in minutes and compare against the selected baseline.
- Backfill ingestion from public pubnet data-lake ledgers stays below TBD lag/duration/resource thresholds.
- Synthetic ingestion from Core-generated ledger files stays below TBD lag/duration/resource thresholds across selected profiles.
- `stellar-rpc-blaster` request runs meet TBD Endpoint RPS, latency, error-rate, and timeout criteria.
- A single release report summarizes pass/warn/fail status and links raw artifacts.

## Acceptance criteria

- A release manager can run one workflow against a `release/**` branch and get a consolidated evaluation report.
- Thresholds that are not known yet are explicit TBD configuration values, not hidden constants.

PR #830, opened 2026-06-30
Add RPC database backfill performance evaluation leg

### LINKED ISSUE #709: Release eval: job to measure data-lake backfill ingestion perf
Parent: #706 
Depends on: #707

## Summary

Add a release-evaluation job that measures RPC database load while backfilling from public pubnet ledger data without relying on live Core nodes.

## Scope

- Start `stellar-rpc` with `--backfill`, `--serve-ledgers-from-datastore`, pubnet network settings, and a public data-lake datastore configuration.
- Use an isolated SQLite DB path on the runner and collect DB size, runtime, ledgers/sec, and final health/latest-ledger state.
- Backfill with a hardcoded `history-retention-window`.
- Validate that no Stellar Core node is required for the backfill ingestion path.

## Acceptance criteria

- The job can ingest a configured pubnet ledger window from public data-lake files and exit deterministically.
- The job fails when metrics exceed TBD thresholds.
- Metrics and logs are uploaded for final reporting.

cjonas9:
### What

Adds a new backfill leg atop the coordinator architecture that spawns an EC2 instance and runs backfill. Metrics for the performance of this backfill are logged through the GHA coordinator alongside metrics for the apply-load ingestion load test.

The core components of this addition are organized canonically in-line with the precendent of the apply-load ingestion test. In particular, this is:
- `perf-eval/backfill-test/`: contains EC2 box script + a runner whose instantiate renders a default pubnet config with a datalake (`testdata/backfill-pubnet.toml.tmpl`), starts the daemon's backfill, detects the completion sentinel, and publishes the elapsed/ledger-count result to S3
- `load-test-coordinator.yml`: contains a roster entry in the coordinator so the leg `"Backfill ingestion"` fans out through the same matrix as apply-load

### Why

See epic #706. This is one of ~4 jobs that will be completed as part of the automated RPC release evaluation framework, and it is necessary for the completion of the leg that evaluates endpoint performance using `stellar-rpc-blaster`.

### Known limitations

Currently, this test measures a backfill of one day of ledgers (=17280 ledgers). This is because Github Actions has a hard-cap of 6 hours per job, and a one-week backfill exceeds this time constraint. Backfill optimizations should be made to allow backfill to execute a one-week backfill in a shorter amount of time.


cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/testdata/backfill-pubnet.toml.tmpl
cjonas9, line 1: This is vaguely inspired by how the integration tests do something similar, I'm not sure if there's a preferred method for this but other options seemed less viable because `datastore_config` is TOML-only.

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -96,6 +105,242 @@ func openSQLiteDB(dbFilePath string) (*db.Session, error) {
...
+		"    application_order INTEGER NOT NULL\n" +
+		")"
+
+	// bulkTransactionsDDL is the indexless bulk-load twin. Same name and
+	// columns as the canonical table, no hash key, so inserts append.
+	bulkTransactionsDDL = "CREATE TABLE transactions (\n" +
+		"    hash BLOB, -- 32-byte binary; key restored by FinalizeBulkLoad\n" +
+		"    ledger_sequence INTEGER NOT NULL,\n" +
+		"    application_order INTEGER NOT NULL\n" +
+		")"
```
Shaptic: Can we get away with just assigning the primary key (via `ALTER TABLE`) after the fact rather than doing an entire table recreation and copy?

cjonas9: I don't think it can literally do that, some brief checking suggests to me that `ALTER TABLE` can't add a PK or any constraint after creation. From [the docs](https://www.sqlite.org/lang_altertable.html):

> SQLite supports a limited subset of ALTER TABLE. The ALTER TABLE command in SQLite allows these alterations of an existing table: it can be renamed; a column can be renamed; a column can be added to it; or a column can be dropped from it.

I do think there's a deliberate trade-off between something of this spirit and something more in line with your suggestion: `CREATE UNIQUE INDEX... ON transactions(hash)` after the load. Doing this on the twin would produce an identical physical structure and query plans. But this doesn't preserve schema shape as byte-identically as rebuilding it with the DDL (`transactionsNeedRestore` actually checks for whether we need this last step by inspecting byte-equivalence there). Just building an index on top of this breaks that shape and ships another table shape here. I'm unsure of whether there's a functional difference that demands we treat the schema here as precious, but suggesting this as a "free win optimization" when it ships a second permanent table shape didn't seem right.

But, one bit of food for thought: filling now takes <2 hours. Finalizing/migrating the twin takes 1 hour for a complete backfill of 3 hours. If we drop this part of that latter step, we _do_ get no-slope-increase-50s/chunk-backfill, and we get it in even less time.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return bail("invalid BACKFILL_DEADLINE %q: %v", deadline, err)
+	}
+	logger.Infof("starting backfill (retention=%s, deadline=%s)", retention, deadline)
+	elapsed, lo, hi, err := runBackfill(ctx, dl, binaryPath, cfgPath)
+	if err != nil {
+		return bail("%v", err)
+	}
+	ingested := hi - lo + 1
+	logger.Infof("backfill complete: %d ledgers [%d -> %d] in %s", ingested, lo, hi, elapsed.Round(time.Second))
```
Shaptic, line 109: Should we validate `ingested` as a value relative to `retention` here?

.github/workflows/load-test-coordinator.yml
```diff
@@ -11,7 +11,9 @@ defaults:
 
 on:
   push:
-    branches: [release/**]
+    # temporary scaffolding: before merge to main, replace with line
+    # branches: [release/**]
+    branches: [release/**, backfill-test]
```
tamirms: This `backfill-test` entry, along with the commented scaffolding on the two lines above it, looks like it needs to come out before this merges. Left in, every push to a `backfill-test` branch triggers the full coordinator and spins up real EC2 boxes, which diverges from the release-only intent. Could we restore `branches: [release/**]` here? If it is useful to keep exercising the branch until the last moment, one option is to leave it and make its removal the final commit so it never lands on main.

.github/workflows/load-test-coordinator.yml
```diff
@@ -53,11 +55,11 @@ jobs:
         run: |
           SHA="$(git rev-parse HEAD)"
           PR="$(gh pr list --repo "${{ github.repository }}" --state open \
-                  --base main --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
+                  --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
```
tamirms: Dropping `--base main` broadens this lookup to match any open PR with this head regardless of base, and `.[0]` then picks arbitrarily if a head ever has PRs to more than one base. Was that intentional, or is it part of the same temporary testing setup as the branch trigger above? If it was only to make the lookup resolve on this branch, it is probably worth restoring `--base main` alongside the trigger change.

.github/workflows/load-test-coordinator.yml
```diff
@@ -74,6 +76,7 @@ jobs:
       target_ref: ${{ needs.plan.outputs.target_sha }}
       run_label: ${{ matrix.leg.run_label }}
       leg_script: ${{ matrix.leg.script }}
+      budget_minutes: ${{ matrix.leg.budget }}
```
tamirms, line 89: Worth noting for whoever owns the AWS role: the backfill leg's `budget: 345` derives a 6h OIDC session in `ec2-leg.yml` via `(B + 15) * 60` seconds, up from the 4h that the apply-load leg's `225` requires. That is within the AWS ceiling and your green run confirms the role already allows it, so no change is needed here. I mention it only because `MaxSessionDuration` on the role is now the binding constraint, and any future budget above 345 would push the derived session past 6h and would need the role raised to match.

.github/workflows/load-test-coordinator.yml
```diff
@@ -53,11 +55,11 @@ jobs:
         run: |
           SHA="$(git rev-parse HEAD)"
           PR="$(gh pr list --repo "${{ github.repository }}" --state open \
-                  --base main --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
+                  --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
           {
             echo "target_sha=$SHA"
             echo "pr_number=$PR"
-            echo 'legs=[{"label":"Apply-load ingestion","run_label":"apply-load","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh"}]'
+            echo 'legs=[{"label":"Apply-load ingestion","run_label":"apply-load","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh","budget":225},{"label":"Backfill ingestion","run_label":"backfill","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/run-backfill.sh","budget":345}]'
```
tamirms: Very minor and pre-existing: this legs array is getting hard to read now that it is two nested objects on one physical line, and it will keep growing as the other planned legs land. At some point a checked-in `perf-eval-legs.json`, or a heredoc piped through `jq -c`, that the plan job reads would make adding a leg a cleaner diff. Fine to leave for this PR.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
```diff
@@ -76,7 +76,10 @@ func Gather(ctx context.Context) error {
 			}
 			return appendOutputs(githubOutput,
 				"found=true",
-				fmt.Sprintf("passed=%t", res.Verdict == "ok"))
+				fmt.Sprintf("passed=%t", res.Verdict == "ok"),
+				"verdict="+res.Verdict,
+				"result_key="+resultKey,
+				"bucket="+bucket)
```
tamirms: As far as I can tell nothing consumes these three new outputs. `ec2-leg.yml` reads only `found` and `passed` from this step, it declares no `workflow_call` outputs so they do not reach the coordinator, and the report job reconstructs the bucket and key deterministically from `PLANNED_LEGS` instead. The pre-existing comment in `ec2-leg.yml` that says the coordinator reads the result location is not accurate today either. Could we either drop `verdict`, `result_key`, and `bucket` here, and the matching pair in `writeTimeoutComment`, or wire them through as real `workflow_call` outputs that the coordinator actually consumes? As written it is two sources of truth for the same fact with only one of them connected.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
```diff
@@ -159,5 +162,8 @@ func writeTimeoutComment(
 	if err := os.WriteFile("/tmp/timeout-comment.md", []byte(b.String()), 0o644); err != nil {
 		return err
 	}
-	return appendOutputs(githubOutput, "found=false")
+	return appendOutputs(githubOutput,
+		"found=false",
+		"result_key="+resultKey,
+		"bucket="+bucket)
```
tamirms: Same as the note on the success path: these `result_key` and `bucket` outputs are not read anywhere, so they can come out together with the others, or be wired through together if we decide the gatherer should be authoritative for the result location.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/run-backfill.sh
```diff
@@ -0,0 +1,9 @@
+# Backfill ingestion leg. Concatenated after bootstrap-common.sh in the rendered
+# EC2 user-data, so it relies on that file's env, helpers, bootstrap_box, and
+# run_leg. It hands off to `runner instantiate`, which builds stellar-rpc and
+# times a `--backfill` run against the pubnet datastore. The other half,
```
tamirms: Small doc nit: the leg triggers backfill through the rendered TOML config (`BACKFILL = true`), not a `--backfill` CLI flag, so this reads a little misleadingly when debugging a failure. Maybe reword to something like "times a backfill run (config `BACKFILL = true`) against the pubnet datastore."

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/main.go
```diff
@@ -0,0 +1,10 @@
+// Command runner drives the backfill ingestion leg. The shared harness owns the
+// dispatch: `runner gather` polls S3 for the result object on the GHA runner,
```
tamirms: This comment describes a `runner gather` / `runner instantiate` dispatch, but `main` unconditionally runs `instantiate`, and `gather` is a separate command under `perf-eval/gather` rather than a subcommand of this binary. It might read more clearly if it just said what this binary does, similar to the one-line comment on the apply-load runner's `main.go`.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness"
+)
+
+// legDir is this leg's path under the repo root, where its config template is
+// checked in (the runner runs with cwd = repo root).
+const legDir = "cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test"
+
+const (
+	corePath   = "/usr/local/bin/stellar-core" // fetched from S3
+	binaryPath = "/data/stellar-rpc-bin"       // built here (the repo checkout is /data/stellar-rpc)
```
tamirms: `binaryPath` is pinned to `/data/...` while the DB and storage paths are derived from `WORK_DIR`, which is already parameterized. If `WORK_DIR` is ever set to anything other than `/data`, the build output and the config paths diverge and the leg cannot find the binary. Low risk since `WORK_DIR` is not overridden anywhere today, but deriving `binaryPath` from `workDir` too would keep them from drifting.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+		case "CAPTIVE_CORE_STORAGE_PATH":
+			return filepath.Join(workDir, "core-storage")
+		case "DB_PATH":
+			return filepath.Join(workDir, "backfill.sqlite")
+		case "STELLAR_CORE_BINARY_PATH":
+			return corePath
+		case "HISTORY_RETENTION_WINDOW":
+			return retention
+		default:
+			return "$" + in // leave unknown placeholders intact
```
tamirms: Two small things here. The default case rewrites an unknown `${FOO}` to `$FOO` rather than leaving it intact as the comment says, so returning `"${" + in + "}"` would match the comment and round-trip faithfully. More broadly, since every placeholder is a simple key lookup, a `map[string]string` plus this fallback would read more directly than the switch. Not important, since no current placeholder hits the default, but it removes a latent surprise if a new `${...}` is added later without a matching case.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+	cmd := exec.CommandContext(runCtx, binary, "--config-path", cfgPath)
+	// hide this box's IMDS creds as the public datalake 403s signed requests
+	cmd.Env = append(os.Environ(), "AWS_EC2_METADATA_DISABLED=true")
+	pr, pw, err := os.Pipe()
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	cmd.Stdout, cmd.Stderr = pw, pw
+
+	start := time.Now()
```
tamirms, line 168: One measurement nuance to be aware of: `start` is set here, before the daemon boots, so the reported wall-clock and ledgers/sec include daemon startup and captive-core construction, not just the backfill. At the 40-minute scale that overhead is negligible, so this is fine as is. Flagging it only in case you ever want the number to be pure backfill time, in which case timing the fill directly, per the in-process note below, would give it to you for free.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+	cfgPath := filepath.Join(workDir, "backfill-rpc.toml")
+	if err := os.WriteFile(cfgPath, []byte(body), 0o644); err != nil {
+		return "", err
+	}
+	return cfgPath, nil
+}
+
+// runBackfill launches the daemon and streams its output (teeing to the box log)
+// until the backfill-complete line fires, recording the wall-clock
+func runBackfill(ctx context.Context, deadline time.Duration, binary, cfgPath string) (time.Duration, int, int, error) {
```
tamirms, line 155: This whole function, the pipe, the scanner, the completion regex, the `cancel()` that kills the daemon before it drifts into live ingestion, and the reaping `Wait()`, exists because we are driving the full daemon binary, which does not stop after backfilling. That also makes success detection depend on the exact INFO-level log string and `LOG_LEVEL = info`. If either changes, the leg does not fail loudly, it runs to the 4h deadline and reports a timeout.

Have you considered measuring backfill in-process, the way the apply-load leg does, rather than shelling the daemon? `RunBackfill` is synchronous: `daemon.MustNew` runs it to completion during construction and only then starts serving, so in-process "backfill done" is simply the call returning, success is the returned error, and the ledger range is a return value or a one-line DB query. `TestIngestSyntheticLedgers` already boots an in-process daemon through `infrastructure.NewTest` and reads its timing from a daemon callback rather than logs, and `TestConfig` already carries `DatastoreConfigFunc` and `HistoryRetentionWindow`, which is most of what a backfill test needs.

Going that way would delete this function and `backfillDoneRe`, and likely the config template and `renderConfig` too, since the datastore config becomes a Go struct rather than rendered TOML, and `run-backfill.sh` would just invoke `go test -run TestBackfillPerf` like the apply-load leg. It needs no new daemon subcommand or flag, at most a small internal completion callback mirroring the existing `OnLedgerIngested` hook if you want backfill time isolated from daemon boot. Not necessarily for this PR, but it may be the cleaner end state, and it removes more than it adds.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/testdata/backfill-pubnet.toml.tmpl
```diff
@@ -0,0 +1,36 @@
...
+CAPTIVE_CORE_STORAGE_PATH = "${CAPTIVE_CORE_STORAGE_PATH}"
+CHECKPOINT_FREQUENCY = 64
+
+DB_PATH = "${DB_PATH}"
+
+# Backfill the retention window from the datastore on startup, then stop (the
+# leg kills the daemon at the completion log line).
+BACKFILL = true
+SERVE_LEDGERS_FROM_DATASTORE = true
+HISTORY_RETENTION_WINDOW = ${HISTORY_RETENTION_WINDOW}
```
tamirms, line 23: `HISTORY_RETENTION_WINDOW` is substituted unquoted as a TOML integer, so a non-numeric `HISTORY_RETENTION_WINDOW` env value would render invalid TOML and fail the daemon at parse time rather than with a clear error. It is operator-controlled with a numeric default, so this is very minor, but validating that `retention` parses as an integer in `instantiate.go` before rendering would fail faster and clearer if it is ever mis-set.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/testdata/backfill-pubnet.toml.tmpl
```diff
@@ -0,0 +1,36 @@
...
+INGESTION_TIMEOUT = "50m0s"
```
tamirms, line 24: For context, `INGESTION_TIMEOUT` here bounds the pre and post gap-verification queries (`verifyDbGapless`), not the fill itself, so 50m is comfortable for the 1-day default and will not cut a backfill short. Worth keeping in mind that if the retention window is ever raised toward the full week, the gap check over a much larger DB could approach this ceiling, so this value may need to scale with the window.

tamirms (commented):
Thorough pass over the backfill leg. It mirrors the apply-load precedent cleanly, the new Go code vets, and the branch's own end-to-end run is green, so nothing here is a functional blocker on the happy path. The one must-fix before merge is the temporary `backfill-test` push trigger. The rest are a few worth-raising items (unused gather outputs, the `--base main` change), some nits, and two design and measurement observations, including whether the backfill could be measured in-process the way the apply-load leg already is. Details inline.

.github/workflows/load-test-coordinator.yml
cjonas9: Yes, this is absolutely the "final commit" edit just so we get verification the review-edits don't mess anything up. Always good to have this flagged, though

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -0,0 +1,207 @@
...
+		return 0, 0, 0, fmt.Errorf("starting daemon: %w", err)
+	}
+	pw.Close() // the child holds the write end and we read until it dies
+	defer pr.Close()
+
+	var elapsed time.Duration
+	var lo, hi int
+	scanner := bufio.NewScanner(pr)
+	scanner.Buffer(make([]byte, 64*1024), 1024*1024)
+	for scanner.Scan() {
```
Shaptic, line 181: Make sure you check the `scanner.Err()` error code in this loop!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
Shaptic, line 1: Claude told me

> gather.go re-adds outputs that main deliberately removed — looks like a mis-resolved merge. The branch's only harness change restores verdict, result_key, and bucket step outputs (success and timeout paths). Main removed exactly these in the "simplify s3 to be the sole reporting channel" cleanup, and nothing consumes them — ec2-leg.yml on this branch is byte-identical to main and reads only found/passed. The branch forked before that simplification, so the merge likely kept the wrong side. Revert gather.go to main's version; otherwise the next reader has to figure out why the "sole reporting channel" design has vestigial outputs again.

Only you can know if this is true or not but I wanna throw it out

Shaptic (commented):
Coupla minor notes but otherwise this looks really good! Love how clean it is to add a new job leg.

.github/workflows/load-test-coordinator.yml
cjonas9: This was intentional. I dropped this a bit ago because I had a PR for this epic that didn't have `--base main`, which led to a silent failure where the results comment didn't make it to the PR. Because that feels like a reasonable use case, I was planning on keeping the `--base main` removal. 
I saw this as needlessly restrictive of what the happy path is, and I'm not sure if I think that PRs should need to have `--base main` for a results comment to land on them. The cost is the one you pointed out, that a head with open PRs to two bases resolves to .[0] arbitrarily. But GitHub forbids duplicate head+base pairs, so that state takes deliberately opening a second PR from the same branch, and the worst case is the comment landing on the newest open PR for that branch.

I find this issue (of resolving the correct PR) a bit more complicated than I thought it'd be, so I'm pretty open to ideas here. We could harden it by filtering on `authorAssociation` or make it deterministic by preferring main-based PRs if there are any and falling back to whatever else. But those things increase the length of this query and I'm unsure of whether they're worth it or not.
Thoughts?

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
cjonas9: Great catch, will remove, these are vestigial from back when S3 wasn't the sole reporting channel

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/run-backfill.sh
cjonas9: This is true. Both are acceptable usage of RPC, in any case, and in favor of reducing overall verbosity, I'm just going to drop the tick marks and "--". We can just say it "times a backfill run," the mechanism by which one chooses to invoke backfill isn't important.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
cjonas9: Tamir caught the same thing, this is totally true, good catch!!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
cjonas9: I totally forgot to have an error check of that altogether, but because scanner.Scan() is the bool we check on (ok or not), it's pretty neat that you can get by with only one error check after the loop instead of during (I think). will add that!

.github/workflows/load-test-coordinator.yml
tamirms: Totally fair, and the silent miss you hit is a real argument against requiring base=main, so I am not pushing to reinstate it.

Stepping back though, I think the base-filter question is a symptom of something one level up. The coordinator fires on a commit-shaped event, a push to `release/**`, but reports to a PR-shaped target, the sticky comment, and there is no reliable function from "a commit that got pushed" to "the one PR this belongs to." That relation is genuinely zero-to-many and changes over time, so every version of this query, with or without `--base main`, with or without a sort, is glue bridging that gap. Your silent miss and the arbitrary `.[0]` are both symptoms of inferring a PR from a head rather than problems with the particular filter. We can keep refining the inference, but the inference itself is what keeps generating these edge cases.

So the question I would actually ask is what the results are anchored to, and there are two clean answers:

1. If the results are about the commit or release, which the push-on-`release/**` trigger implies, we could anchor them to the SHA and stop resolving a PR at all. The report already goes to `$GITHUB_STEP_SUMMARY` unconditionally, so the always-available channel exists today. If we want a persistent artifact, a commit comment on `target_sha` (`POST /repos/{repo}/commits/{sha}/comments`) is unambiguous, always resolvable, and GitHub surfaces it on any PR that contains that commit. No head lookup, no base filter, and the whole silent-miss class disappears because the SHA is always in hand. The tradeoff is that the single evolving comment with prior runs folded in becomes per-commit results instead, which for a release eval is arguably more honest anyway, since we evaluate specific commits rather than a branch in the abstract.

2. If the PR conversation is really where we want it, we could make the PR explicit instead of inferred. Resolving by the exact tested commit via `GET /repos/{repo}/commits/{sha}/pulls` keys on what we actually built rather than a branch label, so it is strictly more correct than `--head`, though it can still return more than one. A `workflow_dispatch` `pr_number` input would also hand it over with zero ambiguity for the manual path.

On the specific options you floated: I would leave `authorAssociation` out. It targets fork and external-contributor PRs, which is a different axis from the multi-base case, and since we only trigger on pushes to in-repo release branches the head is always an internal branch anyway. The main-preference fallback is a fine stopgap if we want to ship now:

```
gh pr list --repo ... --state open --head "$REF" --json number,baseRefName \
  --jq 'sort_by(.baseRefName != "main") | .[0].number // ""'
```

In the normal one-PR case it is a no-op, and it only matters when there are several, so it is a strict improvement over `.[0]`. I would just pair it with making the empty case loud, for example `::warning::no open PR for $REF; skipping results comment` when the lookup returns nothing, since the failure that started this was really the miss being silent rather than the base filter itself.

But I would treat that jq as the interim and the anchor question as the actual fix. As long as the target is "some PR we guessed from the head," I suspect we will be back tuning this query again.

cjonas9: I agree that this has some fragility inherent to it. I'm unsure of whether or not either solution is an improvement. From the beginning, our team has agreed on results being posted in a PR comment as a core piece of the high-level design, so I'd rather not change that now without some discussion with the team (as someone who's been testing using the infrastructure I'm building in these PRs, I can say with certainty that the UX provided by the comment is a piece of operator-friendliness I'd miss sincerely).

Regarding the second suggestion, I put in a new optional pr_number dispatch input, empty/resolving the current way by default. If someone is having trouble getting it to resolve to the correct PR, this can be provided. It's a lightweight change, so I think that's totally reasonable.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
cjonas9: I think this change could be for the better later on! 
I do like how there's none of the sort of inter-ledger timing math we had to deal with in `apply-load` perf eval work, and realistically, I think the broad-strokes wall-clock measuring I have now is appropriate for a test of how long a backfill takes -- smaller timing drifts aren't a very big deal, especially when you consider that the current backfill timing drift that happened without this infra is from 3 hours (in January) to ~8 hours now for a one-week backfill. Definitely agree that it's wise to consider that this does depend on the log lines and keep that in mind should this ever break.
I think that weighing the additional complexity and plumbing involved with this change makes me feel that it is strictly better than what's here, but has a cost of being extra work that's a little out of scope here. But certainly worth filing as a follow-up.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
cjonas9: See comment reply below, thanks!

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
tamirms: Following up now that this is `"${" + in + "}"`: I think I steered this the wrong way in my original note. I suggested making the fallback round-trip faithfully, but the better question is why there is a fallback at all.

I checked, and the daemon's config loader does a plain TOML parse with no `${VAR}` interpolation on values, so a placeholder left in the rendered file is never resolved downstream. It is parsed as the literal string, for example `DB_PATH = "${DB_PATH}"`, which is simply a broken config. Since the template and this mapping are authored together and the mapping covers exactly the template's placeholders, an unresolved placeholder can only mean a new `${...}` was added without a matching case, or a typo. Both are bugs, and leaving them intact turns them into a daemon-startup failure partway into an EC2 run, far from the cause, instead of a clear error at render time on the box.

So I would make the default an error rather than a passthrough. `os.Expand`'s mapping cannot return an error directly, but collecting the unknowns and checking after the expansion works:

```go
var unknown []string
mapping := func(in string) string {
    switch in {
    case "CAPTIVE_CORE_CONFIG_PATH":  return coreCfg
    // ... the other known cases ...
    case "HISTORY_RETENTION_WINDOW":  return retention
    }
    unknown = append(unknown, in)
    return ""
}
body := os.Expand(string(tmpl), mapping)
if len(unknown) > 0 {
    return "", fmt.Errorf("config template has unhandled placeholders: %s", strings.Join(unknown, ", "))
}
```

Minor and not worth blocking on, but it converts a whole class of silent misconfiguration into a loud, self-describing failure right where it happens.

tamirms (approved):

PR #843, opened 2026-07-07
docs(full-history): query routing design (#770)

urvisavla:
## What this delivers

The query-serving design for #770: `design-docs/query-routing-design.md`, the read-side counterpart to the streaming workflow. It covers how each query resolves the chunks of its range to the store serving them (hot database or sealed cold files), and how routing stays correct while ingestion and the lifecycle concurrently add, replace, and delete those stores.

The design in brief:

- **One catalog snapshot per query.** Every request admits by loading the `latest` watermark, the hot handle set, and a RocksDB snapshot of the catalog, then routes from that state for its whole lifetime. Routing metadata has no second copy: queries read the same catalog the lifecycle writes, pinned at admission (H1).
- **Deferred deletion, barrier-backed safety.** Every request runs under an enforced deadline; the lifecycle run demotes resources in the catalog, waits out a grace period T derived from that deadline at the end of the run, and deletes with bounded retries, so queries admitted earlier finish first. Memory safety comes from close discipline rather than timing: the hot database wrapper drains in-flight operations on close, and every cold file (packfiles, events indexes, window `.idx`) is opened and closed by the request using it — no reader tracking, no reference counts (H2, H4).
- **The `latest` watermark advances last.** It moves only after the per-ledger commit and every in-memory serving apply, so a query never observes a partially ingested ledger or advertises a `latestLedger` it cannot serve (H3).
- **Per-method routing** for getLedgers, getTransactions, getTransaction, and getEvents over the existing store packages, with the bounds rule (validate the leading edge, clamp the trailing edge) and cursor requirements.

It asks the streaming workflow for a small set of changes, listed in "Changes to the streaming workflow": router ownership of the hot `hotchunk.DB` handles (the boundary no longer closes the completed chunk; freeze reads through the shared handle supplied to the backfill's `resolveHotSource`), catalog snapshot support in the rocksdb wrapper, and destruction deferred to the end of the lifecycle run. The catalog protocol and storage formats are unchanged.

Open questions flagged for review: serving-floor precision (chunk-aligned vs ledger-precise), the pre-retention datastore fallback (#772), getTransaction probe parallelism, and snapshot release hygiene.

## Notes for review

- Docs only; no code changes. Easiest to review as rendered markdown.
- Per #770, this is the design-review gate: implementation issues get carved from the doc once it is approved.

### DIFF
--- a/design-docs/query-routing-design.md
+++ b/design-docs/query-routing-design.md
@@ -370,7 +370,7 @@ func (v *View) resolve(c chunk.ID, k Kind) (Store, error) {
 
 The resolution order is deterministic. When both hot and cold copies are available, routing selects the cold artifact.
 
-If a chunk has no serving home, routing returns `ErrUnavailable`. This can occur during startup recovery when a required artifact has not yet reached a serving state.
+If a chunk has no serving home, routing returns `ErrUnavailable`.
 
 ### Chunk traversal
 
@@ -396,10 +396,10 @@ tx-hash   ───────────── window .idx ──────
 
 Example request:
 
-- `startLedger = 65,439,500`
-- `limit = 1,000`
+- `startLedger = 65,439,995`
+- `limit = 10`
 
-The request spans chunks 6543 and 6544. Chunk 6543 resolves to the cold ledger store and returns the rest of that chunk. Chunk 6544 resolves to the live hot store and returns ledgers up to the remaining limit. The response is the concatenation of both streams.
+The request spans chunks 6543 and 6544. Chunk 6543 resolves to the cold ledger store and returns the last five ledgers of that chunk. Chunk 6544 resolves to the live hot store and returns the remaining five. The response is the concatenation of both streams.
 
 ### `getTransactions`
 
--- a/design-docs/query-routing-design.md
+++ b/design-docs/query-routing-design.md
@@ -8,7 +8,7 @@ Every query is admitted against a single immutable snapshot of the **serving map
 
 A small in-memory **registry** owns the serving map. When a storage operation changes what is servable, such as freezing a chunk, replacing a transaction hash index, discarding a hot database, or advancing the retention floor, the registry creates a new View, applies the change, and atomically publishes it. Queries already in flight continue using the View they admitted with, while newly admitted queries observe the updated View.
 
-Deletion safety is time-based rather than reader-tracked. Every request runs under a fixed deadline. Once a resource is removed from the current View, no newly admitted query can reach it. Any query that can still reach it must have admitted an older View and must finish before its deadline. The lifecycle therefore delays physical destruction until a grace period longer than the maximum request lifetime has elapsed.
+Deletion is deferred rather than reader-tracked. Every request runs under a fixed deadline. Once a resource is removed from the current View, no newly admitted query can reach it, and physical destruction waits out a grace period longer than the maximum request lifetime, so queries holding an older View can finish against it ([The reaper](#the-reaper)).
 
 The following terms are used throughout this document.
 
@@ -54,11 +54,11 @@ These races occur while ingestion, freezing, index replacement, discard, and pru
 The design uses four mechanisms:
 
 1. Every query resolves against one immutable View of the serving map, so routing decisions never mix storage states (H1).
-2. The registry publishes a new View whenever the serving map changes.
-3. The reaper destroys retired resources only after they have been unreachable for a grace period longer than the maximum request lifetime, so nothing a request can reach is destroyed while it runs (H2, H4).
+2. The registry publishes a new View whenever the serving map changes, for example when a freeze adds cold artifacts or prune advances the retention floor.
+3. The reaper destroys retired resources only after they have been unreachable for a grace period longer than the maximum request lifetime, and every close of a shared handle drains in-flight operations, so nothing a request can reach is destroyed while it runs (H2, H4).
 4. The registry maintains the `latest` watermark so queries never observe partially ingested ledgers (H3).
 
-Together, these mechanisms let ingestion, lifecycle operations, and queries proceed concurrently without requiring readers to coordinate with writers. Queries perform two atomic loads during admission, then use ordinary immutable data for the rest of the request. They do not acquire locks or participate in reference counting.
+Together, these mechanisms let ingestion, lifecycle operations, and queries proceed concurrently without requiring readers to coordinate with writers. Queries perform two atomic loads at admission and never register with the registry, hold reference counts, or block on writers.
 
 ```mermaid
 flowchart TB
@@ -72,25 +72,20 @@ flowchart TB
     CAT[("Catalog")] -.->|"startup rebuild"| V
 
     subgraph REG["registry"]
-        V["current View<br/>floor · hot · cold · indexes"]
+        V["current View<br/>floor · hot · cold · index coverage"]
         W["latest watermark"]
         HOT["hot DB handles"]
-        IDX["window .idx readers"]
-        CACHE["cold reader caches"]
         REAP["reaper"]
     end
 
     Q["Query"] -->|"load latest"| W
     Q -->|"load View"| V
+    Q -.->|"opens per request"| FILES["cold files: .pack, events, .idx"]
 
     V -->|"references"| HOT
-    V -->|"references"| IDX
-    V -.->|"opens on demand"| CACHE
 
-    REAP -->|"retire / close after T"| HOT
-    REAP -->|"retire / close after T"| IDX
-    REAP -->|"evict / close after T"| CACHE
-    REAP -->|"unlink / remove after T"| FS["retired files and dirs"]
+    REAP -->|"close after T"| HOT
+    REAP -->|"unlink after T"| FILES
 ```
 
 ---
@@ -105,7 +100,7 @@ Each View records:
 - the retention floor used by queries admitted with that View,
 - the hot stores,
 - the cold artifacts available for each chunk, and
-- the transaction hash indexes for each window.
+- the transaction hash index coverage for each window.
 
 The registry is rebuilt from the catalog during startup before the server begins accepting requests.
 
@@ -130,11 +125,10 @@ type ColdChunk struct {
 }
 
 type IndexCoverage struct {
-    Window WindowID
+    Window geometry.TxHashIndexID
     Lo, Hi chunk.ID // the frozen .idx's actual coverage, not the window bounds:
                     // Hi trails the tip while the window is current, and Lo is
                     // the retention floor at build time
-    Idx    *txhash.ColdReader
 }
 ```
 
@@ -143,8 +137,8 @@ type IndexCoverage struct {
 Each field exists for a specific reason.
 
 - **`hot` stores shared handles.** Each hot chunk has one registry-owned `hotchunk.DB` instance used by ingestion and queries.
-- **`cold` holds flags, not open readers.** A `ColdChunk` entry only says the chunk's artifact is frozen and servable. The reader object itself is opened on demand through the reader caches ([Open-handle management](#open-handle-management)): with thousands of cold chunks, holding every reader open would cost gigabytes, unlike the handful of hot databases and window indexes the View references directly.
-- **`indexes` stores transaction hash readers.** Each in-retention window has one current `.idx` reader.
+- **`cold` holds flags, not open readers.** A `ColdChunk` entry only says the chunk's artifact is frozen and servable. The reader itself is opened per request ([Open-handle management](#open-handle-management)): with thousands of cold chunks, holding every reader open would cost gigabytes, unlike the handful of hot databases the View references directly.
+- **`indexes` stores coverage metadata, not readers.** Each in-retention window has one current coverage entry; the `.idx` file itself is opened per request ([Open-handle management](#open-handle-management)).
 - **`latest` is stored outside the View.** The serving map changes only at chunk boundaries and lifecycle transitions, while `latest` advances every ledger. Keeping it separate avoids copying the View every few seconds.
 
 ### Admission
@@ -201,7 +195,7 @@ Its rule is:
 
 > A resource that was reachable from a published View may be destroyed only after it has been unreachable for at least grace period **T**.
 
-A resource becomes **unreachable** when it has been removed from the current View and can no longer be returned by reader caches. A resource is **destroyed** when it is permanently closed or removed.
+A resource becomes **unreachable** when it has been removed from the current View. A resource is **destroyed** when it is permanently closed or removed.
 
 ```go
 type Reaper struct {
@@ -216,16 +210,23 @@ Every read handler runs under an enforced request deadline **D**. The grace peri
 T = maximum request timeout + safety margin
 ```
 
-This is enough because every query keeps exactly one admitted View for its entire lifetime. Once a resource is unpublished, no newly admitted query can reach it. Any query that can still reach it must have admitted an older View, and therefore must finish before its deadline **D**. Since **T > D**, the resource is not destroyed until all such queries should have completed.
+The grace period preserves availability. A query that admitted a View before a resource was unpublished may still be routed to that resource for the rest of its lifetime. Since **T > D**, the resource remains open and servable until every such query has passed its deadline, so lifecycle transitions do not fail requests already in flight.
 
-The same rule covers hot databases, transaction hash index readers, and cached cold readers.
+The grace period alone is not sufficient for memory safety, because the deadline bounds the response, not the handler goroutine. On timeout, the request duration limiter cancels the request context and returns an error to the client without waiting for the handler. A handler blocked in a call that cannot observe cancellation, such as a RocksDB read stalled on a slow disk, keeps running past **D** and can still hold a reference after **T**.
+
+Memory safety therefore comes from ownership and drain barriers, not from time:
+
+- **Hot databases**: the only shared handles. `rocksdb.Store` holds a read lock across every C call; `Close` flips a closed flag, then takes the write lock, so it waits for in-flight operations to drain and every later operation fails cleanly with a closed error.
+- **Cold readers and window indexes**: opened and closed by the request that uses them ([Open-handle management](#open-handle-management)). No other component ever closes them, so a close cannot race a read.
+
+A straggler that outlives **T** is harmless: `Close` waits out its in-flight operation, and any later operation fails with a closed error. For retired files, unlink does not disturb descriptors already open: a straggler keeps reading the unlinked file, and the space is reclaimed when it exits.
 
 Deletion follows this sequence:
 
 1. Remove the resource from the serving map.
 2. Record the catalog demotion, such as `"pruning"` or `"transient"`.
 3. Schedule the resource with the reaper.
-4. After **T**, close handles, unlink files, remove directories, and remove the catalog entry.
+4. After **T**: close retired hot databases, unlink retired files, remove directories, and remove the catalog entry.
 
 Only physical destruction is delayed. Catalog demotions still occur during the lifecycle run.
 
@@ -260,9 +261,9 @@ Every write-side transition that changes what is servable gets a registry hook. 
 | `openHotDBForChunk` flips `hot:chunk:{c}` to `"ready"` | Publish `hot[c] = stores` using the shared instance. | Publish after the catalog key flips to `"ready"` and before the chunk's first ledger commits, so the watermark can never enter a chunk the current View does not serve.                                       |
 | Per-ledger ingest cycle: atomic `batch.Commit(sync)` plus in-memory applies | `latest.Store(seq)` | Final step of the cycle, after every serving structure contains the ledger.                                                                                                                                   |
 | `processChunk` flips artifact keys to `"frozen"` | Publish `cold[c].{kind} = true` for each frozen kind. | Publish after the commit. During startup backfill this hook has no effect; the startup scan (last row) covers those chunks.                                                                                   |
-| Transaction index rebuild (`buildTxhashIndex`): a single atomic catalog write freezes the new coverage and demotes its predecessor to `"pruning"` | Replace the window's `IndexCoverage` entry with the new coverage and its reader. Hand the predecessor's reader to the reaper. | Publish after the atomic write. The predecessor's file deletion, previously `buildThenSweep`'s immediate unlink, now waits out the grace period, because queries holding older Views may still be reading it. |
-| `discardHotDBForChunk` | Remove `hot[c]` from the View and hand the hot database to the reaper. | Unpublish before catalog demotion and every destructive step.                                                                                                                                                 |
-| Retention prune | Publish the run's new floor, which also drops below-floor `cold`, `hot`, and `indexes` entries. | Gate, unpublish, demote, then destroy. The floor update removes below-floor resources before every demotion.                                                                                                  |
+| Transaction index rebuild (`buildTxhashIndex`): a single atomic catalog write freezes the new coverage and demotes its predecessor to `"pruning"` | Replace the window's `IndexCoverage` entry with the new coverage. Schedule the predecessor's `.idx` unlink with the reaper. | Publish after the atomic write. The predecessor's file deletion, previously `buildThenSweep`'s immediate unlink, now waits out the grace period, because queries holding older Views may still open and read it. Coverage file names encode their `Lo-Hi` range, so an older View's entry opens its own generation of the index, never the replacement. |
+| Hot database discard (`DiscardHotChunk`) | Remove `hot[c]` from the View and hand the hot database to the reaper. | Unpublish before catalog demotion and every destructive step.                                                                                                                                                 |
+| Retention prune | Publish the run's new floor, which also drops below-floor `cold`, `hot`, and `indexes` entries. | Gate, unpublish, demote, then schedule destruction with the reaper. The floor update removes below-floor resources before every demotion.                                                                                                  |
 | Startup `serveReads()` | Build the initial View from the catalog scan: `"ready"` hot keys, `"frozen"` chunk keys, frozen coverages, and the calculated floor. | Complete before the lifecycle goroutine starts, so no freeze can land between the scan and the hooks becoming live.                                                                                           |
 
 Four ordering notes matter:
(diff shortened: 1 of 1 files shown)


sydneynotthecity:
@codex[agent]+gpt-5.5 review

sydneynotthecity:
@codex review

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+}
+```
+
+Every read handler runs under an enforced request deadline **D**. The grace period is derived from that deadline:
+
+```text
+T = maximum request timeout + safety margin
+```
+
+This is enough because every query keeps exactly one admitted View for its entire lifetime. Once a resource is unpublished, no newly admitted query can reach it. Any query that can still reach it must have admitted an older View, and therefore must finish before its deadline **D**. Since **T > D**, the resource is not destroyed until all such queries should have completed.
```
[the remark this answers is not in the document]
sydneynotthecity: closing as a duplicate of [r3544592318](https://github.com/stellar/stellar-rpc/pull/843#discussion_r3544592318)

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+Full history contains thousands of chunks and continues to grow. Keeping readers open for every cold artifact would consume substantial memory while providing little benefit for infrequently accessed data. Instead, the registry records only whether an artifact is available, and reader objects are managed through bounded per-kind LRU caches.
+
+Ledger and event readers use separate caches because their resource costs differ significantly. Separate caches also prevent heavy event traffic from evicting ledger readers needed by other query paths.
+
+### Interaction with the reaper
+
+Reader caches follow the same lifetime rules as every other serving resource.
+
+When a chunk is removed from the serving map, its cached readers are also retired. The reaper delays closing those readers until the grace period has elapsed, ensuring that no admitted query can still be using them.
```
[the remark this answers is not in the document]
sydneynotthecity: closing as a duplicate of [r3544592331](https://github.com/stellar/stellar-rpc/pull/843#discussion_r3544592331)

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+```mermaid
+flowchart TB
+    I["Ingestion"] --> CAT
+    L["Lifecycle"] --> CAT
+
+    I -->|"publish hot chunk"| V
+    I -->|"advance latest"| W
+    L -->|"publish freeze / index swap / floor / unpublish"| V
+    L -->|"retire resources"| REAP
```
chowbao: Would retired resources actually come from `catalog` when status is `freezing/pruning`?

Basically I'm wondering should `reaper` be connected to `catalog` and get all its retire resource info from `catalog`?

chowbao: Oh I see. This is actually a change to `lifecycle` we don't want the current ingest/lifecycle code to do the actual physical deletes. It will now be handed off to the `reaper`.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+    floor   chunk.ID
+    hot     map[chunk.ID]*hotchunk.DB
+    cold    map[chunk.ID]ColdChunk
+    indexes []IndexCoverage
+}
+
+type ColdChunk struct {
+    Ledgers bool
+    Events  bool
+}
```
chowbao: Do you need a bool for both `Ledgers` and `Events`? I thought a valid cold chunk always contains both

Or is this more about the coldChunk having a ledger reader or event reader?

design-docs/query-routing-design.md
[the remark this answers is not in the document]
chowbao: I don't agree with codex. You would want this to timeout and not wait for retirement or completion

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+    L->>V: Remove from serving map
+    Note over V: New queries cannot reach the resource
+    L->>C: Record "pruning" or "transient"
+    L->>P: Schedule destruction after T
+    P->>FS: Close, unlink, remove
+    P->>C: Remove catalog entry
+```
+
+The reaper has no persistent state. If the process exits before scheduled deletion completes, startup reconstructs pending deletion work from the catalog and runs it immediately. No new grace period is needed because demoted stores never enter the rebuilt View, so no query can reference them.
```
chowbao: Is it possible for a query to still be served while the reaper crashes?

In theory couldn't the restarted reaper delete before the original grace period is complete?

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+Each in-retention window keeps one `txhash.ColdReader` open.
+
+These readers maintain only the state needed to service lookups efficiently. Keeping them open also simplifies index replacement because older Views continue using the superseded reader until the reaper closes it after the grace period.
+
+### Cold readers
+
+Cold readers are opened on demand.
+
+Full history contains thousands of chunks and continues to grow. Keeping readers open for every cold artifact would consume substantial memory while providing little benefit for infrequently accessed data. Instead, the registry records only whether an artifact is available, and reader objects are managed through bounded per-kind LRU caches.
```
chowbao: This is probably an optimization for later depending on user patterns but I could see us just keeping X number of latest ledgers open and anything before X just be normal slow open/reads.

I would wonder how many cache hits you get for LRU

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+The admission protocol is described in [Admission](#admission).
+
+### Bounds
+
+Every request validates its leading edge and clamps its trailing edge against the admitted range `[floorLedger, latest]`.
+
+The leading edge determines where results begin. A range request whose leading edge falls below the admitted retention floor is rejected with an error that carries the available range, matching v1's out-of-range behavior. Silently clamping would drop the first results the caller asked for.
+
+The trailing edge determines where the scan ends. Requests extending beyond `latest` are truncated, and descending scans terminate at the retention floor.
```
chowbao, line 315: Is this what current RPC does? Like does it tell the user that results are truncated cause `latest + #` doesn't exist yet?

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+This proposal instead schedules those resources with the reaper. Physical destruction occurs after the grace period has elapsed.
+
+The catalog protocol is unchanged except for the delayed physical destruction of retired resources.
+
+---
+
+## Open questions
+
+- **Datastore fallback below the floor.** The v1 `getLedgers` handler can fall back to the remote object store for pre-retention ledgers. v2 must either preserve that behavior as an explicit exception to R2 or remove it at cutover. This decision belongs to #772.
+- **Serving floor precision.** This design currently uses a chunk-aligned floor, so `oldestLedger` advances in 10,000-ledger steps. A ledger-precise floor would better match v1’s sliding window, but adds another floor concept.
```
chowbao: I think chunk-aligned makes sense and I think v2 can diverge from v1. There seems to be little benefit from being ledger-precise aside from maybe some memory/disk resource savings

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+- **`getTransaction` probe parallelism.** The proposal uses sequential newest-first probing. If window count or tail latency becomes a problem, parallel probing can be added inside the lookup path without changing the routing model.
+- **Cache sizing.** Ledger and event reader cache capacities are implementation-time tuning parameters.
```
chowbao: Maybe a hot take. Does MVP v2 need a cache?

design-docs/query-routing-design.md
urvisavla: correct.

design-docs/query-routing-design.md
urvisavla: The reason is that there are separate entries in the catalog per chunk per type, one for events and one for ledgers.

If it’s always guaranteed that they’ll be updated together then we can get away with just one.

design-docs/query-routing-design.md
urvisavla: Reaper is part of RPC process, so if anything crashes, RPC will be restarted. In that case, all in-flight queries will be terminated so it should be okay to delete without a grace period on RPC startup.

chowbao:
Does each chunk only have 1 reader per type?

Like for example if there are multiple getLedgers() requests within the same chunk do they each get a different ledgerReader? Or do they share the same ledgerReader and just have different boundaries/cursors?

Basically I'm not sure exactly what does the reading and how multiple queries in the same chunk are served in parallel 😅

design-docs/query-routing-design.md
urvisavla: That’s a fair question. I think cacheing has a lot of upside for query performance without adding too much implementation complexity. But during the implementation phase, if we find that it’s adding unnecessary complexity, we can drop it from the MVP.

design-docs/query-routing-design.md
chowbao: I believe they are always guaranteed to be updated together from how we wrote the streaming logic.

@tamirms if you don't mind confirming

design-docs/query-routing-design.md
tamirms: I actually think there isn't going to be much benefit to caching reader objects (both packfile readers and streamhash indexes), because the OS page cache is effectively the LRU cache for bytes on disk. Say you open a file, read bytes from a specific area, and close the file. If you repeat that workflow and read from the same area, those bytes are still in the page cache and the second read is much faster than the first, whether or not anyone kept the file open in between.

So an LRU cache of readers mostly saves the open/close syscalls plus the small header and index reads each reader does at open which are served from the page cache. I believe the latency reduction isn't going to be significant.

If we did want caching, I think it would need to sit at a higher level to be effective. For example, an in-memory LRU mapping ledger sequence to ledger close meta bytes would avoid re-decompressing records for frequently accessed ledgers. A cache of plain values like that is also much easier to manage than a cache of readers, since evicting bytes is harmless while closing a reader that another request is still using is not.

So if it would simplify the design and implementation, I'd recommend getting rid of the LRU cache of readers, and additionally getting rid of the always-open list of txhash.ColdReaders.

design-docs/query-routing-design.md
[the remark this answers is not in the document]
tamirms: Actually, I think this is a genuine finding by codex. The way we implement timeouts on the request is by setting a deadline on the request's context. However, there is no guarantee that the goroutine which is fulfilling the request will actually terminate once the context's deadline is exceeded. For that to happen, the goroutine needs to periodically check that the context hasn't been cancelled (this is why long running operations in Go take a context argument).

However, the Go rocksdb bindings don't accept a context anywhere, so a goroutine inside a RocksDB `Get` is unreachable until the underlying call returns. The only timeout settings RocksDB itself provides are documented as best-effort, checked periodically between reads and dependent on file system support, so they can bound a long scan but cannot interrupt a read already blocked in the kernel. On a healthy disk this wouldn't be an issue. But on a degraded disk, a read could be blocked for seconds or minutes or longer.

Here's an example which illustrates this failure:

1. t=0s: a request starts probing the hot chunk's RocksDB.
2. t=0.1s: the read enters a CGO call that hits a degraded disk and blocks inside the kernel.
3. t=10s: the deadline fires. The client gets its timeout. The goroutine is still blocked inside the C++ call and has no way to know it was cancelled.
4. t=25s: the grace period T expires. The reaper closes the RocksDB handle and frees the C++ object.
5. t=40s: the disk recovers, the blocked call resumes, and it is now executing C++ code against freed memory.

Step 5 is not an error we can handle. If we're lucky it's a segfault, and a segfault inside C code is not a recoverable Go panic: the entire daemon dies (ingestion included). If we're unlucky it's worse: the resumed call reads and writes through freed pointers (RocksDB reads update block-cache and statistics state internally), and that freed memory may already belong to something else that's live, like another handle or another request's buffers.

I think this is something we should consider seriously and figure out how to fix.

design-docs/query-routing-design.md
tamirms: yes, that should be the case after backfill finishes running

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Explicit reader tracking
+
+Track active readers directly, either by reference-counting Views/resources or by holding a reader lock while a query runs.
+
+This was not chosen because it adds work to every query: each request must acquire and release some reader state. This proposal avoids that by using the request deadline as the bound on reader lifetime. Queries only load `latest` and the current View at admission.
+
+### Filesystem unlink semantics alone
+
+Rely on the filesystem to keep already-open files readable after they are unlinked.
```
tamirms: I had claude explain this to me as I wasn't familiar with this concept and this to me looks really promising. I know you mentioned that this wouldn't work for rocks db handles but I don't think the delayed delete approach would work for rocks db either (see https://github.com/stellar/stellar-rpc/pull/843/changes#r3549598104 ). Perhaps it could be acceptable to treat the hot dbs as a special case?

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+    REAP -->|"unlink / remove after T"| FS["retired files and dirs"]
+```
+
+---
+
+## Registry and View
+
+The catalog remains the durable record of every store and its lifecycle state. The registry is a disposable in-memory projection of the catalog.
+
+A query loads one View when it is admitted and uses that same View for its entire lifetime. Later View updates affect only newly admitted queries.
```
tamirms: I wonder if this could be simplified by using rocks db snapshots. Instead of maintaining an in-memory copy of the catalog as the view abstraction, you could take a rocks db snapshot whenever we start processing a request. The snapshot would be equivalent to a "repeatable read" transaction in postgres: a consistent view of the catalog that doesn't include any mutations that occur during the lifetime of the request. Do you think this approach could eliminate the view abstraction?

design-docs/query-routing-design.md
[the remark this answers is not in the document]
urvisavla: You're right that the timeout doesn't stop the goroutine. The limiter cancels the context and returns; the handler keeps running (verified locally). But the RocksDB step of your timeline can't happen: our wrapper takes a read lock for every C call, and Close waits for those to finish before freeing anything (`storage/rocksdb/rocksdb.go`, pinned by `TestStore_CloseWaitsForInflightIterate`). So the reaper would stall, not free live memory. The real gap was the txhash cold reader: its Close unmaps with no such guard, and racing Gets against Close does segfault the process (reproduced locally). The doc is updated in the latest commit: safety now comes from drain barriers and per-request ownership of cold readers, and the grace period only keeps resources available to in-flight queries; it no longer carries the safety argument.

design-docs/query-routing-design.md
urvisavla: Adopted, and your instinct about the special case was right. The doc now splits it exactly that way: cold files are opened per request and unlinked after the grace period, so unlink semantics carry them with no reader tracking at all; hot databases are the one shared resource, and their safety comes from the wrapper's existing drain-on-close. One correction to the pure-unlink version: immediate unlink breaks a query that opens a chunk for the first time after prune removed it, so the unlink is deferred by the grace period. Details in the revised reaper and open-handle sections.

design-docs/query-routing-design.md
urvisavla: The reader cache was solving a real problem. An open re-reads and re-parses the MPHF and offsets tables, and the page cache doesn't help with that part, so a cached reader saved that work for frequently hit chunks. But its value is all in the parsed state, while its complexity is all in the handles: cached fds need retirement rules, reaper coordination, and safe eviction. That complexity isn't needed for the MVP, so we removed caching entirely; all cold files are opened per request, and the always-open `.idx` list is gone too. If profiling shows the re-parsing cost matters, the doc records the future fix: cache the parsed state keyed by chunk (your values-not-readers point), never the handles.

design-docs/query-routing-design.md
urvisavla: Valid suggestion, and it would partly work: keep the registry for the shared hot DB handles (a snapshot can't provide those) and read floor, frozen state, and coverage through a catalog snapshot. The real pro is that this metadata could never go stale, since it's read from the source of truth at use time. What decides it for me is the frequency mismatch. The serving map changes a few times a day; requests are constant. Between transitions the routing state is identical for every request, so per-request snapshots mean crossing the CGO boundary on every request to recompute an unchanged answer, plus a snapshot to release on every path. The View does the derivation once per transition and requests pay one atomic pointer load. Added this to Alternatives considered in the latest commit.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+    Idx    *txhash.ColdReader
+}
+```
+
+### Design rationale
+
+Each field exists for a specific reason.
+
+- **`hot` stores shared handles.** Each hot chunk has one registry-owned `hotchunk.DB` instance used by ingestion and queries.
+- **`cold` holds flags, not open readers.** A `ColdChunk` entry only says the chunk's artifact is frozen and servable. The reader object itself is opened on demand through the reader caches ([Open-handle management](#open-handle-management)): with thousands of cold chunks, holding every reader open would cost gigabytes, unlike the handful of hot databases and window indexes the View references directly.
```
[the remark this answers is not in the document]
urvisavla: Resolved by the design revision: the reader cache no longer exists. Cold readers are opened per request and closed by the request that opened them, so there is no cache-miss-after-unpublish path. A query holding an older View can still open a pruned chunk's files because their unlink is deferred by the grace period. See the revised open-handle management and reaper sections.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,532 @@
...
+Rely on the filesystem to keep already-open files readable after they are unlinked.
+
+This was not chosen because it only applies to already-open cold files. It does not protect RocksDB handles, cache misses that open a file after unlink, or other resource types that must be closed explicitly. This proposal uses the grace period uniformly for hot databases, cold readers, and index readers.
+
+---
+
+## Related documents
+
+- [full-history-streaming-workflow.md](./full-history-streaming-workflow.md): the write side this design reads from: geometry, the catalog and one write protocol, the lifecycle run, the reader contract, and INV-1 through INV-4.
+- [gettransaction-full-history-design.md](./gettransaction-full-history-design.md): the tx-hash tiers, `.idx` coverage semantics, and the cold-lookup verify chain that `getTransaction` routing drives.
```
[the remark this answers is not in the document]
urvisavla: Correct, those two documents aren't on this base branch yet: they land with the streaming-design docs PR, which is expected to merge before this one. The links resolve once it does.

urvisavla:
> Does each chunk only have 1 reader per type?

Yes — all readers support concurrent reads: positional, no cursor, each call carries its own position. Parallel queries in one chunk never contend, shared or not.

design-docs/query-routing-design.md
urvisavla: Following up: we ended up doing exactly this. The MVP has no reader cache; cold files are opened per request (see the thread with Tamir above and the latest commit). A cache of parsed state is recorded in the doc as a deferred optimization.

design-docs/query-routing-design.md
urvisavla: Keeping the two flags: the catalog tracks per-kind artifact entries and the View mirrors what it reads from the catalog. One bool would bake in an invariant the catalog itself doesn't encode.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,535 @@
...
+
+    L->>V: Remove from serving map
+    Note over V: New queries cannot reach the resource
+    L->>C: Record "pruning" or "transient"
+    L->>P: Schedule destruction after T
+    P->>FS: Close, unlink, remove
+    P->>C: Remove catalog entry
+```
+
+The reaper has no persistent state. If the process exits before scheduled deletion completes, startup reconstructs pending deletion work from the catalog and runs it immediately. No new grace period is needed because demoted stores never enter the rebuilt View, so no query can reference them.
```
tamirms: Suggestion on the reaper: I think the lifecycle run itself should execute the deferred deletions, rather than a separate reaper component. Concretely: as the run's stages demote things in the catalog, they also append each demoted item to a list local to the run. When the stages finish, if the list is non-empty, the run waits out the grace period (max request timeout + margin) once, then deletes each item with a bounded number of retries. If an item can't be deleted (say a close is blocked by a read stuck on a slow disk), it alarms and skips it, and the next run's scans re-discover the demoted key and try again.

The main reason: the scan-based re-discovery path has to exist anyway. The reaper's queue is in memory, so after a crash the pending deletions must be reconstructable from the catalog regardless, which means the catalog-driven path is mandatory in both versions. A separate reaper ends up being a second delivery mechanism layered on top of the one we're required to have, and it needs coordination with the scans: a run that starts while the reaper still holds items would re-discover those same demoted keys, so we'd need dedup between them or deletes that are safe to run twice concurrently, which is fine for an unlink but dicey for a hot DB close.

It also preserves the "idle means settled" property and the INV audits as they are. With deletion inside the run, a run ending means everything it demoted is gone (or alarmed), so the audits keep their current meaning. With a reaper, demoted keys and their files legitimately linger while the daemon is idle, and an audit can't tell "waiting out the grace period" from "a sweep got stuck" without extra state.

The cost is that a run takes a couple of minutes longer at the end, which seems fine against runs that happen every ~14 hours. And it removes the questions a reaper thread would otherwise need answered (its error policy, shutdown behavior, what happens when one stuck close blocks the queue).

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,535 @@
...
+
+The following terms are used throughout this document.
+
+- **Chunk**: 10,000 consecutive ledgers, the unit of storage.
+- **Window**: 1,000 chunks, or 10 million ledgers. Each window has one transaction hash `.idx` file.
+- **Serving map**: The in-memory map of chunks and windows to the stores that serve them, plus the retention floor.
+- **View**: An immutable snapshot of the serving map admitted by a query.
+- **Hot / cold**: Hot data is served from a live RocksDB database, one per hot chunk. Cold data is served from sealed files written once during freezing.
+- **Catalog**: The durable RocksDB record of each store and its lifecycle state.
+- **Retention floor**: The oldest ledger served by a View. A request whose leading edge is below its admitted View's floor is not served (R2).
```
tamirms: Heads up in case you haven't seen it: #859 recently landed on the feature branch and reworked how the retention floor is computed on the write side. The floor is now a pure function of a committed frontier — `geometry.Retention.FloorAt(frontier)` — with startup backfill anchoring on its pass target and the lifecycle run on its boundary chunk (the old ledger-anchored `EffectiveRetentionFloor` is gone). I think the read side should follow the same rule rather than keeping the floor as a separately maintained value.

Why it matters: a query needs two facts that must agree — where the served range starts (the floor), and which chunks the tx index covers. In the current doc they're updated at different times during a run: the index is rebuilt early (covering only chunks from the run's new floor upward), while the floor only advances at the prune stage at the end. A query admitted in between reads a fresh index paired with a stale floor: it's told service starts at chunk 95 while searching an index that starts at 96, so getTransaction returns not-found for chunk-95 data that should still be servable. With bounded retention this window opens on every run.

The fix: stop keeping the floor as its own stored value. A query already sees which chunks exist, and the floor is just "newest chunk minus retention" — `FloorAt`. If each query computes its floor that way, the floor and the index it's paired with can't disagree, because one is calculated from the other.

urvisavla: Agreed — the latest commit does this: the floor is derived per query from `FloorAt`, nothing stores it.

design-docs/query-routing-design.md
urvisavla: Adopted as proposed in the latest commit.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,481 @@
...
+
+> A resource that was servable may be destroyed only after it has been demoted for at least grace period **T**.
+
+Every read handler runs under an enforced request deadline **D**, and the grace period is derived from it:
+
+```text
+T = maximum request timeout + safety margin
+```
+
+As a run's stages demote resources in the catalog, they append each demoted item to a list local to the run. When the stages finish and the list is non-empty, the run waits out the grace period once, then deletes each item: closing discarded hot databases, unlinking superseded index generations and pruned chunk files, and removing their catalog entries. An item that cannot be deleted, such as a hot database whose close is blocked by a read stuck on a slow disk, is alarmed and skipped; the next run's scans re-discover its demoted key and retry.
```
tamirms, line 194: I think there's an inconsistency between two paragraphs in this section. The deletion sequence says a blocked close is "alarmed and skipped," but the drain barrier described below it is blocking: `Close` flips the closed flag and then takes the write lock, which waits indefinitely for in-flight reads. As written, the run has no way to skip a blocked close, because the only close available blocks until the stuck read returns — so one wedged read would stall the run after all.

I'd suggest a non-blocking variant for the deletion path, something like `CloseIfIdle() bool`: set the closed flag, close if no reads are in flight, otherwise return false so the run can alarm, skip, and let the next run's scans retry. By the time the run deletes, the grace period has already outlasted every request timeout, so in the normal case no readers are left and it succeeds immediately; a false return can only mean a read stuck past its deadline, which is exactly the case the alarm is for. For what it's worth, this matches how the Go stdlib handles the same situation for os.File: when an in-flight read can block in the kernel indefinitely, Close deliberately does not wait for it — it marks the fd closed and lets the last reader run the actual close on its way out. Delegating works there because the teardown is one syscall; ours also removes a directory and a catalog key, which is better done by the run's retry than by whichever request goroutine happens to unwedge last.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,481 @@
...
+
+The cursor identifies a transaction within a ledger. The first ledger resumes after the cursor position, while subsequent ledgers begin with their first transaction.
+
+### `getTransaction`
+
+A transaction hash does not identify the chunk containing the transaction, so routing cannot resolve it directly.
+
+Instead, the router probes the transaction indexes in two stages:
+
+1. Probe the hot transaction indexes. A match is definitive.
```
tamirms, line 372: Small gap in the probe rules: candidates below the admitted floor are skipped, but there's no matching gate at the top. The hot probe reads the live chunk's txhash CF through the shared handle, and that data keeps advancing after admission — so the probe can find a transaction committed after the query was admitted. The response would then return a transaction whose ledger is above the `latestLedger` it advertises, and a getEvents at the same moment wouldn't show that ledger's events yet, since `latest` only advances after the in-memory event applies. I'd suggest skipping hot candidates above the admitted `latest`, mirroring the floor gate: a candidate is served iff floor ≤ ledger ≤ latest, both from admission. One comparison, and the answer it produces (not found until `latest` catches up) is what the client's retry would see anyway.

design-docs/query-routing-design.md
```diff
@@ -0,0 +1,481 @@
...
+```go
+func (r *Router) Admit() (*Admission, error) {
+    latest := r.latest.Load()
+    handles := r.handles.Load()
+    snap, err := r.catalog.NewSnapshot()
+    if err != nil {
+        return nil, err
+    }
+    // The frontier is the last complete chunk, the same anchor prune uses.
+    floor := r.retention.FloorAt(geometry.LastCompleteChunkAt(latest))
```
tamirms: A small refinement on the floor derivation: it anchors on `latest`, the first of the three loads, while the routing metadata comes from the snapshot, the last. The prose covers prunes completed before admission ("an admitted floor is never behind a completed prune's floor"), but a run whose index swap and prune both commit between the `latest` load and the snapshot would pair an old floor with newer catalog state: chunks the floor still admits are already demoted in the snapshot and missing from the new index coverage. Reaching that requires a goroutine to stall for a whole run between two adjacent loads, so it's theoretical — but it's also free to close: derive the floor from the snapshot itself, `FloorAt(highest ready hot chunk in the snapshot − 1)`. The live chunk's ready key is always committed before the boundary that triggers a run, so any snapshot containing a given coverage also contains the ready keys that produce the matching floor — the floor and the coverage then come from the same read and can't describe two different states. The only skew left is in the safe direction (a floor briefly ahead of an old, wider coverage, which stops serving an about-to-be-pruned chunk slightly early). It also makes admission fully uniform: everything except `latest` and the handles comes from the snapshot. And it's the same frontier the lifecycle run already anchors on — the boundary chunk it's notified with equals the catalog's highest-ready-minus-one at run start — so the two floor consumers would share one definition, applied to different views of the same catalog.

urvisavla: Adopted — the floor now derives from the snapshot's frontier (latest commit).

design-docs/query-routing-design.md
urvisavla: Adopted — candidates are gated to the admitted `[floor, latest]` (latest commit).

design-docs/query-routing-design.md
urvisavla: Adopted as proposed — `CloseIfIdle`, alarm and skip, next run retries (latest commit).

tamirms (approved):

Issue #841, opened 2026-07-06
full-history: dead pre-wired surface + stale contract docs (~300 lines deletable)


tamirms:
Split out from the #820 review's final pass over pre-existing full-history code. The freeze rework left behind API pre-wired for a handoff design that no longer exists; none of it is reachable from production, and several docs will actively misdirect the #772 read-path work.

**Deletions (~300 lines):**
- The freeze-handoff mirror surface (~95 lines): `ConcurrentBitmaps.Snapshot`/`Len`, `ConcurrentLedgerOffsets.Snapshot`/`EventIDs`, and the Clone-safe/COW ownership contract that exists only for them. `ConcurrentBitmaps.Snapshot`'s doc cites a WriteColdIndex call chain that does not exist — the cold index is built from a single-threaded `Bitmaps` in `ingest/events.go`.
- `Reader.Lookup` (~90 lines): `Query` uses `LookupKeys` exclusively; `Lookup(k) ≡ LookupKeys([k])[0]`, and `ColdReader.Lookup` hand-duplicates the whole resolve pipeline for one key. Deleting it retires `ErrTermNotFound`.
- `events.TermsFor` (~45 lines): test-only; its doc names cold backfill as the consumer, which actually uses `TermsForBytes`.
- The dead `IndexBaseChunk`/`IndexFileName` naming scheme — `OpenColdReader`'s doc tells future wiring to compose a filename that never exists on disk (the real layout is `geometry.Layout.TxHashIndexFilePath`).
- The 20 `streaming:` prefixes on error strings (config_lock ×7, config ×2, catalog ×6, geometry ×5): delete rather than rename — error context comes from the wrap chain, not self-locating strings, and the prefix is the codename #824 retires anyway. The four Info/Debug log prefixes can follow the same decision.
[...]

PR #820, opened 2026-06-24
Streaming daemon: Phase 2 — hot store, lifecycle + live ingestion (closes #816, #808)

chowbao:
**Closes #816, #808** — the complete Phase 2 (hot tier + lifecycle **+ live ingestion**) in one PR.

> **Update:** this PR now contains all of Phase 2. The live-ingestion layer originally split out as **#821 (layer 2)** was folded in here, so #820 alone completes #816 (and #808) rather than the two-PR split. #821 is closed in favor of this PR. Rebased onto the merged Phase 1 (#819) on `feature/full-history`.

Phase 2 introduces the **hot tier + lifecycle machinery + live ingestion** on top of Phase 1's source-blind cold pipeline:

- **Hot storage — the per-chunk `hotchunk` DB**
  - single shared multi-CF RocksDB: ledgers + events + tx-hash
  - `transient`/`ready` state machine; a read-only view serves both the freeze source and the watermark refiner
  - `ingest.HotService` commits each ledger as **one atomic synced WriteBatch across all CFs** (decision (a)) — one `fsync`, a single per-chunk `MaxCommittedSeq`, no per-store frontiers, no fan-out
  - one `ExtractLedgerEvents` walk per ledger feeds both the tx-hash and events CFs (event-ID assignment order unchanged)

- **Backfill integration — hot source by path (no probe seam)**
  - `backfillSource`'s hot branch opens the chunk's hot DB **read-only straight from its `geometry.Layout` path** (`hotchunk.OpenReadOnly`) and yields a `ledgerbackend.LedgerStream`
  - a `ready` chunk whose DB is missing/gutted fails the must-exist open — an ordinary restartable error, **never auto-healed** into a fresh empty DB (no watermark regression)

- **Progress — derived watermark**
  - `LastCommittedLedger(cat, logger)` maxes the cold term (highest fully-durable chunk) vs the highest `ready` hot DB's `MaxCommittedSeq` (one read-only open, which replays any synced WAL after an ungraceful crash) vs the earliest-pin floor — all in the signed domain, never stored; restart re-derives from durable state

- **Lifecycle + live ingestion (the folded-in layer 2)**
  - `run()` transitions from backfill to a **serve+ingest** steady state: start captive core (injected `CoreOpener`) → serve reads (injected) → run the ingestion loop and the lifecycle loop as a **joined `errgroup.WithContext` pair** (whichever returns first tears down the other; both joined before `run` returns — the single-lifecycle-goroutine invariant across supervisor restarts)
[...]

PR #844, opened 2026-07-07
fullhistory: reorganize packages (#824)

### LINKED ISSUE #824: fullhistory/streaming: split the streaming package into purpose-named packages
Follow-up from the round 2 review of #817.

`internal/fullhistory/streaming` has grown into a 13-file catch-all. Once the full-ingestion feature is built and stable, it's worth splitting into packages named for what they do. Figuring out the right split is part of the work, so the layout below is a **starting point, not a final answer**.

## What it should follow
- Name packages after what they do, using the design doc's terms — not "streaming", which is just a codename.
- Drop the `pkg/` folder; it does nothing inside `internal/`.
- Keep imports flowing one direction (no cycles).
- Leave the already-clean packages (`chunk`, the stores, `packfile`) alone.

## A possible layout
```
fullhistory/   the daemon: startup, config, the live ingestion loop
  geometry/    chunk and tx-hash-index math, the key/path naming, fsync helpers
  catalog/     the on-disk state: the write protocol, the sweeps, hot-chunk discard
  ingest/      the write engine: per-type writers, services, RunColdChunk, the sources
  backfill/    loading old history: resolve, build, execute
  lifecycle/   freeze, discard, prune (plus progress and retention)
  stores/{ledger,events,txhash,hot,meta}/   the individual stores
  rocksdb/  packfile/
```

## Still to decide (part of this issue)
- Whether the live ingestion loop goes in the top-level daemon package or its own.
- Whether `chunk` and `geometry` stay separate or merge.
- The exact store folder names (the events store can't just be `events` — there's already an `internal/events` package).
- Whether progress and retention live in `lifecycle` or `catalog`.
- Should there be a `config/`?

Removing the `crashHooks` field in #817 also clears the way to move the catalog into its own package.

### LINKED ISSUE #835: fullhistory: simplify LastCommittedLedger to the design signature LastCommittedLedger(cat)
Follow-up from PR #820 review (thread on backfill/process.go, comment 3515787395).

The nil-logger positional mode of `lifecycle.LastCommittedLedger(cat, logger)` is production-dead (the only production caller, startup.go, passes a real logger; the tick's `LastCommittedLedger(cat, nil)` went away with the #25 clamp deletion). It should be removed, along with the `logger` parameter, landing the design-pseudocode signature `LastCommittedLedger(cat)`.

Blocked on two coupled changes, which is why it wasn't folded into the review round:
1. The `TestLastCommittedLedger` unit tests use the nil-logger mode as a seam to exercise the positional derivation WITHOUT opening a real hot DB. They must be converted to real-DB tests (as `TestDeriveWatermark_RealHotDB_RefinementIsNotStale` already is) or the positional-only cases dropped.
2. The refinement's read-only open needs a logger; drop the param by defaulting nil->nop in `rocksdb.Config` (the same idiom `orNop` applies to MetricSink), or retain a logger on `Catalog`.

Pure cleanup — no behavior change.

chowbao:
Closes #824. Closes #835. The `streaming/` catch-all was already split into `geometry`/`catalog`/`ingest`/`backfill`/`lifecycle` + per-domain stores in the #817→#820 stack; this PR does the leftover moves plus the safe cleanups agreed on the issue.

**This is a pure reorganization — moves + mechanical renames, zero behavior change.** No production logic is altered.

### 1. `pkg/` → `storage/`
`pkg/` is meaningless inside an already-`internal/` tree. Rather than a flat drop (the issue's literal ask), this renames it to `storage/` — a meaningful name for the storage tier that sits *below* the daemon logic, keeping `chunk`, `rocksdb`, and the per-domain `stores/` grouped.

> **Deviation from the issue text**, which says "drop the `pkg/` folder." Flagging explicitly — happy to flatten fully instead if reviewers prefer.

### 2. Extract `config/`
`config.go` + `config_lock.go` (schema, load/parse, `Paths`→`Layout`, root locks) move into a `config/` package. Cycle-safe: only `main.go` imports the root package and it uses `RunDaemon(ctx, path)` — no config symbols.

`config_validate.go` **stays in root**: `validateConfig` depends on `catalog.Catalog` and the root-defined `NetworkTipBackend`, so it's startup orchestration, not static config — moving it would create a `config → root` cycle.

### 3. Extract `durable/` (fsync helpers)
Per [tamirms' review](https://github.com/stellar/stellar-rpc/pull/818#discussion_r3479586855): the 7 os-level durability ops (`BarrierNewFile`, `FsyncDir`, `FsyncParentDirs`, `FsyncNewDirs`, `DeepestExistingDir`, `DeleteFileIfExists`, `RmdirIfEmpty`) had zero geometry/domain references — they only sat in `geometry/paths.go` by accident of sharing the file with the path bijection. They move to a `durable/` leaf imported by `config`, `catalog`, `backfill`, and the daemon. `geometry` is now naming + math only (no `os` import).

### 4. Safe ride-along cleanups
- Removed the `errLockHeld` relay sentinel — `acquireLock` wraps `ErrRootLocked` directly (one producer, one consumer). [issue comment](https://github.com/stellar/stellar-rpc/issues/824#issuecomment-4864164638)
- Swept the `streaming` codename from operator-facing error/log messages: the self-locating prefixes are **deleted outright** (per the #841 plan — an error's location comes from its wrap chain, so a prefix is redundant whether it says `streaming:` or `fullhistory:`); the four Info/Debug log lines follow the same decision. Comments naming a now-nonexistent "streaming package/daemon" reworded. Left domain uses of "streaming", the `fullhistory_streaming` Prometheus subsystem, and metastore test-key literals (data/interface formats, not codenames).

## Open-question resolutions (from #824)
- **chunk + geometry: kept separate** — the dependency graph forbids merging (`chunk` is imported *by* the stores; `geometry` imports the stores → a merge would cycle).
- **hotloop: stays in root** — it's the daemon's runtime loop, coupled to `daemon.go` orchestration.
- **progress → root; retention stays in `lifecycle`** — since the #25 clamp deletion, progress.go's only production caller is `startup.go`, so it moves next to it (unexported `lastCommittedLedger`). Retention remains `lifecycle`'s freeze/prune policy gate.

## 5. Behavioral cleanups (second commit)
The logic refactors from the review, kept in a **separate commit** (`fullhistory: hot-create choreography, OpenReady, test-fixture cleanup`) so the reorg commit above stays a reviewable pure-reorg:

- **`cat.BeginHotCreate` / `cat.FinishHotCreate`** — the hot-DB create choreography (wipe + transient key → `hotchunk.Open` → fsync chain → flip ready) moves behind the catalog, mirroring `MarkChunkFreezing`/`FlipChunkFrozen`. `hotloop.openHotDBForChunk` now sandwiches `hotchunk.Open` between the pair and no longer does raw `os`/fsync itself. [discussion_r3517588234](https://github.com/stellar/stellar-rpc/pull/820#discussion_r3517588234)
- **`hotchunk.OpenReadyWrite` / `OpenReadyView`** — a thin named pair over one unexported `openReady` enforcement site for "ready key ⇒ must-exist, never-creating open," so call sites name the handle instead of passing a naked boolean. `openHotDBForChunk` (write), `refineWithHotDB`, and `tryHotSource` (view) all route through it. `hotchunk` gains a `geometry` import for `HotState` (verified acyclic). [discussion_r3517588278](https://github.com/stellar/stellar-rpc/pull/820#discussion_r3517588278)
- **`DeleteHotKey` → `deleteHotKey`** unexported (only prod caller is same-package `DiscardHotChunk`); its raw-key round-trip test moves into the `catalog` package. `PutHotTransient`/`FlipHotReady` stay exported — they're genuine cross-package test-setup primitives.
- **Shared `fhtest` package** holds the `ZeroTxLCMBytes` fixture, replacing three byte-identical copies.

> **Not done:** the review also named `testSink`/`execRecorder` for consolidation, but each exists in exactly one package — they aren't duplicated, and merging them would only force package-internal interfaces to be exported. Left as-is.

## 6. Fold `metastore` into `catalog` (finishes #824) + `lastCommittedLedger(cat)` (finishes #835)

Per [review comment](https://github.com/stellar/stellar-rpc/pull/844#issuecomment-4914950882): `metastore` was a 115-line rocksdb wrapper with exactly one consumer. The string-KV surface is now catalog-internal (`catalog/kv.go`, unexported), and `NewCatalog(store, …)` becomes `catalog.Open(path, layout, txLayout, logger)` — the catalog owns its backing store, `Close` releases it. Gone with the fold: the two-step daemon wiring, the `metastore.BatchWriter` type threaded through the protocol/sweep closures, the `ErrNotFound` manufacture-then-translate round-trip in `Catalog.get`, and `metastore.New`'s guards that duplicated `rocksdb.New`'s. The KV test suite moved to `catalog/kv_test.go`.

The fold hands over the rest of #835: the catalog now carries a logger, so `refineWithHotDB` uses `cat.Logger()`, the `logger` parameter drops, and `lastCommittedLedger(cat)` lands as the design wrote it. The `deriveCompleteThrough`/`deriveLastCommitted` test-shim aliases collapsed into direct calls (the nil-logger seam #835 worried about was already gone — those tests run against real hot DBs).

One test seam changed shape: backfill's skip-when-coverage-frozen test used to seed raw `"pruning"` chunk keys on the bare store; with no bare store to reach, it now runs the real `SweepChunkArtifacts` — which is exactly the finalized-window scenario it was simulating.

## Verification
`go build`, `go vet`, `go test -short ./cmd/stellar-rpc/internal/fullhistory/...`, and the repo golangci-lint all green (rocksdb cgo) at every commit. Renames are git-mv (history preserved); import rewrites are mechanical (`gofmt -r`, AST-based).


cmd/stellar-rpc/internal/fullhistory/lifecycle/progress.go
```diff
@@ -73,9 +73,12 @@ func LastCommittedLedger(cat *catalog.Catalog, logger *supportlog.Entry) (uint32
 // MaxCommittedSeq is correct even after an ungraceful crash.
 func refineWithHotDB(cat *catalog.Catalog, logger *supportlog.Entry, live int64) (uint32, error) {
 	id := chunk.ID(live) //nolint:gosec // live > cold >= -1, so live >= 0
-	hot, openErr := hotchunk.OpenReadOnly(cat.Layout().HotChunkPath(id), id, logger)
+	// live is the highest "ready" hot chunk (from ReadyHotChunkKeys), so route the
+	// read-only open through the single OpenReady enforcement site: must-exist,
+	// never auto-healed, uniform won't-open error.
+	hot, openErr := hotchunk.OpenReady(geometry.HotReady, cat.Layout().HotChunkPath(id), id, logger, true)
```
tamirms: The PR description resolves "progress + retention: stay in `lifecycle` — freeze/prune policy," but that rationale only fits retention (the tick's gate). Nothing in the lifecycle package has called anything in progress.go since the clamp deletion — its only production caller is `startup.go` — and #835's recorded scope (from your deferral on #820) includes moving it next to that caller in the `fullhistory` root. Since this PR is exactly the place for pure moves, suggest doing it here: `git mv` progress.go to the root, unexport `LastCommittedLedger` (same package as its caller now), and #835 shrinks to just the logger change. If you'd rather not, no problem — but then the resolution text should say "deferred to #835," not "stays."

cmd/stellar-rpc/internal/fullhistory/config/config_lock.go
```diff
@@ -92,29 +87,29 @@ func LockRoots(roots ...string) (*RootLocks, error) {
...
+		return nil, fmt.Errorf("fullhistory: open lock file %q: %w", path, err)
 	}
 	if err := acquireLock(f); err != nil {
 		_ = f.Close()
-		if errors.Is(err, errLockHeld) {
+		if errors.Is(err, ErrRootLocked) {
 			return nil, fmt.Errorf("%w: %q (another daemon is using it)", ErrRootLocked, root)
 		}
-		return nil, fmt.Errorf("streaming: lock %q: %w", path, err)
+		return nil, fmt.Errorf("fullhistory: lock %q: %w", path, err)
```
tamirms: The sweep renamed `streaming:` → `fullhistory:` on the error strings, but the plan recorded on #841 is to delete these prefixes outright: an error's location comes from its wrap chain (callers already wrap with "read config", "lock root", etc.), so a self-locating prefix is redundant — renaming it just moves the problem from a codename to a package name. Since you're already sweeping every one of these strings in this PR, deleting the prefix instead is the same mechanical pass and closes #841's bullet rather than half-addressing it. If you prefer keeping the prefix, that's shippable too — just needs a note on #841 so its bullet shrinks to the four Info/Debug log lines.

cmd/stellar-rpc/internal/fullhistory/storage/stores/errors.go
```diff
@@ -1,9 +1,9 @@
 // Package stores hosts shared per-domain store packages (ledger,
 // txhash, metastore) and the cross-cutting error sentinels they
 // emit. Per-domain stores translate their backing primitive's
-// errors (pkg/rocksdb, pkg/packfile, os) into these sentinels at
+// errors (storage/rocksdb, storage/rocksdb, os) into these sentinels at
```
tamirms: The sweep garbled this line: "(storage/rocksdb, storage/rocksdb, os)" — the second one was `pkg/packfile` in the base (itself stale; packfile lives at `internal/packfile`). Should read "(storage/rocksdb, internal/packfile, os)".

cmd/stellar-rpc/internal/fullhistory/storage/rocksdb/rocksdb.go
tamirms, line 1: One codename straggler the sweep missed, at rocksdb.go:579: "The **streaming** ingestion contract requires…" — that's the retired codename, not the domain word. "The ingestion contract" says it.

cmd/stellar-rpc/internal/fullhistory/storage/stores/hotchunk/hotchunk.go
```diff
@@ -97,6 +98,33 @@ func OpenReadOnly(path string, chunkID chunk.ID, logger *supportlog.Entry) (*DB,
...
+// OpenReady is the single enforcement site for the "ready key ⇒ must-exist,
+// never-creating open" rule. It takes the hot-key state the CALLER already read
+// and refuses to open anything not "ready", so no caller can accidentally open a
+// creating handle for a chunk the catalog considers ready. readOnly=false is
+// ingestion's write handle for a resumed chunk (OpenExisting); readOnly=true is
+// the freeze source's / last-committed refiner's view (OpenReadOnly). Either way
+// a missing or gutted "ready" DB fails the open — never auto-healed into a fresh
+// empty one — wrapped in the uniform won't-open error so every ready-open site
+// reports it identically.
+func OpenReady(
```
tamirms: Optional: the trailing `readOnly bool` makes the call sites read `OpenReady(state, dir, id, logger, false)` — a naked boolean the reader has to look up. Two thin names over an unexported core — `OpenReadyWrite` / `OpenReadyView` (or similar) — would keep the single enforcement site while letting each call site say what it means.

tamirms:
Two more code changes belong in this PR, and they compound:

1. **Fold the metastore package into catalog.** It's a 115-line wrapper over rocksdb with exactly one consumer — the catalog — and folding it was the one #824 item this PR doesn't address (https://github.com/stellar/stellar-rpc/pull/820#discussion_r3514616048). The fold deletes the package, the two-step daemon wiring, and the `metastore.BatchWriter` type threaded through the protocol closures; `NewCatalog(store)` becomes `catalog.Open(path, logger)`.
2. **That fold hands you the rest of #835 for free.** Once the catalog is constructed with a logger, `refineWithHotDB` can use it, the `logger` parameter drops, and `LastCommittedLedger(cat)` lands as the design wrote it. Combined with moving progress.go next to its only caller (other comment), #835 is done — in this PR, in a handful of lines.

With those two, this PR finishes #824 and #835 outright. I'd stop there: the #841 deletions (~300 lines of dead API plus test migrations) are a different kind of change and should be their own PR.

cmd/stellar-rpc/internal/fullhistory/lifecycle/progress.go
chowbao: Done in 8fa6c1a7: `git mv` to the root next to startup.go, `LastCommittedLedger` → unexported `lastCommittedLedger`, and its three test files moved along (the root test helpers already covered the scaffolding, including `openLiveHotDB`). The lifecycle tests that used `deriveCompleteThrough` as a fixture oracle now pass their fixture-known values directly — `runTickForCatalog` collapsed into direct `runLifecycle(…, lastChunk)` calls, so tick inputs are explicit rather than re-derived by the code under test's own family. PR description updated; noted on #835 that its scope shrinks to the logger-parameter change.

cmd/stellar-rpc/internal/fullhistory/config/config_lock.go
chowbao: Done in 8fa6c1a7 — deleted outright rather than renamed: all 19 remaining error strings (config ×2, config_lock ×6, catalog ×6, geometry ×5; the 7th config_lock one was `errLockHeld`, already dropped in this PR), and the four Info/Debug log lines followed the same decision since it's the same mechanical pass. Noted on #841 that its prefix bullet is closed here.

cmd/stellar-rpc/internal/fullhistory/storage/stores/errors.go
chowbao: Fixed in 8fa6c1a7 — now reads "(storage/rocksdb, internal/packfile, os)".

cmd/stellar-rpc/internal/fullhistory/storage/rocksdb/rocksdb.go
chowbao: Fixed in 8fa6c1a7 — "The ingestion contract requires…".

cmd/stellar-rpc/internal/fullhistory/storage/stores/hotchunk/hotchunk.go
chowbao: Done in 8fa6c1a7: `OpenReadyWrite` / `OpenReadyView` over an unexported `openReady` core — the enforcement site stays single, and the three call sites (hotloop resume handle, freeze source, last-committed refiner) now say what they mean without the naked boolean.

chowbao:
Both done in a051a06f (and the progress.go move from the inline thread landed in 8fa6c1a7):

1. **metastore folded into catalog.** The string-KV surface is catalog-internal now (`catalog/kv.go`, unexported), and `NewCatalog(store, …)` became `catalog.Open(path, layout, txLayout, logger)` — the catalog owns its backing store, `Close` releases it. The two-step daemon wiring and the `metastore.BatchWriter` threading through the protocol/sweep closures are gone, and the fold also deleted the `ErrNotFound` manufacture-then-translate round-trip in `Catalog.get` plus `metastore.New`'s guards that duplicated `rocksdb.New`'s. The KV test suite moved to `catalog/kv_test.go`. Net −510 lines.
2. **#835 landed.** `refineWithHotDB` uses `cat.Logger()`, the `logger` parameter dropped, and `lastCommittedLedger(cat)` is the design signature. The `deriveCompleteThrough`/`deriveLastCommitted` shim aliases collapsed into direct calls — the nil-logger seam #835 worried about was already gone (those tests run against real hot DBs).

One test seam changed shape: backfill's skip-when-coverage-frozen test used to seed raw "pruning" chunk keys on the bare store; with no bare store to reach, it now runs the real `SweepChunkArtifacts` — exactly the finalized-window scenario it was simulating.

Agreed on stopping there — the only #841 piece taken here is the error-prefix bullet you flagged inline; the ~300-line dead-API deletions stay a separate PR. PR description updated (`Closes #824` / `Closes #835`).

tamirms (approved):

PR #846, opened 2026-07-09
fullhistory: delete dead pre-wired surface + stale contract docs (closes #841)

### LINKED ISSUE #841: full-history: dead pre-wired surface + stale contract docs (~300 lines deletable)

chowbao:
Closes #841.

Deletes the dead pre-wired surface the freeze rework left behind and corrects the stale contract docs that would misdirect the #772 read-path work. Pure-negative diff plus doc corrections — no behavior change on any production path.

## Deletions

**Freeze-handoff mirror surface.** `ConcurrentBitmaps.Snapshot`/`Len`, `ConcurrentLedgerOffsets.Snapshot`/`EventIDs`, and the Clone-safe/COW ownership contract that existed only for them. `Snapshot`'s doc cited a WriteColdIndex call chain that does not exist — the cold index is built from a single-threaded `Bitmaps` in `ingest/events.go`, and the live freeze re-derives from raw LCMs.

**`Reader.Lookup`.** `Query` uses `LookupKeys` exclusively; `Lookup(k) ≡ LookupKeys([k])[0]`, and `ColdReader.Lookup` hand-duplicated the whole resolve pipeline for one key. Deleting it retires `ErrTermNotFound` — a per-key miss is a nil bitmap from `LookupKeys`, and the closed-store sentinel is unified below.

**`events.TermsFor`.** Test-only; its doc named cold backfill as the consumer, which actually uses `TermsForBytes`.

**Dead `IndexBaseChunk`/`IndexFileName` naming scheme.** `OpenColdReader`'s doc told future wiring to compose a filename that never exists on disk; it now points at the real layout (`geometry.Layout.TxHashIndexFilePath`).

**Census tail.**
- `ColdWriter.ChunkID()`/`Dir()` + backing fields, and the unused `eventstore.ErrWriterClosed` re-export (docs now name `packfile.ErrWriterClosed`).
- The unreachable early-success branch in `ledgerCold.Finalize` + the `appended` latch; a zero-append Finalize now fails loudly via `Commit`, which is the honest shape for driver misuse.
- The unreachable nil/empty guards in `ConcurrentBitmaps.Get` (a published termState always carries a non-nil bm or non-empty ids).
- The redundant closed-store sentinels: `eventstore.ErrClosed` and `txhash.ErrClosed` (no caller distinguished them) are gone; both stores now return `stores.ErrStoreClosed` per the `stores/errors.go` translation contract, matching the ledger store. (The metastore half of this bullet was already resolved by the #835 fold — `catalog.Open` delegates its guards to `rocksdb.New`.)
- The eventstore cold writer no longer mkdirs its own bucket dir; directory creation moved to the ingest layer (`NewEventsColdIngester`), matching the ledger/txhash siblings.
- Tip knobs are no longer double-defaulted: `daemon.go` leaves `StartConfig.TipBackoff`/`TipMaxAttempts` zero for `withDefaults` to fill, and `validateConfig` references the named defaults directly.
- `validateConfig` now returns only `error` — its returned earliest ledger had no consumer (tests read the pin via `cat.EarliestLedger()`).
- Test-only exports: `txhash.ReadColdBin` (tests use an `fhtest` helper / in-package copy), `TxHashIndexLayout.ChunksPerIndex` (in-package tests read `cpi`), `events.NewConcurrentBitmaps` (tests convert from an empty `Bitmaps`), `ledger.ColdReader.FirstSeq`.

## Doc corrections (the #772-misdirection set)

- `Reader.All` no longer claims the freeze loop uses it (the freeze re-derives from raw LCMs); `Reader.Offsets` no longer describes the retired hot-side Snapshot allocation (it's a `View` now).
- The three contract comments naming the deleted `IngestLedgerEvents` as the single writer now name the actual writer: `HotStore.applyLedger` via the post-commit hook (`bitmaps.go`, `concurrent_bitmaps.go`, `concurrent_ledgeroffsets.go`).
- `db/event.go` no longer claims full-history stores no per-event index — it stores `EventIdx` explicitly per the byte-stable-ID decision.
- `rocksdb.CFOptions.BlockSize`: RocksDB's BBTO default is 4 KiB, not 16 KiB — so hot_store's explicit 4-KiB values pin the default rather than override it, and future CF sizing isn't reasoned from a 4× wrong baseline.
- The txhash package doc no longer promises a future "RecSplit-backed" cold reader; it names the existing streamhash-MPHF-backed one.
- `cold_bin.go` credits `scanBinHeader` (the production on-disk guard), and no longer claims duplicate truncated keys are "preserved" — the streamhash build fails on them.
- The three packfile-laziness comments in `eventstore/cold_reader.go` were false: `packfile.Open` starts each open (and holds its fd) in a background goroutine immediately; only the metadata decode is deferred. The ledger store's description was already accurate.
- `TxReader.GetTransaction`'s "lookup incomplete" (rather than not-found) shape for an unverifiable candidate is now documented as deliberate for #772.
- The stale `query.go:248` line ref in `ConcurrentBitmaps.Get`'s borrow-guard note names the guard instead of a line number.

## Notes

- The `streaming:` error/log-prefix bullet was already closed by #844 (commit 8fa6c1a7).
- Tests exercising deleted API were ported to the surviving surface (e.g. warmup mirror comparison now derives expected term keys via `TermsForBytes` and compares per-key lookups); tests that existed only to pin a deleted symbol's own contract were removed.

### DIFF
--- a/cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
+++ b/cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
@@ -164,10 +164,7 @@ func TestConcurrentBitmaps_GetReturnsImmutableSnapshot(t *testing.T) {
 // TestConcurrentBitmaps_ConcurrentGetIsSafe runs many concurrent
 // Get callers against the same store. Get is lock-free past the
 // brief map-lookup RLock and returns an immutable snapshot, so
-// concurrent reads should not race. Snapshot is intentionally NOT
-// exercised here — its contract is single-caller (called only at
-// freeze time after ingest stops), so multi-goroutine Snapshot
-// would be out-of-contract. Run under -race.
+// concurrent reads should not race. Run under -race.
 func TestConcurrentBitmaps_ConcurrentGetIsSafe(t *testing.T) {
 	s := newTestConcurrentBitmaps()
 	const nTerms = 200
--- a/cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
+++ b/cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
@@ -8,8 +8,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// newTestConcurrentBitmaps builds an empty ConcurrentBitmaps via the
+// only remaining constructor (production always converts from a
+// warmup/backfill-built Bitmaps).
+func newTestConcurrentBitmaps() *ConcurrentBitmaps {
+	return NewConcurrentBitmapsFromBitmaps(NewBitmaps())
+}
+
 func TestConcurrentBitmaps_AddToAndGet(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("transfer"), FieldTopic0)
 
 	s.AddTo(key, 0)
@@ -26,7 +33,7 @@ func TestConcurrentBitmaps_AddToAndGet(t *testing.T) {
 }
 
 func TestConcurrentBitmaps_GetMissing(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("missing"), FieldTopic0)
 	bm, err := s.Get(key)
 	require.NoError(t, err)
@@ -34,7 +41,7 @@ func TestConcurrentBitmaps_GetMissing(t *testing.T) {
 }
 
 func TestConcurrentBitmaps_ListMode(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("sparse"), FieldTopic0)
 
 	for i := range uint32(promotionThreshold - 1) {
@@ -58,7 +65,7 @@ func TestConcurrentBitmaps_ListMode(t *testing.T) {
 }
 
 func TestConcurrentBitmaps_Promotion(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("dense"), FieldTopic0)
 
 	for i := range uint32(promotionThreshold) {
@@ -74,7 +81,7 @@ func TestConcurrentBitmaps_Promotion(t *testing.T) {
 }
 
 func TestConcurrentBitmaps_AddAfterPromotion(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("dense"), FieldTopic0)
 
 	for i := range uint32(promotionThreshold) {
@@ -90,77 +97,8 @@ func TestConcurrentBitmaps_AddAfterPromotion(t *testing.T) {
 	assert.True(t, bm.Contains(2000))
 }
 
-func TestConcurrentBitmaps_Len(t *testing.T) {
-	s := NewConcurrentBitmaps()
-	assert.Equal(t, 0, s.Len())
-
-	keyA := ComputeTermKey([]byte("a"), FieldTopic0)
-	keyB := ComputeTermKey([]byte("b"), FieldTopic1)
-
-	s.AddTo(keyA, 0)
-	assert.Equal(t, 1, s.Len())
-
-	s.AddTo(keyA, 1) // same term
-	assert.Equal(t, 1, s.Len())
-
-	s.AddTo(keyB, 2)
-	assert.Equal(t, 2, s.Len())
-}
-
-// TestConcurrentBitmaps_SnapshotIsUniquelyOwned pins that Snapshot
-// returns a Bitmaps whose bitmaps are independent of the source
-// ConcurrentBitmaps. Mutating a snapshot bitmap must not affect a
-// subsequent Snapshot of the same source.
-func TestConcurrentBitmaps_SnapshotIsUniquelyOwned(t *testing.T) {
-	s := NewConcurrentBitmaps()
-	key := ComputeTermKey([]byte("dense"), FieldTopic0)
-	for i := range uint32(promotionThreshold) {
-		s.AddTo(key, i)
-	}
-
-	snap1 := s.Snapshot()
-	bm1 := snap1[key]
-	require.NotNil(t, bm1)
-	bm1.Add(999_999)
-
-	snap2 := s.Snapshot()
-	bm2 := snap2[key]
-	require.NotNil(t, bm2)
-	assert.False(t, bm2.Contains(999_999),
-		"mutating a snapshot bitmap must not bleed into the source ConcurrentBitmaps")
-
-	// And the original index is still clean.
-	live, err := s.Get(key)
-	require.NoError(t, err)
-	assert.False(t, live.Contains(999_999),
-		"snapshot mutation must not affect the live ConcurrentBitmaps")
-}
-
-func TestConcurrentBitmaps_SnapshotIncludesSparseAndDense(t *testing.T) {
-	s := NewConcurrentBitmaps()
-
-	sparseKey := ComputeTermKey([]byte("sparse"), FieldTopic0)
-	denseKey := ComputeTermKey([]byte("dense"), FieldTopic0)
-
-	// Sparse: stays in list mode.
-	s.AddTo(sparseKey, 0)
-	s.AddTo(sparseKey, 1)
-
-	// Dense: promoted to bitmap mode.
-	for i := range uint32(promotionThreshold + 10) {
-		s.AddTo(denseKey, 100+i)
-	}
-
-	snap := s.Snapshot()
-	assert.Len(t, snap, 2)
-	require.NotNil(t, snap[sparseKey])
-	require.NotNil(t, snap[denseKey])
-	assert.Equal(t, uint64(2), snap[sparseKey].GetCardinality())
-	assert.Equal(t, uint64(promotionThreshold+10), snap[denseKey].GetCardinality())
-}
-
 func TestConcurrentBitmaps_BatchAddTo(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("batch"), FieldTopic0)
 
 	s.AddTo(key, 0, 1, 2, 3, 4)
@@ -174,7 +112,7 @@ func TestConcurrentBitmaps_BatchAddTo(t *testing.T) {
 }
 
 func TestConcurrentBitmaps_BatchAddToPromotion(t *testing.T) {
-	s := NewConcurrentBitmaps()
+	s := newTestConcurrentBitmaps()
 	key := ComputeTermKey([]byte("batch-promote"), FieldTopic0)
 
 	// Single batch call that crosses threshold.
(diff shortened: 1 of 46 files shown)


cmd/stellar-rpc/internal/events/concurrent_ledgeroffsets.go
```diff
@@ -114,8 +90,8 @@ func (m *ConcurrentLedgerOffsets) EndLedger() uint32 {
 // View returns a *LedgerOffsets sharing the live backing array,
```
tamirms, line 90: The type comment over in `ledgeroffsets.go` (lines 11 and 18) still names `ConcurrentLedgerOffsets.Snapshot`: once as the freeze-handoff example in the build-then-read list, and once as the recommended conversion path ("call Snapshot to convert when handing off to the cold reader's build-then-read shape"). With Snapshot deleted here, both now point at API that no longer exists, which is the same misdirection class this PR is clearing out. That file isn't in the diff, so I'm flagging it on the deletion instead.

Suggested rewording: drop the Snapshot example from the list (cold backfill and DecodeLedgerOffsets are the two remaining build-then-read producers), and replace the second sentence with something like: "For live ingest paths (HotStore) use ConcurrentLedgerOffsets; queries borrow a read-only view of it via View."

cmd/stellar-rpc/internal/events/index.go
```diff
@@ -50,54 +50,6 @@ func ComputeTermKey(value []byte, field Field) TermKey {
 	return key
 }
 
-// TermsFor returns the 16-byte hashed term keys (contractID + topics
```
tamirms, line 53: One reference to the deleted function survives outside this diff: `eventstore/query.go:47` says "(see events.TermsFor)" in the `Filter` doc. Pointing it at `events.TermsForBytes` instead would keep the topic-count rationale reachable. `query.go` isn't part of the diff, so I can't anchor a comment there; noting it on the deletion instead.

cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore/cold_index.go
```diff
@@ -34,9 +34,9 @@ import (
 // A zero-term bitmaps (an eventless chunk, e.g. a pre-Soroban
 // backfill range) produces the EMPTY index: a zero-length index.hash
 // plus a zero-record index.pack. The cold reader resolves every
-// lookup against it to ErrTermNotFound through the ordinary path, so
-// neither readers nor orchestrators need a pack-without-index special
-// case.
+// LookupKeys entry against it to a nil-bitmap miss through the
+// ordinary path, so neither readers nor orchestrators need a
+// pack-without-index special case.
```
tamirms, line 39: A few lines below this hunk (line 54), the comment still says both builders re-derive terms via "per-event events.TermsFor + Bitmaps.AddTo". With `TermsFor` deleted in this PR, this should now name `events.TermsForBytes`, which is what the cold backfill ingester actually calls (`ingest/events.go`).

cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
```diff
@@ -231,7 +169,7 @@ func TestConcurrentBitmaps_GetReturnsImmutableSnapshot(t *testing.T) {
 // freeze time after ingest stops), so multi-goroutine Snapshot
```
tamirms: This comment still explains why Snapshot is intentionally not exercised by the test, but Snapshot no longer exists after this PR. The sentence starting "Snapshot is intentionally NOT exercised here" can go; the remaining rationale (Get is lock-free and returns an immutable snapshot, run under -race) stands on its own.

cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore/hot_warmup_test.go
```diff
@@ -42,20 +42,33 @@ func TestWarmup_RebuildsMirrorFromIngestedRows(t *testing.T) {
...
+	// Derive the terms each seeded payload contributes (the same
+	// events.TermsForBytes the ingest path indexed by), then snapshot
+	// their pre-close cardinalities straight from the live mirror.
+	var seededKeys []events.TermKey
+	for _, p := range []events.Payload{p1, p2} {
+		keys, err := events.TermsForBytes(p.ContractEventBytes)
+		require.NoError(t, err)
+		seededKeys = append(seededKeys, keys...)
+	}
 	expected := make(map[events.TermKey]uint64)
```
tamirms, line 58: Small observation on the port, not blocking: the old Snapshot-based comparison checked both directions, so it would also have caught warmup inventing terms that were never seeded. This version checks that every seeded term survives with the right cardinality, but a phantom extra term in the reopened mirror would now pass unnoticed. I don't see a way to restore that direction without giving ConcurrentBitmaps term enumeration back, which would defeat the point of the deletion, so maybe just a sentence in the test comment acknowledging the narrower scope.

tamirms (commented):
Verified the deletion claims against the code rather than the PR body: the `ConcurrentBitmaps.Get` invariant holds on every publish path, the zero-append `Finalize` branch is unreachable because `drain` errors unless the full chunk range was consumed, the tip-knob single-defaulting is behavior-preserving (`withDefaults` fills the same constants), and the query path touches none of the deleted surface (`Query` resolves exclusively through `LookupKeys` at `query.go:231`, misses are nil slots, and offsets flow through `Offsets()` -> `View()` -> `LedgerOffsets.EventIDs`). The doc corrections I spot-checked are all accurate, including the packfile.Open background-open description and the 4 KiB RocksDB BBTO default.

The comments below are all one class: a few stale references of exactly the kind this PR exists to remove survived, mostly in files the diff didn't touch. None are blocking on their own, but since the PR's whole purpose is killing misdirection ahead of #772, it seems worth sweeping them in the same pass.

The failing Build / dependency-sanity-checker / complete checks fail identically on #844, so they look pre-existing on the base branch rather than caused by this PR.

cmd/stellar-rpc/internal/events/concurrent_ledgeroffsets.go
chowbao: Done in a83d4a3f — dropped the Snapshot example from the build-then-read list (cold backfill and DecodeLedgerOffsets remain) and replaced the second paragraph with your suggested wording: live ingest uses ConcurrentLedgerOffsets, queries borrow a read-only view via View.

cmd/stellar-rpc/internal/events/index.go
chowbao: Done in a83d4a3f — query.go:47 now points at events.TermsForBytes. Grepped the tree for other surviving `events.TermsFor` references; this and cold_index.go were the only two.

cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore/cold_index.go
chowbao: Done in a83d4a3f — the cold_index.go builder comment now names events.TermsForBytes, matching what the cold backfill ingester actually calls.

cmd/stellar-rpc/internal/events/concurrent_bitmaps_test.go
chowbao: Done in a83d4a3f — removed the stale Snapshot sentence; the comment now ends with the Get lock-free/immutable-snapshot rationale and the -race note.

cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore/hot_warmup_test.go
chowbao: Agreed there's no way to restore the reverse direction without re-adding term enumeration, which would defeat the deletion. Added a sentence to the test comment in a83d4a3f acknowledging the check is one-directional by design (phantom extra terms in the reopened mirror would pass unnoticed).

tamirms (approved):

PR #847, opened 2026-07-09
fullhistory: hardening + contract notes from the pre-existing-code review (closes #840)

### LINKED ISSUE #840: full-history: hardening and contract notes from the pre-existing-code review
Split out from the #820 review's final pass over pre-existing full-history code. None of these is a live bug; each is a cheap guard, a false claim, or a contract that will bite later if left unwritten.

- **Storage roots.** The per-store roots are load-bearing (tiered storage — distinct volume types per data family), so no schema change. Fix: pairwise distinct-and-non-nested validation over the configured roots in `validateConfig` (~15 lines — catches `hot == catalog` and nesting, which today silently puts one store's tree inside another's and lets a retention prune delete it); fix `config_lock.go`'s header, which claims a concurrent-daemon fail-fast that doesn't hold for nested roots (lock dedup is exact-path only); optionally, an ancestor-walk flock probe in `LockRoots` so the second daemon of a cross-daemon nested pair fails fast (for any nesting, the deeper daemon's root has the shallower daemon's lock file at an ancestor).
- **Snapshot-distribution guards** (matter when cold trees get copied between machines): the non-empty index.pack/index.hash pair has no chunk-identity binding (events.pack has one — `eventstore/cold_reader.go:587`), so a misrouted index silently returns an incomplete subset of matches; and `loadMeta` never cross-checks `offsets.TotalEvents()` against the trailer item count, so a mispaired offsets blob silently truncates tail events. One cross-check each, at load time.
- **eventstore never verifies `Trailer().Format`** — `cold_format.go:70-78` claims readers dispatch on it, and the ledger store actually checks (`ledger/cold_reader.go:83`); a wrong-format pack currently fails mid-query with an opaque zstd error instead of loudly at open.
- **`Query`'s `MaxEvents` cap is applied before `postFilter`** (`eventstore/query.go:322-335`), so a short page does not mean exhaustion — a false positive on a capped page shrinks it below the limit while in-range matches remain. Whoever wires getEvents (#772) must not use the `len < limit ⇒ done` idiom; document the contract or add a top-up loop.
- **`InsertEvents`' deferred close-error join can never propagate** (`db/event.go:77,92-95` — the deferred `err = errors.Join(...)` assigns a local against an unnamed `error` return). Latent today (the SDK reader's Close returns nil unconditionally); a named return fixes it.
- **The warmup tamper tripwire wraps at `eventID == MaxUint32`** (`eventstore/hot_store.go:631`): `indexUpperBound = eventID+1` wraps to 0 and passes the `> total` check — the exact adversarial input the check exists for. Evasion of the tripwire only; queries stay clipped by the pinned count.
- **The two backends diverge on a future ContractEvent body version**: SQLite hard-fails the whole ledger (`db/event.go:263`) while full-history indexes contractID-only and continues (`events/index.go:84`). Neither is wrong alone; pick one behavior before a protocol bump picks for us.
- **`ColdReader.Close` does synchronous events.pack metadata I/O for eventless chunks** — the common case across early history — with the error discarded (`eventstore/cold_reader.go:198-219`). The sentinel validation belongs on the Lookup path. (Related open-cost territory: #834.)
- Nit: `FetchRange`'s count==0 short-circuit contradicts its out-of-bounds doc (`eventstore/reader.go:174` vs both implementations).

chowbao:
Closes #840. Hardening from the pre-existing-code review: cheap load-time guards, corrected doc claims, and contracts written down before #772 builds on them. No schema or on-disk format changes.

## Storage roots

- **Per-root flocks dropped** (review round): single-process enforcement is `catalog.Open`'s own exclusive RocksDB LOCK — a second daemon on the same catalog fails to start. Cold artifact trees are write-once behind the one-write protocol (a shared tree is redundant work, not corruption) and each hot chunk DB carries its own RocksDB LOCK. `LockRoots`' load-bearing other half survives as `PrepareRoots`: create missing roots and fsync the new directory chain, so a fresh-deploy crash cannot lose a tree the synced catalog already references. `RootsToLock` → `Roots`.
- `Paths.ValidateRoots()`: distinct roots only — two roots at the same path would write into each other's trees (resolved via `filepath.Abs`, so unclean-spelling duplicates are caught). Nesting is deliberately allowed: prune and discard delete per-chunk paths, never a whole root.

## Snapshot-distribution guards (cold trees copied between machines)

One load-time cross-check each, mirroring the chunk-ID binding events.pack already had:

- `loadMeta` cross-checks `offsets.TotalEvents()` against the trailer item count — a mispaired offsets blob otherwise silently truncates tail events off every per-ledger range.
- index.hash's MPHF key count must equal index.pack's record count (halves of one build), and a non-empty index on an eventless events.pack is rejected (converse of the empty-index check). The pair carries no chunk ID of its own, so these are the strongest bindings available without a format change.
- Both eventstore packs verify `Trailer().Format` at open — `cold_format.go` claimed readers dispatch on it, but only the ledger store checked; a mis-pointed pack now fails at open instead of mid-query with an opaque zstd error.
- The MPHF handle is separate from its verdict (review round): `waitMPHF` loads, `validateMPHF` is an error-only gate the lookup path runs — `Close` always gets the real handle to release, and an eventless chunk's teardown (the common case across early history) does no events.pack metadata I/O.

## Contracts and point fixes

- `Query`'s `MaxEvents`-before-`postFilter` contract documented at `QueryOptions` and the cap site: a short page does not mean exhaustion, so the #772 getEvents wiring must not use `len < limit ⇒ done`. Documented rather than a top-up loop — the consumer doesn't exist yet.
- `InsertEvents`: named error return so the deferred close-error join reaches the caller (previously assigned a dead local).
- The warmup tamper tripwire no longer wraps at `eventID == MaxUint32`: the index upper bound is uint64, so the exact adversarial input the check exists for can't wrap `max+1` to 0 and slip past.
- **`TermsForBytes` now hard-fails a non-V0 ContractEvent body**, matching the SQLite backend, instead of silently indexing contractID-only (topic queries would miss real events with no signal). Flagging for review — the old behavior was defensible as forward-compat; this picks the loud option before a protocol bump picks for us.
- `FetchRange`'s doc matches both implementations: `count == 0` is a no-op regardless of `start`.

Tests: `ValidateRoots` + `PrepareRoots` tables, the five cold-reader guard rejections, the MaxUint32 tripwire row, the body-version hard-fail (raw-XDR discriminant patch — the SDK can't marshal a body the protocol doesn't have yet), and the daemon contention test now pins the catalog's RocksDB-LOCK failure.


chowbao:
~TODO: update when https://github.com/stellar/stellar-rpc/pull/858 is merged in~

Edit: Done

chowbao:
@codex[agent] review

cmd/stellar-rpc/internal/fullhistory/config/config.go
```diff
@@ -247,6 +249,53 @@ func (p Paths) RootsToLock() []string {
...
+			}
+		}
+	}
+	return nil
+}
+
+// isAncestorDir reports whether inner lives strictly inside outer. Both
+// must already be absolute, cleaned paths (filepath.Abs does both).
+func isAncestorDir(outer, inner string) bool {
+	return strings.HasPrefix(inner, outer+string(filepath.Separator))
```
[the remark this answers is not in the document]
chowbao: Fixed in e61214db: isAncestorDir now only appends the separator when the cleaned path doesn't already end in one (true only for the filesystem root), so `ledgers = "/"` is rejected. Added a `filesystem root nests everything` case to the ValidateRoots table.

cmd/stellar-rpc/internal/fullhistory/config/config_lock.go
```diff
@@ -14,7 +14,7 @@ import (
 // Single-process enforcement (design "Single-process enforcement"). The daemon
 // holds an exclusive OS lock on a LOCK file under EVERY independently configurable
 // storage root — the catalog, each immutable artifact tree, AND the
-// hot-storage tree. A second daemon touching any shared root fails fast.
+// hot-storage tree. A second daemon pointing at the SAME root path fails fast.
```
tamirms: Do we need the per-root flocks? RocksDB already takes an exclusive lock when `catalog.Open` opens the catalog read-write, so a second daemon on the same catalog already fails to start on its own. The flocks only add coverage for the cold artifact trees, and those are write-once: two daemons freezing the same chunk write identical bytes and the atomic rename resolves it, so a shared cold tree is redundant work, not corruption. Suggest dropping them and leaning on the catalog lock.

Same for `ValidateRoots`: the prune is per-chunk (`eligiblePruneOps` deletes `{root}/{chunkID}/`, never a whole root), so a nested root is harmless. The only config that loses data is two roots at the same path, so a one-line distinct-roots check is enough and `isAncestorDir`/`rootNames`/the nesting cases can go.

cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore/cold_reader.go
```diff
@@ -161,24 +165,61 @@ func OpenColdReader(chunkID chunk.ID, bucketDir string, opts ColdReaderOptions)
 		m, err := openMPHF(indexHashPath)
 		ch <- mphfResult{idx: m, err: err}
 	}()
-	c.waitMPHF = sync.OnceValues(func() (*mphf, error) {
+	c.waitMPHFRaw = sync.OnceValues(func() (*mphf, error) {
```
tamirms: The raw/validated split works, but the leak is avoided by convention: `waitMPHF` returns `(nil, err)` on a validation failure, so `Close` has to remember to use `waitMPHFRaw` or it would skip `m.Close()` and leak the handle. Cleaner to separate the handle from the verdict: one load accessor (`waitMPHF`, no validation) plus an error-only gate (`validateMPHF`). Then `Close` always gets the real handle and a nil can't reach it.

chowbao: Done in 14603d1b: `waitMPHF` is now the validation-free load accessor and `validateMPHF` is a `sync.OnceValue` error-only gate; `Close` always gets the real handle, and `LookupKeys` runs the gate before using it.

cmd/stellar-rpc/internal/fullhistory/config/config_lock.go
chowbao: Done in 7cd54162. Verified both premises before cutting: the catalog is RocksDB (`catalog.go:42`) so a second daemon fails at `catalog.Open` on its exclusive LOCK, and every whole-directory deletion is per-chunk (`BeginHotCreate`/`DiscardHotChunk` remove `HotChunkPath(chunkID)` only) — nothing ever removes a root. Flocks are gone; `LockRoots`' load-bearing other half (create missing roots + fsync the new dir chain, the fresh-deploy crash-safety step) survives as `PrepareRoots`, and `RootsToLock` is renamed `Roots`. `ValidateRoots` is now the one-line distinct-roots check (still on resolved absolute paths so unclean-spelling duplicates are caught); nesting cases, `isAncestorDir`, and `rootNames` are deleted, and the daemon contention test now pins the RocksDB-LOCK failure instead of `ErrRootLocked`.

cmd/stellar-rpc/internal/fullhistory/config/config.go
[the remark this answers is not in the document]
chowbao: Superseded by 7cd54162: nesting detection (including this helper) is deleted entirely — nesting is now allowed since no operation removes a whole root; ValidateRoots only rejects duplicate roots. See the config_lock.go thread.

cmd/stellar-rpc/internal/fullhistory/config/config_lock.go
chowbao: `PrepareRoots` seems good for creating new dirs/roots on startup

cmd/stellar-rpc/internal/fullhistory/config/config_roots.go
```diff
@@ -0,0 +1,52 @@
...
+	seen := make(map[string]struct{}, len(roots))
+	for _, root := range roots {
+		if root == "" {
+			continue
+		}
+		abs, err := filepath.Abs(root)
+		if err != nil {
+			return fmt.Errorf("resolve storage root %q: %w", root, err)
+		}
+		if _, dup := seen[abs]; dup {
```
tamirms: `PrepareRoots`'s `seen` dedup is redundant: `ValidateRoots` runs first in `runDaemonWith` and already rejects any two roots resolving to the same path, so this loop never sees a duplicate. `MkdirAll` is idempotent anyway, so a repeat would be harmless. Could drop it, minor.

tamirms (approved):
LGTM. Verified locally: the tree builds, `go vet` is clean, and the config, eventstore, daemon, and full e2e suites pass, including the new second-daemon test confirming the RocksDB catalog lock preserves single-process enforcement. One minor nit inline.

cmd/stellar-rpc/internal/fullhistory/config/config_roots.go
chowbao: Dropped in 4652e378 — the loop is now a plain range with the empty-root skip; the test's repeated-root case stays as a harmless-no-op check.

PR #849, opened 2026-07-10
fullhistory: scope the txhash tuning knobs to the txhash CF (#838)

### LINKED ISSUE #838: fullhistory: scope the txhash Tuning knobs to the txhash CF (per-CF options in rocksdb.CFOptions)
Split out from review of #820 (thread on `hotchunk.go`: https://github.com/stellar/stellar-rpc/pull/820#discussion_r3511861048).

The design says per-CF options keep tuning independent, but the shared hot-chunk config applies txhash's `Tuning` — 12-bit bloom filters, 64 MB write buffers, its compaction triggers — to the ledgers and events CFs too; eventstore's overrides reclaim only compression and block size. The ledgers CF is never probed for missing keys, so it pays for bloom filters nothing consults, and the write buffers are sized for txhash's write pattern rather than its own.

Scoping the `Tuning` fields to only the txhash CF needs the shared `rocksdb.CFOptions` to grow those per-CF knobs (it currently carries only compression + block size) and touches the shared block-cache / filter allocation in `applySharedTableOptions` — agreed in the thread to do this as its own change rather than piecemeal in the shared rocksdb layer (https://github.com/stellar/stellar-rpc/pull/820#discussion_r3514615299).

chowbao:
Closes #838.

The shared hot-chunk config applied txhash's `Tuning` (12-bit bloom, 64 MB write buffers, no-compaction triggers) to the ledgers and events CFs too. Per the issue's sharpened scope, this is a mechanical port of each store's pre-unification standalone tuning (verified against `c445557d^`).

**Commit 1** moves the memtable/compaction/bloom knobs from DB-wide `Tuning` into per-CF `rocksdb.CFOptions` (each still zero-means-default). `Tuning` keeps only the genuinely DB-wide resources: block cache, WAL cap, background jobs, open-file cap. The bloom filter is still constructed fresh per CF, now only for CFs that opt in.

**Commit 2** scopes the values:

| CF | after (== pre-unification standalone) |
|---|---|
| txhash | keeps everything: bloom 12, 64 MB × 2 memtables, no-compaction triggers, 64 MB target files |
| ledgers | RocksDB defaults, no bloom |
| events | own ZSTD + block-size overrides only, no bloom |

The DB-wide values (512 MB cache, 1 GB WAL, 8 jobs, 10k files) are now defined in `hotchunk.dbTuning()` — the DB owner configures the DB, each facade configures only its own CFs; `txhash.Tuning()` is gone. The values keep the standalone txhash calibration, the only pre-unification instance that set them.

Tests: `TestConfig_PerCFOptionRouting` and `TestConfig_DBWideTuningStaysShared` pin the routing and shared values at the config seam. Full `fullhistory` `-short` suite and `-race` on `rocksdb`/`hotchunk` pass.

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
+++ b/cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
@@ -94,12 +94,7 @@ func CFOptions() map[string]rocksdb.CFOptions {
 			// 64 MB target file matches WriteBufferMB so one memtable
 			// flush produces one ~64 MB SST — fewer bloom checks per
 			// query at no-compaction scale.
-			// MaxBytesForLevelBaseMB is set explicitly even though it's
-			// irrelevant under DisableAutoCompactions (compaction never
-			// promotes past L0); explicit > implicit so a future reader
-			// doesn't have to derive that it's a no-op.
-			TargetFileSizeMB:       64,
-			MaxBytesForLevelBaseMB: 256,
+			TargetFileSizeMB: 64,
 
 			// 12 bits/key bloom (~0.4% false-positive) is tighter than
 			// the standard 10 bits/key because every false positive at
--- a/cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
+++ b/cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
@@ -44,42 +44,14 @@ func NewWithStore(store *rocksdb.Store) *HotStore {
 // the hotchunk shared-DB opener can register it alongside the other CFs.
 func CFNames() []string { return []string{txhashCF} }
 
-// Tuning returns this facade's RocksDB tuning, applied to the shared per-chunk
-// DB by the hotchunk opener. The hot txhash workload is write-once /
-// point-lookup; the cross-knob interactions below are non-obvious enough that
-// they get an explicit per-stanza rationale. The other facades ride on RocksDB
-// defaults by contrast — only this workload earned the calibration.
+// Tuning returns the DB-wide RocksDB knobs the shared per-chunk DB adopts —
+// resources that were per-instance before the hot stores collapsed into one
+// multi-CF DB and are now shared across every CF. The txhash instance was the
+// only one that set them; the ledger and events instances rode on RocksDB
+// defaults, so keeping txhash's values is the faithful consolidation of the
+// three instances' combined footprint.
 func Tuning() rocksdb.Tuning {
 	return rocksdb.Tuning{
-		// 64 MB memtable so one flush produces one ~64 MB SST under
-		// uniform writes.
-		WriteBufferMB:        64,
-		MaxWriteBufferNumber: 2,
-
-		// L0 triggers pinned high + DisableAutoCompactions=true:
-		// compaction would re-write the same data with no reordering
-		// benefit (txhash is write-once, random-key, point-lookup).
-		// The L0 999s match DisableAutoCompactions so even if a future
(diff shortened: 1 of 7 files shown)


cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
```diff
@@ -100,6 +66,50 @@ func Tuning() rocksdb.Tuning {
 	}
 }
 
+// CFOptions returns the calibration the hotchunk opener installs on the
+// txhash CF alone. The workload is write-once / point-lookup; only it
+// earned this tuning — the other CFs ride on RocksDB defaults.
+func CFOptions() map[string]rocksdb.CFOptions {
```
chowbao, line 50: Are there any options worth removing from here or `tuning.go`?

Note actual values used for RPC v2 can be updated/corrected when full v2 benchmarking is run

chowbao:
@codex[agent] review

tamirms (changes requested):
The per-CF split looks right. One thing I would like addressed in this PR before it merges.

`hotchunk.config()` now assembles the whole `rocksdb.Config`, pulling the per-CF options from each facade that owns those CFs (`eventstore.CFOptions()` and `txhash.CFOptions()`). But it still sources the DB-wide half from a single tenant:

```go
Tuning: txhash.Tuning(),
```

The `Tuning` fields (block cache, WAL cap, background jobs, open files) describe the shared multi-CF DB that `hotchunk` owns, not the txhash facade. The catalog, which is the other RocksDB database here, already configures its own DB inline rather than borrowing from a store:

```go
store, err := rocksdb.New(rocksdb.Config{Path: path, Logger: logger})
```

The tell is change-locality: if getEvents later wants a larger block cache, you would be editing `txhash.Tuning()` to size a cache for an events-driven reason. Some of these values are already not txhash-specific anymore, since the background jobs now drive the ledger and events compaction this PR re-enables, and the cache serves every CF.

This is the PR that closes #838, whose whole point is to stop txhash from configuring things that are not its CF, and the shared DB is the largest such thing. As it stands the change fixes the per-CF half of that coupling but leaves the DB-wide half in place, so it reads as half-done against its own goal. I would define the DB-wide `Tuning` in `hotchunk` (in or beside `config()`, with a comment crediting the txhash calibration as its origin), keep `txhash.CFOptions()`, and drop `txhash.Tuning()`. Then the rule is uniform: each facade configures its own CFs, and the DB owner configures the DB, matching the catalog.

chowbao:
Addressed in 5b018a08: the DB-wide `Tuning` now lives in `hotchunk.dbTuning()` beside `config()` (with the txhash-calibration origin credited in its doc), and `txhash.Tuning()` is dropped. Each facade now configures only its own CFs and the DB owner configures the DB, matching the catalog. `TestConfig_DBWideTuningStaysShared` pins the values at the new seam.

cmd/stellar-rpc/internal/fullhistory/storage/stores/txhash/hot_store.go
chowbao: Two removals landed since: `MaxBytesForLevelBaseMB` (a documented no-op under `DisableAutoCompactions`) was dropped from `CFOptions` end-to-end in 4e0c912e, and `txhash.Tuning()` itself is gone in 5b018a08 (the DB-wide knobs moved to `hotchunk.dbTuning()`, per the review). The remaining fields are all live: each is consumed by `applyCFOverride`/`applyDBTuning` and set by at least one facade. Value re-calibration deferred to v2 benchmarking as noted.

tamirms (approved):

Issue #839, opened 2026-07-04
full-history daemon: readiness and health signals from the last committed ledger


tamirms:
Split out from review of #820 (threads https://github.com/stellar/stellar-rpc/pull/820#discussion_r3517545236, https://github.com/stellar/stellar-rpc/pull/820#discussion_r3521089759).

The streaming daemon needs two operational signals, both derivable from state it already tracks:

**Readiness** — the pipeline has proven itself once: startup completed and the ingestion loop has committed its first ledger. A misconfigured deployment (broken hot tier, bad core binary or config — exec-time failures included) never reports ready, so the rollout fails at deploy time, while the failure is still attributable to the deploy. This replaces the design's original "open hot DB + start core before serving" ordering: #820 landed the hot-DB half (cheap — the open was needed anyway), and the core half is deliberately not gated, because holding reads behind core boot and catchup after every supervised restart trades away availability that local data can serve correctly. Readiness carries the deploy-time signal instead, with no SDK `Start` hook.

**Health** — the pipeline is still moving: the age of the last committed ledger against a threshold (the existing `getHealth` semantic). This catches what no startup gate can: config that rots mid-life, a permanently failing source, a wedged lifecycle (boundaries advancing while ticks stall — `chunk_boundaries_total` vs `phase_duration_seconds{phase="discard"}`). The ingestion loop already owns the per-ledger `last_committed_ledger` gauge, which is the same signal.

Useful from the first ingest-only deployment, before any reads exist. When #772 cuts reads over to the v2 stores, `getHealth` should answer from the same signal — the last committed ledger.

PR #851, opened 2026-07-10
fullhistory: readiness/health signal from the last committed ledger (#839)

chowbao:
Progresses #839 (the HTTP surface is deferred to #772).

The daemon gains its readiness/health **signal**, derived from state the ingestion loop already owns — the last committed ledger — held as one atomic close-time (0 = none yet), fed from the same per-ledger commit that sets the `last_committed_ledger` gauge. No new metrics, no new tracking machinery, no new config.

**The signal** — `healthState`, fed by the ingestion loop per commit:
- **Readiness** latches true on the first committed ledger ("proven itself once"). A misconfigured deployment — broken hot tier, bad core binary or config, exec-time core failures — never commits, so it never reports ready.
- **Last-commit close time** is the raw input for the health/staleness check (the v1 `getHealth` close-time-vs-now semantic). The judgment itself (`time.Since(closeTime)` vs a latency threshold) is check policy and deliberately does NOT live here.

**The seam** — `HealthSignal` (`Ready() bool`, `LastCommitClose() (time.Time, bool)`), created in `runDaemonWith` outside the supervised run loop so readiness stays latched across restarts. When #772 builds the read server, it serves the checks from this signal (as handlers or methods, the way v1's `getHealth` is a method on the main server) and picks the surface topology then — per review, landing an HTTP contract plus a second listener now would pre-empt that decision.

Tests: readiness latching (including staying latched on stale close times), nil-safety of the loop-side hook, and the ingestion loop feeding the signal from real commits (drives `runIngestionLoop` with real ledgers). Full `fullhistory` `-short` suite green; health tests `-race` clean; golangci clean vs base.

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/daemon.go
+++ b/cmd/stellar-rpc/internal/fullhistory/daemon.go
@@ -90,6 +90,12 @@ func runDaemonWith(ctx context.Context, configPath string, opts daemonOptions) e
 		}
 	}
 
+	// Readiness/health signal, fed by the ingestion loop per commit; both signals
+	// derive from the last committed ledger. Created outside the supervised run
+	// loop so it survives restarts (readiness stays latched across them).
+	// TODO(#772): serve it from the read server (as HealthSignal).
+	hs := &healthState{}
+
 	paths := cfg.ResolvePaths()
 
 	// --- Reject shared roots, then create + fsync any missing ones. Single-
@@ -180,7 +186,7 @@ func runDaemonWith(ctx context.Context, configPath string, opts daemonOptions) e
 
 	// --- Assemble the StartConfig and run the supervised run loop. ---
 	start := startConfig(
-		cfg, cat, logger, backend, core, serveReads, metrics, sink)
+		cfg, cat, logger, backend, core, serveReads, metrics, sink, hs)
 
 	backoff := opts.RestartBackoff
 	if backoff <= 0 {
@@ -195,7 +201,7 @@ func runDaemonWith(ctx context.Context, configPath string, opts daemonOptions) e
 func startConfig(
 	cfg config.Config, cat *catalog.Catalog, logger *supportlog.Entry,
 	backend backfill.Backend, core CoreOpener, serveReads func(context.Context) error,
-	metrics observability.Metrics, sink ingest.MetricSink,
+	metrics observability.Metrics, sink ingest.MetricSink, hs *healthState,
 ) StartConfig {
 	exec := backfill.ExecConfig{
 		Catalog:    cat,
@@ -213,6 +219,7 @@ func startConfig(
 		RetentionChunks: deref(cfg.Retention.RetentionChunks),
 		Core:            core,
 		ServeReads:      serveReads,
+		health:          hs,
 	}
 }
 
--- a/cmd/stellar-rpc/internal/fullhistory/health.go
+++ b/cmd/stellar-rpc/internal/fullhistory/health.go
@@ -1,0 +1,63 @@
+package fullhistory
+
+import (
+	"sync/atomic"
+	"time"
+)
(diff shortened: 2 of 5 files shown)


tamirms:
I think this is doing more than #839 needs right now, and the extra part is the HTTP surface. The valuable, durable piece is the signal: `healthState` derived from the last committed ledger, fed by the ingestion loop. The premature piece is turning that signal into an HTTP contract, `newProbeMux`/`writeProbe`/`startProbeServer` on a second listener with its own `[health]` config.

The reason I would hold the surface is that there is no RPC read server yet, `serveReads` is still a `TODO(#772)` no-op, so this stands up a second HTTP server purely for probes. When #772 builds the read server, readiness and health want to live on it as handlers or methods, the way v1's `getHealth` is a method on the main server. So the separate listener and its `[health].endpoint` are scaffolding that gets migrated or deleted, and landing a `/ready` `/health` contract plus a second port now pre-empts a topology decision that belongs to #772.

The PR's argument for the separate server is real: liveness has to answer during the initial backfill, which can run for days, and health on a post-backfill server could not do that without killing the long first backfill. But that is a question about where the always-on health surface lives, which is #772's to settle, and the `healthState` mechanism supports whichever topology it picks. So the sequencing I would prefer is: land the signal now, decide the surface with the server.

Concretely, I would reduce this to:
- Keep `healthState` (`observe`/`ready`/`lastCommitClose`) and the ingestion-loop feed. This is the mechanism, and it exposes exactly the raw signal, the readiness latch and the last close time, while the staleness judgment (the `time.Since(closeTime) > maxLatency` comparison) is check policy that lives with #772.
- Expose it as the plug-in seam, a small `HealthSignal` interface (`Ready() bool; LastCommitClose() (time.Time, bool)`) or a daemon accessor, so #772 wires it into the read server. Otherwise the read methods have no production caller yet and read as dead.
- Drop `newProbeMux`, `writeProbe`, `startProbeServer`, the probe-server block in `runDaemonWith`, the `[health]` config (both `endpoint` and `max_ledger_latency`), and the httptest handler tests. Keep the latch, nil-safety, and loop-feed tests.

Separately from scope, this branch is off pre-#850 feature/full-history (it still carries `networkTip`/`NetworkTipBackend`), so it will need a rebase onto the current tip machinery.

chowbao:
Done in 5f7186a (and rebased onto current `feature/full-history`):

- **Dropped the HTTP surface**: `newProbeMux`, `writeProbe`, `startProbeServer`, the probe block in `runDaemonWith`, the whole `[health]` config (endpoint + `max_ledger_latency`), and the httptest handler tests. The staleness comparison went with it — that judgment is check policy for #772.
- **Kept the signal**: `healthState` (`observe`/`Ready`/`LastCommitClose`) and the ingestion-loop feed, plus the latch, nil-safety, and loop-feed tests.
- **Added the seam**: a `HealthSignal` interface (`Ready() bool; LastCommitClose() (time.Time, bool)`); `runDaemonWith` builds the `healthState` outside the supervised run loop (so readiness stays latched across restarts) and a `TODO(#772)` marks where the read server picks it up.

On the rebase: #850 is still open, so the branch is now on today's `feature/full-history` tip (6974c13); it'll pick up the new tip machinery whenever #850 lands.

Updated the PR title/body to match the reduced scope.

tamirms (approved):

chowbao:
Rebased onto `feature/full-history` now that #850 is merged (7bd1fee). One adaptation: #850 removed `networkTip`/`NetworkTipBackend` from `startConfig`, so the health wiring now rides the new signature — `healthState` itself is untouched. Also squashed the branch to a single commit matching the reduced scope. Full `fullhistory` + `config` `-short` suites green, health tests `-race` clean, golangci clean vs base.

PR #852, opened 2026-07-10
fullhistory: halve cold per-ledger extraction with one ExtractLedgerEvents walk (#836)

### LINKED ISSUE #836: fullhistory: halve cold per-ledger extraction (one ExtractLedgerEvents walk) + thin the ColdIngester seam
Follow-up from PR #820 review (comment 3516278906).

#820 part 1 halved the HOT per-ledger extraction to a single `ExtractLedgerEvents` walk. The same optimization applies to the COLD path, with a bigger payoff — backfill runs it across the whole history on the CPU-parallel worker pool — but `txhashCold` and `eventsCold` still run the two-walk pattern per ledger (`ExtractTxHashes` + `LCMViewToPayloads`→internal `ExtractLedgerEvents`).

Part 1's factoring already exports the pieces: extract once in `ColdService` (or the drain), feed txhash entries from each element's `.Hash` and events via `events.PayloadsFromLedgerEvents`; the ledgers writer never needed the walk.

Deletes the wrapper stack that exists only for today's shape: `eventPayloads`, `LCMViewToPayloads` (this would be its last production caller), and `txhashCold`'s extraction block.

Knock-ons (why it's its own PR — it reshapes the cold ingester seam + metrics):
- The `ColdIngester` seam thins toward per-type writers, converging with the hot shape.
- Cold `stageExtract` gets the same ledger-scoped treatment as hot's extract phase.

Pure optimization/refactor — the cold artifacts and event-ID assignment are unchanged (same `PayloadsFromLedgerEvents` ordering).

chowbao:
Closes #836.

The cold path ran two per-ledger walks over the same borrowed view — `ExtractTxHashes` in the txhash ingester and `LCMViewToPayloads` (internal `ExtractLedgerEvents`) in the events ingester. Backfill runs this across all of history on the CPU worker pool, so the redundant walk was pure waste.

**One walk.** `ColdService.Ingest` now extracts each ledger once (`ExtractLedgerEvents`), mirroring the hot path's `IngestLedger`, and hands the shared result to the per-type ingesters as a `ledgerData` bundle: txhash reads each element's `.Hash`, events shapes the slice with `events.PayloadsFromLedgerEvents`, the ledgers writer takes the raw bytes with no walk. The substitution is exact: both SDK extractors iterate the same `TxProcessing()` and take `txProcessingHash` per element — one element per transaction, event-less or not, V0–V4 alike — so the `.Hash` read is byte-identical, in the same order, to the deleted `ExtractTxHashes` call. A ledgers-only chunk skips the walk entirely.

**Byte-identity is proven, not asserted:** `TestWriteColdChunk_ByteIdentity_SharedWalk` drives the real `WriteColdChunk` over a chunk with event-rich sentinel ledgers and compares the artifacts to independent references — the txhash `.bin` against sorted/truncated `ExtractTxHashes` output (the OLD extractor, a separate walk), and the events term bitmaps + count against `PayloadsFromLedgerEvents` shaping with chunk-relative IDs. Same tx-hash entries, same event-ID assignment.

**Seam thinning.** `ColdIngester.Ingest` takes the pre-extracted `ledgerData` instead of the raw view; the shared walk is metered once per ledger as a ledger-scoped extract stage (`data_type="shared"`, mirroring the hot path's single extract phase) instead of once per consuming type, and the cold stage pair set shrinks 8→7.

**Wrapper stack deleted:** `events.LCMViewToPayloads` (this was its last production caller; its tests keep a local helper) and `ingest.eventPayloads`.

**Writer tuning wired.** Both cold pack writers opt into batch tuning — `Concurrency: 4`, `BytesPerSync: 1 MiB` — instead of the zero-value serial defaults, per the writer's own docs ("bump for large backfills where zstd encoding is CPU-bound"; background writeback smooths the final fsync). `WriteColdChunk` is their sole production caller and is always a batch freeze/backfill. Concurrency stays modest because the backfill pool already runs GOMAXPROCS chunks concurrently; the construction site notes the downgrade path if profiling ever shows contention.

Tests: the byte-identity guard above plus the existing ingest suite migrated to the new seam. Full `fullhistory` + `events` `-short` suites green; `-race -count=1` on `ingest`/`events` green.

Behavioral note: a mid-walk extraction failure now surfaces from `ColdService` before any ingester runs, so it no longer bumps a per-`data_type` `cold_errors_total` — the error still propagates and `ColdChunkTotal` still fires, matching how the hot path attributes decode failures to its data-type-less extract phase. No test asserted the old attribution.

### DIFF
--- a/cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
+++ b/cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
@@ -148,7 +148,9 @@ func WriteColdChunk(
 		return berr
 	}
 	logger.Debugf("cold ingest chunk %d [%d, %d]", uint32(chunkID), chunkID.FirstLedger(), chunkID.LastLedger())
-	service := NewColdService(ings, sink)
+	// txhash and events share the single ExtractLedgerEvents walk; a ledgers-only
+	// chunk needs no walk at all.
+	service := NewColdService(ings, sink, cfg.Txhash || cfg.Events)
 	defer func() {
 		if cerr := service.Close(); cerr != nil {
 			err = errors.Join(err, fmt.Errorf("close: %w", cerr))
--- a/cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
+++ b/cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
@@ -13,30 +13,16 @@ import (
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/storage/chunk"
 )
 
-// closeColdAll closes every cold ingester built so far, joining each Close error
-// into err. Used when a LATER constructor fails mid-build. The already-built
-// ingesters never ingested or finalized, and Close emits no per-ingester
-// ColdIngest sample, so a rolled-back build produces no phantom-success sample — no
-// abort bookkeeping needed here.
-func closeColdAll(ings []ColdIngester, err error) error {
-	for _, ing := range ings {
-		if cerr := ing.Close(); cerr != nil {
-			err = errors.Join(err, fmt.Errorf("close: %w", cerr))
-		}
-	}
-	return err
-}
-
-// drain feeds each of the chunk's raw ledgers (as a borrowed view) to the
-// service on a local sequence counter, then verifies the full [first,last] range
-// was consumed — for cold this runs before Finalize, so a short stream never
-// finalizes a truncated artifact. The in-order contract is enforced at the SOURCE
+// drain feeds each of the chunk's raw ledgers (as a borrowed view) to the cold
+// chunk on a local sequence counter, then verifies the full [first,last] range
+// was consumed — this runs before finalize, so a short stream never finalizes a
+// truncated artifact. The in-order contract is enforced at the SOURCE
 // (packStream reads positionally by key; hotLedgerStream key-checks its own
 // keyspace; the SDK backends validate their own output), so drain trusts the
 // counter rather than re-parsing every view's sequence. Cancellation is the
 // iterator's job (RawLedgers errors on a canceled ctx), so there is no ctx poll
 // here.
-func drain(ctx context.Context, ledgers iter.Seq2[[]byte, error], chunkID chunk.ID, svc *ColdService) error {
+func drain(ledgers iter.Seq2[[]byte, error], chunkID chunk.ID, cc *coldChunk) error {
 	first, last := chunkID.FirstLedger(), chunkID.LastLedger()
 	seq := first
 	for raw, serr := range ledgers {
@@ -49,7 +35,7 @@ func drain(ctx context.Context, ledgers iter.Seq2[[]byte, error], chunkID chunk.
 			return fmt.Errorf("ingest: stream for chunk %d yielded a ledger past %d (chunk overrun)",
 				uint32(chunkID), last)
 		}
-		if err := svc.Ingest(ctx, seq, xdr.LedgerCloseMetaView(raw)); err != nil {
+		if err := cc.ingest(seq, xdr.LedgerCloseMetaView(raw)); err != nil {
 			return err
 		}
 		seq++
@@ -72,50 +58,20 @@ type ColdDirs struct {
 	EventsDir  string
 }
 
-// buildColdIngesters opens one ColdIngester per enabled type at its resolved path.
-// Single definition site of the ctor table, order, and rollback.
-func buildColdIngesters(dirs ColdDirs, chunkID chunk.ID, sink MetricSink, cfg Config) ([]ColdIngester, error) {
-	ctors := []struct {
-		enabled  bool
-		dataType string
-		path     string
-		open     func(string, chunk.ID, MetricSink) (ColdIngester, error)
-	}{
-		{cfg.Ledgers, dataTypeLedgers, dirs.LedgerPack, NewLedgerColdIngester},
-		{cfg.Txhash, dataTypeTxhash, dirs.TxhashBin, NewTxhashColdIngester},
-		{cfg.Events, dataTypeEvents, dirs.EventsDir, NewEventsColdIngester},
-	}
-	ings := make([]ColdIngester, 0, len(ctors))
-	for _, c := range ctors {
-		if !c.enabled {
-			continue
-		}
-		if c.path == "" {
-			return nil, closeColdAll(ings, fmt.Errorf("ingest: %s enabled but its ColdDirs path is empty", c.dataType))
-		}
-		ing, err := c.open(c.path, chunkID, sink)
-		if err != nil {
-			return nil, closeColdAll(ings, fmt.Errorf("open %s cold ingester: %w", c.dataType, err))
-		}
-		ings = append(ings, ing)
-	}
-	return ings, nil
-}
-
 // WriteColdChunk materializes ONE chunk's cold artifacts at the resolved paths
 // named by dirs, in a single pass, from the already-opened raw ledger iterator. It is
 // SOURCE-BLIND: the caller (backfill) resolves the chunk's ledger source — the
 // local frozen .pack or the bulk backend — and hands its RawLedgers iterator here,
 // so the cold materializer never learns where the bytes came from and is faked in
-// tests with a literal slice iterator. The ingesters overwrite any crashed
+// tests with a literal slice iterator. The writers overwrite any crashed
 // partial, so this is the freeze protocol's re-materialization. On any failure the
 // attempt is abandoned — leftover files are inert scratch (see the package doc's
 // artifact model) and a retry's overwrite is the cleanup.
 //
 // Source resolution (pack-stat, coverage wait) runs in the caller BEFORE this, so
 // a pack-missing or coverage-timeout failure is metered there rather than as a
-// ColdChunkTotal attempt here. The only pre-service failures left to meter here
-// are a canceled ctx and a cold-ingester constructor failure.
+// ColdChunkTotal attempt here. The only pre-open failures left to meter here
+// are a canceled ctx and a cold-writer open failure.
 func WriteColdChunk(
 	ctx context.Context,
 	logger *supportlog.Entry,
@@ -130,37 +86,29 @@ func WriteColdChunk(
 	}
 	sink = orNop(sink)
 
-	// Pre-service failures (ctx and the constructor failure below) emit the
-	// chunk's single ColdChunkTotal here: the ColdService that normally owns that
-	// aggregate isn't built yet, but the invariant is "exactly one ColdChunkTotal
-	// per chunk attempt, including failures."
+	// One deferred emit covers every return below exactly once — the invariant
+	// "exactly one ColdChunkTotal per chunk attempt, including pre-open
+	// failures" is a property of the control flow, not a runtime latch. It sits
+	// after validate, so a config error still emits nothing. The window is
+	// deliberately the whole attempt — open, the drain of the source stream,
+	// every ingest, finalize, and the deferred close (see coldBuckets).
 	start := time.Now()
+	defer func() { sink.ColdChunkTotal(time.Since(start)) }()
+
 	if cerr := ctx.Err(); cerr != nil {
-		sink.ColdChunkTotal(time.Since(start))
 		return cerr
 	}
-
-	ings, berr := buildColdIngesters(dirs, chunkID, sink, cfg)
-	if berr != nil {
-		// A constructor failure is still a chunk attempt: emit the aggregate
-		// (closeColdAll rolled back the built ingesters with no per-ingester emit).
-		sink.ColdChunkTotal(time.Since(start))
-		return berr
+	cc, oerr := openColdChunk(dirs, chunkID, sink, cfg)
+	if oerr != nil {
+		return oerr
 	}
+	defer func() { err = errors.Join(err, cc.close()) }()
 	logger.Debugf("cold ingest chunk %d [%d, %d]", uint32(chunkID), chunkID.FirstLedger(), chunkID.LastLedger())
-	// txhash and events share the single ExtractLedgerEvents walk; a ledgers-only
-	// chunk needs no walk at all.
-	service := NewColdService(ings, sink, cfg.Txhash || cfg.Events)
-	defer func() {
-		if cerr := service.Close(); cerr != nil {
-			err = errors.Join(err, fmt.Errorf("close: %w", cerr))
-		}
-	}()
 
-	if derr := drain(ctx, raw, chunkID, service); derr != nil {
+	if derr := drain(raw, chunkID, cc); derr != nil {
 		return derr
 	}
-	// drain verified the full range was consumed, so Finalize never commits a
+	// drain verified the full range was consumed, so finalize never commits a
 	// truncated artifact.
-	return service.Finalize(ctx)
+	return cc.finalize(ctx)
 }
--- a/cmd/stellar-rpc/internal/fullhistory/ingest/events.go
+++ b/cmd/stellar-rpc/internal/fullhistory/ingest/events.go
@@ -7,17 +7,19 @@ import (
 	"os"
 	"time"
 
+	sdkingest "github.com/stellar/go-stellar-sdk/ingest"
+
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/events"
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/storage/chunk"
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/storage/stores/eventstore"
 )
 
-// ───────────────────────── Cold ingester ─────────────────────────
+// ───────────────────────── Cold writer ─────────────────────────
 
-// eventsCold models the backfill path: per-ledger view → payloads → term-index
-// accumulate + cold append, then chunk-end Finish + WriteColdIndex. No HotStore
-// is involved — it maintains an in-memory events.Bitmaps mirror via NewBitmaps
-// + per-event TermsForBytes, and an events.LedgerOffsets to assign
+// eventsCold models the backfill path: shared-walk output → payloads →
+// term-index accumulate + cold append, then chunk-end Finish + WriteColdIndex.
+// No HotStore is involved — it maintains an in-memory events.Bitmaps mirror via
+// NewBitmaps + per-event TermsForBytes, and an events.LedgerOffsets to assign
 // chunk-relative event IDs.
 type eventsCold struct {
 	chunkID   chunk.ID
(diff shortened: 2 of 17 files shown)


cmd/stellar-rpc/internal/fullhistory/ingest/ingester.go
```diff
@@ -21,14 +46,30 @@ import (
...
-// the current iteration step — an implementation must copy any bytes it retains.
-// ColdService drives the per-ledger Ingest calls sequentially, so each view is
-// fully consumed before the next.
+// Input: l carries one ledger's shared pre-extracted data (see ledgerData). Its
+// borrowed byte slices are valid only for the current call — an implementation
+// must copy any bytes it retains. ColdService drives the per-ledger Ingest calls
+// sequentially, so each ledger is fully consumed before the next.
 type ColdIngester interface {
-	Ingest(ctx context.Context, seq uint32, lcm xdr.LedgerCloseMetaView) error
+	Ingest(ctx context.Context, l ledgerData) error
```
tamirms: `ColdIngester` is exported but this method now takes the unexported `ledgerData`, so nothing outside the package can implement it, yet it stays part of the public surface. Combined with the fact that `ColdIngester`, `ColdService`, and `NewColdService` have no references outside this package (only `WriteColdChunk` is called externally, and `HotService` is the seam that genuinely needs export), I read this as a sign the generic fan-out is buying extensibility we never use. The unexported `ledgerData` is really the tell: there was never going to be an external ingester.

Since the set is fixed at three types and mirrors what `hotchunk.DB.IngestLedger` already does inline (extract once, then three concrete writes), what do you think about collapsing the whole seam to a concrete `coldChunk` and dropping the interface, `ledgerData`, `buildColdIngesters`, and `closeColdAll`?

```go
type coldChunk struct {
	ledgers *ledgerCold // nil if not requested
	txhash  *txhashCold
	events  *eventsCold
	sink    MetricSink
}

func (c *coldChunk) ingest(seq uint32, lcm xdr.LedgerCloseMetaView) error {
	if c.ledgers != nil {
		if err := c.ledgers.write(seq, []byte(lcm)); err != nil {
			return err
		}
	}
	if c.txhash == nil && c.events == nil {
		return nil
	}
	txEvents, err := sdkingest.ExtractLedgerEvents(lcm)
	if err != nil {
		return fmt.Errorf("extract ledger events seq %d: %w", seq, err)
	}
	closedAt, err := lcm.LedgerCloseTime()
	if err != nil {
		return fmt.Errorf("ledger close time seq %d: %w", seq, err)
	}
	if c.txhash != nil {
		if err := c.txhash.write(seq, txEvents); err != nil {
			return err
		}
	}
	if c.events != nil {
		if err := c.events.write(seq, closedAt, txEvents); err != nil {
			return err
		}
	}
	return nil
}
```

Each writer takes exactly the inputs it uses (`ledgers.write(seq, raw)`, `txhash.write(seq, txEvents)`, `events.write(seq, closedAt, txEvents)`), which matches the fields each one reads today, so the bag disappears without loss. `finalize` and `close` are written out over the three nil-checked pointers in fixed order. The shared extract is metered right after the walk; how it should be labeled is the subject of the metrics note on `WriteColdChunk`. The walk-or-not decision becomes the structural `if c.txhash == nil && c.events == nil`, so the `extract` flag goes away too (see the note on `NewColdService`).

The one real tradeoff is testing: the current interface lets the orchestration tests inject fakes (`failingCold`, `finalizeErrCold`, and friends). Concrete writers push those failure-path tests toward driving real writers with adversarial inputs, which `TestColdIngester_Failure_RecordsErrorMetric` already does with an out-of-order seq. If we would rather keep injectable fakes, the smaller alternative is to keep a fan-out but unexport the seam and split the input by role (a `rawConsumer` writing from `raw`, an `eventConsumer` writing from the shared extraction), which also removes the flag and this private-type-in-exported-interface issue. I lean toward the concrete version since it reads like the hot path, but either works.

cmd/stellar-rpc/internal/fullhistory/ingest/service.go
```diff
@@ -79,25 +80,46 @@ func (s *HotService) Ingest(_ context.Context, seq uint32, lcm xdr.LedgerCloseMe
...
-// case where no Ingest follows is an already-errored short/empty stream, where
-// the timing sample is meaningless anyway.
-func NewColdService(ingesters []ColdIngester, sink MetricSink) *ColdService {
-	return &ColdService{ingesters: ingesters, sink: orNop(sink), start: time.Now()}
+// sink defaults to NopSink. extract selects whether Ingest runs the shared
+// per-ledger ExtractLedgerEvents walk: true iff txhash or events is enabled
+// (a ledgers-only chunk needs no walk). The per-chunk aggregate timer starts
+// here; the only case where no Ingest follows is an already-errored short/empty
+// stream, where the timing sample is meaningless anyway.
+func NewColdService(ingesters []ColdIngester, sink MetricSink, extract bool) *ColdService {
```
tamirms: What `extract` means here is a fair question: it is whether `Ingest` should run the shared `ExtractLedgerEvents` walk, set at the one call site to `cfg.Txhash || cfg.Events`. The trouble is it encodes a fact the writer set already determines, and the two stay consistent only by both being derived from the same `cfg` in `WriteColdChunk`. If they ever drift so an events ingester is present with `extract=false`, `PayloadsFromLedgerEvents(nil, ...)` returns zero payloads with no error, so the chunk records every ledger as event-free and silently drops its events. Ledgers-only is a reachable config (a chunk re-materializing just its ledgers pack via `ingestConfigFor`), so this is a live silent-data-loss path rather than a purely hypothetical one.

I would treat removing the flag as the one blocking item here, independent of how far we take the rest of the cleanup. Under the concrete `coldChunk` in the note on `ColdIngester` it becomes structural (`if c.txhash == nil && c.events == nil { return }`), and you cannot reach an event consumer without going through the extraction. If the fan-out stays, deriving the decision from the writers (a small capability signal) instead of a passed bool gets the same guarantee.

tamirms (commented):
A few suggestions on the cold ingest path, all about the shape of the extraction-sharing change rather than its correctness (the byte-identity looks solid and the suite is green locally, including -race). There is one blocking item (the extract-flag footgun on NewColdService), one larger design direction (a concrete coldChunk), a metrics-organization cleanup, and a stale comment.

cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
```diff
@@ -148,7 +148,9 @@ func WriteColdChunk(
 		return berr
 	}
 	logger.Debugf("cold ingest chunk %d [%d, %d]", uint32(chunkID), chunkID.FirstLedger(), chunkID.LastLedger())
-	service := NewColdService(ings, sink)
+	// txhash and events share the single ExtractLedgerEvents walk; a ledgers-only
+	// chunk needs no walk at all.
+	service := NewColdService(ings, sink, cfg.Txhash || cfg.Events)
```
tamirms: Two metric-organization cleanups worth doing together while this seam is open.

First, the aggregate. `ColdChunkTotal` is emitted from three places today (twice up in `WriteColdChunk` for the ctx and constructor pre-failures, once in `ColdService` via finalize-or-close) and deduplicated with the `totalEmitted` latch, with the timer living in the service constructor. `WriteColdChunk` is the one function with the natural start and end of an attempt and already owns two of the three emits, so a single deferred emit covers every return exactly once:

```go
func WriteColdChunk(...) (err error) {
	if verr := cfg.validate(); verr != nil {
		return verr
	}
	sink = orNop(sink)
	start := time.Now()
	defer func() { sink.ColdChunkTotal(time.Since(start)) }()

	if cerr := ctx.Err(); cerr != nil {
		return cerr
	}
	cc, oerr := openColdChunk(dirs, chunkID, sink, cfg)
	if oerr != nil {
		return oerr
	}
	defer func() { err = errors.Join(err, cc.close()) }()

	if derr := drain(ctx, raw, chunkID, cc); derr != nil {
		return derr
	}
	return cc.finalize(ctx)
}
```

That lets `emitChunkTotal`, `totalEmitted`, and the `start` field come out of the service, and emitted-exactly-once becomes a property of the control flow rather than a runtime guard. The `defer` sits after `validate` so a config error still emits nothing, matching today.

Second, the shared extract stage. It is metered as `IngestStage(dataTypeShared, stageExtract, ...)`, and `data_type="shared"` is a synthetic value that exists only in `cold_stage_duration_seconds`. The sibling cold vectors (`cold_ingest_duration_seconds`, `cold_items_total`, `cold_errors_total`) are registered for `{ledgers, txhash, events}` only, so `data_type` has two different domains inside one subsystem, and any stage-to-ingester join by `data_type` gets a `shared` bucket with no counterpart. The hot tier already models this exact walk cleanly, as a type-less `hot_phase_duration_seconds{phase="extract"}`, so this is the one place cold does not mirror hot even though the walk is the whole point of the change. Cleaner: give it its own label-less metric, `cold_extract_duration_seconds` (plus `cold_extract_items_total` for the volume if we want it), emitted by a dedicated `ColdExtract(d, items)` sink method from `coldChunk.ingest`, and delete `dataTypeShared`, `stageExtract`, and their `coldStagePairs` entry (seven pairs down to six).

One consequence to make explicit either way: after moving extraction into the shared walk, the per-stage durations no longer partition `cold_ingest_duration_seconds` for txhash and events. Extraction left those ingester totals, and each grew an unstaged per-ledger remainder that used to sit inside its old `extract` stage (txhash's truncate-and-append, events' `PayloadsFromLedgerEvents` shaping). For txhash especially, `cold_ingest_duration_seconds{txhash}` now measures only the cheap accumulate plus the finalize sort and write, not the dominant walk, so its meaning shifted. I would document that the per-ingester total is now a post-extraction figure and the stages are a partial breakdown, rather than re-staging the remainders (which would fight the one-walk goal). The `cold_stage_duration_seconds` help text ("extract, term_index, write, finalize; not every data type emits every stage") also needs a tweak, since after this `extract` is emitted by no real data type at all.

cmd/stellar-rpc/internal/fullhistory/ingest/events.go
```diff
@@ -148,10 +142,11 @@ func (e *eventsCold) ingestSeq(seq uint32, lcm xdr.LedgerCloseMetaView) (int, er
...
 	// and Finalize refuses afterwards: recovery means abandoning the chunk via
 	// Close, not resuming mid-chunk. An empty-payload ledger (genuinely zero
-	// events, or a V0 ledger handled by eventPayloads) runs zero iterations but
-	// still emits term_index/write samples and advances offsets below, so every
-	// ledger contributes exactly one sample to each of the three cold stage
-	// histograms — a consumer can divide a stage total by the ledger count.
+	// events, or a V0 ledger that PayloadsFromLedgerEvents shapes to zero payloads)
+	// runs zero iterations but still emits term_index/write samples and advances
+	// offsets below, so every ledger contributes exactly one sample to each of the
+	// two per-ledger events stage histograms — a consumer can divide a stage total
```
tamirms, line 152: You updated this comment from three histograms to two, but the matching claim just below on the `offsets.Append` block (around L170) was missed and now contradicts it: it still says "extract + term_index + write partitions the per-ledger observe window with no unexplained remainder." After this change that is off on two counts. The events ingester no longer emits an `extract` stage at all, and the `PayloadsFromLedgerEvents` shaping now runs inside this observe window without its own stage, so there is an unstaged remainder. Suggested rewrite of that block:

```
// offsets.Append LAST: it is the commit point for the ledger. Its cost folds
// into the write stage, so term_index and write are the two per-ledger stages
// this ingester emits. The PayloadsFromLedgerEvents shaping at the top of the
// function folds into the ColdIngest total without its own stage; the shared
// ExtractLedgerEvents walk is metered once, ledger-scoped, by ColdService.
// uint32(len(payloads)) is 0 for an empty ledger: an explicit Append(seq, 0)
// that records the empty ledger.
```

(If the `cold_extract_duration_seconds` change from the metrics note lands, "by ColdService" becomes "by the shared extract metric," but the partition point stands either way.)

cmd/stellar-rpc/internal/fullhistory/ingest/ingester.go
chowbao: Went with the concrete `coldChunk` in 3fafa153 — agreed the unexported `ledgerData` in an exported interface was the tell. `ColdIngester`, `ledgerData`, `ColdService`, `NewColdService`, `buildColdIngesters`, and `closeColdAll` are all gone; `openColdChunk` is now the single definition site of the type order and build rollback, and each writer takes exactly what it uses (`ledgers.write(seq, raw)`, `txhash.write(seq, txEvents)`, `events.write(seq, closedAt, txEvents)`).

On the testing tradeoff: took the adversarial-real-writers route. The fakes' coverage moved to `TestColdChunk_MidChunkFailure_NoArtifact` (garbage LCM mid-chunk fails the shared walk *after* the raw ledger append already landed — exercising the abandon-with-no-phantom-sample path against two real writers) and `TestColdChunk_Finalize_FirstErrorStopsRemaining` (a directory planted at the .bin path fails txhash's finalize after ledgers committed and before events published — covering both first-error-stops and earlier-artifact-stays in one test). `TestDrain_OverrunPastChunk` now proves the guard via the per-ledger write-stage count instead of a counting fake.

cmd/stellar-rpc/internal/fullhistory/ingest/service.go
chowbao: Fixed in 3fafa153. The flag is gone with the fan-out: under the concrete `coldChunk` the decision is `if c.txhash == nil && c.events == nil { return nil }`, so the walk runs iff a consumer of its output exists — the drift you describe (events writer present, walk skipped, `PayloadsFromLedgerEvents(nil, ...)` recording every ledger as event-free) is now unrepresentable.

cmd/stellar-rpc/internal/fullhistory/ingest/driver.go
chowbao: Both done in 3fafa153.

Aggregate: `WriteColdChunk` now owns the emit as a single `defer` registered right after `validate` (config errors still emit nothing), so every later return — ctx, open failure, drain error, finalize — is covered exactly once by control flow. `totalEmitted`, `emitChunkTotal`, and the service-held `start` are gone with `ColdService` itself. The window now also spans the deferred `close`, a slight widening of the old failure-path window; noted in the `ColdChunkTotal` doc.

Extract metric: the walk is now `cold_extract_duration_seconds`, a label-less histogram fed by a dedicated `MetricSink.ColdExtract(d, items)` from `coldChunk.ingest`; `dataTypeShared`, `stageExtract`, and their `coldStagePairs` entry are deleted (six pairs), so `data_type` has one domain again. I skipped `cold_extract_items_total` — the walk's item count is the chunk's transaction count, which `cold_items_total{txhash}` already carries; the count stays on the sink method for the CSV bench sink, mirroring how `IngestStage` items work.

The partition consequence is documented rather than re-staged, as you suggested: `cold_ingest_duration_seconds` help now says post-extraction (pointing at the extract metric), `cold_stage_duration_seconds` help drops `extract` and calls the stages a partial breakdown, and the `ColdExtract` interface doc states the per-type totals shifted meaning.

cmd/stellar-rpc/internal/fullhistory/ingest/events.go
chowbao: Fixed in 3fafa153 with your suggested text, taking the "shared extract metric" variant since `cold_extract_duration_seconds` landed — the block now names term_index + write as the two per-ledger stages, calls out the shaping folding into the ColdIngest total without its own stage, and points the walk at the ColdExtract signal.

cmd/stellar-rpc/internal/fullhistory/ingest/ingester.go
```diff
@@ -2,33 +2,191 @@ package ingest
...
+			return err
+		}
+	}
+	if c.txhash == nil && c.events == nil {
+		return nil
+	}
+	start := time.Now()
+	txEvents, err := sdkingest.ExtractLedgerEvents(lcm)
+	if err != nil {
+		return fmt.Errorf("extract ledger events seq %d: %w", seq, err)
```
tamirms, line 109: This branch, and the `ledger close time` one just below, surfaces a shared-walk failure with no error-carrying metric. `ColdExtract` fires only on the success path below, no writer runs, and `ColdChunkTotal` has no error dimension, so a walk failure lands in no counter at all, only in logs.

That is a step down from the behavior before the shared walk. Previously extraction lived inside the writers (`ExtractTxHashes`, the events shaping), and a failure went through `observe(dur, 0, err)` to `ColdIngest(dataType, err)`, so it bumped `cold_errors_total{data_type}`. Moving the walk out here dropped that, and nothing downstream picks it up: `processChunk` returns the error to `executePlan`, which retries via `withRetries`, and there is no chunk-failure or retry counter in `observability.Metrics`. So a persistently failing extract during a backfill is invisible in metrics through the whole retry loop and only surfaces as an eventual whole-pass failure and daemon restart. Backfill is an unattended batch over all of history, which is where log-only visibility is weakest.

It is also the one spot where cold stops mirroring the hot path, which is the stated goal. Hot emits `HotPhase(PhaseExtract, dur, 0, err)` on a decode failure, so `hot_phase_errors_total{phase="extract"}` increments immediately. Mirroring that here means emitting on failure, not only on success.

A small fix that fits the `ColdExtract` shape you just added: give it an error, `ColdExtract(d time.Duration, items int, err error)`, and emit `c.sink.ColdExtract(time.Since(start), 0, err)` on both this branch and the `LedgerCloseTime` one before returning. That also records the partial walk duration on failure, matching hot's "a phase that blocked then failed is signal," which the current code drops. Then register a label-less `cold_extract_errors_total` next to `cold_extract_duration_seconds`.

Calibration: this is observability, not correctness, and the counter is near-always zero because the inputs are trusted (a local pack that already parsed once, or a validated bulk backend). Its value is as a canary that spikes on a systemic problem, an SDK regression or a corrupt pack, which is exactly what you want alertable in an unattended batch. Fine to leave only if backfill health is tracked purely through logs and the daemon-restart signal, though that would sit oddly next to bothering to meter the hot extract failure.

cmd/stellar-rpc/internal/fullhistory/ingest/events.go
```diff
@@ -70,31 +63,34 @@ func NewEventsColdIngester(bucketDir string, chunkID chunk.ID, sink MetricSink)
...
 // referencing event IDs past the offsets commit point.
-func (e *eventsCold) Finalize(ctx context.Context) error {
+func (e *eventsCold) finalize(ctx context.Context) error {
 	start := time.Now()
 	if e.failed {
-		// Ingest already metered and latched this failure; refuse to finalize a
+		// write already metered and latched this failure; refuse to finalize a
 		// chunk whose mirror/pack may be ahead of the offsets commit point.
-		return fmt.Errorf("events cold ingester for chunk %s: Finalize after failed Ingest", e.chunkID)
+		return fmt.Errorf("events cold writer for chunk %s: Finalize after failed Ingest", e.chunkID)
```
tamirms: Small leftover from the rename: this refusal still says "Finalize after failed Ingest" with the old capitalized method names, while the rest of the file moved to `finalize`/`write`.

```suggestion
		return fmt.Errorf("events cold writer for chunk %s: finalize after failed write", e.chunkID)
```

tamirms (commented):
Two items remain after the rewrite, which addressed all four earlier threads cleanly. One is an observability regression worth weighing (the shared walk dropped the error signal that extract failures used to have), the other a trivial rename leftover.

cmd/stellar-rpc/internal/fullhistory/ingest/ingester.go
chowbao: Done in 5683b490, exactly as sketched: `ColdExtract(d, items, err)`, emitted with the partial walk's duration and a zero item count on both the `ExtractLedgerEvents` and `LedgerCloseTime` failure branches, plus a label-less `cold_extract_errors_total` registered next to the duration histogram (help text names it the cold mirror of `hot_phase_errors_total{phase="extract"}`). Agreed on the calibration — the inputs are trusted so it should sit at zero, but backfill being an unattended batch is precisely where a log-only canary is worthless, and it would have been odd to meter the hot decode failure and not this one.

`TestColdChunk_MidChunkFailure_NoArtifact` (which already drives a garbage-LCM walk failure) now also asserts the error-carrying `ColdExtract` fires for the failed ledger after the first ledger's clean one, and the success-path test asserts none of its extracts carry an error.

cmd/stellar-rpc/internal/fullhistory/ingest/events.go
chowbao: Applied in 5683b490 (and updated the test that matched on the old string).

tamirms (approved):

PR #854, opened 2026-07-10
Optimize backfill

cjonas9:
### What

This PR introduces several changes to backfill for the sake of drastically reducing its runtime. **These changes flatten the curve of ledger insertion time vs. count of ledgers inserted**, a curve that grew explosive over the last ~6 months as ledger complexity increased. These changes are:
- Deferred index builds: during `backfill` only, `transactions` is recreated with its hash key as an explicit unique index rather than a `PRIMARY KEY` (the two shapes are physically identical in SQLite + indistinguishable to readers, writers, and the query planner). That index plus the events secondary indexes are dropped for the load and their DDL recorded in the meta table. `finalize` builds whatever the meta table says is still pending, then a final frontfill catches up ledgers that closed during the finalize phase (since it can take ~30 minutes). Populated DBs are never touched and keep the PK shape.
    - `finalize`'s CREATE INDEX passes run on a dedicated single-connection session with `_cache_size=-32768` and `PRAGMA threads=4` (this sizes the sorter's memory budget, and larger caches were found to hurt in both time and residual memory consumption)
- Fixed-arity batching: tx/event writers accumulate rows and flush in fixed-size batches (e.g. 3,000 events), keeping the `INSERT`'s bind-variable count constant so SQLite reuses the prepared statement instead of re-parsing variable-arity SQL
- Decoding pipeline: ledger XDR is pre-decoded into a channel feeding `ingestRange,` overlapping decode CPU work with DB writes instead of decoding inline

Of these changes, the only "truly touches hot-ingest path" change is fixed-arity batching, but functionally, this introduces no behavioral differences. The only thing it changes is when within the transaction the `INSERT`s execute, not what the transaction commits. Before, each ledger's events were `INSERT`ed immediately during `ingestLedgerCloseMeta` in variable-sized batches (≤1000 rows). We just fix the batch size so rows accumulate in the in-memory pending slice and are written by `flushPending()` at `Commit` (plus any full 3000-row batches along the way). This still commits the same rows in the same order in the same write txs.

### Why

Backfill has demonstrated to have an explosive tail latency as ledgers are ingested into the DB due to the indexing overhead incurred by ever-compounding tx, event, and other inserts. It was never intended to take 10+ hours in the first place, and this is a regression that's snuck up on us as ledger composition has grown more complex. 

To summarize the reduction in time, measured on a `c5.2xlarge` (16 GiB), to perform a 7-day/120,960 ledger pubnet window starting from an empty DB:
- vanilla main `backfill` takes **11 h 34 m**
    - per-chunk timing climbs from 132 s to ~400s plateau, with the final frontfill filling chunks at 11–13 minutes/chunk
    - an 11-hour backfill has to fill extra chunks to chase checkpoints that closed mid-run
- this branch takes an average of **2hr 15 m**
    - ~55 s flat per-chunk timing, one ~30 minute finalize at the end
    - measured across six runs

These timing gains are strictly necessary if we wish to maintain architectural work done for the performance evaluation epic: GHA imposes a 6 hour cap on job length, and backfill takes way too long. Because we intend on using the DB backfilled in the backfill test (see PR #830) for the RPC stood up for the `stellar-rpc-blaster` endpoint evaluation test, we must optimize backfill first -- these CI jobs are inherently serial.

Further, as a tangential gain from this branch, this also allows backfill to be used for backfills much longer than one week: insertion is now a constant-time operation, and though sorting isn't, it's far faster than building indices inline.

### Known limitations

N/A


cmd/stellar-rpc/internal/db/db.go
```diff
@@ -78,13 +79,22 @@ func (d *DB) ResetCache() {
...
+
+// Backfill DSN pragmas. synchronous=OFF is safe since backfill is restartable
+// and gap-checked; the 1 GiB cache leaves most RAM to the OS page cache.
+const backfillSQLitePragmas = "_journal_mode=WAL&_wal_autocheckpoint=0&_synchronous=OFF" +
+	"&_cache_size=-1048576" // 1 GiB page cache (negative value = KiB)
+
+// Index-build DSN pragmas. A large cache feeds the CREATE INDEX sorter;
+// default temp_store spills large sorts to disk.
+const indexBuildSQLitePragmas = "_journal_mode=WAL&_wal_autocheckpoint=0&_synchronous=NORMAL" +
+	"&_cache_size=-2097152" // 2 GiB page cache (negative value = KiB)
```
Shaptic: would it be dtm to do

```suggestion
const indexBuildSQLitePragmas = serveSQLitePragmas + "&_cache_size=-2097152" // 2 GiB page cache (negative value = KiB)
```

cmd/stellar-rpc/internal/ingest/service.go
```diff
@@ -273,12 +273,35 @@ func (s *Service) ingestRange(ctx context.Context, backend backends.LedgerBacken
...
+	defer cancelFetch()
+	fetches := make(chan fetched, 8) //nolint:mnd
+	go func() {
+		defer close(fetches)
+		for seq := seqRange.From(); seq <= seqRange.To(); seq++ {
+			lcm, err := backend.GetLedger(fetchCtx, seq)
+			select {
+			case fetches <- fetched{lcm, err}:
+			case <-fetchCtx.Done():
+				return
```
Shaptic, line 291: Should we check `fetchCtx.Err()` somewhere?

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
```diff
@@ -76,7 +76,10 @@ func Gather(ctx context.Context) error {
 			}
 			return appendOutputs(githubOutput,
 				"found=true",
-				fmt.Sprintf("passed=%t", res.Verdict == "ok"))
+				fmt.Sprintf("passed=%t", res.Verdict == "ok"),
```
Shaptic: This seems extraneous to the rest of the PR; is it? Nbd but just for my reviewing sanity

cmd/stellar-rpc/internal/ingest/service.go
```diff
@@ -273,12 +273,35 @@ func (s *Service) ingestRange(ctx context.Context, backend backends.LedgerBacken
...
 
+	// Fetch (and decode) ledgers ahead of the DB writes
+	type fetched struct {
+		lcm xdr.LedgerCloseMeta
+		err error
+	}
+	fetchCtx, cancelFetch := context.WithCancel(ctx)
+	defer cancelFetch()
+	fetches := make(chan fetched, 8) //nolint:mnd
+	go func() {
```
Shaptic, line 284: Just fyi the PR says

> ledger XDR is pre-decoded on separate goroutines

but this is just one

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -96,6 +106,276 @@ func openSQLiteDB(dbFilePath string) (*db.Session, error) {
...
+
+const idxLedgerSequenceName = "index_ledger_sequence"
+
+// deferredIndexNames are the secondary indexes dropped during a fresh-DB
+// backfill bulk-load and rebuilt by FinalizeBulkLoad. Their DDL comes from
+// migratedSchemaSQL; a migration touching them or the transactions table must
+// tolerate their absence/twin form during an unfinalized backfill.
+//
+//nolint:gochecknoglobals // effectively-constant name list
+var deferredIndexNames = []string{"idx_id_contract_id", "idx_id_topic1", idxLedgerSequenceName}
```
Shaptic: We should add a similar comment to the `sqlmigrations/` folder in a readme or something, or maybe even a commit hook

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -218,10 +235,46 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
...
 	}
+	// Restore the canonical schema after a bulk-load, including one interrupted
+	// by a crash. Must finish before ingestService.Start to avoid starving it.
+	finalizeStart := time.Now()
+	if err := db.FinalizeBulkLoad(context.Background(), daemon.db, cfg.SQLiteDBPath, logger); err != nil {
+		logger.WithError(err).Fatal("failed to finalize backfill bulk-load")
+	}
+	// The backfill perf-eval runner keys off this line; keep it stable
+	logger.WithField("duration", time.Since(finalizeStart).String()).Info("Bulk-load finalize complete")
+	// EXPERIMENT (do not commit): top-up frontfill after finalize so captive
```
Shaptic: Just marking this with a comment so you don't forget to fix this

cmd/stellar-rpc/internal/db/db.go
cjonas9: I think this is definitely the right way to go other than just being cleaner. I think the separate variable names/structure for this PR generally should give enough of a tell that "serve mode ingest, backfill ingest, and index-building have different needs" but I think this change reads as more intuition-forward

cmd/stellar-rpc/internal/ingest/service.go
cjonas9: I'm fine with adding it as a redundancy check to the following err check (i.e. changing the line from `if err != nil` to `if errors.Join(err, fetchCtx.Err()); err != nil`), but I asked Claude about this, and the argument that we don't have to is fairly convincing:

> The fetchCtx.Err() != nil clause is redundant because cancellation is already caught on every path — it just arrives one step later, as err:
> 1. Canceled before/during GetLedger: GetLedger(fetchCtx, seq) returns a context error, so err != nil fires (after the error is forwarded to the consumer, which is what you want).
> 2. Canceled during the send: the case <-fetchCtx.Done() in the select returns.
> 3. Canceled after a successful send — the only window fetchCtx.Err() could uniquely catch: without the clause, the loop just goes around once more, calls GetLedger with an already-canceled context, and gets an immediate context error, landing back in case 1.

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness/gather.go
cjonas9: Ah, good catch! This PR was cut off as a separate section from the backfill perf eval PR, so these edits must have rode along from there, and they were dropped from the upstream as unnecessary, but then stuck along for the ride on this PR. Dropping this

cmd/stellar-rpc/internal/db/db.go
cjonas9: This is totally reasonable. I think a commit hook looks a bit heavy implementation-wise to ride along with this PR (let me know if I'm mistaken), so I added comments in the relevant .sql files and a test that enforces this by calling `bulkLoadDDLs` and runs the embedded migrations against a throwaway in-memory DB and asserts the set of secondary indexes on `transactions` and `events` is exactly `deferredIndexNames`.

.github/workflows/load-test-coordinator.yml
```diff
@@ -56,14 +58,14 @@ jobs:
...
           PR: ${{ inputs.pr_number }}
+        # TODO: Restore relevant legs before merging (removed apply load leg)
         run: |
           SHA="$(git rev-parse HEAD)"
           if [ -z "$PR" ]; then
             PR="$(gh pr list --repo "${{ github.repository }}" --state open \
                     --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
           fi
           LEGS='[
-            {"label":"Apply-load ingestion","run_label":"apply-load","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh","budget":225},
```
cjonas9, line 81: Marking with a comment:
Undo this removal before merging

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -205,22 +205,31 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
 	var ingestCfg ingest.Config
 	daemon.ingestService, ingestCfg = createIngestService(cfg, logger, daemon, feewindows, historyArchive, rw)
 	if cfg.Backfill {
-		backfillMeta, err := ingest.NewBackfillMeta(
```
cjonas9, line 208: All of this was folded into `mustBackfill()` down below because the structure got more complex and now there's two backfill launches: the first, which backfills the twin, then there's the `Prepare`/`FinalizeBulkLoad()` steps to convert the twin, then there's the second backfill, which catches the twin up closer to the tip. I do imagine that there's further, more clever refactoring that can be done here to take some extra heft out of `MustNew()`

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -96,6 +97,244 @@ func openSQLiteDB(dbFilePath string) (*db.Session, error) {
...
+var deferredIndexNames = []string{"idx_id_contract_id", "idx_id_topic1", "index_ledger_sequence"}
+
+// migratedSchemaSQL runs the embedded migrations against a throwaway in-memory
+// DB and returns each named object's sqlite_master DDL.
+func migratedSchemaSQL(ctx context.Context, names ...string) (map[string]string, error) {
+	ref, err := sql.Open("sqlite3", ":memory:")
+	if err != nil {
+		return nil, fmt.Errorf("could not open schema reference DB: %w", err)
+	}
+	defer func() { _ = ref.Close() }()
```
cjonas9: flagging this to remind myself to check for the idiomatic `Close()` pattern here

tamirms:
The whole twin apparatus exists because hash uniqueness on `transactions` is expressed as a `PRIMARY KEY`, and SQLite can't drop or rebuild a PK's hidden index separately from the table. If that uniqueness were an explicit index instead, transactions would work exactly like events already does in this PR: drop indexes before the bulk load, `CREATE INDEX` at finalize, done. No twin, no restore copy.

Concretely, the shape for bulk-loaded DBs would be:

```sql
CREATE TABLE transactions (
    hash BLOB NOT NULL,
    ledger_sequence INTEGER NOT NULL,
    application_order INTEGER NOT NULL
);
CREATE UNIQUE INDEX idx_transactions_hash ON transactions(hash); -- deferred during bulk load
CREATE INDEX index_ledger_sequence ON transactions(ledger_sequence);
```

This is physically identical to the current schema: a BLOB primary key on a rowid table is already stored as a hidden unique index (`sqlite_autoindex_transactions_1`) plus rowid-keyed rows. The reshape just gives that index a name so it can be dropped and rebuilt. Query plans, insert behavior, and constraint enforcement are unchanged.

The nice part is that no migration is needed for existing deployments. `PrepareBulkLoad` already requires an empty DB, so it can apply the reshape itself: recreate the empty `transactions` table in the no-PK shape, stash the DDL of the deferred indexes (the unique hash index plus the two events indexes) in the meta table, and drop the events indexes. `FinalizeBulkLoad` builds whatever the meta table says is outstanding and clears the record. Populated DBs are never touched; they keep the PK shape until an operator rebuilds them, and since the two shapes are indistinguishable to the readers and writers, nothing else needs to know which shape a given DB has.

What this deletes from the PR: `bulkDDLFromCanonical`, `transactionsNeedRestore`, `restoreTransactionsTable` with its `INSERT ... SELECT ... ORDER BY hash` copy, `migratedSchemaSQL` and the in-memory migration replay, the `deferredIndexNames` coupling with the .sql files, and the tripwire test guarding that coupling. `index_ledger_sequence` also stops moving entirely (today it gets created on the twin at prepare and rebuilt again after the restore; here it's created once by migrations and left alone).

What's preserved: the insert path during bulk load is byte-for-byte the same as the twin's (a rowid table with only `index_ledger_sequence` live), so the measured backfill numbers should carry over unchanged. Finalize should only get cheaper, since `CREATE UNIQUE INDEX` writes just the index instead of rewriting the whole table body plus its autoindex in hash order, though that deserves one validation run. Duplicate hashes still fail closed, just at `CREATE UNIQUE INDEX` time instead of at the restore's insert.

The one genuine cost is that the migrations directory would keep describing the PK shape while bulk-loaded DBs carry the explicit-index shape, so the reshape DDL needs a well-commented home in code, and we'd want a round-trip test covering prepare, fill, crash points, and finalize (which this PR needs anyway per the earlier review comment). If the fork in shapes bothers us long term, a later stamped migration could converge everyone, but nothing forces that decision now.

Curious whether you already went down this road and hit something I'm not seeing.

cmd/stellar-rpc/internal/db/transaction.go
```diff
@@ -109,37 +119,41 @@ func (txn *transactionHandler) InsertTransactions(lcm xdr.LedgerCloseMeta) error
...
-				Columns("hash", "ledger_sequence", "application_order")
-			rowsInBatch = 0
-		}
+		txn.pending = append(txn.pending, pendingTx{hash, lcm.LedgerSequence(), tx.Index})
 	}
-	if rowsInBatch > 0 {
-		if _, err = query.RunWith(txn.stmtCache).Exec(); err != nil {
+	// Full fixed-size batches keep the SQL text constant so the statement
+	// cache reuses one prepare; the remainder is flushed at Commit.
+	for len(txn.pending) >= maxTxRowsPerBatch {
```
tamirms, line 127: The remainder flush at Commit covers the `writeTx` path, but these writers have one more caller: the guarded data migrations. `transactionTableMigration.Apply` calls `InsertTransactions` on a handler it constructs itself (transaction.go:370), and the migration commit path (`MultiMigration.Commit` → `guardedMigration.Commit`) only writes the guard bool and commits the SQL transaction; nothing ever calls `flushPending`. So on an upgrade that runs the `TransactionsTable` or `EventsTable` migration, rows that don't fill a complete batch stay in `pending` and are silently dropped when the migration commits: up to 9,999 transactions and 2,999 events at the tail of the migrated range, which is the most recent history. Since the transactions migration truncates the table first and the guard bit prevents a re-run, the loss is permanent. `eventTableMigration.Apply` (event.go:465) has the same shape.

One fix that keeps the batching win would be to expose the flush on the writer interfaces and call it from the migration commit path before the guard is set. A migration test over a range whose row count isn't a multiple of the batch size would lock this in.

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
-	github.com/stellar/go-stellar-sdk v0.6.1-0.20260625225930-6181cdf8bda5
+	github.com/stellar/go-stellar-sdk v0.6.1-0.20260724075953-80ad52de7fcf
```
tamirms: Anchoring the description's TODO here so it stays on the merge checklist: `v0.6.1-0.20260724075953-80ad52de7fcf` resolves to stellar/go-stellar-sdk@80ad52de ("prevent double catchup after backfill"), which currently exists only on the `patch-double-peak` branch. Merging as is would pin main to an unreviewed branch commit, and if the branch gets rebased or deleted the pseudoversion becomes unfetchable for anyone without a warm module proxy cache. I think this needs the SDK change landed on master and the pin moved to a master pseudoversion first. The bump also spans about a month of SDK history (6181cdf to 80ad52de), so it's worth a quick look at what else rides along.

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -78,13 +79,13 @@ func (d *DB) ResetCache() {
 	d.cache.firstLedgerCloseTime = 0
 }
 
+const (
+	serveSQLitePragmas      = "_journal_mode=WAL&_synchronous=NORMAL"
```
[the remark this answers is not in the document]
tamirms: +1 on this one. `writeTx.postCommit` still issues `wal_checkpoint(TRUNCATE)` after every commit, so with the default autocheckpoint re-enabled a large commit can checkpoint twice back to back. If the auto checkpointer is helping somewhere (the finalize session, bulk chunk commits), that rationale deserves a comment like the one the old connection string carried; otherwise restoring `_wal_autocheckpoint=0` keeps the checkpoint policy in one place.

cjonas9:
> The whole twin apparatus exists because hash uniqueness on `transactions` is expressed as a `PRIMARY KEY`, and SQLite can't drop or rebuild a PK's hidden index separately from the table. If that uniqueness were an explicit index instead, transactions would work exactly like events already does in this PR: drop indexes before the bulk load, `CREATE INDEX` at finalize, done. No twin, no restore copy.
> 
> Concretely, the shape for bulk-loaded DBs would be:
> 
> ```sql
> CREATE TABLE transactions (
>     hash BLOB NOT NULL,
>     ledger_sequence INTEGER NOT NULL,
>     application_order INTEGER NOT NULL
> );
> CREATE UNIQUE INDEX idx_transactions_hash ON transactions(hash); -- deferred during bulk load
> CREATE INDEX index_ledger_sequence ON transactions(ledger_sequence);
> ```

> Curious whether you already went down this road and hit something I'm not seeing.

@tamirms Yes, I actually did go down this road earlier on! You're right that it works, and I could certainly see the code winding up there eventually. What made me implement this as I did was the blast radius of the two approaches: the explicit-index shape forks the schema across the fleet, so then we'd have bulk-loaded DBs carry DDL that the migrations directory no longer describes, and every future migration that touches transactions has to be written with both shapes in mind. All of that feels a bit invasive for a code path that only ever runs against a fresh, empty DB.

The twin apparatus is somewhat uglier, but it's quarantined so that everything lives in prepare/finalize, and by the time the node is serving, the DB is byte-for-byte the canonical schema. Every DB in existence remains logically identical, nobody ever has to ask "which shape is this one," and there's no hand-crafted DDL living in Go that has to be considered.

Since the perf win is already banked with the current approach, I'd rather not expand the schema surface area in this PR. I do think your version is a solid follow-up, but I'd probably want much more team input before moving forward on that.

cmd/stellar-rpc/internal/db/transaction.go
cjonas9: Good catch! Implemented in latest commit.

cmd/stellar-rpc/internal/db/db.go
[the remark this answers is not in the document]
cjonas9: I meant to leave a comment about this, and I'm extremely surprised to say this, but from everything I can tell, mattn's go sqlite3 driver genuinely does not support `wal_autocheckpoint` as an ordinary DSL parameter. Which is to say: **this code never did anything and this is a documentation fix.** I've done some more robust testing of this, and I would love to discover that I'm wrong, but to borrow the code tracing from Claude:
> **How the param gets silently dropped** (line refs are v1.14.17, the version pinned in go.mod):
> - mattn parses the DSN query string ([sqlite3.go#L1099](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L1099)) and consumes only a hardcoded allowlist of `_` params (`_journal_mode` at [L1291](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L1291), `_synchronous`, `_cache_size`, …), executing pragmas for those after open (e.g. [L1687](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L1687)). `_wal_autocheckpoint` appears **zero times** in sqlite3.go — never read, never executed. (Still true on master.)
> - For `file:` DSNs like ours, the unconsumed query string is passed through ([L1437-L1439](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L1437-L1439)) to `sqlite3_open_v2` with `SQLITE_OPEN_URI` ([L53-L54](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L53-L54), [L1450](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L1450)) — and SQLite's URI parser [silently ignores unrecognized query parameters](https://sqlite.org/uri.html).
> - The bundled C amalgamation *does* contain `wal_autocheckpoint` — that's the pragma implementation ([sqlite3-binding.c#L135017](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3-binding.c#L135017)) and the compiled-in default `SQLITE_DEFAULT_WAL_AUTOCHECKPOINT=1000` ([L13513](https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3-binding.c#L13513)) — i.e. the C library has the feature; there's just no DSN plumbing to reach it.
>
> **Empirical check** (mattn v1.14.17):
> 
> ```go
> db, _ := sql.Open("sqlite3", "file:/tmp/x.sqlite?_journal_mode=WAL&_wal_autocheckpoint=0")
> db.SetMaxOpenConns(1)
> db.QueryRow("PRAGMA wal_autocheckpoint").Scan(&ac) // querying the pragma returns the live threshold
> db.QueryRow("PRAGMA journal_mode").Scan(&jm)
> ```
> 
> ```
> no params                                          journal_mode=delete wal_autocheckpoint=1000
> _journal_mode=WAL only                             journal_mode=wal    wal_autocheckpoint=1000
> _journal_mode=WAL&_wal_autocheckpoint=0 (old DSN)  journal_mode=wal    wal_autocheckpoint=1000
> ```
> 
> The recognized param takes effect; `_wal_autocheckpoint=0` does not. `PRAGMA compile_options` on the same connection confirms the threshold is the compiled default (`DEFAULT_WAL_AUTOCHECKPOINT=1000`).

I am still baffled to be the one to catch this, but if we want this behavior, it's an external behavior modification. One way to get that is by running the following as ordinary SQL:
> _, err := db.ExecContext(ctx, `PRAGMA wal_autocheckpoint = 0`)

But on the secondary concern: when a commit crosses 1000 WAL pages, the autocheckpoint runs in PASSIVE mode and backfills the frames at commit, so postCommit's `wal_checkpoint(TRUNCATE)` finds almost nothing left to copy and just truncates the file. Every timing number in this PR was measured in exactly this state, so I don't feel extremely worried about this.

cmd/stellar-rpc/internal/daemon/daemon.go
```diff
@@ -205,22 +205,31 @@ func MustNew(cfg *config.Config, logger *supportlog.Entry) *Daemon {
...
+	}
+
+	// Restore the canonical schema after a bulk-load, including one interrupted
+	// by a crash. Must finish before ingestService.Start to avoid starving it.
+	finalizeStart := time.Now()
+	if err := db.FinalizeBulkLoad(context.Background(), daemon.db, cfg.SQLiteDBPath, logger); err != nil {
+		logger.WithError(err).Fatal("failed to finalize backfill bulk-load")
+	}
+	// The backfill perf-eval runner keys off this line; keep it stable
+	logger.WithField("duration", time.Since(finalizeStart).String()).Info("Bulk-load finalize complete")
```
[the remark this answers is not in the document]
cjonas9: Resolved, total wall-clock in the table now includes finalize, both lines are searched for now

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -96,6 +97,244 @@ func openSQLiteDB(dbFilePath string) (*db.Session, error) {
...
+	}
+	for _, stmt := range stmts {
+		if _, err := session.ExecRaw(ctx, stmt); err != nil {
+			return fmt.Errorf("transactions restore failed on %q: %w", stmt, err)
+		}
+	}
+	// Hash order builds the key's B-tree append-only; the WHERE applies the
+	// retention floor.
+	result, err := session.ExecRaw(ctx,
+		"INSERT INTO "+transactionTableName+" SELECT hash, ledger_sequence, application_order "+
```
tamirms: nit: since the twin is staying, naming the target columns here is cheap insurance:

```sql
INSERT INTO transactions (hash, ledger_sequence, application_order) SELECT hash, ledger_sequence, application_order FROM transactions_bulk ...
```

The SELECT side already names its columns, but the insert is positional, so it silently depends on the canonical DDL's column order. With the target list explicit, the copy is order-independent and self-documenting, and anyone adding a column to `transactions` later sees exactly what the restore carries over when this line shows up in their grep.

cmd/stellar-rpc/internal/db/migration.go
```diff
@@ -177,6 +177,12 @@ func (g *guardedMigration) Commit(ctx context.Context) error {
 	if g.alreadyMigrated {
 		return nil
 	}
+	// flush rows still buffered by a batching writer before setting the guard
+	if flusher, ok := g.migration.(interface{ flushPending() error }); ok {
+		if err := flusher.flushPending(); err != nil {
+			return err
```
tamirms, line 186: One pre-existing wrinkle this new error path runs into: when a `Commit` here fails, `MultiMigration.Commit` joins the error with a rollback but then unconditionally returns `mm.db.Commit()`, so the joined error (including whatever this flush reported) is dropped and the caller sees a confusing commit-after-rollback failure instead of the root cause:

```go
func (mm MultiMigration) Commit(ctx context.Context) error {
	var err error
	for _, m := range mm.migrations {
		if localErr := m.Commit(ctx); localErr != nil {
			err = errors.Join(err, localErr, mm.db.Rollback())
		}
	}
	return mm.db.Commit()
}
```

Returning `err` when it's non-nil would let a flush failure surface directly. It predates this PR, so equally fine as a follow-up.

tamirms:
Coming back to this one because the blast-radius framing rests on a premise worth examining: the costs you named are all forward-looking, and I don't think the transactions schema has a forward.

The strongest cost in your list is that every future migration touching `transactions` would have to be written with both shapes in mind. That cost is paid per future schema change, and with rpc v2 fast approaching, the v1 sqlite store is effectively frozen: the realistic number of further migrations touching this table before the store is retired is zero. If that premise holds, the fork's forward cost is never paid at all. What remains is a divergence that no reader, writer, query plan, or constraint can observe, and it retires itself as v1 DBs get rebuilt or decommissioned.

The freeze also shrinks the hand-crafted-DDL concern. The reason DDL in Go feels dangerous is drift against the migrations, but a frozen schema has nothing to drift toward, and the round-trip test pins the constants functionally: if the hardcoded DDL were wrong, `GetTransaction` and the UNIQUE assertion fail in CI. That's a stronger guarantee than text-level syncing. And one mechanical point I under-emphasized the first time: the explicit-index shape halves the transactions share of finalize write volume. The restore rewrites the entire table body plus its hidden PK index in hash order; with the index-only variant, the table body is written once at ingest (arrival order is already its final rowid order) and finalize sorts and writes only the index keys. It also makes transactions uniform with how this PR already treats events: insert with indexes absent, build everything at the end.

Given that, I think this belongs in this PR rather than in a follow-up, and the deciding argument is what deferral ships in the meantime. If this merges as-is, the first release mints twin-state DBs in the wild (any crash mid-bulk-load leaves one), so the follow-up can never simply delete the restore path; it has to keep recognizing and restoring twin-shaped tables for at least a release cycle, and the codebase carries both mechanisms at once. Doing the swap here means the twin never exists anywhere, and `migratedSchemaSQL`, `bulkDDLFromCanonical`, `transactionsNeedRestore`, and `restoreTransactionsTable` are never shipped instead of shipped and then unwound.

On tractability: this shrinks the PR rather than expanding it. Prepare and finalize reduce to a handful of literal, idempotent statements, the state check reduces to index-name lookups, and the insert path is physically identical to the twin's, so the banked per-chunk numbers carry over; finalize only gets cheaper. One run of the perf harness validates that, and the round-trip test you just added covers the new shape with minor adjustments. On team input: the question is narrower than "expand the schema surface area", it's whether DBs created by the bulk path may carry hash uniqueness as an explicit unique index instead of a PK, given that nothing can observe the difference and the schema is frozen. That feels scoped enough to settle on this PR, and I'd rather we spend the discussion here than merge machinery we already agree might wind up deleted.

cjonas9:
For anyone's reference, run 8 in the comment should show a run completed with 1 week of ledgers backfilled, where this DB was then used for a `stellar-rpc-blaster` test. Run 9 is likely to show these results with the changes suggested by @tamirms integrated.

@tamirms, definitely hear you on these points. I still would love to get @Shaptic 's points, but for now, I'm quite happy to make these changes and see how it works -- I definitely think that I was hedging on many points related to RPC migrations/updating in the future, but you're right that this isn't going to be some every day issue. Today I hope to experiment with including these changes as I think that this direction would likely be faster and cleaner, and unless @Shaptic has strong thoughts or good points to the contrary that require us to wind this back to the smaller-blast-zone version, I'm looking forward to seeing how this plays out. Thanks for the idea!

For expanded visibility, here are the results of **test 8** (pre- these changes):
**Commit:** `b1d5d2feb613`
Metric | Value |
|---|---|
| Ledgers ingested | 120960 (`[63681216 -> 63802175]`) |
| Retention window | 120960 |
| Wall-clock (total) | 2h51m33s |
| Ingest phase | 2h14m56s |
| Bulk-load finalize phase | 36m37s |
| Ledgers/sec (ingest) | 14.9 |

tamirms (approved):

cmd/stellar-rpc/internal/db/db.go
```diff
@@ -96,6 +97,143 @@ func openSQLiteDB(dbFilePath string) (*db.Session, error) {
...
+		return fmt.Errorf("could not commit bulk-load prepare: %w", err)
+	}
+	logger.Infof("Reshaped empty DB for backfill bulk-load, deferring %d indexes", len(deferredIndexes))
+	return nil
+}
+
+// FinalizeBulkLoad builds the indexes deferred by PrepareBulkLoad and clears
+// the record.
+func FinalizeBulkLoad(ctx context.Context, d *DB, dbFilePath string, logger *log.Entry) error {
+	pendingJSON, err := getMetaValue(ctx, d, pendingIndexesMetaKey)
```
Shaptic, line 173: Do we really need to pass the indices through SQLite via JSON through the prepare and finalize steps given that they are sourced from an array of hardcoded constants at runtime rather than just `CREATE INDEX IF NOT EXISTS` in the `Finalize` step?

cmd/stellar-rpc/internal/db/sqlmigrations/06_topic_indices.sql
```diff
@@ -2,6 +2,8 @@
 DROP INDEX idx_contract_id;
 DROP INDEX idx_topic1;
 
+-- Both indexes are deferred during a backfill bulk-load, so changing them
+-- here affects deferredIndexes in db.go.
 CREATE INDEX idx_id_contract_id ON events (contract_id, id);
 CREATE INDEX idx_id_topic1 ON events (topic1, id);
```
Shaptic, line 9: Does it make sense to actually do the drop primary key and create unique index in a new migration rather than have the DDL divergence in the backfill case, or do you guys (cc @tamirms) think it's not worth the unnecessary downtime on the next RPC startup?

cmd/stellar-rpc/internal/db/db.go
cjonas9: No, I don't think it's that we really _need_ to, but this is more of a design choice than a clean-win simplification. `FinalizeBulkLoad` is run unconditionally, and to distinguish between a bulk loaded DB and a legacy DB (with a PK-shaped `transactions` table), the presence of the pending-indexes record in the metadata table lets us distinguish on whether this `FinalizeBulkLoad` rebuilds indexes or not. We could do this with some boolean marker, which probably would make this less confusing/hidden, but then we're replacing one two-function statement with two one-function statements.
If you think we should veer towards the latter, then by all means it's not a very heavy change to make.

cmd/stellar-rpc/internal/db/sqlmigrations/06_topic_indices.sql
cjonas9: I think I'm generally convinced enough by the case made by @tamirms that, even though RPCv1 is far from being legacy, there're likely to be few necessary migrations that would justify this going forward. And should those ever approach, or work that necessitates them comes up, this conversation could be revived, but causing downtime to protect against a case that's unlikely to happen feels to me like jumping the gun a bit.

cmd/stellar-rpc/internal/db/db.go
Shaptic: Ahh yes of course - since the indices are different in each case, we need to know if bulk loads happened. But

> `FinalizeBulkLoad` is run unconditionally

I didn't realize this from the diff, indentation is hard lol. I thought it was conditioned on `cfg.Backfill`; why does it need to run unconditionally? Well, more importantly, if it *did* run conditionally, would we be able to avoid the save/load element of this thread?

cjonas9: I can see that, but yes, check out line 224 of daemon.go -- we do our bulk-load backfill, an unconditional FinalizeBulkLoad, and then a little old-style (pre-optimization) backfill to get closer to the current ledger. 

It runs unconditionally because of the crash cases. Say someone runs this with `--backfill` and it crashes before finalization is complete, and then they restart without `--backfill`. This would lead to the indexes never being built. (but also, it's guarded internally to be basically free if this happens. It's first step is to read the `BulkLoadPendingIndexes` meta key and immediately return nil if the key is absent).

Regarding whether we'd be able to get rid of the save/load element, this is starting to get somewhat involved, but I don't think we could and preserve the same functionality. If we just conditioned it on `cfg.Backfill`, then we'd run on cases that didn't really need it, and some legacy DBs would end up having a redundant second hash index built for it. That doesn't break anything, but it takes time and is less "correct," and I don't feel we lose that much by saving/loading a row in the metadata table -- the row tells us that we're mid-backfill and have some finalization work to do before this DB is all good to go. I like how the row _not_ being there tells us there's nothing else to do to the DB (this is more of an aesthetic comment, but I find the behavior there intuitive as opposed to overloading the meaning of the `--backfill` flag being on).

Shaptic: All of this justification makes perfect sense to me :) I was trying to attack from a "simplify, simplify," angle, but this thread has convinced me that tracking intermediate state does make sense, and pushing the issue on how that's actually done is not worth the add'l churn + complexity.

.github/workflows/load-test-coordinator.yml
```diff
@@ -11,7 +11,9 @@ defaults:
 
 on:
   push:
-    branches: [release/**]
+    # temporary scaffolding: before merge to main, replace with line
+    # branches: [release/**]
+    branches: [release/**, optimize-backfill]
```
Shaptic: reminder to remove this :)

Shaptic (approved):

Issue #856, opened 2026-07-11
Full-history query benchmark commands — reads across all data types 


marwen-abid:
## TL;DR
#771 was re-scoped to ingestion benchmarks; this issue carries the query half. Rewire the rpc-hack read benches onto the **production store readers** as a `bench-query` subcommand next to `bench-ingest`, reusing #771's CSV sink and reporting format. Together with #771 this completes the original scope (and #730's).

## Scope
- The eight read benches from the prototype: **{hot,cold} × {ledgers, txpage, txhash, events}**.
- Measure the production read paths in `internal/fullhistory/storage/stores/` — e.g. `eventstore/query.go` + `cold_reader.go`, `ledger/cold_reader.go` + `hot_store.go`, `txhash/read_assembly.go` + `cold_reader.go` — not the prototype's own reader implementations (same rewiring principle as #771).
- Cold event/txhash benches must measure the **complete correct read, including the MPHF post-filter** — an index probe alone benchmarks a correctness bug, not a query.
- `--query-concurrency` sweeps and the grid runner (concurrency × data type matrix).
- Events query-corpus auto-generation: derive realistic contract-ID/topic terms from the ingested data rather than hand-picked constants.
- XDR-views only (the decided path; no views-vs-decode comparison).
- CSV output in #771's schema (`stage,n,n_items,total_ns,p50_ns,p90_ns,p99_ns,max_ns`) so #775 can consume ingest and query reports uniformly.
- Land on `feature/full-history` at mergeable quality.

## Source
rpc-hack `scripts/bench-fullhistory/`: `bench_{cold,hot}_*.go`, `bench_grid.go`, `bench_concurrent_runner.go`, `corpus.go`, and the cache-eviction helpers (cold read numbers are meaningless against a warm page cache). Carry the harness and methodology; leave the prototype's reader code behind.

## Reuses from #771
`internal/fullhistory/bench/` — the cobra subcommand pattern, the CSV/percentile sink, and the fixture/source plumbing. Benchmark inputs (a populated hot DB / frozen cold chunks) can be produced by `bench-ingest` itself.

## Out of scope
- CI harness and baselines (#775 — blocked on this + #771).
- Ingestion benchmarks (#771).
- Scenario suite (#776).

## Dependencies
- #771's bench package (subcommand plumbing + CSV sink) — PR pending.
- Production cold/hot readers in `storage/stores/` (landed).

Issue #775, opened 2026-06-08
Run full-history benchmark commands in CI to catch perf regressions


tamirms:
## TL;DR
Run the full-history benchmark commands (#771) automatically in GitHub CI and compare results across versions, so performance regressions are caught before they land.

## Scope
- A CI job that runs the benchmark commands on a defined dataset.
- Compare results against a baseline and flag regressions.

## Depends on
- #771 (the benchmark commands).

*(Full-history analog of #708, which does the same for the v1 Go benchmarks.)*

marwen-abid:
Design diagram for Benchmarking CI. 
<img width="1051" height="646" alt="Image" src="https://github.com/user-attachments/assets/06d0529b-9646-4023-b032-59dadce83c4e" />

PR #857, opened 2026-07-12
fullhistory: add bench-ingest subcommand for ingestion benchmarks (#771)

### LINKED ISSUE #771: Full-history benchmark commands — ingest across all data types
## TL;DR
The single home for full-history **ingestion** benchmark commands: ingest benchmarks across ledgers, events, and txhash, on both hot and cold tiers. Built by rewiring the rpc-hack bench prototypes onto the production ingesters from #765, with a CSV metrics sink. Subsumes #730's ingest half.

> Re-scoped to ingestion only; the query benchmarks are split out to #856.

## Scope
- Benchmark commands for ingest, per data type (ledger / events / txhash), hot and cold tiers.
- Run against #765's production ingesters (CSV metrics sink), not the rpc-hack prototype's own ingest path.
- XDR-views only — drop the views-vs-struct-decode comparison the rpc-hack prototype carried; views is the committed path.
- Land on `feature/full-history` at mergeable quality.

## Source
rpc-hack `scripts/bench-fullhistory/` (~31 files) — the prototype to rewire.

## Dependencies
- Production ingesters from #765.

marwen-abid:
### What

Adds a `bench-ingest` subcommand to the `stellar-rpc` binary for benchmarking full-history ingestion. Both subcommands drive the daemon's own production ingestion code over a benchmark-controlled ledger source instead of reimplementing it, so the numbers keep tracking what actually runs as the ingestion internals evolve:

- `stellar-rpc bench-ingest cold` runs the daemon's backfill (`backfill.RunBackfill`) over a chunk range: the per-chunk freezes (ledger packs, txhash `.bin`s, events segments), the cross-chunk txhash index build, and the concurrent scheduling that ties them together.
- `stellar-rpc bench-ingest hot` runs the daemon's live ingestion loop over a chunk range: the per-ledger synced `WriteBatch` and the chunk-boundary hot-DB rotation (close the filled DB → open the next → rebuild the service).

**Ledger sources.** Both plug into the same `backfill.Backend` seam the production backfill reads through:

- `--source=pack` (default) reads a frozen local ledgers tree (`--pack-dir`, the full-history system's own `{bucket:05d}/{chunk:08d}.pack` output). Local and fully repeatable.
- `--source=bsb` streams through the production BSB source over `--datastore-type=GCS|S3|Filesystem`. GCS/S3 pull real network data; `Filesystem` reads a local lake directory for deterministic offline runs.

**Run shape.**

- Cold backfills `--num-chunks` chunks starting at `--start-chunk`, with a `--workers`-sized pool shared by chunk freezes and index builds. It always materializes the full artifact set — no type selection, which is what real backfill does.
- Hot ingests the same `--start-chunk`/`--num-chunks` range; `--num-chunks>1` crosses a chunk boundary and exercises the hot-DB rotation. `--num-ledgers` caps the run from the range start (hot ingestion fsyncs once per ledger, so full runs are slow by design — a cap gives cheap smoke runs without changing what is measured per ledger).
- `--cpuprofile`/`--memprofile` capture pprof profiles.

**Output.** One CSV per cold data type (`ledgers.csv`, `txhash.csv`, `events.csv`, one row per stage: `term_index`/`write`/`finalize`), `hot.csv` (one row per per-ledger phase), and `driver.csv` for run-level rows — the cold scheduler's `backfill_wall` and `index_rebuild`, the per-chunk `chunk_total`, the per-type `<type>_total` and the shared per-ledger `cold_extract` walk, plus the hot `ingest_total` (per-ledger end-to-end, reconstructed from each ledger's phase burst) and `run_wall`. All share one schema:

```
stage,n,n_items,total_ns,p50_ns,p90_ns,p99_ns,max_ns
```

If a run fails or is interrupted partway, CSVs for the completed work are still written (marked PARTIAL).

**Supporting production-code edits** (kept minimal):

- `fullhistory` exports the ingestion loop behind `RunBoundedIngestionLoop`/`BoundedIngestConfig`; the loop body is unchanged — its terminal `return` just names a new `errStreamEnded` sentinel that a bounded run maps to success.
- `ingest` exports its `DataType*`/`Stage*` label constants so the ingesters and the CSV sink share one label vocabulary.
- `geometry` adds a package-level `LedgerPackPath(ledgersRoot, chunk)` so a caller holding only a ledgers tree derives pack paths through the one formula home.
- `main.go` registers the `bench-ingest` subcommand.

**Example**

```
# cold: run the daemon's backfill over a chunk range from a GCS/S3 ledger archive
stellar-rpc bench-ingest cold --source=bsb --datastore-type=S3 --region=us-east-2 \
    --bucket-path=aws-public-blockchain/v1.1/stellar/ledgers/pubnet \
    --start-chunk=6100 --num-chunks=4 --workers=4 \
    --cold-out-dir=/tmp/cold --out=/tmp/bench-out
```

```
# hot: run the daemon's ingestion loop from a local pack tree, capped for a cheap
# smoke run (drop --num-ledgers for a full run; --num-chunks>1 crosses a chunk
# boundary and exercises the hot-DB rotation)
stellar-rpc bench-ingest hot --source=pack --pack-dir=/data/ledgers \
    --start-chunk=6100 --num-ledgers=1000 --hot-dir=/tmp/hot --out=/tmp/bench-out
```

### Why

We need baseline ingestion numbers for the full-history storage engine — cold per-stage throughput and hot per-ledger latency — for three consumers: tuning the engine (#836), the apply-load measurement campaign, and the benchmark-in-CI gate (#775), which parses these CSVs.

Query benchmarks are split out to #856.

Closes #771.


tamirms:
The goal is to benchmark backfill and cold-ingestion performance, so the benchmark should run the same code the daemon runs. Instead of looping the low-level `WriteColdChunk` per chunk, drive `backfill.RunBackfill` (what the daemon runs at startup to backfill history) over a chunk range.

That way it exercises the real backfill: the per-chunk materialization plus the cross-chunk txhash index build (streamhash) and the concurrent scheduling. And since it is the production path, it keeps measuring what actually runs as ingestion internals change, rather than a hand-maintained loop that drifts.

It is mostly wiring. `RunBackfill` calls `WriteColdChunk` with the same `MetricSink`, so the per-stage cold CSVs are unchanged; the index build reports through `observability.Metrics.Rebuild`, so the recorder implements that interface too. On top of the current harness it needs:

- a scratch catalog: `catalog.Open(scratchDBPath, geometry.NewLayout(outRoot), txhashLayout, logger)`, fresh per run so every run is a clean backfill from empty. It carries the output layout too, so this sets where artifacts land.
- an `ExecConfig` wiring that catalog into Process and Build, the `csvSink` as `Process.Sink`, the source as `Process.Backend`, `Workers`, and `MaxRetries: 0`.
- the source stays your abstraction; add `Filesystem` to `--datastore-type` for deterministic local runs, keep BSB for real-world.

CLI changes:

- `--types` goes away: `RunBackfill` builds the full artifact set (ledgers, events, txhash, index), which is what real backfill does. You still get per-type CSVs, just not one type in isolation.
- `--chunk` and `--num-chunks` become the `rangeStart, rangeEnd` range.
- `--chunk-workers` becomes `ExecConfig.Workers`, now one pool shared by chunk and index builds.
- `--cold-out-dir` becomes the layout root plus a scratch catalog dir.

`runCold`'s worker loop collapses into a `RunBackfill` call; the recorder, sources, and reporting stay.

tamirms:
Same shape as the cold suggestion: run the loop the daemon actually runs. `driveHot` reimplements the per-ledger loop; instead drive the real `runIngestionLoop` (hotloop.go) over the range.

`ingestionLoopConfig` is already fully parameterized, so this is mostly supplying what `startup.go` does, minus the live parts:

- `Stream`: your existing source, bounded to the range (a bounded stream ends, so the loop terminates instead of running forever).
- `Catalog`: the same scratch catalog the cold path needs.
- `HotDB`: a fresh hot DB, which the bench already opens.
- `Boundary`: a nop publisher, so completed chunks are not handed to freeze, which keeps the hot number isolated from the cold path.
- `Sink`: the same `csvSink`; the loop already emits `HotPhase` through it, so the recorder is unchanged.

`runIngestionLoop` is unexported, so it needs exporting (or a thin wrapper). That is the whole change.

It buys the same as the cold change: the benchmark tracks production instead of drifting, and since it now crosses chunk boundaries it exercises the DB rotation a single-chunk run never hits. The hot subcommand takes a chunk range now, matching the cold shape.

`read_blocked` folds away, since the real loop pulls the stream itself and does not meter the between-ledger wait. That does not matter for per-ledger latency, and against a local fixture it is near zero anyway.

Concretely, `driveHot` collapses into a `runIngestionLoop` call; the recorder, source, and reporting stay.

cmd/stellar-rpc/internal/fullhistory/bench/scratch.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/catalog"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/fullhistory/geometry"
+)
+
+// openScratchCatalog opens a fresh, run-scoped catalog in a temp dir bound to
+// layout, and returns a release func that closes it and deletes the dir. The
+// catalog records completed work and skips it on re-run, so a fresh one per run
+// keeps every run a clean backfill from empty.
+func openScratchCatalog(layout geometry.Layout, logger *supportlog.Entry) (*catalog.Catalog, func(), error) {
+	dir, err := os.MkdirTemp("", "bench-ingest-catalog-")
```
tamirms: The scratch catalog is created in `/tmp` (`MkdirTemp("")`), while the artifacts go under `--cold-out-dir` / `--hot-dir`. The catalog is a RocksDB the backfill writes and fsyncs to as it runs, so the speed of the disk it sits on is part of the measured time. `/tmp` is whatever the machine backs it with, tmpfs (RAM, unrealistically fast), an EBS volume (slow), or the local NVMe, so the cold number shifts with the `/tmp` mount rather than the code you are benchmarking.

A `--catalog-dir` flag defaulting to the output root fixes it, and matches `--cold-out-dir` / `--hot-dir` already being flags: the catalog then lands on the disk you chose for the run, not whatever `/tmp` happens to be. Keep the flag a base dir with `MkdirTemp` inside it (`os.MkdirTemp(catalogBase, "bench-catalog-")`) rather than the exact catalog path, so it keeps the fresh-per-run dir, the isolation, and the `RemoveAll` on release, and never wipes a path someone named.

cmd/stellar-rpc/internal/fullhistory/bench/profile.go
```diff
@@ -0,0 +1,65 @@
...
+			return fmt.Errorf("start CPU profile: %w", err)
+		}
+		defer func() {
+			pprof.StopCPUProfile()
+			_ = f.Close()
+			logger.Infof("wrote CPU profile to %s", p.cpuProfile)
+		}()
+	}
+	err := fn()
+	if p.memProfile != "" {
```
tamirms, line 46: The heap profile is written after `fn()` returns, but `runCold` / `runHot` close the catalog and backend in their own defers before they return, so by here the RocksDB block cache, write buffers, and BSB prefetch buffers are already freed. The snapshot then under-reports the memory ingestion actually used, which is what `--memprofile` is for. Taking it inside `run`, before the deferred store closes, would capture the real footprint.

cmd/stellar-rpc/internal/fullhistory/ingest/metrics.go
```diff
@@ -12,22 +12,22 @@ import (
 // subdirectory names used on disk. (The hot tier keys its per-ledger phases by
 // hotchunk.Phase, not by data type — see MetricSink.HotPhase.)
 const (
-	dataTypeLedgers = "ledgers"
-	dataTypeTxhash  = "txhash"
-	dataTypeEvents  = "events"
+	DataTypeLedgers = "ledgers"
```
tamirms: These were unexported before this PR; the only reason to re-export them is the bench's `fileSpecs` row ordering. But the `csvSink` keys its rows on the actual emitted `data_type` / `stage` strings, and `withUnknown` already reports anything it does not recognize, so bench-local ordering constants would work and this production change would go away entirely. The cost is losing the compile error if a label is ever renamed.

tamirms (approved):
this looks good! I left a few comments for some minor remaining issues

cmd/stellar-rpc/internal/fullhistory/bench/scratch.go
marwen-abid: Good catch! Addressed in ec91c01f348e22aaa022071a0f9a9b2a2c850fa5

cmd/stellar-rpc/internal/fullhistory/ingest/metrics.go
marwen-abid: Makes sense! Addressed in ec91c01f348e22aaa022071a0f9a9b2a2c850fa5

cmd/stellar-rpc/internal/fullhistory/bench/profile.go
marwen-abid: I tried doing what's suggested here but I wasn't happy with the end result. Profiling was mixed in with benchmarking and the separation of concerns got a little muddied. Went a different way. Please flag if you'd rather I do what you suggested: 

I kept the snapshot in the `around` wrapper (out of the drivers) and instead documented `--memprofile` as an **allocation** profile — read with `-alloc_space`. That view is cumulative (counts freed allocations too), so writing it after the stores close loses nothing; only `inuse_space` is teardown-sensitive. 

If you want footprint, I'll add an in-run sampler (still out of the driver code).

chowbao (approved):

cmd/stellar-rpc/internal/fullhistory/bench/profile.go
```diff
@@ -0,0 +1,72 @@
...
+
+// writeHeapProfile writes the heap profile to --memprofile. Read it as an
+// ALLOCATION profile (`go tool pprof -alloc_space`): that view is cumulative —
+// every byte the run allocated, whether or not it was later freed — so writing
+// it here, after the run's stores have closed, still captures the full picture.
+// The default `-inuse_space` view is a live snapshot taken after teardown and
+// under-reports, so it is not what this profile is for. No GC is forced:
+// alloc_space is unaffected by collection, so there is nothing to gain by
+// perturbing the run.
+func (p *profileFlags) writeHeapProfile(logger *supportlog.Entry) {
```
tamirms, line 60: The alloc profile is right for allocation churn, but it is Go-only: RocksDB's block cache, memtables, and write buffers are C++ allocations off the Go heap, so no pprof view sees them, and they are what a backfill's memory is dominated by. So it does not answer how much RAM a backfill holds, which is the number capacity and a memory-regression gate care about.

That number is `VmHWM`, the kernel's peak resident set for the process. `github.com/prometheus/procfs` is already a dependency (indirect, via client_golang) and reads it from `/proc/self/status` pre-normalized to bytes: `procfs.Self()`, then `NewStatus()`, then `.VmHWM`. Report it once at run end as an always-on `driver.csv` row (no flag), alongside `backfill_wall`, since it is a single comparable value you can threshold and it counts everything the process held, Go heap plus the RocksDB C++ caches plus the buffers. It complements `--memprofile` rather than replacing it.

procfs reads `/proc`, so `Self()` returns an error off Linux; handle it by skipping the row, which only happens on the macOS dev machine, never on the Linux gate. It still compiles everywhere, so nothing breaks the macOS build.

cmd/stellar-rpc/internal/fullhistory/bench/hot.go
```diff
@@ -0,0 +1,165 @@
...
+	last := (opts.StartChunk + chunk.ID(uint32(opts.NumChunks-1))).LastLedger()
+	// Overflow-safe cap: compare against the range's span rather than adding
+	// a flag-supplied count to a ledger sequence.
+	if span := last - first + 1; opts.NumLedgers > 0 && opts.NumLedgers < span {
+		last = first + opts.NumLedgers - 1
+	}
+
+	sink := newCSVSink()
+	start := time.Now()
+	err = fullhistory.RunBoundedIngestionLoop(ctx, fullhistory.BoundedIngestConfig{
```
tamirms, line 136: The hot mode ingests ledgers back-to-back, as fast as they stream. That measures throughput, the catch-up case where the daemon is draining a backlog. It does not answer the steady-state question that usually matters for hot: if ledgers closed faster than they do now, say one every 600ms, could ingestion keep up?

A pacing mode would answer it. Give each ledger a due time at a fixed close-time interval: ledger 1 at 0, ledger 2 at 600ms, ledger 3 at 1,200ms, and so on. After ingesting a ledger, sleep until the next one's due time. When you are on pace that is a real wait, you sit idle until the ledger would have closed. When a ledger runs long, say 3 seconds, the next several due times have already passed, so you do not sleep and instead ingest them back-to-back to drain the backlog, then resume sleeping once you have caught up. So the result to report is whether the backlog drains or grows: one slow ledger is fine if the ones after it catch back up, and you fall behind only if ingestion is slower than the close time on average.

Back-to-back also charges GC to the ingest window, since there is no idle time for it to run in. Pacing lets GC run in the gaps between ledgers, which is what production has, so the paced number is the truer keep-up figure and the back-to-back run overstates the cost.

cmd/stellar-rpc/internal/fullhistory/bench/csvsink.go
```diff
@@ -49,6 +49,21 @@ const (
...
+	// hot only, paced runs (--close-interval > 0): per-ledger lag behind the
+	// close schedule at commit. Its "duration" columns carry a lag time; a
+	// lag that grows across the run means the backlog is growing (see
+	// recordPaceLag).
+	driverPaceLag = "pace_lag"
+	// hot only, paced runs: the final ledger's lag behind the close schedule,
+	// one sample per run. The pace_lag percentiles discard sample order, so
+	// they cannot show whether the backlog had drained or was still growing
+	// when the run ended; this row records that ending state.
+	driverPaceLagFinal = "pace_lag_final"
```
marwen-abid: @tamirms thoughts about this one ? I think it makes sense to capture it because it tells us if the backlog had drained or was still growing when the stream ran out. That info is lost if we just rely on `pace_lag`.

cmd/stellar-rpc/internal/fullhistory/bench/hot.go
marwen-abid: Great point. Addressed here : fef3448a96b69c8344499c696ca6fc5c65f6b37a

cmd/stellar-rpc/internal/fullhistory/bench/profile.go
marwen-abid: TIL! addressed here : fef3448a96b69c8344499c696ca6fc5c65f6b37a

cmd/stellar-rpc/internal/fullhistory/bench/csvsink.go
tamirms: Sorry for the churn on this. Working through what the pacing output actually tells us, I realized the verdict is answering a question we haven't decided yet.

`backlogGrew` calls a run failed when the last ledger ends up a full close interval behind schedule (`final >= interval`). That threshold is really a statement about how far behind tip is too far, and `final >= interval` is a guess at a number we've never actually chosen. What we care about is freshness, how far ingestion falls behind the tip, and we don't know the tolerance for it yet.

So instead of judging each run against a guessed line, I think we should measure the lag now and decide the tolerance later, once real runs show us what the numbers look like. Concretely:

- Keep `pace_lag` recording each ledger's lag, and keep reporting it as the percentile row it already produces (p50/p90/p99/max across the run). That row is what shows how far behind we fall, and what we'd read a tolerance off of later.
- Drop `backlogGrew`, the `logPaceSummary` verdict line, and `pace_lag_final`. Those collapse the lag into a single pass/fail, which is the part that needs a threshold we don't have.

On units: no change to the recorded values, `pace_lag` stays a duration. "Ledgers behind tip" is just lag divided by the close interval, a way to read the same numbers when we pick a tolerance, since that's the unit it would be in.

PR #858, opened 2026-07-13
fullhistory: build empty cold indexes for zero-key coverages (closes #826)

### LINKED ISSUE #826: fullhistory/backfill: zero-transaction window stalls the daemon (ErrEmptyBuildSet leaves coverage freezing)
## Summary

When a tx-hash index coverage spans only zero-transaction chunks, the cold index build returns `ErrEmptyBuildSet` and writes no `.idx`. `buildTxhashIndex` treats that as a hard failure, so the coverage stays `freezing`, the next tick re-plans the same build, and it fails again. **The full-history daemon cannot advance past that window** — it's a stall, not just a missed optimization.

Tracked follow-up from the #818 review: https://github.com/stellar/stellar-rpc/pull/818#discussion_r3482213691 (agreed it's out of scope for #818).

## Root cause

A zero-transaction chunk produces a structurally valid but key-less `.bin`. When every `.bin` in a coverage is like that, `BuildColdIndex` returns `ErrEmptyBuildSet`:

- `cmd/stellar-rpc/internal/fullhistory/pkg/stores/txhash/cold_index.go:76-77` — after `scanAndValidate`, `total == 0` ⇒ `return ErrEmptyBuildSet` (streamhash can't build an index over zero keys).

`buildTxhashIndex` runs the build inside the one-write protocol's `create` step (`backfill/txindex.go`, the `txhash.BuildColdIndex(...)` call). `ErrEmptyBuildSet` propagates out of `OneWrite` **after** the keys were marked `freezing` (step 1) but **before** the commit (step 4). So:

1. coverage key is left at `StateFreezing` (no `.idx` on disk, no commit);
2. the next resolve/tick re-plans the identical `[lo, hi]` build;
3. it hits the same empty input set and fails again → infinite re-plan loop, daemon wedged at that window.

## When it happens

The rolling tx-hash index builds `[lo, current]`, so a window that merely **starts** with a quiet (zero-tx) chunk trips this on its first build. That makes it routine, not exotic:

- **testnet / futurenet / standalone** networks (quiet windows are common);
- **pubnet early history** during a full-history backfill (sparse early ledgers).

It is *not* a concern for pubnet steady-state, which is why it was correctly kept out of #818.

## Proposed fix

**Preferred (upstream, no special-casing in backfill):** have the txhash cold format / streamhash treat a **zero-key index as a valid empty result** — a real (empty) `.idx` whose reads resolve to not-found — instead of returning `ErrEmptyBuildSet`. Then `buildTxhashIndex` commits normally and needs no awareness of the empty case.

**Stopgap (in `buildTxhashIndex`), if the streamhash change lands later:** treat `ErrEmptyBuildSet` from the build step as a *frozen-empty coverage* — skip the barrier/`.idx` and still commit the coverage `frozen` so the daemon advances. This unblocks the stall without the format change, at the cost of a special case the upstream fix would later remove.

## Related / also verify

- The events cold format has the analogous guard (`pkg/stores/eventstore/cold_format.go:236`, `ErrEmptyBuildSet` from `buildMPHF` on zero keys). Worth confirming whether a zero-event window can stall the events freeze path the same way, or whether that path already tolerates it — fix both consistently if so.
- Parent work: #815 (Phase 1 — Backfill), epic #808.

tamirms:
## Summary

Closes #826. A window of only zero-transaction chunks stalled the full-history daemon: the txhash cold index build returned `ErrEmptyBuildSet`, so the coverage stayed `freezing` and the daemon re-planned the same window on every restart. streamhash now accepts zero-key builds (stellar/streamhash#12), so both cold stores build a real empty index for a zero-key coverage instead of special-casing it.

## What changed

**txhash (the stall).** `BuildColdIndex` drops both `ErrEmptyBuildSet` returns and builds a valid empty index for a zero-key coverage (no inputs, or inputs that hold no entries). It skips the k-way merge when there is nothing to feed. `buildTxhashIndex` then commits normally, so an all-empty window freezes instead of re-planning every restart. Removed the `TODO(#826)` at `resolveWindow`.

**events (fixing both stores consistently).** The events store already tolerated empty chunks via a zero-length `index.hash` sentinel, so it was not stalling, but this migrates it off the sentinel so both stores handle empty identically. `WriteColdIndex` builds a real streamhash empty index over zero terms and writes a zero-record `index.pack` through the ordinary path; `writeEmptyColdIndex` is deleted, `openMPHF` always opens the file, and `isEmpty()` is `NumKeys() == 0`.

## Testing

New `TestBuildTxhashIndex_AllEmptyWindowFreezes` proves the backfill flow (an all-empty window freezes rather than stalling), and the six tests that encoded `ErrEmptyBuildSet` or the zero-length sentinel were reworked to the real-empty-index behavior, including asserting the empty txhash index carries correct `[minLedger, maxLedger]` coverage metadata. All green: txhash, eventstore, backfill, lifecycle, and the full `TestE2E_DaemonLifecycle`. Build, vet, and gofmt clean. The change deletes more than it adds, since the sentinel and `ErrEmptyBuildSet` machinery is gone.

## Two dependencies before this can merge

1. **streamhash stellar/streamhash#12 must merge first.** `go.mod` points streamhash at that PR's commit so this builds and tests now. Re-pin it to a merged commit on streamhash `main` before landing.
2. **The events half overlaps #847.** It reworks `openMPHF`, `isEmpty`, and the cold-reader empty cross-check, which #847 also touches. Whichever lands second needs a rebase, so sequence the two deliberately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stellar/go-stellar-sdk v0.6.1-0.20260618191317-308407eca8c6
-	github.com/stellar/streamhash v0.0.0-20260622155330-9e08e05c5fa6
+	github.com/stellar/streamhash v0.0.0-20260713164615-c72a4e6f578d
 	github.com/stretchr/testify v1.11.1
 )
 
--- a/cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
+++ b/cmd/stellar-rpc/internal/fullhistory/backfill/resolve.go
@@ -116,10 +116,6 @@ func resolveTxHashIndex(
 			needs[cs.Chunk] = needs[cs.Chunk].Add(geometry.KindTxHash)
 		}
 	}
-	// TODO(#826): an all-empty range (no transactions — a quiet test network, or an
-	// early-history tail) still schedules a build here, which fails with ErrEmptyBuildSet
-	// and re-plans every restart. The txhash cold store needs an empty-index sentinel (as
-	// the eventstore has) before such a range can resolve to a covered, no-retry state.
 	return IndexBuild{Index: w, Lo: desired.Lo, Hi: desired.Hi}, true, nil
 }
 
--- a/cmd/stellar-rpc/internal/fullhistory/backfill/txindex_test.go
+++ b/cmd/stellar-rpc/internal/fullhistory/backfill/txindex_test.go
@@ -151,6 +151,30 @@ func TestBuildTxhashIndex_BuildsQueryableCoverage(t *testing.T) {
 	assertCoverageQueryable(t, cat, []txEntry{e0a, e0b, e1a})
 }
 
+// TestBuildTxhashIndex_AllEmptyWindowFreezes pins the #826 fix: a window whose
+// chunks all hold zero transactions (empty .bin files) builds a valid empty
+// index and freezes the coverage, instead of failing the build and re-planning
+// the same window on every restart.
+func TestBuildTxhashIndex_AllEmptyWindowFreezes(t *testing.T) {
+	cat, _ := smallTxHashIndexCatalog(t, 4)
+	cfg := testBuildConfig(cat)
+
+	// Chunks 0 and 1 are frozen with empty .bin files (no transactions).
+	freezeChunkBin(t, cat, 0, nil)
+	freezeChunkBin(t, cat, 1, nil)
+
+	require.NoError(t, buildTxhashIndex(context.Background(), 0, 0, 1, cfg))
+
+	// The coverage freezes rather than stalling in StateFreezing.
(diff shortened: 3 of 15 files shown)


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stellar/go-stellar-sdk v0.6.1-0.20260618191317-308407eca8c6
-	github.com/stellar/streamhash v0.0.0-20260622155330-9e08e05c5fa6
+	github.com/stellar/streamhash v0.0.0-20260713140807-9f6eeb7acecf
```
chowbao: TODO: bump when https://github.com/stellar/streamhash/pull/12#pullrequestreview-4686654791 is merged (one possible change in the PR which would change the commit hash)

chowbao (approved):

tamirms:
@codex[agent]+gpt-5.5 review

tamirms:
@codex review

PR #859, opened 2026-07-14
fullhistory: derive the retention floor from a committed frontier

tamirms:
## What

Fixes a startup-backfill retention-floor bug, and generalizes the fix into one shared retention-floor primitive.

The bug: startup backfill derived `rangeStart` from the tip anchor (`max(tip, lastCommitted)`) while `rangeEnd` used the post-exclusion target, so the two ends anchored on different frontiers. A mid-chunk near-tip restart rebuilt the tx-hash window index twice and briefly dropped a chunk from coverage (a false not-found once #772 wires reads).

## Changes

- **One floor primitive.** `geometry.Retention{size, earliestChunk}` with `FloorAt(frontier int64) chunk.ID`. Deletes the old `lifecycle` floor API (`RetentionFloor`, `RetentionFloorAt`, `Excludes`, `FirstChunk`, `EffectiveRetentionFloor`) and the now-orphaned `geometry.ChunkFirstLedger`.
- **Startup backfill.** `passTarget` quarantines all tip logic and returns the post-exclusion target; the floor derives from that target. Regression test: `TestBackfill_FloorAnchorsOnTargetNotTip`.
- **Lifecycle tick and eligibility scans.** `FloorAt(lastChunk)` and plain `c < floor` comparisons.
- **`validateConfig`.** Returns the validated `earliest`, so the daemon binds the retention policy from the validated path instead of re-reading the pin. Extracted `openCatalog` / `resolveCore`.

`FloorAt` takes a signed `int64` frontier so `LastCompleteChunkAt`'s `-1` (nothing complete) flows through one clamp, and no caller branches on the young-store case.

## Notes

Rebased on top of #850 (archive-tip backfill), and adopts its fatal-on-tip-error semantics. All fullhistory test suites and lint pass.


chowbao (approved):
LGTM

PR #861, opened 2026-07-15
Add Go benchmark leg in release performance evaluation suite

cjonas9:
### What

Adds leg of performance evaluation suite that runs the existing Stellar RPC Go benchmarks and compares their outputs from version to version. In particular, this PR introduces a leg that is spawned by the coordinator as an EC2 instance, runs `BenchmarkGetLedgers`, `BenchmarkGetEventsTopicFilters`, `BenchmarkGetEvents`, `BenchmarkJSONTransactions`, `BenchmarkGetProtocolVersion`, `BenchmarkGetPreflight`, `BenchmarkComputeFeeDistribution`, `BenchmarkGetLedgerRange`, `BenchmarkBatchGetLedgers`, and `BenchmarkTransactionFetch` on **both** the candidate ref and the previous release ref, captures the go benchmark logs and `benchstat`, then ships the results to S3 for the coordinator.

Also, just for posterity, some minor cleanup edits are made to DRY multiple legs that have shown themselves to require some common code.

### Why

This is part of epic #706 and mostly seeks to ensure dependable benchmark execution, artifact capture, and comparison. Nothing else ever ran the benchmark suite, and having it as part of our pre-release evaluation is a sensible place to check.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/load-test-coordinator.yml
+++ b/.github/workflows/load-test-coordinator.yml
@@ -54,10 +54,30 @@ jobs:
           SHA="$(git rev-parse HEAD)"
           PR="$(gh pr list --repo "${{ github.repository }}" --state open \
                   --base main --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
+
+          # go-bench baseline: latest published release, falls back to the highest v* tag.
+          BASELINE="$(gh api "repos/${{ github.repository }}/releases/latest" --jq .tag_name 2>/dev/null || true)"
+          if [ -z "$BASELINE" ]; then
+            BASELINE="$(git ls-remote --tags --refs origin 'v[0-9]*' \
+              | sed 's|.*refs/tags/||' | sort -V | tail -n 1 || true)"
+          fi
+          [ -n "$BASELINE" ] || echo "::warning::no baseline release found; the go-bench leg will fail"
+          echo "go-bench baseline: ${BASELINE:-<none>}"
+
+          # Per-leg sizing rides in the roster; ec2-leg.yml receives it verbatim.
+          LEGS="$(jq -nc --arg baseline "$BASELINE" '[
+            {label: "Apply-load ingestion", run_label: "apply-load",
+             script: "cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh",
+             instance_type: "c5.2xlarge", root_volume_gb: 500, budget_minutes: 225, extra_env: ""},
+            {label: "Go endpoint benchmarks", run_label: "go-bench",
+             script: "cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/run-go-bench.sh",
+             instance_type: "c5.2xlarge", root_volume_gb: 50, budget_minutes: 90,
+             extra_env: ("export BASELINE_REF=" + $baseline)}
+          ]')"
           {
             echo "target_sha=$SHA"
             echo "pr_number=$PR"
-            echo 'legs=[{"label":"Apply-load ingestion","run_label":"apply-load","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh"}]'
+            echo "legs=$LEGS"
           } >> "$GITHUB_OUTPUT"
 
   # The fan-out: one matrix entry per leg.
@@ -74,6 +94,10 @@ jobs:
       target_ref: ${{ needs.plan.outputs.target_sha }}
       run_label: ${{ matrix.leg.run_label }}
       leg_script: ${{ matrix.leg.script }}
+      instance_type: ${{ matrix.leg.instance_type }}
+      root_volume_gb: ${{ matrix.leg.root_volume_gb }}
+      budget_minutes: ${{ matrix.leg.budget_minutes }}
+      extra_env: ${{ matrix.leg.extra_env }}
 
   report:
     name: Aggregate + report
--- a/.github/workflows/load-test-coordinator.yml
+++ b/.github/workflows/load-test-coordinator.yml
@@ -72,11 +72,11 @@ jobs:
           echo "go-bench baseline: ${BASELINE:-<none>}"
 
           # Unquoted heredoc so $BASELINE expands into the go-bench entry;
-          # jq -c compacts and validates.
+          # Temporarily disabled while iterating on go-bench (restore before merge):
+          #   {"label":"Apply-load ingestion","run_label":"apply-load","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/ingest-load-test/run-load-test.sh","budget":225},
+          #   {"label":"Backfill ingestion","run_label":"backfill","script":"cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/backfill-test/run-backfill.sh","budget":345},
           LEGS="$(jq -c . <<EOF
           [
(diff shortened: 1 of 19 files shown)


cmd/stellar-rpc/internal/db/transaction_test.go
```diff
@@ -381,13 +381,14 @@ func BenchmarkTransactionFetch(b *testing.B) {
 	require.NoError(b, write.Commit(lcms[len(lcms)-1], nil))
 	reader := NewTransactionReader(log, db, passphrase)
 
-	randoms := make([]int, b.N)
-	for i := 0; b.Loop(); i++ {
```
cjonas9, line 385: Claude flagged this as incorrect/outdated and it looks pretty valid to me. Citing the std lib:

> [benchmark.go:428](https://go.dev/src/testing/benchmark.go) — on the first b.Loop() call, the framework sets b.N = 0 (comment: "Within a b.Loop loop, we don't use b.N (to avoid confusion)").
[benchmark.go:453](https://go.dev/src/testing/benchmark.go) — b.N only gets its real value (b.N = int(b.loop.n)) after Loop returns false and the run is done.
The doc comment at [benchmark.go:488](https://go.dev/src/testing/benchmark.go): "After Loop returns false, b.N contains the total number of iterations that ran" — i.e. it's meaningless before then.

It intuitively makes some sense to me as I don't think b.Loop() even knows how many times it's gonna run in advance 😅

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -0,0 +1,276 @@
...
+		CandidateFails: candidateFails,
+		RawLogsPrefix:  "s3://" + env["BUCKET"] + "/" + path.Dir(env["RESULT_KEY"]) + "/",
+	}
+	if err := os.WriteFile(env["RESULTS_FILE"], []byte(renderMarkdown(report)), 0o644); err != nil {
+		return bail("writing results markdown: %v", err)
+	}
+	// The shell wrapper publishes the fail result from resultsFile on non-zero exit.
+	if len(candidateFails) > 0 {
+		return fmt.Errorf("candidate benchmarks failed in %s", strings.Join(candidateFails, ", "))
+	}
```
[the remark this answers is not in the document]
cjonas9: intended behavior

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness"
+)
+
+const legTitle = "Go endpoint benchmarks"
+
+// auditedBenches is the roster of benches for this test (excludes benchmarks
+// that are integration tests).
+var auditedBenches = []struct {
```
Shaptic: I'm confused why we need to name them individually instead of running them all at once

cjonas9: Oops, good catch, this is wasted space. I wasn't up on being able to make it a skip-based list (since we have to skip the integration test benchmarks + there was one that was a little broken) so probably just defaulted to this initially and didn't circle back to it between this + blaster test.
Gonna fix this now!

cmd/stellar-rpc/internal/db/transaction_test.go
Shaptic: Yeah that makes sense to me, too! Maybe the author (might be me honestly) thought the first b.Loop set the stage for the next loop to iterate the same amount.

cjonas9: Funny enough, this actually came back to bite me today as I was trying to get CI passing for this. Since it's not merged to main and this test does a candidate vs. current comparison, this test is always gonna fail until the current release has a fix to this in place.
Obviously this just means that for now, this one bench test can live in the deny list, but thought it was worth bringing up here so there wasn't a question about that.

.github/workflows/load-test-coordinator.yml
```diff
@@ -62,14 +62,29 @@ jobs:
             PR="$(gh pr list --repo "${{ github.repository }}" --state open \
                     --head "$REF" --json number --jq '.[0].number // ""' 2>/dev/null || true)"
           fi
-          LEGS='[
```
cjonas9, line 65: revert before merge to main: this was done to reduce resource waste on tests we don't care about (that are already merged and passing)

Shaptic (approved):
👏

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -0,0 +1,295 @@
...
+// and returns the checked-out SHA.
+func checkoutBaseline(ctx context.Context, dir, repo, ref string) (string, error) {
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return "", err
+	}
+	for _, args := range [][]string{
+		{"init", "-q"},
+		{"remote", "add", "origin", "https://github.com/" + repo + ".git"},
+		{"fetch", "--depth", "1", "origin", ref},
+		{"checkout", "--detach", "FETCH_HEAD"},
```
Shaptic, line 156: `git clone` not good enough for ya boss? 😆

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -0,0 +1,295 @@
...
+			continue
+		}
+		key := prefix + "/" + name
+		if _, err := client.PutObject(ctx, &s3.PutObjectInput{
+			Bucket:      &bucket,
+			Key:         &key,
+			Body:        bytes.NewReader(body),
+			ContentType: aws.String("text/plain"),
+		}); err != nil {
+			logger.Warnf("uploading s3://%s/%s: %v", bucket, key, err)
```
Shaptic: would be good to log this in either case so you get a sense of progress

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/main.go
```diff
@@ -0,0 +1,10 @@
+// Command runner runs the audited endpoint Go benchmarks on the box for the
+// release candidate and a baseline release, and publishes a benchstat
+// comparison of the two.
+package main
+
+import "github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/harness"
+
+var logger = harness.NewLogger()
```
Shaptic, line 8: Unused? Am I trippin

Shaptic (approved):

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
cjonas9: I wish it was here! `git clone` takes `--branch <ref>` for a branch or tag name ref, but our ref here is an arbitrary commit SHA

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/go-bench/runner/main.go
cjonas9: I don't get an unused warning here oddly enough, and I don't see that the linter picked anything like that up. Maybe a stale cache in your IDE or something?

PR #875, opened 2026-07-20
Karthik/merge main into feature full history 2026 07 20

karthikiyer56:
## TL;DR

- Merges 17 commits from `origin/main` into `feature/full-history` and
  brings the whole branch onto go1.26. Only two spots were still on go1.25:
  the Docker build stage (`cmd/stellar-rpc/docker/Dockerfile`,
  `GO_VERSION=1.25.8`) and an EC2 bootstrap script that came in with the
  merge itself (`.../perf-eval/bootstrap-common.sh`, `GO_VERSION=1.25.11`)
  — both bumped to `1.26.5`. `go.mod`, CI's `setup-go` action, and
  `codeql.yml` were already on go1.26 before this PR.
- Along the way, found and fixed **5 stacked, pre-existing macOS CI bugs**
  that had silently broken `Build (macos-latest, ...)` on *every* PR merged
  into this branch since 2026-06-30 — none caused by this merge, all
  verified end-to-end on real hardware before pushing.
- Dropped `windows-latest` from the Build matrix — it never actually had
  working RocksDB support and isn't a real deployment target (see the
  "Windows build" section for the full story).
- **All CI checks now pass**, including the required-status gate.
- **Post-review follow-up (`00530d6b`):** addressed every inline review comment, hardened the native-lib CI against runner-image / toolchain rotation, adopted `grocksdb_clean_link` (drops the snappy/lz4/zlib link dependency outright — supersedes macOS fixes 4–5), and de-duplicated the zstd/RocksDB install scripts.

### Review follow-up (commit `00530d6b`)

Post-approval changes addressing @tamirms's inline comments and hardening the
CI/CD flow against unrelated future churn. Verified end-to-end on macOS arm64
(native), macOS x86_64 (cross), and Linux amd64/arm64 (Docker) before pushing.

- **Runner-rotation immunity:** export `ImageOS` on macOS so the
  libzstd/librocksdb caches invalidate when the runner image rotates; hash
  `install-zstd.sh` into the librocksdb cache key so a zstd bump forces a
  matching rebuild.
- **`grocksdb_clean_link`:** build the grocksdb bindings with
  `-tags grocksdb_clean_link` via `GOFLAGS` (Makefile + `setup-go`), dropping
  the default `-lsnappy -llz4 -lz`. Our librocksdb is built with zstd only
  (verified with `otool`/`ldd`), so those libraries were never used — every
  install of them is removed (CI action, scripts, Dockerfile). **Supersedes
  macOS fixes 4–5 below and the review note about the implicit Homebrew search
  path:** there is no snappy/lz4 left to find.
- **Other review comments:** normalize the cache-key arch to Go naming so jobs
  with/without `go-arch` share one cache; drop the SIP-stripped
  `DYLD_LIBRARY_PATH` on macOS; declare the previously-dead `go-version` action
  input; pin the three cgo-lib jobs (`golangci-lint`, `codeql`,
  `dependency-sanity-checker`) to `ubuntu-24.04`; trim comments that narrated
  this PR's history.
- **Script simplification:** collapse the duplicated Linux/macOS branches in
  `install-zstd.sh` / `install-rocksdb.sh` into one flow, drop the
  source-tarball checksum verification, and add `WITH_TOOLS` to `install-rocksdb.sh` so the Docker
  RocksDB stage calls the script instead of duplicating the build.

### What

- Merges `origin/main` (17 commits, including the v27.1.0/v27.1.1 releases, the
  golang.org/x/crypto and golang.org/x/net CVE patches, and the new database
  backfill / ingestion load-test performance-evaluation infrastructure) into
  `feature/full-history`.
- Brings the whole branch onto go1.26 (see below) — `go.mod` was already there,
  but two spots that install their own Go toolchain had been missed.
- Fixes a pre-existing, unrelated macOS CI break, and drops Windows from the
  Build matrix (see below).

### Conflict resolution

- `go.mod` / `go.sum`: kept `go 1.26` and this branch's RoaringBitmap/stellar-sdk
  pins, merged in main's new AWS SDK (S3/SSM) dependencies, then regenerated
  both files with `go mod tidy` rather than hand-picking transitive versions.
- `Cargo.lock`: regenerated with `cargo check --workspace` against the merged
  `Cargo.toml` (this branch never touched `Cargo.toml`, so main's stellar-xdr
  v27 / soroban-env-host v27 bump applied cleanly).
- `cmd/stellar-rpc/internal/rpcdatastore/ledger_reader_test.go`: both branches
  added an identical `GetLedgerRaw` mock independently; kept this branch's
  version since it also has an explanatory doc comment.

### go1.26 bump

- `cmd/stellar-rpc/docker/Dockerfile`: build-stage `GO_VERSION` `1.25.8` → `1.26.5`.
- `cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/bootstrap-common.sh`:
  EC2 bootstrap `GO_VERSION` `1.25.11` → `1.26.5` (this script came in with the
  merge itself).
- Everything else (`go.mod`, `.github/actions/setup-go`, `codeql.yml`) was
  already on go1.26.

### macOS CI fix (pre-existing, unrelated to the merge)

`Build (macos-latest, ...)` has failed on every PR merged into
`feature/full-history` since 2026-06-30 (bisected against `#818` passing /
`#819` failing) — unrelated to this merge or the go1.26 bump. Five stacked
problems, each found only after fixing the last:

1. **Headers not found.** `setup-go` installed libzstd/librocksdb via
   Homebrew on macOS but never set `CGO_CFLAGS`/`CGO_LDFLAGS`, relying on
   Homebrew's prefix being on clang's default search path by accident. That
   broke when GitHub rotated the `macos-latest` runner image (confirmed via
   runner-image metadata: `macos-26-arm64` on the last passing run vs.
   `macos-15-arm64` on the first failing one). Fixed by resolving the
   Homebrew prefix explicitly.
2. **Wrong RocksDB version.** Macos installed RocksDB via unpinned `brew
   install rocksdb`, which now pulls Homebrew's current 11.1.2 — but
   `grocksdb v1.10.7` calls a C API function that RocksDB 11.x removed.
   Fixed by building RocksDB from source pinned to 10.9.1, same as Linux.
3. **Wrong architecture.** Disabling `fail-fast` (below) let
   `x86_64-apple-darwin` reach its own conclusion for the first time,
   revealing that `macos-latest` runners are arm64 hardware — that matrix
   leg cross-compiles, but the from-source RocksDB build (and Homebrew's
   zstd) only produced arm64 libraries. Fixed by passing
   `-DCMAKE_OSX_ARCHITECTURES` (computed from the job's `go-arch`) through
   to both libraries, and moving zstd to a pinned from-source build too.
4. **Missing snappy/lz4.** Fixing (3) initially dropped `brew install
   snappy lz4` since RocksDB's own cmake build never references them — true,
   but `grocksdb`'s Go bindings hard-code `-lsnappy -llz4 -lz` in their own
   `#cgo LDFLAGS` unconditionally regardless of what the linked library
   supports, so the final Go link fails outright without them present at
   all (even though nothing calls into them). Reinstated.
5. **Same missing-snappy failure, different cause.** The reinstated `brew
   install snappy lz4` lived inside the cache-gated "build RocksDB from
   source" step, so it only ran on a cache *miss*. The moment a warm
   RocksDB cache from an earlier push in this same PR got restored on a
   fresh runner, that whole step was skipped — taking the brew install
   with it — and the exact same link error came back. Moved it to its own
   unconditional step, mirroring the pattern Linux already used for
   exactly this reason.

Verified end-to-end locally on real macOS/arm64 hardware for fixes 1–4 —
including cross-compiling `GOARCH=amd64` against x86_64-targeted
from-source libraries and actually **running** the resulting binary through
Rosetta (not just linking it). Fix 5 was a CI cache-interaction bug specific
to GitHub Actions' cache restore behavior, not something reproducible
locally; verified instead via two consecutive full CI runs — the first
confirmed the regression, the second (after the fix) came back fully green.

### Windows build — what actually happened, in plain terms

`Build (windows-latest, ...)` also started failing once `fail-fast` came
off (below) — but it's a different, separate, pre-existing issue, not
something this PR broke. This PR drops it from the matrix rather than
fixing it. Explained simply:

- This project uses RocksDB, a C++ database library, to store history data.
  Since it's not Go code, something has to compile it separately and hand
  it to the Go build. On Linux and macOS, our CI scripts always did that
  compiling themselves.
- **On Windows, nothing ever did.** CI has never had a step that builds or
  installs RocksDB for Windows — not recently, not ever, going back to when
  RocksDB support was first added to this repo in April.
- This didn't show up as a failure for months, for two stacked reasons:
  - Until late June, the Windows build never actually needed to compile the
    RocksDB-using code at all — that code existed in the repo, but wasn't
    yet wired into the parts of the program that get built every time. So
    there was nothing for the missing RocksDB setup to break.
  - Once that wiring landed (around 2026-06-30, the same window the macOS
    break above started in), Windows needed RocksDB for the first time —
    and would have failed immediately. But it never got the chance to: the
    macOS build was *also* failing (for its own, unrelated reason — the
    Homebrew/runner-image issue above), and GitHub's Build matrix was set
    to cancel every other platform the instant any one of them failed
    (`fail-fast: true`). Windows kept getting cancelled before it could run
    long enough to show its own problem — every single time, for three
    straight weeks, on every PR merged into this branch.
- Fixing the macOS build required turning `fail-fast` off, so that one
  platform failing stops hiding whether the others are actually fine. The
  moment that happened, Windows finally got to run all the way through for
  the first time — and immediately hit the "RocksDB isn't installed" wall
  that had been there all along, just never visible.
- Separately, and more fundamentally: this project's own design docs
  already note that Windows doesn't reliably support a filesystem trick
  ("deferred unlink" — deleting a file while it's still memory-mapped) that
  the full-history storage engine depends on, with no workaround ever
  built for it. So Windows was never really a supported target for this
  feature in the first place — CI just hadn't gotten around to reflecting
  that.
- **Net: Windows was never solid.** It looked fine only because it was
  never asked to do the hard part, and once it was, its failure got masked
  by a coincidental, unrelated failure on another platform. There's no
  regression to fix here — just a gap that was always there. Given it was
  default CI-matrix boilerplate rather than an actual deployment target (no
  Windows packaging or release artifact exists anywhere in this repo),
  this PR removes the `windows-latest` leg entirely instead of building out
  real RocksDB support for it.

### CI matrix fix (also pre-existing, also unrelated)

Set `fail-fast: false` on the Build matrix so every platform's actual
result is visible instead of being hidden by whichever platform happens to
fail first — this is also what let macOS fixes (3) and (4) above surface,
and what revealed the Windows situation described above. Kept even after
removing Windows: it's independently good practice for the remaining
platforms, not just a Windows workaround.

### Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (integration tests self-skip without
      `STELLAR_RPC_INTEGRATION_TESTS_ENABLED`)
- [x] `cargo check --workspace` / `cargo test --workspace`
- [x] Verified the `go1.26.5` Docker tarball URL resolves and confirmed the
      Dockerfile's `ARG`/curl substitution produces a working `go1.26.5`
      toolchain in a container
- [x] Reproduced macOS CI fixes 1–4 end-to-end locally before pushing each
      one (see above), including a cross-compiled build+run for the
      architecture/snappy fixes
- [x] Confirmed no other workflow's `setup-go` call site runs on Windows
      before simplifying its now-dead Windows conditional
- [x] **All CI checks pass**, including the `complete` required-status
      gate: both `Build (macos-latest, ...)` legs, both
      `Build (ubuntu-24.04, ...)` legs, `Integration tests (P26)` and
      `(P27)`, `Unit tests`, `golangci-lint`, `dependency-sanity-checker`


.github/actions/setup-go/action.yml
```diff
@@ -16,63 +16,114 @@ runs:
...
 
+    # grocksdb's Go bindings hard-code `-lsnappy -llz4 -lz` in their own
+    # #cgo LDFLAGS unconditionally, regardless of what librocksdb itself
+    # was actually compiled with — see scripts/install-rocksdb.sh for
+    # detail. Same reasoning as the Linux step above for why this can't
+    # live inside the cache-gated build step below.
+    - name: Install RocksDB link deps (macOS)
+      if: runner.os == 'macOS'
+      shell: bash
+      run: brew install snappy lz4
```
tamirms: After the switch to from-source builds, nothing on the link path points at Homebrew anymore: the macOS `CGO_LDFLAGS` below lists only `~/.zstd/lib` and `~/.rocksdb/lib`. So the `-lsnappy` and `-llz4` that grocksdb emits resolve only because the current runner image's toolchain happens to search the Homebrew prefix, which is the same implicit dependency that broke zstd/RocksDB when the image rotated (your fix 1), just moved to a different pair of libraries. Appending `-L"$(brew --prefix)/lib"` to the macOS `CGO_LDFLAGS` would make it explicit; appending rather than prepending keeps the pinned `~/.zstd` build winning over any zstd the image's brew already has.

A related nuance: the x86_64 leg links against these arm64-only bottles, which works only because the linker tolerates wrong-architecture dylibs whose symbols are never referenced. That holds today, but it is one more reason to make the search path explicit rather than inherited.

Longer term, grocksdb's `grocksdb_clean_link` build tag drops the `-lsnappy -llz4 -lz` requirement entirely, but that would touch every `go build` invocation (Makefile, Dockerfile, dev setup), so a follow-up at most.

.github/actions/setup-go/action.yml
```diff
@@ -16,63 +16,114 @@ runs:
...
+
     # --- libzstd ---
     - name: Cache libzstd
-      if: runner.os == 'Linux'
+      if: runner.os == 'Linux' || runner.os == 'macOS'
       id: cache-zstd
       uses: actions/cache@v4
       with:
         path: ~/.zstd
         key: libzstd-1.5.7-${{ runner.os }}-${{ env.ImageOS }}-${{ inputs.go-arch || runner.arch }}-${{ hashFiles('scripts/install-zstd.sh') }}
```
tamirms: `ImageOS` is set on macOS runners too (for example `macos15`), but the export step above is gated to Linux, so on macOS this key carries an empty segment where the Linux keys have the image version. Since the original break was triggered by a runner image rotation, extending the export to macOS would let these native-lib caches invalidate automatically when the image (and its SDK/toolchain) rotates, instead of restoring a build made on the previous image.

.github/actions/setup-go/action.yml
```diff
@@ -16,63 +16,114 @@ runs:
...
+      run: brew install snappy lz4
+
     - name: Cache librocksdb
-      if: runner.os == 'Linux'
+      if: runner.os == 'Linux' || runner.os == 'macOS'
       id: cache-rocksdb
       uses: actions/cache@v4
       with:
         path: ~/.rocksdb
         key: librocksdb-10.9.1-${{ runner.os }}-${{ env.ImageOS }}-${{ inputs.go-arch || runner.arch }}-${{ hashFiles('scripts/install-rocksdb.sh') }}
```
tamirms: Small cache inefficiency, applies to the zstd key above too: jobs that call this action without `go-arch` fall back to `runner.arch`, which uses GitHub's naming (`X64`), while the Build legs pass Go's naming (`amd64`). The Unit tests job and the Build amd64 leg therefore build and cache the same ubuntu libraries twice under different keys. Mapping `runner.arch` to Go naming in the fallback would let them share.

.github/actions/setup-go/action.yml
```diff
@@ -16,63 +16,114 @@ runs:
...
         echo "CGO_LDFLAGS=-L$HOME/.zstd/lib -L$HOME/.rocksdb/lib" >> $GITHUB_ENV
         echo "LD_LIBRARY_PATH=$HOME/.zstd/lib:$HOME/.rocksdb/lib" >> $GITHUB_ENV
 
+    - name: Configure CGO for native libs (macOS)
+      if: runner.os == 'macOS'
+      shell: bash
+      run: |
+        echo "CGO_CFLAGS=-I$HOME/.zstd/include -I$HOME/.rocksdb/include" >> $GITHUB_ENV
+        echo "CGO_LDFLAGS=-L$HOME/.zstd/lib -L$HOME/.rocksdb/lib" >> $GITHUB_ENV
+        echo "DYLD_LIBRARY_PATH=$HOME/.zstd/lib:$HOME/.rocksdb/lib" >> $GITHUB_ENV
```
tamirms: Worth knowing: SIP strips `DYLD_*` variables whenever a protected binary like `/bin/bash` is exec'd, so this value most likely never reaches the step processes at all. It is harmless today because the macOS legs only compile and nothing in CI loads these dylibs at runtime, but if macOS ever runs tests, this line will not be doing what it appears to do. Either dropping it or noting the limitation would save someone a confusing debugging session later.

.github/workflows/stellar-rpc.yml
```diff
@@ -37,6 +37,13 @@ jobs:
   build:
     name: Build
     strategy:
+      # One platform failing shouldn't hide whether the others actually
+      # pass. This used to matter a lot here: the Windows leg that used to
+      # be in this matrix (see below) had never had zstd/RocksDB CI
+      # support, and stayed invisibly broken for weeks because it kept
+      # getting cancelled by macOS failing first, before fail-fast was
+      # turned off.
+      fail-fast: false
```
tamirms, line 41: The first sentence carries all the ongoing value here; the rest narrates this PR's history, which already lives in the commit message and PR description, and will read as a puzzle once the Windows story fades from memory. Same thought for the Windows note below, which is worth keeping so nobody re-adds the leg but could be halved, and for the "as of writing" phrasings in scripts/install-rocksdb.sh, which will age poorly.

tamirms (commented):
I reviewed the conflict resolutions and the commits on top of the merge. The merge checks out mechanically: I re-ran it from the same parents and the result is byte-identical to the clean auto-merge everywhere outside the four conflicted files. `go mod tidy` at the head is a no-op, main's go-stellar-sdk pin is an ancestor of the pin this branch kept (so nothing from main's SDK bump was dropped), the regenerated Cargo.lock is byte-identical to main's, and the x/crypto + x/net CVE bumps survive. The CI fixes read well and the one-problem-per-commit split made them easy to verify. Inline comments below, nothing blocking.

tamirms (approved):
PR looks good overall. I left some comments about some minor findings from claude

karthikiyer56:
> PR looks good overall. I left some comments about some minor findings from claude

I was doing a second fresh claude pass and it flagged some of the same things as you. 
Ima update the pr and then merge it rightaway, since you already approved

.github/actions/setup-go/action.yml
karthikiyer56: Done in `00530d6b` — took the longer-term path you flagged rather than the explicit `-L`: grocksdb is now built with `-tags grocksdb_clean_link` (via `GOFLAGS` in the Makefile + this action), which drops `-lsnappy -llz4 -lz` from the link entirely. Our librocksdb is zstd-only (confirmed with `otool`/`ldd`), so there's no implicit search path left to depend on, and `brew install snappy lz4` / `apt install libsnappy-dev …` are gone from CI, the scripts, and the Dockerfile. It also removes the x86_64 leg's reliance on the linker tolerating wrong-arch bottles.

.github/actions/setup-go/action.yml
karthikiyer56: Done in `00530d6b` — the ImageOS export now runs on macOS too, so the libzstd/librocksdb cache keys carry `macosNN` and invalidate on an image rotation instead of restoring libs built on the previous image.

.github/actions/setup-go/action.yml
karthikiyer56: Done in `00530d6b` — added a `Resolve target architecture` step that maps `runner.arch` (X64/ARM64) to Go's naming (amd64/arm64) and uses it in both native-lib cache keys, so the Unit tests job and the amd64 Build leg now share one cache instead of building the libs twice.

.github/actions/setup-go/action.yml
karthikiyer56: Done in `00530d6b` — dropped `DYLD_LIBRARY_PATH` (SIP strips `DYLD_*`, and CI only links the binary, never runs it). Left a one-line note in its place so nobody re-adds it expecting it to work.

.github/workflows/stellar-rpc.yml
karthikiyer56: Done in `00530d6b` — trimmed the fail-fast and Windows comments to the durable reasons and dropped the PR-history narration, and removed the "as of writing" phrasings in `scripts/install-rocksdb.sh`. The rest of the CI/script comments got the same pass.

PR #893, opened 2026-07-23
Reorganize stellar-rpc into shared/rpcv1/rpcv2 trees and build both binaries (#876)

karthikiyer56:
## TL;DR

- Implements #876: one repo → shared code + two binaries.
  - `stellar-rpc` (v1, SQLite) and `stellar-rpc-v2` (v2, full-history/RocksDB), each with its own Make target and Dockerfile.
- Mostly mechanical relocation (`git mv` + import rewrites). Real code change is confined to:
  - a small set of seams pulled out of v1-owned packages (tabled below)
  - post-reorg cleanup commits (dead code, shared `version` command, layout alignment)
- Enforced guarantees:
  - v1 binary's dep closure is grocksdb-free (`go list -deps` = 0)
  - trees can't import each other; shared imports neither — depguard, negative-tested
- Downstream unaffected: binary name `stellar-rpc`, `build-stellar-rpc` target (alias kept for the Jenkins packager), `Dockerfile.release`.

## The new layout

All paths in this document are relative to `cmd/stellar-rpc/`.

| Tree | Path | Holds |
|---|---|---|
| shared | `internal/` | backend-agnostic code:<br>`internal/methods`, `internal/store`, `internal/jsonrpc`,<br>`internal/host`, `internal/network`, `internal/preflight`, ... |
| v1 | `internal/rpcv1/` | `config`, `daemon`, `ingest`,<br>`feewindow`, `sqlitedb`, `integrationtest` |
| v2 | `internal/rpcv2/` | everything that was `internal/fullhistory/`<br>+ v2-only primitives |
| v1 binary | `rpcv1/` | → `stellar-rpc` |
| v2 binary | `rpcv2/` | → `stellar-rpc-v2` (new) |

## Shared seams pulled out of v1 packages

| New shared home | Pulled out of | What & why |
|---|---|---|
| `internal/store` | `internal/db` | read interfaces: `LedgerReader(Tx)`, `TransactionReader`, `EventReader`<br>+ plain types (`Transaction`, topic filters, `LedgerRange`, `LedgerInfo`)<br>→ the contract both backends implement |
| `internal/store/feestats.go` | `internal/feewindow` | `FeeDistribution` + new 2-method `FeeStats` interface<br>→ `getFeeStats` was the one handler on a concrete v1 type<br>→ #881's v2 fee windows implement the same contract |
[...]

Issue #887, opened 2026-07-21
EventReader adapter


karthikiyer56:
## What

A concrete type in `internal/rpcv2/adapters` satisfying `store.EventReader` (one method, `GetEvents`), backed by the per-chunk event engine (`stores/event/query.go`), reached through a `query.ReadView`.

`EventReaders` does the chunk split, the per-chunk clip, and the reader opens. `event.Query` does all the filtering. This ticket is the glue between them plus the cursor handling — and **the cursor handling is the only part that can fail silently.**

## Hard invariant this ticket relies on

- **A `getEvents` call never spans more than 2 chunks.** `LedgerScanLimit = 10000` equals `LedgersPerChunk`, so the window covers at most two. `EventParts` therefore returns 1 or 2 parts, and resume is only ever "in part 1" or "in part 2".
- Enforced in `methods/get_events.go` today; keep it enforced when #889 wires the handler.

## What the read view provides

| Call | File | What it does for us |
|---|---|---|
| `view.EventReaders(dir, lo, hi)` | `query/scan.go` | Clamps to the view's range, splits by chunk, **clips each chunk's bounds**, opens each reader up front. Returns `[]event.BoundedReader{Reader, From, To}` in scan order. |
| `view.Events(c)` | `query/resolve.go` | The uniform `event.Reader` across tiers. Hot handles are read-write, so the events facade is warmed (a read-only open has none). |

## The six steps

```go
// 1. FILTERS: store.TopicFilters + contractIDs -> []event.Filter
// 2. RANGE:   EventReaders(Ascending, cr.Start.Ledger, cr.End.Ledger-1)   // End is EXCLUSIVE
// per part:
// 3. IDS:     ofs := p.Offsets(); clip; event.IDRangeForLedgers(ofs, lo, hi)
// 4. ENGINE:  event.QueryPage(ctx, p.Reader, filters, QueryOptions{Range: idRange, MaxEvents: batch})
//             in a pull loop. batch (512) is a FETCH SIZE, not a result budget: the handler's
//             limit lives inside f and never crosses store.EventReader. QueryPage reports
//             NextStart + Exhausted — resume state only the engine can know (payloads carry
//             no event IDs, and MaxEvents applies before its defensive post-filter) — so
//             pages never overlap and exhaustion is engine-reported, never inferred.
// 5. CURSOR:  build protocol.Cursor per payload; skip cursor < cr.Start; stop at >= cr.End
[...]

Issue #886, opened 2026-07-21
TransactionReader adapter


karthikiyer56:
## What

A concrete type in `internal/rpcv2/adapters` satisfying `store.TransactionReader` (one method, `GetTransaction`), backed by the query router's by-hash probe surface (#865, implemented in PR #894; design in PR #843): probe the hot transaction-hash indexes first, then each window's frozen index, verify candidates against the full hash, and gate the result to the read view's `[floor, latest]` range.

The probe algorithm already exists in `stores/txhash/read_assembly.go` — this ticket wires it up and gates it.

## What the read view provides

| Call | File | What it does for us |
|---|---|---|
| `view.HotTxHashIndexes()` | `query/txlookup.go` | `[]txhash.HashIndex` for every published hot chunk, newest first. Ready to pass straight into `NewTxReader`. Registry-owned — do not close. |
| `view.ColdTxHashIndexCoverages()` | `query/txlookup.go` | Frozen coverage **keys**, newest `Hi` first, read through the snapshot so a concurrent index rebuild can't change the probe set mid-request. |

## What this ticket owns

- **Open the cold `.idx` files.** Coverages are keys, not readers. Per coverage: `txhash.OpenColdReader(layout.TxHashIndexFilePath(cov))`. These are not view-owned — we close them. (Unless we add `ColdTxIndexes()` — see the open decision below.)
- **`txhash.LedgerSource`** — one method, `GetLedgerRaw(seq)`. Plain per-candidate resolution: `view.Ledgers(chunk.IDFromLedger(seq))`, one open per candidate. No caching layer: `TxReader.scan` takes one candidate per index, index ranges are disjoint, and the probe stops at the first verified match, so no chunk is ever probed twice.
- **The floor/latest gate.** `HotTxHashIndexes` is deliberately unfiltered (its doc comment says so): a hot handle can predate the view's floor, and a cold index hit is a **fingerprint**, not an exact match. A resolved ledger outside `[OldestLedger(), LatestLedger()]` must become `store.ErrNoTransaction`, not an error and not a served result.
[...]

Issue #885, opened 2026-07-21
LedgerReader + LedgerReaderTx adapter


karthikiyer56:
## What

A concrete type in `internal/rpcv2/adapters` satisfying `store.LedgerReader` (from `internal/store`), backed by the v2 query router (#865, implemented in PR #894; design in the merged PR #843, "docs(full-history): query routing design"). Its `NewTx` opens one `query.ReadView`; the returned handle satisfies `store.LedgerReaderTx`, and `Done()` releases that read view.

## Why this one interface serves four endpoints

getLedgers, getLatestLedger, getTransactions, and getEvents all take a `store.LedgerReader` — build it once here rather than once per endpoint.

## Scope

- `GetLedger`, `GetLedgerRange`, `StreamLedgerRange`, `NewTx`, `GetLatestLedgerSequence` — the interface as extracted in #876 (`StreamAllLedgers` was deleted there outright; `GetLedgerCountInRange` stayed behind as a v1-only concrete method).
- The `LedgerReaderTx` returned by `NewTx`: `GetLedger`, `GetLedgerRange`, `BatchGetLedgers`, `Done`.

## What the read view provides

| Call | File | What it does for us |
|---|---|---|
| `registry.NewReadView()` / `view.Release()` | `query/registry.go` | One consistent view per request: frozen latest ledger, frozen handle set, catalog snapshot. `Release` also closes every reader the view opened. |
| `view.ScanLedgers(lo, hi)` | `query/scan.go` | Clamps to the view's range, splits by chunk, **clips each chunk's sub-range**, and returns one flat `iter.Seq2[ledger.Entry, error]`. **Windowed:** holds at most two readers regardless of range — chunk `i` plus a one-ahead open of `i+1`, closing each as the walk passes. The first two chunks resolve at call time so their failures surface there; **a later chunk's failure surfaces mid-stream.** At ≤2 chunks (see the 10k cap below) everything fails up front. |
| `view.OldestLedger()` / `view.LatestLedger()` | `query/bounds.go` · `registry.go` | The two sequence numbers `GetLedgerRange` needs. |
| `view.ClampRange(dir, lo, hi)` | `query/bounds.go` | Returns `*query.RangeError{Requested, Oldest, Latest}` for a leading edge below the floor — maps onto v1's out-of-range error. |

## What this ticket owns

- **`GetLedger(ctx, seq)` — the point read.** No gated single-ledger call exists on `ReadView`. Three guards, all ours:
[...]

PR #894, opened 2026-07-23
Query routing: serving foundation (write + read side)

urvisavla:
## Query routing: serving foundation (write + read side)

Implements the [query routing design](https://github.com/stellar/stellar-rpc/blob/feature/full-history/design-docs/query-routing-design.md): catalog-snapshot read views, the serving registry (latest ledger + shared hot-database handles), hot-DB ownership transfer, lifecycle-executed deferred deletion, and the read-side routing primitives. The registry runs live against real ingestion and lifecycle activity with **zero user-visible behavior** — no endpoint reads it yet, so the concurrency orderings can be validated before any query depends on them. The per-method endpoint adapters and the read server (#772) follow.

> **Naming:** the code and the design doc now share one vocabulary: `ReadView`/`NewReadView` (formerly *Admission*), `Registry` (formerly *Router*), `latestLedger` (formerly the *watermark*). The doc update is part of this PR.

### Write-side foundation

1. **rocksdb primitives** — `NewSnapshot`/`ReleaseSnapshot` + snapshot-pinned `GetAsOf`/`IterateAsOf`, and the non-blocking `CloseIfIdle` drain barrier.
2. **catalog snapshot reads** — snapshot-pinned `*AsOf` accessors; a nil snapshot reads live, so existing call sites are unchanged.
3. **serving registry / handle set / read views** — `Registry` (latest ledger + copy-on-write hot handles), `NewReadView` (three loads: latest ledger, handles, catalog snapshot last — the order is pinned by a snapshot-seam test). Floor from the highest ready hot chunk; an **empty ready scan fails the acquisition** (it marks a broken catalog, and erroring beats serving the widest possible range from broken state).
4. **ownership transfer** — the ingestion loop advances the latest ledger and hands each completed hot DB to the registry instead of closing it at the boundary; the freeze reads completed chunks through the shared handle. Discard demotes then destroys via `CloseIfIdle` at end of run; a reader-busy handle is retained in the registry and the close retried across runs.
5. **deferred deletion** — discard and prune demote in-stage and destroy at end of run behind one grace wait. The **stages run independently** (`errors.Join`) and the destroys are always reached, so a persistently failing freeze — a full disk is the sharp case — cannot wedge the reclamation that would clear it; a failed freeze's own chunk stays protected by eligibility, not stage ordering. Every exported destroy re-checks the durable key before deleting, so an un-demoted ref is skipped (with a warning) instead of deleting a live resource.
6. **startup bootstrap** — the registry publishes handles for ready hot chunks and seeds the live chunk's handle + latest ledger before serving; `Registry.Close` flushes handles on shutdown. Demotions a crashed run left behind need no startup pass: the boundary seed fires a lifecycle run right after boot, and its scans re-collect every demoted key — the same re-discovery that recovers a crash mid-run.

### Read-side foundation

7. **resolution + per-request cold opens** — `ReadView.resolveTier` is the single R1/cold-wins routing decision; `ReadView.Ledgers`/`Events` return a request-owned cold reader or the registry's hot facade, or `ErrUnavailable`. Cold readers are **view-owned**: `Release` closes whatever the view opened, so a request has exactly one mandatory teardown (page limits cap a request at two chunks, so per-chunk early close bought nothing).
8. **bounds** — `OldestLedger` + `ClampRange` (leading edge below floor → `RangeError`; trailing edge truncated at latest / terminated at floor; inverted input → `ErrInvertedRange`).
9. **range helpers** — `ScanLedgers(lo, hi)`, a flat ascending iterator with the per-chunk intersect inside (a caller cannot read past the admitted latest), and `EventParts(dir, lo, hi)`, the per-chunk {chunk, reader, from, to} parts the events engine consumes. Both open the (at most two) chunk readers up front, so the border chunk's open overlaps the first chunk's streaming, and the chunk border is tested once here instead of once per endpoint. `HotTxHashIndexes`/`TxHashIndexCoverages` supply what getTransaction needs from the read view.

### Out of scope

- The **endpoint adapters** — getLedgers, getTransactions, getTransaction, getEvents — and their cursors. With the read-side foundation here, each becomes a thin loop over `ScanLedgers`/`EventParts` or `HotTxHashIndexes`+`TxHashIndexCoverages`.
- Deferred to **#772 (cutover)**: the read server, `getLatestLedger`/`getHealth` mappings, serving metrics (including the deferred-destroy failure counter and the broken-catalog alarm), deriving the real grace period (5m placeholder for now), and the v1→v2 handler cutover + SQLite removal.

### Notes

- Rebased onto the rpcv2 reorganization (#893): `internal/fullhistory` → `internal/rpcv2`, `storage/` flattened, `eventstore` → `stores/event`.
- Full `rpcv2` package green (~220s), including the concurrent-lifecycle e2e; lint clean in the PR's files.


cmd/stellar-rpc/internal/rpcv2/serving/registry.go
```diff
@@ -0,0 +1,248 @@
...
+// the live, still-ingesting chunk, so the one below it is the last complete one).
+// Returns -1 when no hot chunk is ready, meaning nothing is complete yet — the
+// signed convention Retention.FloorAt expects.
+func hotFrontier(cat *catalog.Catalog, snap *rocksdb.Snapshot) (int64, error) {
+	ready, err := cat.ReadyHotChunkKeysAsOf(snap)
+	if err != nil {
+		return 0, err
+	}
+	if len(ready) == 0 {
+		return -1, nil
```
tamirms: Two different situations end up at -1 here. A young store (chunk 0 is ready, nothing complete yet) correctly gets -1 from `ready[last]-1` and serves from the pin. But an empty scan (no ready key at all) can't happen in a working daemon: the resume chunk goes ready before serving starts, and the live key is never demoted. If we ever see it, something is broken, and deriving a floor from it makes a broken store serve with the widest possible range. I'd have the empty case return an error (and count it on an alarm counter, now or with the #772 metrics) instead of -1. On tests: the "no ready hot chunk" case in `TestAdmit_FloorDerivation` currently pins the fallback behavior; I'd split it into one test per case.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// Admit captures a query's view of serving state with three loads, in this order:
+// latest first, the handle set second, the catalog snapshot last. The order makes
+// the snapshot's metadata the newest of the three, so any skew between the handle
+// set and the snapshot resolves safely (see the design's Admission section).
+//
+// The caller MUST call Release when the request completes, including on error
+// paths.
+func (r *Router) Admit() (*Admission, error) {
```
tamirms: The latest-before-snapshot load order here is load-bearing, but no test can catch a swap: `TestAdmit_CapturesStateAtAdmissionInstant` passes either way. One small seam would pin it: a `snapshotFn` field that defaults to `catalog.NewSnapshot`, and a test whose hook advances the watermark from inside the snapshot call. With the right order the admitted `Latest()` is the pre-hook value; with the loads swapped it picks up the new one. Worth adding while this is fresh?

cmd/stellar-rpc/internal/fullhistory/lifecycle/deletion.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+	p.add("index "+cov.Key, func() error { return cat.DestroyTxHashIndexKey(cov) })
+	return nil
+}
+
+// run waits the grace period once, then runs every queued destroy. A destroy that
+// fails is logged and left for the next run's scan to re-discover via its still-
+// demoted key.
+func (p *pendingDeletions) run(ctx context.Context, cfg Config) {
```
tamirms: Failed destroys are logged here but not counted, so a handle that stays stuck across runs only shows up in logs. A small failure counter in `observability.Metrics` would make that watchable. Fine to land it with the #772 metrics work.

cmd/stellar-rpc/internal/rpcv2/lifecycle/startup_sweep.go
```diff
@@ -0,0 +1,93 @@
...
+func StartupSweep(cat *catalog.Catalog, liveChunk chunk.ID) error {
+	if err := sweepTransientHotChunks(cat, liveChunk); err != nil {
+		return err
+	}
+	if err := sweepIndexDebris(cat); err != nil {
+		return err
+	}
+	return sweepPruningChunkArtifacts(cat)
+}
+
```
tamirms: I think we can delete this file. Startup seeds the boundary signal, so a lifecycle run fires right after boot, and that run's scans already re-find everything this sweep cleans up: the discard scan re-collects transient hot chunks, and the prune scan re-collects "pruning" coverages and chunk refs, including a crashed terminal `.bin` sweep. That re-discovery is already how the design recovers from a crash mid-run. All the startup sweep buys is deleting those leftovers a few minutes earlier (the run's destroys wait out the grace period first), and for that it keeps a second deletion path alive. Dropping it also lets `SweepTxHashIndexKey` and the composite `DiscardHotChunk` be deleted once the eager `.idx` sweep is gone (`SweepChunkArtifacts` stays for the `.bin` sweep). The doc's two startup-sweep mentions would then point at the scan re-discovery it already documents.

cmd/stellar-rpc/internal/fullhistory/lifecycle/lifecycle.go
```diff
@@ -156,27 +181,42 @@ func runLifecycle(ctx context.Context, cfg Config, cat *catalog.Catalog, lastChu
...
-		return fmt.Errorf("prune op: %w", err)
+		return fmt.Errorf("prune demote: %w", err)
 	}
 	if prunedArtifacts > 0 {
 		logger.WithField("pruned", prunedArtifacts).Info("lifecycle prune stage complete")
 	}
+
+	// End of run: destroy everything demoted this run — discarded hot handles and
+	// pruned cold files — after one grace wait (design: wait once, then delete).
+	pending.run(ctx, cfg)
```
tamirms: `pending.run` executes only when all three stages succeed, and a failed stage 1 also skips the stage 2/3 scans that re-collect leftover demotions, so a persistently failing run reclaims nothing. Worst case is a full disk: the freeze fails, the scans and destroys never run, the space held by already-demoted resources is never freed, and every later tick fails the same way. The discard and prune scans read only durable state and don't depend on the plan having succeeded, so I'd run all three stages even when one errors, join the errors, and always reach the end-of-run destroys; ctx cancel still aborts everything immediately, including the grace wait. Ingestion keeps running while a failed run waits out the grace (the error only propagates on return), so the cost is a delayed restart of an already-failing lifecycle, and reclaim can't be wedged by a failing stage.

cmd/stellar-rpc/internal/fullhistory/serving/bounds.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if lo > hi {
+		return nil, nil // entirely beyond latest
+	}
+	return chunkSeq(chunk.IDFromLedger(lo), chunk.IDFromLedger(hi), dir), nil
+}
+
+// chunkSeq returns the inclusive chunk ids from first..last in scan order
+// (first <= last). Descending counts down without underflowing at chunk 0.
+func chunkSeq(first, last chunk.ID, dir Direction) []chunk.ID {
```
tamirms: Tiny one: the descending branch counts down with an underflow guard at chunk 0. Building the slice ascending and calling `slices.Reverse` when the direction is descending does the same thing without the tricky loop.

cmd/stellar-rpc/internal/fullhistory/serving/txlookup.go
```diff
@@ -0,0 +1,61 @@
...
+// HotTxIndexes returns the transaction hash index of every published hot chunk,
+// newest chunk first. A hot match is exact and definitive, so the newest indexes
+// are probed first. The returned indexes are router-owned handles; the caller does
+// not close them.
+//
+// It is deliberately unfiltered — every published handle, regardless of the
+// admitted floor/latest. A match can therefore name a ledger in a chunk below the
+// floor (a handle that predates this admission); the lookup's floor/latest gate on
+// the resolved ledger is the only thing that keeps such a match from being served.
+func (a *Admission) HotTxIndexes() []txhash.HashIndex {
```
tamirms: Tiny one: `slices.Sorted(maps.Keys(a.handles.byChunk))` plus `slices.Reverse` replaces the manual append-and-sort for the newest-first id list.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
```diff
@@ -0,0 +1,248 @@
...
+//
+// The Router does not own the catalog; the daemon constructs and closes it.
+type Router struct {
+	catalog   *catalog.Catalog
+	retention geometry.Retention
+
+	// watermark is the newest fully ingested ledger visible to queries. The ingest
+	// loop advances it as the final step of each per-ledger cycle. Queries read a
+	// frozen copy at admission (Admission.Latest), never this live value.
+	watermark atomic.Uint32
```
tamirms: The doc names this value `latest`: its glossary entry and the Router sketch both use it, and "watermark" only appears in the doc as a descriptive word. The code also ends up with two names for one thing, since `Admission` already calls it `latest`. I'd rename the field and methods to `latestLedger` / `SetLatestLedger` / `LatestLedger`, use the same name on the view's accessor, and drop "watermark" from the comments, so there is one term everywhere, matching the concrete ledger names the other docs settled on (`lastCommittedLedger`, `lastCompleteChunk`). Worth updating the doc's glossary entry to `latestLedger` in the same pass, so the doc and the code share the term.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
```diff
@@ -0,0 +1,248 @@
...
+
+// Release releases the admission snapshot back to the catalog.
+func (a *Admission) Release() { a.catalog.ReleaseSnapshot(a.snap) }
+
+// hotFrontier is the last-complete-chunk anchor the floor is derived from: the
+// highest ready hot chunk in the snapshot minus one (the highest ready chunk is
+// the live, still-ingesting chunk, so the one below it is the last complete one).
+// Returns -1 when no hot chunk is ready, meaning nothing is complete yet — the
+// signed convention Retention.FloorAt expects.
+func hotFrontier(cat *catalog.Catalog, snap *rocksdb.Snapshot) (int64, error) {
```
tamirms: Small naming one: this derives the same value `geometry.LastCompleteChunkAt` derives elsewhere (highest complete chunk, signed, -1 when nothing is complete), just from a snapshot's ready keys instead of a ledger number. Naming it in that family, something like `lastCompleteChunkAsOf`, keeps one name for the concept; "frontier" then stays where it already lives, as `FloorAt`'s parameter.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
tamirms: Naming: `Admission` reads oddly as an object a request holds and calls methods on; what it holds is a consistent read-only view of serving state. I'd rename the type to `ReadView` and the constructor to `NewReadView`, mirroring the `NewSnapshot` / `ReleaseSnapshot` pair this PR adds one layer down, with `Release` unchanged. Call sites then read naturally: `view, err := r.NewReadView()`, `defer view.Release()`, `view.LedgerReader(c)`. Worth touching the doc's glossary in the same pass so the "Admission snapshot" entry becomes the read view's snapshot.

cmd/stellar-rpc/internal/fullhistory/serving/resolve.go
```diff
@@ -0,0 +1,114 @@
...
+			return tierHot, db, nil
+		}
+	}
+	return tierNone, nil, nil
+}
+
+// LedgerReader resolves chunk c's ledger store for this request. The returned
+// close releases a cold reader; for the hot tier it is a no-op (the router owns
+// the handle). Returns ErrUnavailable when c has no serving home.
+func (a *Admission) LedgerReader(c chunk.ID) (LedgerReader, func() error, error) {
```
tamirms: A page never spans more than two chunks (the getEvents design caps every query at 10,000 ledgers and says so; getLedgers' limit is 200), which suggests two simplifications together. First, make readers view-owned: whatever a view opens, `Release` closes, and `LedgerReader`/`EventReader` stop returning closers. The request already has exactly one mandatory teardown, so this folds reader cleanup into it instead of adding a second discipline, and with at most two readers per request, per-chunk early close buys nothing. Second, two range helpers that open both chunks' readers immediately, so the border chunk's async open (three files plus the MPHF load for events) overlaps the first chunk's streaming: `ScanLedgers(lo, hi)`, a flat entry iterator for getLedgers/getTransactions with the per-chunk intersect against the clamped range inside, so reading past the admitted `latest` can't be written; and `EventParts(dir, lo, hi)`, a plain one-or-two element slice of {chunk, reader, from, to}, which is exactly what the events engine consumes. Both stores defer validation to first use, so a border chunk's open failure surfaces at its own position, and the borrow rules are unchanged. Adapters reduce to "range and emit" or "loop the parts and run the engine," and the border behavior gets pinned by one test here instead of one per endpoint.

tamirms (commented):
**`backfill/txindex.go:146`** (not part of this diff, but load-bearing for it): The design moved this deletion to the end of the run: the old `.idx` is query-visible, so a request admitted before the rebuild can still open it, and the file has to survive the grace period. That loop deletes it immediately after the swap instead. I think we can just delete the loop: the rebuild already marks the old coverage "pruning", the prune scan picks that up, and the end-of-run destroy removes the file. The eager reclaim only ever bought disk back early during long bootstraps, and losing it costs a handful of superseded files lingering until the first run's destroys. The `.bin` sweep below stays as is, since no query ever reads a `.bin`.

---

**`lifecycle/lifecycle.go:129`** (outside this diff): This PR adds a second floor anchor: the run uses the notified boundary chunk there, and admission now uses the snapshot's ready frontier. They are the same value (the live chunk's ready key commits before the boundary publish), so I'd unify them while both are in front of us: one derivation over a catalog view, used by the run against the live catalog and by admission against its snapshot. The boundary signal then carries no value and becomes a pure wake-up. That leaves one implementation and one test, instead of two anchors that have to stay equal.

---

A few naming nits, batched. `pendingDeletions.run`: "run" is already the lifecycle's central concept, so `pending.run(ctx, cfg)` reads like it executes a lifecycle run; something like `destroyAll` says what it does. `hotHandles`: the doc's glossary calls this the handle set, so `handleSet` keeps the code searchable by the doc's term. `getRO`/`iterateRO`: "RO" means ReadOptions here, but next to `OpenReadOnly` it reads as read-only; `getWith`/`iterateWith` avoids that, and the `ro` parameter already names the options.

---

One note for the adapters PR: the `floor <= ledger <= latest` probe gate and driving every scan by the clamped range are deferred by this PR's scope and carried by `txlookup.go`'s doc comment; flagging them here so they get re-checked when the endpoints land.

cmd/stellar-rpc/internal/fullhistory/startup.go
```diff
@@ -119,11 +120,38 @@ func run(ctx context.Context, cfg StartConfig) error {
...
 	}
 
+	// The serving router holds the query watermark, advanced by the ingestion loop
+	// below. It also owns the shared hot-database handles: the loop publishes them,
+	// the freeze reads completed chunks through them (HotHandle), and the lifecycle
+	// unpublishes them at discard (Router) for deferred deletion to close. It will
+	// back the read server in later work; for now nothing reads it. Constructed per
+	// run — no query survives a restart, so a fresh router each run is fine.
+	router := serving.NewRouter(cat, cfg.Retention)
+	cfg.Exec.Process.HotHandle = router.Handle
```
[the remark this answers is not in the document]
urvisavla: Done in 085b5d72.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
urvisavla: Done in d60b4d48; the alarm metric lands with #772.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
urvisavla: Done in bc76f014.

cmd/stellar-rpc/internal/fullhistory/lifecycle/deletion.go
urvisavla: Deferred to the #772 metrics batch as suggested.

cmd/stellar-rpc/internal/rpcv2/lifecycle/startup_sweep.go
urvisavla: Done in 0cce4ee4; DiscardHotChunk went with it, and SweepTxHashIndexKey followed in 79fb5dcd.

cmd/stellar-rpc/internal/fullhistory/lifecycle/lifecycle.go
urvisavla: Done in c44ee69f.

cmd/stellar-rpc/internal/fullhistory/serving/bounds.go
urvisavla: Done in 5045e073.

cmd/stellar-rpc/internal/fullhistory/serving/txlookup.go
urvisavla: Done in 5045e073.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
urvisavla: Done in 2afb7af3; doc glossary updated in da9d1eeb.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
urvisavla: Done in 5045e073; folded into catalog.LastCompleteChunkAsOf in 6a62c746.

cmd/stellar-rpc/internal/fullhistory/serving/router.go
urvisavla: Done in 2afb7af3 and 5045e073 (NewReadView); the doc section is retitled in da9d1eeb.

cmd/stellar-rpc/internal/fullhistory/serving/resolve.go
urvisavla: Done: 70a74475 (view-owned readers) and 8e98945c (ScanLedgers/EventParts).

cmd/stellar-rpc/internal/rpcv2/query/registry.go
```diff
@@ -0,0 +1,262 @@
...
+	floor        chunk.ID
+	handles      *handleSet
+	snap         *rocksdb.Snapshot
+	catalog      *catalog.Catalog
+
+	// closers releases every cold reader this view opened (hot facades are
+	// registry-owned and never appear here). Appended by the resolve methods,
+	// drained by Release — a view's resources live exactly as long as the view.
+	// A ReadView serves one request on one goroutine; no locking.
+	closers []func() error
```
karthikiyer56, line 221: Could `ReadView` keep a per-chunk lookup alongside `closers`, so `Ledgers`/`Events` return a reader already resolved for this request?

**`registry.go` — `ReadView`:**

```go
ledgers map[chunk.ID]LedgerReader   // new: lookup, cold AND hot
events  map[chunk.ID]event.Reader
closers []func() error              // unchanged: cold only
```

Both maps initialised in `Registry.NewReadView`, same place the other fields are set.

**`resolve.go` — `ReadView.Ledgers`:** an early return, and a store in each branch of the existing switch.

```go
if r, ok := a.ledgers[c]; ok {
	return r, nil
}
// resolveTier unchanged
switch t {
case tierCold:
	// existing OpenColdReader + closers append, then:
	a.ledgers[c] = cr
	return cr, nil
case tierHot:
	hot := db.Ledgers()
	a.ledgers[c] = hot        // no closer — the registry owns this one
	return hot, nil
}
```

Same shape in `ReadView.Events`.

**`registry.go` — `ReadView.Release`:** unchanged. `closers` still holds exactly what it held before, and it has to stay a callback list regardless, since `LedgerReader`/`event.Reader` don't expose `Close`. Hot readers are safe in the map because they never enter `closers`.

**Why:** `getTransactions` fetches its ledgers one at a time, not as a range. A request covers at most 2 chunks but can walk a few hundred ledgers to fill a page, so it asks for a chunk's reader a few hundred times — repeatedly for the same one or two chunks. Each ask re-runs `resolveTier` (two snapshot reads) and, on the cold path, opens a new reader: ~400 catalog reads and up to 200 cold readers held until `Release`, to serve 2 chunks. The `resolveTier` half is paid even when both chunks are hot.

tamirms: The repeated-ask pattern doesn't come up: getTransactions rides `ScanLedgers` (added in this PR) rather than fetching ledgers one at a time. The adapter asks for the whole clamped range once, `ScanLedgers` resolves each chunk once and opens each reader once, and the page streams all its ledgers out of those one or two readers. The only single-ledger fetch in the system is getTransaction's candidate verify, and even that never repeats a chunk: each probed index yields at most one candidate, the indexes cover disjoint chunk ranges, and probing stops at the first verified match. So no request path ever asks for the same chunk twice, and the lookup would never find an entry to reuse.

cmd/stellar-rpc/internal/rpcv2/query/registry_test.go
```diff
@@ -0,0 +1,368 @@
...
+	_, err := r.NewReadView()
+	require.ErrorIs(t, err, catalog.ErrNoReadyHotChunk)
+}
+
+// TestNewReadView_LoadOrderPinned pins the three-load order the design's skew
+// argument depends on: the latest ledger and the handle set are loaded BEFORE
+// the catalog snapshot. The hook mutates both from inside the snapshot call —
+// with the correct order the view holds the pre-hook values; with the loads
+// swapped it would observe the mutations and this test fails.
+func TestNewReadView_LoadOrderPinned(t *testing.T) {
```
tamirms, line 348: The seam pins snapshot-last but cannot see the latest-vs-handles half: swapping the two loads at registry.go:227-228 passes the whole suite, and that swap re-opens a real window (a boundary between the loads leaves latestLedger pointing into a chunk the handle set predates, so resolveTier returns ErrUnavailable for an in-range tip ledger). A second seam closes it: a loadHandles field defaulting to r.handles.Load, with a test hook that advances latestLedger from inside; with the correct order the view still holds the pre-hook value. If that feels like too much seam, the smaller fix is scoping the comment at registry.go:52, which currently claims the seam pins "the ordering the design's skew argument depends on" when it pins the snapshot-last half only.

cmd/stellar-rpc/internal/rpcv2/stores/hotchunk/hotchunk.go
```diff
@@ -248,6 +248,11 @@ func (d *DB) Source() ledgerbackend.LedgerStream {
 // concurrently with in-flight reads/writes.
 func (d *DB) Close() error { return d.store.Close() }
 
+// CloseIfIdle is the non-blocking Close deferred deletion uses to reclaim a
```
tamirms, line 249: Doc drift, batched. Ledgers/Txhash/Events just above (lines 214-227) still say "no production caller yet", but this PR's read side now calls all three (resolve.go:82 and :111, txlookup.go:35); and Source() says the freeze opens the DB read-only, which the shared-handle path made the fallback. In bounds.go, ClampRange's leading/trailing wording misses that a descending leading edge above latest is truncated rather than erroring; the behavior is right and tested, the comment predicts the wrong result. Two "admitted latest" leftovers at scan.go:21 and scan_test.go:73. In the design doc, "It changes only when a hot database opens or is discarded" reads as the registry but means the handle set, and the acquisition pseudocode still shows a free function where Catalog.LastCompleteChunkAsOf shipped. hotloop.go's nil-Registry comments call the other caller "the bounded backfill loop", but backfill never drives this loop; the only caller of RunBoundedIngestionLoop is the ingest bench, and saying "the bounded bench/test loop" would prevent the misreading that catch-up ingestion runs through here. Last, TestBuildThenSweep_RollingPredecessorDemotedAndSwept's name says swept while its body correctly asserts demoted-not-swept; the name predates the fix.

cmd/stellar-rpc/internal/rpcv2/query/scan_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	require.NoError(t, cat.FlipHotReady(c))
+	r.PublishHandle(c, db)
+}
+
+// borderFixture seeds two adjacent hot chunks, each with its first two ledgers
+// committed (the events CF requires ingestion to start at the chunk's first
+// ledger). latest is pinned to c1's FIRST ledger — one below c1's last committed
+// one — so clamping at latest is distinguishable from data simply running out.
+func borderFixture(t *testing.T) (*Registry, chunk.ID, chunk.ID) {
```
tamirms, line 38: Three coverage gaps in the new code. The Events cold branch (resolve.go:100-109) is never exercised: no test routes a frozen events chunk to OpenColdReader, so a wrong bucket dir or a dropped closer would pass. No scan test crosses a cold chunk: the border fixture seeds two hot chunks, so the cold-to-hot border and cold-wins at the scan level are unpinned, and the assertions check sequence numbers but never payload bytes, so a border read served from the wrong tier would pass. And the deferred- deletion busy-retry composition is untested end to end: errReaderInFlight leaving the transient key for the next run's scan to re-collect and retry the close is the recovery seam the design leans on, and each half is pinned in isolation while the wiring between them is not.

cmd/stellar-rpc/internal/rpcv2/catalog/catalog.go
```diff
@@ -111,54 +138,81 @@ func (c *Catalog) ChunkArtifactKeys() ([]ArtifactRef, error) {
...
 
+// ReadyHotChunkKeysAsOf is ReadyHotChunkKeys read through a read view's snapshot.
+func (c *Catalog) ReadyHotChunkKeysAsOf(snap *rocksdb.Snapshot) ([]chunk.ID, error) {
+	return c.hotChunkKeysWith(snap, isReadyHot)
+}
+
+// ErrNoReadyHotChunk means a catalog view held no ready hot chunk at all. That
+// cannot happen in a working daemon (the live chunk's key is created before
+// serving starts and is never demoted), so it marks a broken catalog.
+var ErrNoReadyHotChunk = errors.New("catalog: no ready hot chunk")
```
tamirms, line 182: Could `NewSnapshot` return a catalog type instead of the raw `*rocksdb.Snapshot`? Right now every snapshot read is a second copy of a live method with the snapshot passed in on each call (`StateAsOf(snap, ...)`, `HotStateAsOf(snap, ...)`, `AllTxHashIndexKeysAsOf(snap)`), `LastCompleteChunkAsOf(nil)` uses nil to mean "read live", and the rocksdb type ends up in serving, which imports the rocksdb package just to hold the pointer between catalog calls. If `NewSnapshot` returns a `*catalog.Snapshot` with the reads as methods, all of that goes away:

```go
snap, err := cat.NewSnapshot()
st, err := snap.State(c, kind)
hst, err := snap.HotState(c)
covs, err := snap.AllTxHashIndexKeys()
lc, err := snap.LastCompleteChunk()
snap.Release()
```

The seven `*AsOf` methods get deleted: these four become methods on `Snapshot`, and `FrozenTxHashIndexAsOf` and `TxHashIndexKeysAsOf` have no callers outside the catalog's own tests, so they can go without replacement. The live methods don't change at all; the lifecycle loop's `LastCompleteChunkAsOf(nil)` becomes a plain `cat.LastCompleteChunk()`. `ReadView` then holds one `*catalog.Snapshot`, its `Release` calls `snap.Release()`, and serving stops importing rocksdb. (Pebble's snapshot API works this way: `db.NewSnapshot()` returns an object you call `Get` on and close when done.)

The same shape fits one layer down if you want both layers consistent: `rocksdb.Snapshot` could carry `Get`/`Iterate`/`Release` itself (it already wraps the grocksdb type opaquely, and the `AsOf` bodies exist as the `getWith`/`iterateWith` helpers), which deletes the store-level `GetAsOf`/`IterateAsOf`/`ReleaseSnapshot`. Optional; after the catalog change its only visible effect is inside catalog/kv.go.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
```diff
@@ -0,0 +1,262 @@
+// Package serving is the query read side: it routes each requested chunk to its
```
tamirms: Package-name question: `ingest` names the write side's machinery (the loop that drives it lives in the daemon package), `lifecycle` names the housekeeping, and this package is the read side's machinery, for which the doc's noun is "query": query workers, and "Query Routing" as the subsystem's title. Should this be package `query`? `query.ReadView` reads as what it is, and the #772 handlers will be its main importers. The rename is import paths only, and it gets more expensive the moment the adapters start importing it.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
```diff
@@ -0,0 +1,262 @@
...
+	return r
+}
+
+// PublishReadyHandles opens and publishes a handle for every ready hot chunk
+// except liveChunk, which the ingestion loop opens and publishes itself. These are
+// completed chunks a prior run left ready (not yet discarded); queries read them
+// hot until the freeze covers them cold. They are opened read-write so the events
+// facade is warmed (a read-only open is ledgers-only), and the registry closes them
+// at discard. Runs at startup before any read view is acquired.
+func (r *Registry) PublishReadyHandles(liveChunk chunk.ID, logger *supportlog.Entry) error {
```
tamirms: Initialization question: bringing the registry up takes four ordered calls in run() (NewRegistry, PublishReadyHandles, PublishHandle for the live chunk, SetLatestLedger), and nothing enforces the order or completeness; a registry that exists but is half-initialized is representable. Could the constructor do it?

```go
registry, err := serving.OpenRegistry(cat, cfg.Retention, hotDB, lastCommitted)
```

It opens and publishes the ready chunks below the live one, publishes the passed-in live handle, seeds latestLedger, and returns a serving-ready registry or fails startup. The live chunk comes from `hotDB.ChunkID()` (passing it separately could disagree with the handle), and the logger from `cat.Logger()`. The live DB stays the caller's to open, since openHotDBForChunk creates the chunk and flips its catalog key on a fresh boundary, which is ingestion's transition, not the read side's. lastCommitted has to be passed: on a fresh boundary the live DB is empty, and the value comes from the whole-catalog derivation in run(). PublishReadyHandles then drops out of the exported surface, and the boundary-time PublishHandle/SetLatestLedger stay as the loop's runtime calls.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
```diff
@@ -0,0 +1,262 @@
...
+// reader is still in flight; the caller leaves the chunk's transient key so a
+// later run re-collects it and retries — the handle stays in closing until it
+// drains. Deferred deletion calls this after the grace period, before unlinking
+// the chunk's files.
+//
+// The close runs under mu, so a successful close's memtable flush briefly blocks a
+// concurrent boundary PublishHandle. Read-view acquisition is unaffected (it loads
+// handles without mu), and the stall is rare (a discard overlapping a boundary)
+// and bounded (one memtable), so it is not worth closing outside the lock.
+func (r *Registry) CloseDiscarded(c chunk.ID) bool {
```
tamirms: Tiny rename: `TryCloseHandle`. `Try` is the stdlib's convention for a non-blocking attempt whose bool reports success (`sync.Mutex.TryLock`), which is exactly this contract, and it stops the bool reading as "was there one pending". The discarded-handles-only precondition moves to the doc comment: it only closes a handle DiscardHandle already unpublished, and never touches a published one.

cmd/stellar-rpc/internal/rpcv2/serving/scan.go
```diff
@@ -0,0 +1,85 @@
...
+// open failure surfaces at its own position. Page limits cap every request at
+// two chunks; the code handles any count.
+
+// ScanLedgers returns a flat ascending iterator over the raw ledgers in
+// [lo, hi] clamped to the view's range. The per-chunk intersect lives here, so
+// a caller cannot read past the admitted latest or below the floor; the
+// intersect also satisfies the cold reader's coverage requirement, since a
+// frozen pack covers its whole chunk. Entry.Bytes follows the hot store's
+// borrow contract: valid only until the next iteration step.
+func (a *ReadView) ScanLedgers(lo, hi uint32) (iter.Seq2[ledger.Entry, error], error) {
```
tamirms: ScanLedgers opens every chunk's reader up front, which came from my suggestion and is right for two chunks, but the two-chunk assumption isn't structural for getTransactions: `ScanLedgers(cursor, latestLedger)` on a deep-history node with an old cursor resolves and opens thousands of cold readers immediately (each open starts a background goroutine and holds an fd), all view-owned until Release, while the page fills from the first chunk. getTransactions' natural call is exactly that shape (hi = latest, break when the page fills). Suggest the iterator open lazily, one chunk ahead: open chunk i on entry, launch chunk i+1's open, and close chunk i as the walk advances past it, with Release as the backstop for whatever is open at an early break. This restores the "released per-chunk as wide scans advance" property the original per-chunk closers carried, moved inside the walker instead of being caller discipline. At one or two chunks this is identical to today (both opens still start at scan start); at any larger count the scan holds at most two readers regardless of range. That also makes the "the code handles any count" comment true on cost, not just correctness.

cmd/stellar-rpc/internal/rpcv2/serving/scan.go
```diff
@@ -0,0 +1,85 @@
...
+					return
+				}
+			}
+		}
+	}, nil
+}
+
+// EventPart is one chunk's slice of a clamped range: the reader plus the
+// intersected ledger bounds — the input shape the events query engine consumes.
+type EventPart struct {
```
tamirms: EventPart's name and contents both have a smell: `Chunk` duplicates what `Reader.ChunkID()` already answers, and "Part" was a fallback because "window" is taken by the tx-index unit. Embedding fixes both: `type BoundedReader struct { event.Reader; From, To uint32 }`, returned by `EventReaders(dir, lo, hi)`. The chunk id comes from the embedded reader, the struct says what it is (a reader plus the bounds to query it with), and the adapter loop reads as `engine.Query(r, r.From, r.To, ...)`. Worth defining it in the event package rather than serving: its long-term consumer is the events engine (the adapters PR will grow a cross-chunk driver there), serving already imports event, and that placement saves moving it later. Also fine to defer this whole change to the getEvents endpoint implementation, since that's when the struct gains its real consumer; nothing in this PR depends on it.

cmd/stellar-rpc/internal/rpcv2/query/txlookup.go
```diff
@@ -0,0 +1,58 @@
...
+// HotTxHashIndexes returns the transaction hash index of every published hot chunk,
+// newest chunk first. A hot match is exact and definitive, so the newest indexes
+// are probed first. The returned indexes are registry-owned handles; the caller does
+// not close them.
+//
+// It is deliberately unfiltered — every published handle, regardless of the
+// view's floor/latest. A match can therefore name a ledger in a chunk below the
+// floor (a handle that predates this view); the lookup's floor/latest gate on
+// the resolved ledger is the only thing that keeps such a match from being served.
+func (a *ReadView) HotTxHashIndexes() []txhash.HashIndex {
```
tamirms, line 29: Two things on this pair. `HotTxIndexes` returns probe-ready indexes, but the cold half stops at metadata: a coverage can't be turned into an open index through the view (the path composition and `txhash.OpenColdReader` live below it), so the lookup adapter would need its own catalog reference and manual closer management, unlike every other cold reader. A `ColdTxIndexes() ([]txhash.HashIndex, error)` that opens each frozen coverage newest-first with view-owned closers completes the surface, feeds `NewTxReader`'s existing slice contract directly, and lets `TxHashCoverages` become internal. Fine to defer to the getTransaction endpoint implementation, which is when it gains its consumer. And if `TxHashCoverages` does stay exported meanwhile, `ColdTxIndexCoverages` would make the hot/cold pair read as a pair.

cmd/stellar-rpc/internal/rpcv2/hotloop.go
```diff
@@ -92,45 +94,55 @@ type ingestionLoopConfig struct {
 	Metrics  observability.Metrics
 	Sink     ingest.MetricSink
 	Health   *healthState
+	// Registry, when set, receives the served latest ledger after each commit.
+	// The bounded backfill loop leaves it nil: backfill serves no queries.
+	Registry *serving.Registry
```
tamirms: The nil-Registry mode could become a second implementation instead of branches. A small interface for what the loop needs (`PublishHandle`, `SetLatestLedger`); the daemon passes the real Registry, and the bench/tests pass a ~15-line closingSink whose PublishHandle closes the previously published DB and holds the new one. The loop then has one ownership story, publish everything and the sink owns closing: the three nil checks and the boundary's close-vs-transfer branch all delete, and the bench still closes each completed chunk at the same boundary moment it does today, so its measurements don't change.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
karthikiyer56: this is cleaner.
`query.ReadView` and `query.Registry` both read well, and it lines up with the doc's own noun.
If we can settle this, I can update the writeup in #885 #886  and #887

cmd/stellar-rpc/internal/rpcv2/query/registry.go
karthikiyer56: thats fair.
I can rely on semantics on scanLedgers in the getTransactions(startLedger, endLedger) usecase instead of trying to make it comport with what v1 does.
NNT make this map change

cmd/stellar-rpc/internal/rpcv2/serving/scan.go
karthikiyer56: Is this optimization even needed?

It only matters because `getTransactions` can be handed any range. In v1 a client can ask for ledgers 10,000 to 100,000 — that's 10 chunks — and the handler just walks it until the page fills. 
If my adapter passes that range straight through, `ScanLedgers` opens 10 readers before the first one streams. 

But v2 doesn't have to allow that range in the first place. 
Lets give `getTransactions` the same rule `getEvents` already has for v2: don't scan more than 10k ledgers to fill a page. 
The adapter then asks for `ScanLedgers(start, min(latest, start+10000))`, which is at most two chunks, always. I'd rather put that at the call site — it knows its own page semantics — than push it into the read view. 

And it isn't new behaviour for the API, since getEvents already works this way via `LedgerScanLimit`: a sparse range returns a short page and the client pages again.

On closing each chunk as the walk passes it: you already say the eager open is right for two chunks, and that the worry is the two-chunk assumption not being structural.
Capping at the call site makes it structural. 
Then the early close isn't buying anything, and closing stays in one place instead of being split between the walker and `Release`.
(It wouldn't be unsafe either way — `packfile.Reader.Close` is a `sync.Once`, so `Release` closing an already-closed reader is a no-op.)
Calling it in the walker and then again in the `Release()` feels weird.

cmd/stellar-rpc/internal/rpcv2/query/registry_test.go
urvisavla: Done in cfda857f (loadHandles seam; the seam comment now states what each hook pins).

cmd/stellar-rpc/internal/rpcv2/stores/hotchunk/hotchunk.go
urvisavla: Done in cfda857f.

cmd/stellar-rpc/internal/rpcv2/query/scan_test.go
urvisavla: Done in 63a9fc7f (cold-events routing, cold-to-hot border by payload bytes with cold-wins against a published hot handle, busy-retry wired end to end).

cmd/stellar-rpc/internal/rpcv2/catalog/catalog.go
urvisavla: Done in 3c42ac88; the rocksdb-layer variant deferred as you suggested.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
urvisavla: Done in 04512c39.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
urvisavla: Done in 0481fc64.

cmd/stellar-rpc/internal/rpcv2/serving/registry.go
urvisavla: Done in cfda857f.

cmd/stellar-rpc/internal/rpcv2/serving/scan.go
urvisavla: Done in 6a3c33ef; first two chunks still resolve at call time, later failures surface at their position.

cmd/stellar-rpc/internal/rpcv2/serving/scan.go
urvisavla: Deferring to the getEvents endpoint PR as suggested; it will open with the event.BoundedReader shape.

cmd/stellar-rpc/internal/rpcv2/query/txlookup.go
urvisavla: Deferring ColdTxIndexes to the getTransaction endpoint PR as suggested; the ColdTxHashIndexCoverages rename landed in cfda857f.

cmd/stellar-rpc/internal/rpcv2/hotloop.go
urvisavla: Done in 0481fc64 (handleSink + closingSink; the bench closes at the same boundary moment).

karthikiyer56:
this looks good to me.
I am still not convinced we need the optimization about closing readers presumptively on the off chance that we pass in a range that spans several chunks.
I can retrofit the code in #772 's sub issues to work with this implemetnation of query router

karthikiyer56 (approved):

cmd/stellar-rpc/internal/rpcv2/query/scan.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/chunk"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/stores/event"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/stores/ledger"
+)
+
+// The range helpers the endpoint adapters build on. Both clamp the raw request
+// range themselves (RangeError / ErrInvertedRange surface here) and resolve the
+// overlapping chunks. Page limits cap every request at two chunks, but the
+// two-chunk assumption is not structural for every caller — getTransaction's
```
tamirms: A batch of residues from this round, all drift rather than behavior. In code: `TestCloseDiscarded_BusyRetainsThenRetryDrains` and `TestHotTxIndexes` kept old method names; this header comment says "getTransaction's scan shape" where it means getTransactions, and its "the helpers must be cheap at any count" sentence covers EventParts, which is deliberately open-all (getEvents' 10k protocol cap makes its count structural), so the sentence should scope to ScanLedgers; scan_test.go's "readers are resolved eagerly" comment states the pre-walker contract (only the first two chunks resolve eagerly now); lifecycle.go's Loop comment still cites `LastCompleteChunkAsOf`, deleted this round; the error strings in bounds.go and resolve.go and the log prefix in registry.go still say "serving:" after the package became query; NewRegistry's doc says benches use it, but only tests do (the bench gets a closingSink inside RunBoundedIngestionLoop); the closers field doc says "appended by the resolve methods" but ScanLedgers appends the walk's backstop directly; the txindex_test section banner still says "demoted AND swept" above the test now named DemotedNotSwept; the loop's initial publish keys the live handle by `chunk.IDFromLedger(cfg.Resume)` where `hotDB.ChunkID()` would extend OpenRegistry's own can't-disagree rationale; and deleting `bootstrapServing` deleted the only in-code explanation of why crash leftovers need no startup pass; a sentence of it belongs near the `boundary.Publish()` seed in run(). In the design doc: the ReadView struct snippet still shows `snap *rocksdb.Snapshot` (the code holds `*catalog.Snapshot`, and the doc's own later section says so), the startup row still names `bootstrapServing`, and the getTransaction section cites `TxHashIndexCoverages`, renamed to ColdTxHashIndexCoverages.

cmd/stellar-rpc/internal/rpcv2/query/scan_test.go
```diff
@@ -0,0 +1,223 @@
...
+	for e, err := range scan {
+		require.NoError(t, err)
+		seqs = append(seqs, e.Seq)
+	}
+	return seqs
+}
+
+// TestScanLedgers_CrossesChunkBorder pins the flat iterator across a chunk
+// border: one ascending sequence, no seam visible to the caller.
+func TestScanLedgers_CrossesChunkBorder(t *testing.T) {
```
tamirms, line 66: Three coverage holes in this round's code. Every ScanLedgers test spans at most two chunks, so the walker's lazy branch is never exercised: the mid-stream open failure surfacing at its own position, the dropped-then-retried one-ahead error, and the at-most-two-readers property are all unpinned, and the closeAll backstop only ever runs against empty maps (no test breaks out of a scan while cold readers are still held). OpenRegistry's error path is uncovered: nothing makes a ready chunk unopenable to pin the error wrap and the close of already-opened handles. And closingSink's core behavior is pinned only indirectly through the fence test; its same-handle republish guard is unexercised.

tamirms (approved):
pr looks good! I posted 2 more minor comments from claude. feel free to merge the PR after they've been addressed

cmd/stellar-rpc/internal/rpcv2/query/scan.go
urvisavla: Done in 31bc7656.

cmd/stellar-rpc/internal/rpcv2/query/scan_test.go
urvisavla: Done in c73a3988.

Issue #774, opened 2026-06-08
Implement the v2 events endpoint and deprecate the existing getEvents


tamirms:
## TL;DR
Following the `getEvents` product pass in #426, implement the redesigned v2 events endpoint and back-compat support for the existing v1 `getEvents` API.

## Scope
In priority order:

- [x] 1. **Implement the v2 events endpoint** per the direction settled in #426.
https://github.com/stellar/go-stellar-sdk/pull/5971
https://github.com/stellar/stellar-rpc/pull/914 
https://github.com/stellar/stellar-rpc/pull/934 
https://github.com/stellar/stellar-rpc/pull/939 
https://github.com/stellar/stellar-rpc/pull/954 

- [x] 2. **v1 back-compat.** Figure out how to serve the existing v1 `getEvents` API so current clients keep working unchanged, and implement it. The deprecation timeline is out of scope: it is a product decision. 
https://github.com/stellar/stellar-rpc/pull/904 
https://github.com/stellar/stellar-rpc/pull/960

Post-MVP work from this issue has been moved to https://github.com/stellar/stellar-rpc/issues/995

## Depends on
- #426's product direction being settled.
- Method naming convention for the v2 endpoints, tracked in #777 (getEventsV2 as the working name).

tamirms:
On the migration story: v1 `getEvents` can be served from the v2 event store as a shim, and most of it maps onto the store as it is. v1 cursors correspond to the store's native (ledger, tx, op, event) positions, and every response field derives from the stored event payload. (`inSuccessfulContractCall` is deprecated and will be removed from getEvents in a release before the shim ships, so transaction success is not needed in the event path.) Two v1 features touch the storage layer, and because completed chunks are immutable, both need to be decided before the initial backfill; adding them later means reingesting the events store.

**1. Event type term.** v1 filters on event type (`contract` or `system`). The index currently covers contractId and topic0..topic3 only, so type is reachable only in the post-filter stage, and a type-only query (for example, all system events) degenerates to a full scan over exactly the rarest data. Add one index term per event, keyed on the type value. The two bitmaps partition each chunk with extreme skew (contract near-full, system near-empty), so both run-length compress to a few KB. This also lets getEvents v2 gain a type filter later as an API-only change.

[...]

PR #904, opened 2026-07-28
fullhistory: index event type and topic count for getEvents v1

tamirms:
## What

Adds two index term families to the event store: one keyed on an event's type, one keyed on how many topics it carries.

Both come out of the storage analysis in #774. getEvents v1 filters on event type, and its topic filters carry arity semantics: a filter ending in `**` matches events with at least as many topics as it names, and one that does not matches events with exactly that many. Neither is answerable from an index over contract IDs and topic values, so today a type-only query degenerates to a full scan over exactly the rarest data. Arity is worse than a missing value term, because it is an absence property and a value term can only say what an event has.

Completed chunks are immutable, so adding a term family later means reingesting the events store. That is why this lands before the initial backfill rather than alongside the v1 shim that will use it.

## The overflow bucket

Topic-count buckets are exact for every count a filter can name, plus one overflow bucket shared by every event carrying more.

The overflow bucket is load bearing. Events can carry more than four topics: `ContractEventV0.topics` is an unbounded XDR vector and the Soroban host records topics without a count check. The design doc already notes this on the value side ("Additional topics beyond 4 are stored but not indexed"). With a bucket per exact count and nothing above, a five-topic event would fall outside every bucket an "at least n" union can reach and would silently drop out of results it belongs in. The post-filter cannot undo that, since it only removes candidates.

Keeping the top bucket exact rather than saturating also matters for pagination. A saturating top bucket would make the index return a superset for any exact count a v1 filter can express, and because `MaxEvents` is applied before the post-filter, a page could come back empty with matches still ahead and no cursor to resume from. With the overflow bucket, the only superset left is an exact count above what any filter can name, which neither API can produce.

## Query side

`Filter` grows an `EventType` pointer and a `TopicCount` field whose zero value is the wildcard. `TopicCountFilter{Count, Exact}` maps directly onto v1's two shapes.

Terms are now grouped by field: bitmaps within a group are OR-ed and the groups are AND-ed. Only the topic-count group ever holds more than one term, and every other field is a single-element group, so the old flat plan is the degenerate case. The match-all short-circuit is now read off the term groups themselves rather than a separate predicate, so the two cannot drift.

One planner shortcut: an "at least" count already implied by a constrained topic position fetches no bucket at all, since a topic term is only indexed for events that carry that position. That covers the standard `["transfer", "**"]` shape, and skips OR-ing chunk-sized bitmaps for a constraint that was a no-op. The post-filter enforces the count either way.

## Cost

Measured with `BenchmarkEventIndex_10M` at 9M events, with adds grouped per ledger the way `HotStore.applyLedger` batches them: the two families add roughly 7% build time and 1% heap. The type bitmaps are near-degenerate and compress away; the topic-count buckets are the real cost and are still small against the ~57MB of bitmap data per chunk. Including them is the low-regret call, since an unused term costs disk and nothing else, while a missing one costs a reingest.

## Notes for review

- No format identifier bump. Chunks built before this have no bitmaps for either family, so they need reingesting; that is consistent with how this branch treats pre-release storage changes.
- `TermsForBytes` now rejects an event type outside the enum, which fails ingestion of that ledger. This matches the existing hard-fail on an unsupported `ContractEvent` body version, and is now commented and tested rather than implicit.
- `protocol.MaxTopicCount` is pinned by a compile-time assertion. The bucket space is written into artifacts that are never rewritten, so growing that constant would silently redefine what the overflow bucket in every existing chunk means.
- The v1 shim itself is not here. This is the storage and query support it needs.

## Follow-ups, not in this PR

- Nothing detects an under-indexed chunk: a query against one reads the missing bitmaps as "no matching events" rather than failing. Both new families are universal, one term per event, so per-chunk bucket cardinalities must sum to the event count. That is a cheap open-time tripwire if we want one.
- `indexPackItemsPerRecord = 128` was tuned assuming every bitmap is small. The topic-count buckets are around a megabyte each, so a handful of records per chunk become large and an unrelated term landing in one pays that read. Worth measuring.
- A type-only filter no longer takes the sequential match-all path; it resolves through a near-full term and does point gets. Nothing detects "this group covers essentially the whole range" and falls back.

https://claude.ai/code/session_017GGKYLKf29KsdFRtvWXVNf

### DIFF
--- a/cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
+++ b/cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
@@ -82,10 +82,10 @@ func (f TopicCountFilter) matches(n int) bool {
 	return n >= f.Count
 }
 
-// termKeys returns the topic-count buckets whose union covers f. An
-// exact count above what a filter can name shares the overflow bucket
-// with every other such count, so it is a superset the post-filter
-// narrows.
+// termKeys returns the topic-count buckets whose union covers f. Every
+// count validateFilters admits has a bucket of its own, and an "at
+// least" union is closed by the overflow bucket, so the union never
+// holds an event f does not match.
 func (f TopicCountFilter) termKeys() []events.TermKey {
 	if f.isWildcard() {
 		return nil
@@ -453,6 +453,17 @@ func validateFilters(filters []Filter) error {
 			return fmt.Errorf(
 				"events: filter[%d].TopicCount.Count must be non-negative, got %d",
 				fi, f.TopicCount.Count)
+		}
+		// Above MaxTopicCount the index cannot answer a count exactly: every
+		// such count shares the overflow bucket, so the terms would return a
+		// superset and the post-filter would narrow it. MaxEvents is applied
+		// before the post-filter, so that page can come back empty while
+		// matches remain, leaving no delivered position to resume from. A
+		// getEvents filter cannot name that many topics anyway.
+		if f.TopicCount.Count > protocol.MaxTopicCount {
+			return fmt.Errorf(
+				"events: filter[%d].TopicCount.Count must be at most %d, got %d",
+				fi, protocol.MaxTopicCount, f.TopicCount.Count)
 		}
 	}
 	return nil
--- a/cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
+++ b/cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
@@ -7,9 +7,9 @@ package event
 //
 // Semantics:
 //
-//   - A Filter is a conjunction of equality constraints on the
-//     indexed fields (contract ID + topics 0..3). A field left
-//     nil/empty is a wildcard.
+//   - A Filter is a conjunction of constraints on the indexed fields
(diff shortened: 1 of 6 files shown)


tamirms:
@codex[agent] review

tamirms:
@codex[agent]+gpt-5.5 review

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
```diff
@@ -39,43 +39,109 @@ import (
...
+		if keys := f.TopicCount.termKeys(); len(keys) > 0 {
+			groups = append(groups, keys)
+		}
+	}
+	return groups
+}
+
+// impliesTopicCount reports whether f's constrained topic positions
+// already guarantee its topic-count bound.
+func (f *Filter) impliesTopicCount() bool {
```
urvisavla, line 135: Could you add these tests?

1. Two more rows in `TestQuery_TypeAndCountIndexTerms`: an `Exact` count together with a constrained topic position, and `Topics[0]` with `Count: 2`.
2. A post-filter test that feeds `matchesAnyFilterView` an event with the wrong type, and one with the wrong topic count, and checks both are rejected.

urvisavla (approved):

Issue #884, opened 2026-07-21
Serve getLedgerEntries, simulateTransaction, sendTransaction on v2


karthikiyer56:
## Context

`getLedgerEntries`, `simulateTransaction`, and `sendTransaction` read current ledger state via captive core (the bucket list), not the chunk/catalog stores. A package-level audit (done for the repo reorg tickets) found that v1's actual implementation of these — `internal/preflight`, `internal/ledgerentries` — has **zero SQLite coupling**: both are already written against the interfaces in `internal/host` (`Daemon`, `CoreClient`, `FastCoreClient` — the package was renamed from `internal/daemon/interfaces` during #876's implementation), which describe "talk to captive core" in the abstract, not "talk to *v1's* captive core setup" specifically. All three stay put directly under `internal/` as shared packages in the reorg (#876).

## What this means for scope

v2 does **not** need to reimplement preflight execution or ledger-entry fetching. It needs one new thing: a concrete type satisfying `Daemon` from `internal/host` (and its `CoreClient`/`FastCoreClient` sub-interfaces) backed by v2's own captive-core process. Then it calls the *same* `internal/preflight` and `internal/ledgerentries` handler-construction code v1 already uses, unmodified.

`sendTransaction` is simpler still — a pure write proxy to Core via `CoreClient.SubmitTransaction`, needing only the current ledger sequence for its response (from whichever `store.LedgerReader` v2 has).

## Design notes (settled 2026-07-27)

**Who talks to what** — core runs two HTTP servers when its toml enables them; everything the three endpoints need reduces to those ports plus a version string:

| Endpoint | Path to core | What core provides |
|---|---|---|
| sendTransaction | `CoreClient` → admin port (`tx?blob=...`) | validation + flooding; the only true submit path |
| getLedgerEntries | `FastCoreClient` → query port | bucket-list ledger entries (read-only) |
| simulateTransaction | preflight worker pool → Rust `libpreflight` (CGO, in-process); the Rust host fetches each ledger entry it needs back through `FastCoreClient` → query port | read-only state lookups — the simulation itself runs in the Rust library, nothing is submitted |

**What `host.Daemon`'s members actually require of v2:**

| Member | Real requirement | Needs the live core's Go object? |
|---|---|---|
| `CoreClient()` | plain `stellarcore.Client{URL, HTTP}` at the admin port | No — needs the port enabled on the core PROCESS |
| `FastCoreClient()` | plain HTTP client at the query port | No — same |
[...]

Issue #865, opened 2026-07-17
Implement concurrent query serving: ingestion changes and the query router


tamirms:
## TL;DR
Implementation follow-up to #770. Implement the approved concurrent query serving design: the ingestion-side changes it requires, and the query router component that every request is threaded through.

## Scope
- **Ingestion changes.** Implement the store-set mutations the design assigns to the ingestion and lifecycle paths (hot-store append; freeze hot to cold; prune cold stores past the retention window; delete a hot store once its freeze completes) against the design's reader-safety contract, so mutations are safe under live queries.
- **Query router.** Implement the component every request flows through: resolve the query's ledger range against the live set of hot and cold stores for its data type, run the per-chunk engine against each covered store, merge results to the requested order and limit, and enforce the reader-safety contract: a query in flight never reads a store the lifecycle thread is removing, and sees a consistent view of which stores cover its range.

## Depends on
- The #770 design being reviewed and signed off.

karthikiyer56:
Hi @urvisavla
it would be nice if the query router exposes the raw capability these 4 v1 interfaces need, so I can implement them myself as a thin wrapper on top, without anything needing to be renamed or restructured on your side:

- [`LedgerReader`](https://github.com/stellar/stellar-rpc/blob/b5df41dcb9f2010e516cc3f5d39a5c1162816e7f/cmd/stellar-rpc/internal/db/ledger.go#L25-L33) — point lookup by ledger sequence, ranged read/stream, current range, latest sequence
- [`LedgerReaderTx`](https://github.com/stellar/stellar-rpc/blob/b5df41dcb9f2010e516cc3f5d39a5c1162816e7f/cmd/stellar-rpc/internal/db/ledger.go#L35-L40) — returned by `LedgerReader.NewTx`; needs the same range/point-lookup capability, held consistently for one request's lifetime
- [`TransactionReader`](https://github.com/stellar/stellar-rpc/blob/b5df41dcb9f2010e516cc3f5d39a5c1162816e7f/cmd/stellar-rpc/internal/db/transaction.go#L50-L53) — lookup by transaction hash
- [`EventReader`](https://github.com/stellar/stellar-rpc/blob/b5df41dcb9f2010e516cc3f5d39a5c1162816e7f/cmd/stellar-rpc/internal/db/event.go#L47-L57) — filtered event scan over a cursor range

[...]

PR #908, opened 2026-07-31
Serve the v2 read path: every method except the events reads (#885, #886; subsumes #889's wiring)

### LINKED ISSUE #885: LedgerReader + LedgerReaderTx adapter

### LINKED ISSUE #886: TransactionReader adapter

### LINKED ISSUE #887: EventReader adapter

### LINKED ISSUE #889: Wire the v2 binary's JSON-RPC handlers (all except getEvents and getEventsV2)
## What

Make the v2 binary serve JSON-RPC. Two parts: assemble the method table, then put an HTTP server in front of it.

**1. Method table** — instantiate `internal/methods`'s existing, unmodified handler constructors (`NewGetLedgersHandler`, `NewGetLatestLedgerHandler`, `NewGetTransactionHandler`, `NewGetTransactionsHandler`, `NewGetEventsHandler`, `NewGetFeeStatsHandler`, plus the cheap ones the epic also cuts over: `NewHealthCheck`, `NewGetNetworkHandler`, `NewGetVersionInfoHandler`) with the v2 router adapters instead of the SQLite ones, inside `internal/rpcv2/jsonrpc.go` (the thin wrapper analogous to `internal/rpcv1/jsonrpc.go`), calling `internal/jsonrpc`'s generic method-table builder with v2's own config-derived method specs. The entries for #884's endpoints (getLedgerEntries, simulateTransaction, sendTransaction) are assembled here too, from the pieces #884 wires. No handler code changes — composition only.

**2. HTTP server** — `internal/jsonrpc.Handler` embeds `http.Handler`; it is a handler, not a server. Nothing binds a port today, so v2 currently serves nothing at all. This ticket also mirrors v1's `daemon.setupHTTPServers`:

- a listener on `[service].endpoint` and an `http.Server` over the method table;
- the admin endpoint (`[service].admin_endpoint`) with pprof and `/metrics` over the daemon's existing registry;
- replacing `runDaemonWith`'s `serveReads` no-op with the real thing, keeping its contract (return promptly, do not block) and closing the handler + listeners on shutdown — after widening its signature (next section).

The `TODO(#889)` markers in `internal/rpcv2/daemon.go` mark the three attachment points (server, health signal, metrics).

## Widen the ServeReads seam — the registry is per-run

`ServeReads`'s current signature cannot reach the query router. The daemon has two lifetimes, and the pieces this ticket wires sit on opposite sides:

| Piece | Built in | Lifetime |
|---|---|---|
| `ServeReads` closure, `corestate`, preflight pool, prometheus registry, `healthState` | `runDaemonWith` | once per process (outside `supervise`) |
| `query.Registry` (#894) — what the #885–#887 adapters need for read views | `run()` | per supervised attempt; "no query survives a restart" |

- `ServeReads` is `func(ctx context.Context) error`, assembled before `supervise` starts — no parameter can carry the per-attempt registry into it, so the router-backed handlers are unbuildable through the seam as-is.
- **Widen it**: `ServeReads(ctx context.Context, reg *query.Registry) error` (or a narrower view-provider interface), called by `run()` after `query.OpenRegistry` — the call site already sits after the registry is built.
- Decide the server's lifetime alongside: either launch listener + method table per attempt inside `ServeReads` (simplest; the per-attempt shutdown already closes them), or bind the listener once per process and swap the registry behind an atomic holder. The per-process pieces (`corestate`, preflight pool, metrics registry) stay outer either way — they address core's HTTP ports and register collectors once.

## Cap getTransactions' ledger span at 10,000

`getTransactionsByLedgerSequence` loops to `ledgerRange.LastLedger.Sequence` and breaks only on the *transaction* count — `ValidatePagination` bounds the limit and the start ledger, not the span. So a sparse range can walk arbitrarily far.

- **Cap the span at 10,000 ledgers**, mirroring `methods.LedgerScanLimit` for getEvents. The adapter then calls `ScanLedgers(start, min(latest, start+10_000))`.
- **Why it matters beyond cost:** `ScanLedgers` is windowed and only resolves its first two chunks at call time. At ≤2 chunks an unroutable chunk fails before any row is emitted; past that it fails mid-stream, after a partial response is already written.
- **Behaviour change to record:** a sparse range returns a short page and the client pages again. Consistent with getEvents, but different from v1's walk-until-full.

## Derive the deletion grace period from the request deadline

`lifecycle/lifecycle.go` carries `defaultGrace = 5 * time.Minute` and a `TODO(#772)` to derive it from the read server's request deadline (`T = max request timeout + margin`, boot-validated).

**Why it belongs here.** Deletion is deferred: a lifecycle run demotes a resource (hot chunk → `transient`, cold artifact → `pruning`), which makes it invisible to *new* read views, then waits `Grace` **once** at end of run before unlinking. There is **no reader refcount** — `query/` has no knowledge that deletion exists. The grace period is the only thing separating an in-flight query from `os.Remove`.

| Resource | Guard at destroy time |
|---|---|
| Cold artifacts (packfiles, `.idx`) | Grace only — `DestroyChunkArtifacts` has no idle check. In practice POSIX also covers us: once the reader holds the fd, `unlink` leaves the inode alive. |
| Hot handles (RocksDB) | Grace **plus** `CloseIfIdle`, but that is a `TryLock` at one instant, not a reservation. A view between two reads gets closed underneath it and the next read returns `ErrStoreClosed`. |

**What wiring it buys.** With `Grace = max request timeout + margin` and a boot-time validation that the margin holds, no request can outlive the grace, and the whole class of "resource deleted under an in-flight query" becomes impossible by construction rather than improbable. v2 already has the deadline: `MaxExecutionDuration` per method in `MethodConfig`/`PaginatedMethodConfig`. The request timeout and `Grace` must be decided together.

## Error mapping

| Query-layer error | JSON-RPC shape |
|---|---|
| `*query.RangeError` | out-of-range (v1-compatible) |
| `query.ErrUnavailable` | retryable `-32002` — reachable only via the read-view acquisition race (handle set loads before the catalog snapshot); heals on immediate retry. A normal freeze never produces it: discard demotes only after cold coverage commits, and cold-wins covers the overlap |
| `stores.ErrStoreClosed` | retryable `-32002`; server-side counter — reaching a client means a request outlived the grace period |

## Metrics that ride along

- A failed-destroy counter in `observability.Metrics` — a handle that stays stuck across lifecycle runs currently only shows up in logs. Deferred from the #894 review to this epic's metrics work; this is the ticket with the `/metrics` endpoint.

## Cleanup that rides along

This ticket makes `[service]` live, so remove the "dormant until the read server exists" framing from `config.go`'s `ServiceConfig` comment, the sample-config header, and the design doc's `[service]` section. Note #884 already narrowed that framing — `[service.preflight]` is live as of #884, so what is left to remove is the wording about the *read server* and the method tables, not the whole caveat.

## Depends on

The v2 `LedgerReader`/`TransactionReader`/`EventReader` adapter tickets (#885, #886, #887), the fee-stats feed and warm-up tickets, the `[service]` config ticket (#882), the core-state ticket (#884 — the shared builder's `jsonrpc.Params` takes a `host.Daemon`, which #884's `internal/rpcv2/corestate` implements, and `NewGetVersionInfoHandler` reads `daemon.CoreVersion()`), and the repo reorg (#876) that establishes the shared packages and the v1 pattern to mirror.

karthikiyer56:
# What this PR does

The v2 binary now serves JSON-RPC over the hot + cold chunk stores, through the #894 query router. Every method is live except the events reads (#774).

| Method | Backed by |
|---|---|
| getLedgers, getLatestLedger | `adapters.LedgerReader` over the query router |
| getTransaction, getTransactions | `adapters.TransactionReader` + the ledger adapter's walk |
| getHealth, getNetwork, getVersionInfo | the ledger adapter + registry close-time stamps + config + `corestate` |
| getFeeStats | the daemon-owned `feewindow.FeeWindows` (#917) |
| getLedgerEntries, simulateTransaction, sendTransaction | #884's captive-core pieces; sendTransaction through the txsub metrics wrapper |
| getEvents, getEventsV2 | stubs — explicit not-implemented errors naming #774 |

## Request model

| Piece | Shape |
|---|---|
| One read view per request | <ul><li>The per-request wrapper acquires a read view, puts it on the context (`adapters.WithView`), and releases it after the handler returns.</li><li>Every adapter read in the request goes through that one view. A response can never mix two states of the serving world.</li><li>A context without a view is an error.</li></ul> |
| Error handling | <ul><li>No adapter error is remapped. A store-boundary failure reaches the client as the shared handler's generic internal error — the same as v1.</li><li>The conditions that could justify a "retry" answer are unreachable within the serving model. Each one is counted at the site that detects it (see *Operator metrics*), and any non-zero rate is an alarm.</li></ul> |
| One per-request scan bound | `chunk.LedgersPerChunk` (10,000) is the single source of "a request never scans more than one chunk's span": the adapters' walk cap and the events pager derive from it, and `methods.LedgerScanLimit` (shared v1 code) is pinned equal by a test |
| Deletion grace derived from the timeouts | <ul><li>Grace = max(all method budgets, global cap) + 30s margin; 55s with defaults, computed at boot.</li><li>The same limits map builds the method table, so a raised budget widens the grace automatically.</li><li>After a demotion, this grace is the only thing between an in-flight query and `os.Remove`.</li></ul> |

## The ledger adapter (#885)

| Piece | What it does |
|---|---|
| `adapters.LedgerReader` | <ul><li>Satisfies the shared `store.LedgerReader` interface, so the shared handlers run unchanged.</li><li>`inWindow` is the single servable-window gate for every point read; it also rejects sub-genesis sequences.</li></ul> |
| `ledgerReaderTx` | <ul><li>Serves getTransactions' ascending contiguous walk from one primed `ScanLedgers` iterator, capped at one chunk's span.</li><li>Checks its context per ledger, so an abandoned handler stops within one decode of its deadline.</li></ul> |
| Zero-decode headers | <ul><li>`BatchGetLedgers` slices the marshaled header straight out of the raw LCM bytes; nothing is unmarshaled.</li><li>`LedgerMetadataChunk` carries `HeaderRaw + Lcm`; the XDR path base64s the bytes directly.</li><li>Uses `LedgerCloseMetaView.LedgerHeader()`, exported upstream for this (go-stellar-sdk#5982; dependency bumped).</li></ul> |
| Close times in memory (#865) | <ul><li>The ingest loop stamps `(seq, closeTime)` per commit; the oldest edge is a read-through cache the view writes (`ReadView.RecordOldestCloseTime` stamps its own floor, so a wrong sequence cannot be recorded).</li><li>`adapters.SeedCloseTimes` stamps both window edges at startup, before serving begins.</li><li>`getLedgerRange` does no disk reads for close times after boot; the request-path point reads remain as backstops.</li></ul> |

## The transaction adapter (#886)

| Piece | What it does |
|---|---|
| `adapters.TransactionReader` | <ul><li>Probes the hot tx-hash indexes first; the frozen window indexes are enumerated only when every hot index misses, so the polling hot path never touches frozen-window metadata.</li><li>Verifies candidates against the full transaction hash.</li><li>Copies only the transaction's bytes out of the SDK view, so the multi-MB ledger buffer frees at once.</li><li>Covers both V1 and V2 `LedgerCloseMeta` shapes.</li></ul> |
| `windowGatedIndex` | <ul><li>The view's index accessors (`HotTxHashIndexes`, `ColdTxIndexes`) return every index already wrapped in the servable-window gate, applied before the candidate's ledger is fetched. A caller cannot forget the gate.</li><li>Cold rationale: a frozen index outlives its chunks' ledger files, so a below-floor hit must read as a plain miss — the client sees v1's NOT_FOUND, never "lookup incomplete".</li><li>Hot rationale: a request racing a prune can hold a handle its snapshot already retired; gated, that hit is the clean miss it truly is.</li></ul> |

## Serving (#889)

| Piece | Shape |
|---|---|
| JSON-RPC server | <ul><li>Per supervised attempt: `ServeReads(ctx, registry)` returns a stop function plus a channel that reports the server dying on its own.</li><li>`run()` stops the server before the registry closes, and a dead accept loop fails the attempt so the supervisor rebinds.</li></ul> |
| Admin server (pprof, `/metrics`) | Per process. Log-only on failure, matching v1: dead pprof must not take down a serving node |
| `[service]` config fully live | Per-method queue limits and execution budgets (v1's defaults), `endpoint`, `admin_endpoint`, getNetwork's `friendbot_url`; getEventsV2 carries getEvents' full knob set, validated at startup |
| getHealth `ledgerRetentionWindow` | Computed by `geometry.Retention.RetentionWindow()`; full history reports `0` — the documented convention |

**#889 is closed by this PR** — the events work is #774's; the readiness follow-up moved to #937 (I1).

## Events carve-out

- getEvents and getEventsV2 belong to #774. Both serve explicit not-implemented stubs naming it — never an empty success, which would tell a paging client "no events exist".
- Events handlers are per-daemon by design: `HandlerDeps.GetEventsHandler` is required; #774's v2 handlers will call the merged engine (`query.QueryEvents`) directly. The #934 cross-chunk pager and #939 term-budget merges are absorbed here.
- getEventsV2 is the one method v2 serves that v1 does not.
- #887 closed.

## Shared with v1 (one copy, both daemons)

| What | Where it lives |
|---|---|
| The method spec table | `internal/jsonrpc.BuildHandlerSpecs` + `LimitsByMethod.Apply` — a served method missing from the limits map, or an unserved entry, fails that daemon's startup |
| Admin mux, XDR decode limits, HTTP read timeout | `internal/jsonrpc` |
| Serving-metrics registration | `internal/jsonrpc.NewHandlerMetrics` — once per process; collectors survive supervised restarts |
| txsub metrics wrapper | `internal/host.NewCoreClientWithMetrics` — both daemons publish identical `soroban_rpc_txsub_*` series |
| rpcv2 test fixtures | `internal/rpcv2/rpcv2test` — LCM builders, catalog/seeding helpers, the raw JSON-RPC test poster (`PostRPC`) |

The dividing rule: **share the protocol surface, never the lifecycle.** Wire behavior, limits validation, and metric series are shared. Daemon lifecycles, config packages, and restart models stay separate, because their failure modes differ by design.

## v1-visible behavior changes (shared handlers; reach main when this branch merges)

| Change | Effect on v1 clients |
|---|---|
| getTransactions scan bound | <ul><li>Scans at most 10,000 ledgers per request — the same bound getEvents already uses. Before: an unbounded ledger-by-ledger walk toward the tip.</li><li>A sparse range can return a short page and a cursor; clients page by comparing the cursor's ledger part against `latestLedger`.</li><li>Documented in the CHANGELOG.</li></ul> |
| Stricter strkey parsing in getEvents filters | Came in with the SDK bump; malformed contract-ID strkeys are now rejected. Documented in the CHANGELOG |

## Operator metrics (new)

| Counter | Meaning |
|---|---|
| `failed_destroys_total` | A deferred destroy failed and was left for the next lifecycle run; a stuck hot handle recounts every run |
| `store_ops_after_deferred_close_total` | An operation hit a store already closed by deferred deletion. Each count is a caller that outlived the deletion grace. Teardown closes never count |
| `unavailable_chunk_resolves_total` | A read found no serving store for a chunk (plus a log naming the chunk). Unreachable within the serving model — any count is an alarm |
| `missing_cold_pack_opens_total` | A cold ledger pack's file was gone on first read — any count is an alarm |
| `tx_index_inconsistencies_total` | An exact hot tx index disagreed with the ledger store — any count means corruption |

## Deferred on purpose

| What | Where it is tracked |
|---|---|
| Two pre-existing getTransactions cursor bugs (v1 parity): cursor `"0"` on an above-tip request; a fully-consumed cursor regressing | #937 F1/F2 — one small standalone PR; this PR is docs-only on cursor semantics |
| Per-method config lives in four hand-synced tables; cold `.idx` files reopen per request | #937 H1/H2 |
| A per-run readiness signal for getHealth (v1 has the identical behavior) | #937 I1 |

## Testing

- End-to-end over HTTP: the method table against a seeded registry; server rebind across supervised attempts.
- Request-view semantics: frozen at acquisition, shared across adapters, `Done` leaves the request's view alive, a bare context errors.
- The four invariant counters, each at its site; grace derivation; the walk span cap; header slices equal a full decode over V1 and V2 LCMs; V1-LCM adapter coverage; txsub series land after a v2 submission.
- Full `internal/methods` + `internal/network` + `internal/rpcv2/...` sweep green; lint (`--new-from-rev origin/feature/full-history`) 0 issues.


cmd/stellar-rpc/internal/rpcv2/jsonrpc.go
```diff
@@ -0,0 +1,202 @@
...
+//
+// A store closed under an in-flight request also counts on the
+// StoreClosedServed metric: reaching a client at all means the request
+// outlived the deletion grace period, which operators must see.
+func wrapAdapterRequest(h jrpc2.Handler, metrics observability.Metrics) jrpc2.Handler {
+	return func(ctx context.Context, req *jrpc2.Request) (any, error) {
+		ctx, mark := adapters.WithErrorMark(ctx)
+		ctx, releaseView := adapters.WithSharedView(ctx)
+		result, err := h(ctx, req)
+		releaseView()
```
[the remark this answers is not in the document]
karthikiyer56: Done in 3ce93eff — the release is now deferred, so it runs during the panic unwind the duration limiter recovers from. Pinned by TestWrapAdapterRequest_PanicReleasesSharedView (a handler that acquires the view and panics; the catalog must close with no unreleased snapshots).

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -0,0 +1,334 @@
...
+	scan, err := tx.view.ScanLedgers(start, end)
+	if err != nil {
+		return nil, markErr(ctx, err)
+	}
+	out := make([]store.LedgerMetadataChunk, 0, end-start+1)
+	for entry, err := range scan {
+		if err != nil {
+			return nil, markErr(ctx, err)
+		}
+		header, err := headerFromLCMBytes(entry.Bytes)
```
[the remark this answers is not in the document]
karthikiyer56: Done in c78a1d14 — BatchGetLedgers checks its context per scanned ledger, matching the checks its siblings (GetLedger, StreamLedgerRange) already had. Pinned by TestLedgerReaderTx_BatchGetLedgersStopsOnCanceledContext.

cmd/stellar-rpc/internal/rpcv2/serve.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			handler.Close()
+			return nil, fmt.Errorf("read server listen on %q: %w", deps.cfg.Service.Endpoint, err)
+		}
+
+		server := &http.Server{Handler: handler, ReadTimeout: jsonrpc.DefaultHTTPReadTimeout}
+		go func() {
+			if serr := server.Serve(listener); serr != nil && !errors.Is(serr, http.ErrServerClosed) {
+				deps.params.logger.WithError(serr).Warn("read server exited")
+			}
```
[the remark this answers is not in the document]
karthikiyer56: Done in 8f7c69b4 for the read server: ServeReads now returns a channel reporting a non-graceful Serve exit, and run() joins it into the attempt's errgroup — a dead accept loop fails the attempt and the supervisor rebinds (v1 treated this as fatal; this restores the equivalent, in-process). Pinned both ways: TestRun_ReadServerDeathFailsTheAttempt, plus a graceful-stop-reports-nothing assertion in the rebind test.

Line 93 (admin server) is log-only on purpose, matching v1 (rpcv1/daemon Run uses Error, not Fatal, for the admin server): a dead pprof/metrics endpoint must not take down a node still serving reads. Upgraded that log from Warn to Error and documented the choice.

cmd/stellar-rpc/internal/rpcv2/config/config.go
```diff
@@ -122,14 +120,17 @@ type MethodsConfig struct {
...
 	GetVersionInfo  MethodConfig          `toml:"getVersionInfo"`
 	GetLatestLedger MethodConfig          `toml:"getLatestLedger"`
 	GetTransaction  MethodConfig          `toml:"getTransaction"`
 	GetTransactions PaginatedMethodConfig `toml:"getTransactions"`
 	GetLedgers      PaginatedMethodConfig `toml:"getLedgers"`
 	GetEvents       PaginatedMethodConfig `toml:"getEvents"`
-	GetFeeStats     MethodConfig          `toml:"getFeeStats"`
+	// GetEventsV2 is the one method v2 serves that v1 does not; it defaults to
+	// getEvents' values.
+	GetEventsV2 PaginatedMethodConfig `toml:"getEventsV2"`
```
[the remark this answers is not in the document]
karthikiyer56: Done in 72cc5845 — getEventsV2 added to both validation lists (the queue/duration list and validatePaginatedMethods), with invalid-value test cases alongside the other methods'.

cmd/stellar-rpc/internal/rpcv2/adapters/errormark.go
```diff
@@ -0,0 +1,68 @@
...
+//	if err != nil && errors.Is(mark.Err(), query.ErrUnavailable) {
+//		// answer "temporarily unavailable, retry" instead of the
+//		// handler's generic internal error
+//	}
+//
+// The mark is only consulted when the handler ultimately fails, so an error an
+// adapter hit but the handler recovered from cannot change a successful
+// response. It CAN mislabel a failure whose final cause differs from the
+// recorded one — accepted: the mapped conditions are rare, and the worst case
+// is telling a client to retry an error it would otherwise see as internal.
```
tamirms: We traced the reachability of all three conditions this channel classifies and ended up
concluding the mechanism has no beneficiary, so I want to propose deleting it rather than
hardening it.

The reachability chain: `*query.RangeError` cannot reach the wrapper because every handler
validates its range against the same snapshot the scan uses. `query.ErrUnavailable` is
closed by the design itself: query-routing-design.md:144 proves both skew directions
(coverage is published before discard and routing prefers cold; the admitted latest never
points into an unpublished chunk), and this PR's `windowGatedIndex` closes the remaining
prune-direction path through a stale hot index. `stores.ErrStoreClosed` requires a request
to outlive the deletion grace, and since grace is derived as max(all budgets) + 30s, any
request old enough to hit it had its context canceled and its client answered by the
duration limiter (-32001) at least 30 seconds earlier. So in every single-failure world the
-32002 remap is delivered to a closed connection, and the other two arms are unreachable
outright.

The PR's own deferred list already identifies the observability halves of this (the D2/D3
rows) and the enforcement weakness (the E2 row), so most of this is your analysis carried
one step further. The one step: the funnel direction E2 sketches was also evaluated and
fails on the probe's soft-error semantics (recording at the seams lets a routine trailing
index miss clobber a soft ErrUnavailable, losing exactly the classification the mark
preserves), which is what pushed the conclusion from "enforce the marking" to "delete the
remap and keep only the counting". What remains real is the observability, and that does
not need the mark: each condition can be counted at its production site, where the error
and its cause are both known.

- Store-closed: count in the rocksdb store's op guard, gated on which close set the flag.
  `Close` (teardown) versus `CloseIfIdle` (deferred deletion) is exactly the distinction
  the current live-context check approximates, and it needs no context plumbing. This also
  counts abandoned-straggler hits, which are real margin-violation signals the current
  filter discards, and excludes restart noise precisely instead of by proxy.
- `ErrUnavailable`: an unconditional counter plus a structured log (chunk id in the log,
  not a label) at `resolveTier`. Under the corrected model every occurrence is a violated
  invariant, which resolves #937-D2 in its simplest form.
- Cold tier: an ENOENT-class counter at the cold reader. Routing only sends in-window reads
  at coverages the snapshot holds, so any occurrence is alarm-worthy. This closes #937-D3
  and the ENOENT-on-cold-open alarm item carried since the #894 review.
- `txhash.ErrInconsistent`: one `errors.Is` count at its single consumption site in
  `GetTransaction`. Hot index/store disagreement is corruption operators should see. With
  the sharedview change this is the one dependency TransactionReader keeps besides the
  passphrase; the other three counters live below the adapters.

A single counter with a low-cardinality cause label would give operators one
flatline-at-zero alert rule; separate counters match the existing style. Either works.

Concretely this deletes: this file and its tests, the classification switch in
rpcv2/jsonrpc.go (the wrapper's one remaining job is the per-request view scope), every
`markErr` wrap, and `network.ErrTemporarilyUnavailable` — v1 never emits -32002, so with
the remap gone v2 introduces no wire error code v1 lacks. Follow-ups: #937-E2 dissolves (nothing left to enforce), D2/D3 become the
two counters, and the comments that teach the race model would be reworded to "unreachable
within the model; counted if ever wrong" (this file, the windowGatedIndex rationale's
"-32002" sentence — reworded in the same edit that relocates it into query per the
txlookup thread — hotloop.go:159-164, and the PR body's "only the read-view acquisition
race produces it" parenthetical). Test fallout: the TestMapAdapterErrors_* trio,
errormark_test.go, and the e2e -32002 assertions get replaced by counter-increment
assertions. Since the counting semantics change from "served to a client" to "observed",
StoreClosedServed deserves a rename too; nothing has shipped. One more #937 line while it
is being restructured: the per-request snapshot leak gauge from the #894 F5 list is still
unowned and could live there. Happy to talk through any part of the reachability argument
if it looks off.

cmd/stellar-rpc/internal/rpcv2/query/txlookup.go
```diff
@@ -24,8 +25,9 @@ import (
 //
 // It is deliberately unfiltered — every published handle, regardless of the
 // view's floor/latest. A match can therefore name a ledger in a chunk below the
-// floor (a handle that predates this view); the lookup's floor/latest gate on
-// the resolved ledger is the only thing that keeps such a match from being served.
+// floor (a handle that predates this view); the caller's window gate, wrapped
```
tamirms: Could the window gate move into these accessors, so `HotTxHashIndexes` and `ColdTxIndexes`
return gated indexes and this "deliberately unfiltered" caveat is deleted? The view already
knows its window, the adapter is the only consumer of both accessors (verified by sweep),
and the wrap type relocates verbatim with its rationale comment. The F5 probe-gate
obligation from the #894 review then stops being a per-consumer convention and becomes a
property of the accessors, which matters before #774 adds the next lookup consumer.

Two details worth preserving in the move. The gate must stay at the index layer rather than
fetch time: the exact-index probe classifies a fetch-level ErrNotFound as ErrInconsistent
(corruption), so gating only at fetch would turn the below-floor hot-handle race into an
internal error where NOT_FOUND is the truth. And viewLedgerSource's sub-genesis guard stays
as panic defense for corrupt index data. On the adapter side, `gateIndexes` and the
provider closure both delete, since the accessor is already the provider signature:

    probe, err := txhash.NewTxReader(
        view.HotTxHashIndexes(), view.ColdTxIndexes,
        &viewLedgerSource{view: view}, r.passphrase)

The txlookup tests would pin the gated contract (a published below-floor handle's hit
answers ErrNotFound from the accessor); the mid-prune adapter test stays as the end-to-end
pin.

cmd/stellar-rpc/internal/rpcv2/adapters/sharedview.go
```diff
@@ -0,0 +1,59 @@
...
+
+// WithSharedView returns ctx carrying an empty holder, plus the release the
+// serving wrapper must call after the handler returns:
+//
+//	ctx, release := adapters.WithSharedView(ctx)
+//	result, err := handler(ctx, req)
+//	release()
+//
+// release is a no-op if no adapter call ever acquired a view.
+func WithSharedView(ctx context.Context) (context.Context, func()) {
```
tamirms: What would you think of the wrapper owning the view eagerly, with the context carrying an
immutable value instead of this lazily filled holder?

    func withRequestView(h jrpc2.Handler, reg *query.Registry) jrpc2.Handler {
        return func(ctx context.Context, req *jrpc2.Request) (any, error) {
            view, err := reg.NewReadView()
            if err != nil {
                return nil, err
            }
            defer view.Release()
            return h(adapters.WithView(ctx, view), req)
        }
    }

The mutable slot, the three-way acquireView, and the test fallback all collapse into a
WithView/viewFrom pair, and the panic-safety defer carries over unchanged. What laziness
buys today is a microsecond-scale catalog snapshot avoided on the three lowest-volume
methods, which does not seem to pay for the extra states. Knock-on: the adapters stop
needing the registry at all (together with the ledger_reader.go cache-surface suggestion),
so `NewLedgerReader()` takes nothing and `NewTransactionReader` takes the passphrase plus
the metrics handle for the corruption counter (errormark thread), and
`ledgerReaderTx.Done` stops owning any release, which the TxDoneLeavesTheRequestViewAlive
test already pins as the intent. Tests that use adapters directly get a small helper that
wraps a context instead of the implicit fresh-view branch.

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -0,0 +1,331 @@
...
+
+// ledgerReaderTx satisfies store.LedgerReaderTx over one read view. GetLedger
+// serves getTransactions' ascending, contiguous per-ledger walk by pulling from
+// a single ScanLedgers iterator primed on the first call; GetLedgerRange and
+// BatchGetLedgers read through the same view but never touch that iterator.
+type ledgerReaderTx struct {
+	view    *query.ReadView
+	release func()
+	// registry outlives the view; GetLedgerRange writes the oldest-close-time
+	// cache through it.
```
tamirms: The cache mechanism itself is exactly right and worth keeping verbatim: read-through with
staleness-by-seq-equality means a moved floor invalidates by construction, and the
last-write-wins argument is sound. The suggestion is only about its surface: this registry
field exists for one call, and `RecordOldestCloseTime(seq, ct)` asks the caller to supply a
seq the view already knows. If the view carried the write (an unexported pointer to the
registry's atomic, `view.RecordOldestCloseTime(ct)` with the seq implied by the view's own
floor), a mismatched record becomes inexpressible, this field and getLedgerRange's second
parameter delete, and the decode stays here in the adapters so query stays xdr-free. For
the record, a zero-read optimum exists (each chunk record carrying its first close time,
written at first commit and carried through freeze and backfill), but it is a
four-subsystem schema change to save one cold read per floor move; it becomes worth it only
if a future feature such as by-time range queries needs per-chunk time metadata anyway, and
then this cache should dissolve into it.

tamirms (commented):
Review pass over the serving wiring and the adapters; 9 inline comments below. The serve.go thread carries the branch-level lifecycle decision context the others assume.

cmd/stellar-rpc/internal/rpcv2/query/registry.go
```diff
@@ -93,7 +110,10 @@ func OpenRegistry(
 		return nil, err
 	}
 	r.PublishHandle(live.ChunkID(), live)
-	r.SetLatestLedger(lastCommitted)
+	// lastCommitted comes from the catalog, which has no close times, so the
+	// stamp starts with close time 0 (unknown): adapters fall back to one point
+	// read until the first ingested ledger stamps a real value.
+	r.SetLatestLedger(lastCommitted, 0)
```
tamirms, line 121: The PR's deferred list already flags the boot-window tip reads and sketches a seq-guarded
cache as the follow-up (#937-B2). A simpler shape that covers both edges: one point read
per edge at startup (guarded by the empty-catalog case), stamping both close times before
serving begins. No new cache mechanism, the deferred item closes instead of growing one,
and with retention_chunks = 0 pinning the floor, the oldest-edge fallback is dead for the
process lifetime since the floor never moves in full-history mode. The request-path
fallbacks then survive purely as backstops. With the cache write moved onto the view
(the ledger_reader thread), the seed writes through a short-lived view, which it needs for
the point read anyway, and the registry-level write method goes fully unexported.

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -0,0 +1,331 @@
...
+		})
+	}
+	return out, nil
+}
+
+// headerFromLCMBytes decodes only the LedgerHeaderHistoryEntry out of raw LCM
+// bytes. The consumer (getLedgers) reads identity fields from the header and
+// sends the metadata as the raw bytes themselves, so unmarshaling the whole
+// LedgerCloseMeta — potentially MBs of transactions per ledger — is pure waste.
+func headerFromLCMBytes(raw []byte) (xdr.LedgerHeaderHistoryEntry, error) {
```
tamirms: This decode exists to fill `LedgerMetadataChunk.Header`, and the handler's XDR path
immediately re-encodes it (`header.MarshalBinary` at get_ledgers.go:283) to base64 the same
bytes this function was already holding. Since the wire needs raw header bytes plus three
scalars, and the SDK view exposes the scalars (`LedgerSequence`, `LedgerHash`,
`LedgerCloseTime`), would it make sense to finish the #854 reshape one field further:
`LedgerMetadataChunk{HeaderRaw []byte; Lcm []byte}`? The handler base64s HeaderRaw
directly, decodes only on the JSON path (which decodes everything anyway), the datastore
fallback marshals the header beside the LCM it already marshals, and this function shrinks
to the view slice plus a clone. Wire output is byte-identical for both daemons, and
get_ledgers.go is the type's only consumer, so the change is contained.

cmd/stellar-rpc/internal/rpcv2/hotloop.go
```diff
@@ -197,18 +199,20 @@ func runIngestionLoop(ctx context.Context, cfg ingestionLoopConfig) error {
 		// The tick must not touch it — its chunk-aligned value would regress it.
 		metrics.LastCommitted(seq)
 
+		// A committed ledger's close time always decodes; on the near-impossible
+		// decode error, publish 0 (readers fall back to a point read) rather
+		// than fail an already-durable commit.
+		closeUnix, cerr := view.LedgerCloseTime()
```
tamirms: `IngestLedger` already decodes the close time and rejects the ledger when it does not
decode (hotchunk.go:381-386), so by the time the loop gets here the decode cannot fail:
this fallback branch is dead and the close time is decoded twice per commit. Could the
ingest path return `closedAt` (it is in hand at hotchunk.go:381) so the loop stamps the
returned value? The 0-means-unknown convention stays for the boot seed only.

cmd/stellar-rpc/internal/rpcv2/serve.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// newServeReads returns the production ServeReads: per supervised attempt it
+// builds the method table over that attempt's registry, binds
+// [service].endpoint, and serves until the returned stop function runs. run()
+// calls stop before its registry closes, so the port is released before the
+// next attempt binds and no handler outlives its stores. A Serve failure that
+// is NOT the graceful shutdown (v1 treats these as fatal) lands on the
+// returned channel, failing the attempt so the supervisor rebinds.
+func newServeReads(deps readServerDeps) func(context.Context, *query.Registry) (func(), <-chan error, error) {
```
tamirms, line 43: Branch-level decision, landing as a small precursor PR before this one merges: the
supervised restart loop is being deleted in favor of crash-only. run() is called once and
becomes the process body; failures past the existing bounded retries exit, and Kubernetes
owns restarts. The short rationale: backfill already carries three bounded retry layers and
nothing serves during its phase, so crash-looping through an outage there is harmless;
steady-state write failures are environmental or deterministic (OOM, disk full, a stale
core binary at a protocol upgrade), which is the crash class; and a coupling sweep found
the codebase is already process-boot-shaped (fee replay, watermark derivation, boundary
seeding, the e2e restart tests, and bench all recover from durable state).

For this file that means bind-once/build-once beside the admin server, and the following
delete outright rather than needing review: the (func(), <-chan error, error) seam, the
died channel with its buffered-send, sender-filter, and nil-channel-for-tests conventions,
the readsDied errgroup leg in startup.go (a dead accept loop becomes process-fatal, which
is v1's classification — this file's own comment notes v1 treats it as fatal; the admin
server stays log-only), newServeReads/readServerDeps with the handlerParams
copy-then-overwrite, the HandlerMetrics carrier in internal/jsonrpc (build-once makes
registration trivially once; the rebind-across-attempts test and the port-clash failure
mode it tests also go). One sub-choice to record in the precursor: serving can begin
before initial backfill, honestly by construction, since the window gates already answer
for ledgers that do not exist yet.

cmd/stellar-rpc/internal/jsonrpc/specs.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/methods"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcdatastore"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/store"
+)
+
+// SpecDeps carries every input the shared internal/methods constructors take.
+// Each daemon fills it from its own config and wiring and hands it to
+// BuildHandlerSpecs, so the method table — which method runs which handler
+// with which decode limits — is written once instead of once per daemon.
+type SpecDeps struct {
```
tamirms: Three naming suggestions, best batched with other edits rather than standalone churn.
`SpecDeps` names its consumer rather than its contents; `HandlerDeps` says what it holds
(the shared handlers' inputs) and keeps the family on its existing root
(HandlerSpec/BuildHandlerSpecs/NewHandler). `SpecLimits` next to `MethodLimits` makes the
reader check which is the map; `LimitsByMethod` names the key and `LimitsByMethod.Apply`
still reads cleanly. And `GetEventsHandler` is an override, not a dependency; calling it
`GetEventsOverride` makes that visible for the field's remaining lifetime until #774 builds
getEvents uniformly.

---

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
```
urvisavla, line 5: Are these rpcv2 related changes? If so, it’s probably premature to add them to the changelog and if not, you’ll want to create a separate pr against the main branch with these changes (?).

cmd/stellar-rpc/internal/rpcv2/adapters/errormark.go
karthikiyer56: Done in 258f3457. errormark.go, every markErr wrap, the classification switch, and network.ErrTemporarilyUnavailable are deleted. The wrapper keeps only the per-request view scope. Four counters landed: store_ops_after_deferred_close_total in the rocksdb op guard, gated on CloseIfIdle vs Close (replaces store_closed_reads_total); unavailable_chunk_resolves_total plus a structured log with the chunk id at the routing decision; missing_cold_pack_opens_total at the cold reader; tx_index_inconsistencies_total counted in GetTransaction. Each counter has a test. The inconsistency test builds a freeze-overlap pack that lacks the transaction. #937 is restructured: E2 dissolved, D2/D3 marked resolved by the counters, and the snapshot-leak gauge added as row G1.

cmd/stellar-rpc/internal/rpcv2/query/txlookup.go
karthikiyer56: Done in 847a91c8. HotTxHashIndexes and ColdTxIndexes now return gated indexes. windowGatedIndex moved into query/txlookup.go with its rationale comment; the -32002 sentence was reworded when 258f3457 removed that code. The adapter-side gateIndexes helper and the cold closure are deleted. The gate stays at the index layer. TestWindowGatedIndex_OutOfWindowHitIsAMiss pins the contract in query; the mid-prune adapter test stays as the end-to-end pin.

cmd/stellar-rpc/internal/rpcv2/adapters/sharedview.go
karthikiyer56: Done in 0bb11c94. wrapAdapterRequest acquires the view before the handler runs and releases it after the handler returns. The context carries the view (WithView / viewFrom). The lazy holder and its three-way acquire path are deleted. An adapter call errors when the context has no view. The panic-safety release keeps its test.

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
karthikiyer56: Done in e08f92ce. The cache cell and its seq-guarded read are unchanged. The write moved onto the view: ReadView.RecordOldestCloseTime stamps the view's own floor ledger, so a caller cannot record a wrong sequence. The registry method is deleted. The adapters no longer hold a registry pointer: NewLedgerReader takes nothing, and NewTransactionReader takes the passphrase (plus the metrics sink after 258f3457).

cmd/stellar-rpc/internal/rpcv2/query/registry.go
karthikiyer56: Done in 5164c1b6. adapters.SeedCloseTimes runs at startup, after OpenRegistry and before ServeReads. It stamps both window edges with one point read each. An empty catalog is a no-op. A seed failure is a restartable startup error. The request-path fallbacks stay as backstops. This closes #937-B2.

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
karthikiyer56: Done in 5af684bc. LedgerMetadataChunk now carries HeaderRaw + Lcm; the adapter slices the header bytes with no decode, the XDR path base64s HeaderRaw directly, and the JSON path converts from bytes. The datastore fallback marshals the header beside the LCM, and the v1 SQL reader slices the header segment it already walks. Wire output is byte-identical.

cmd/stellar-rpc/internal/rpcv2/hotloop.go
karthikiyer56: Done in b3b8f56f. IngestLedger reports the close time on its LedgerReport; Ingest returns it and the loop stamps the returned value. The dead fallback branch is gone, and 0-means-unknown survives only for the boot seed. The unused ctx parameter on Ingest was dropped in the same commit.

cmd/stellar-rpc/internal/rpcv2/serve.go
karthikiyer56: Acknowledged — holding this PR's serve/startup shape as is.
can I request that this PR go in first and you do undo the run() in a loop to a single run(). - one that, if it fails, halts ingestion and crashes. 
that is probably cleaner.
I am not sure why we had the retry loop in the first place. 
prapbaly because rpcv1 has similar structure - we shud revisit that too in a future issue

cmd/stellar-rpc/internal/jsonrpc/specs.go
karthikiyer56: Renames done in 8f027ea5: SpecDeps is now HandlerDeps, and SpecLimits is now LimitsByMethod (the local helper and the test helpers follow). GetEventsHandler keeps its name: since 553739e2 it is a required per-daemon dependency, not an override. The two structural findings from the same review round were deferred to #937 as rows H1 (four parallel per-method tables) and H2 (per-view cold .idx reopen), each with a fix direction written up there.

CHANGELOG.md
karthikiyer56: These entries document changes to the shared internal/methods handlers.
so, once this PR lands,  they are v1-visible behavior changes, not rpcv2 ones — that's why they belong in the changelog: the getTransactions scan bound applies to v1 today, and the stricter strkey parsing in getEvents filters came in through the SDK bump v1 also uses. 

On placement: they sit on this feature branch, and the changelog is inherited by main whenever feature/full-history folds in. 
So no separate PR against main is needed atm


On why the scan bound lives in shared code rather than rpcv2 only: this does change semantics for existing v1 clients — once this branch merges into main, a sparse range that used to come back in one call can return a short page and a cursor, and clients have to page. That said, scoping the bound to v2 would have left v1 unbounded and v2 capped for the identical request, and I'd argue the stipulation is worth taking in v1 whenever that merge happens rather than living with that split later.

tamirms (approved):

PR #909, opened 2026-07-31
Remove the deprecated Soroban Tools e2e workflow

marwen-abid:
### What
Delete `.github/workflows/e2e.yml`.

### Why
The workflow calls a reusable workflow from `stellar/system-test`, which was transferred to the `stellar-deprecated` org and archived.

All the runs have been failing for a while https://github.com/stellar/stellar-rpc/actions/workflows/e2e.yml


Shaptic (approved):

Shaptic:
Sad to see this go, it honestly had so much potential but fell by the wayside.

PR #914, opened 2026-08-03
rpcv2: add the getEvents v2 event cursor and codec

urvisavla:
### What

The server-minted, opaque, versioned cursor behind getEvents v2 pagination ([proposal](https://github.com/orgs/stellar/discussions/1872)): `EventCursor` (canonical query, optional position of the last delivered event, scanned-ledger watermark) encoded as a `gec1_`-prefixed base64url token over a fixed-layout binary body. Decode checks well-formedness only and returns typed errors (`ErrCursorMalformed`, `ErrCursorUnknownVersion`) for the handler to map; validation against the serving window and budgets stays with the handler.

### Why

v2 pagination needs a cursor that carries the whole query and resume state, works identically over a 7-day window and full history, and never expires. This PR is the codec foundation; the cross-chunk pager and the getEventsV2 handler build on it in follow-ups.

### Notes for review

- The format version is cleartext in the token prefix, so version skew is greppable in logs and a future body format is a clean version bump; every version's decoder stays supported forever.
- One envelope has one encoding: presence bits for optional fields, exact-consume decode, strict base64. Decode-then-encode is byte-identical, pinned by the fuzz target. A cleared max-ledger bit is the proposal's open upper bound (ascending follows the tip) — no sentinel value.
- Every decode read checks remaining length first, and the filter cap applies before allocation, so a forged cursor cannot make decode allocate beyond its own size. Encode enforces the same checks as decode, so a minting bug fails on the page that has it rather than on the resume.
- Golden vectors pin the v1 layout as literal bytes; the malformed table's rows are mutation-verified — each guard has an input only that guard can reject.
- The worst-case measurement (256 full filters) encodes to ~102 KiB, well under the 1 MiB cap.

### DIFF
--- a/cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
+++ b/cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
@@ -1,0 +1,320 @@
+package query
+
+// The getEvents v2 cursor envelope and its codec: the versioned, opaque,
+// self-contained structure behind a response's cursor string. Only the server
+// creates or decodes it; clients hold the encoded string and send it
+// back unchanged. The codec checks well-formedness only; validating a query
+// against the serving window and budgets is the handler's job.
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	protocol "github.com/stellar/go-stellar-sdk/protocols/rpc"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/stores/event"
+)
+
+// ErrCursorMalformed rejects a cursor string that does not decode to a
+// well-formed envelope.
+var ErrCursorMalformed = errors.New("query: malformed cursor")
+
+// ErrCursorUnknownVersion rejects a cursor whose version byte this build does
+// not implement. Distinct from ErrCursorMalformed so the handler can tell a
+// corrupt cursor from one minted by a newer server.
+var ErrCursorUnknownVersion = errors.New("query: unknown cursor version")
+
+// cursorVersion is the envelope's format version. The byte sits outside the
+// JSON so a future body format is a clean version bump.
+const cursorVersion = 1
+
+// maxCursorBytes caps the encoded string length DecodeEventCursor accepts,
+// bounding decode work on hostile input. Encode enforces the same cap so an
+// oversized envelope fails at mint rather than on the resume that follows.
+// Sized above anything the request pipeline can produce: the serving stack
+// caps request bodies at 512 KiB (internal/jsonrpc), and an envelope
+// re-wraps roughly the request's filter content.
+const maxCursorBytes = 1 << 20
+
+// maxCursorFilters caps the filter count in both codec directions. It
+// matches the getEvents v2 proposal's 256-filter request cap: no encoder
+// mints more, so a cursor beyond it is forged or corrupt. Cursors never
+// expire, so this constant can only grow; lowering it would reject
+// outstanding cursors minted at the old cap.
+const maxCursorFilters = 256
+
+const contractIDLen = 32
+
+// EventPosition is the ledger-denominated identity of a delivered event: the
+// TOID fields (Ledger, Tx, Op) plus the event index within the operation, and
+// K, the within-ledger index over all stored events of the ledger. K is
+// portable across nodes only while ingest preserves the close meta's
+// within-ledger stream order and the store's inclusion rule stays fixed.
+type EventPosition struct {
+	Ledger, Tx, Op, Event, K uint32
+}
+
+// EventCursorQuery is the canonical (engine) form of the original query the
+// cursor pins: ledger bounds, direction, and the engine's filters.
+type EventCursorQuery struct {
+	MinLedger, MaxLedger uint32
+	Dir                  Direction
+	Filters              []event.Filter
+}
+
+// EventCursor is the envelope: the server-minted structure inside the opaque
+// cursor string, holding the original query, the position of the last
+// delivered event, and the scanned-ledger watermark. A nil Position means no
+// event has been delivered yet, so resume starts from the watermark alone.
+type EventCursor struct {
+	Query         EventCursorQuery
+	Position      *EventPosition
+	ScannedLedger uint32
+}
+
+// The structs below define the envelope's JSON body. Their tags are the
+// version-1 format: cursors never expire, so once v1 ships the key names are
+// frozen and any change to them is a version bump. event.Filter carries no
+// JSON tags and stays serialization-free; these conversions own the body
+// shape instead. []byte fields serialize as standard base64 via
+// encoding/json.
+
+type envelopeJSON struct {
+	Query    queryJSON     `json:"query"`
+	Position *positionJSON `json:"position,omitempty"`
+	Scanned  uint32        `json:"scanned"`
+}
+
+type queryJSON struct {
+	Min     uint32       `json:"min"`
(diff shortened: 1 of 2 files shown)


cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
```diff
@@ -0,0 +1,320 @@
...
+
+const contractIDLen = 32
+
+// EventPosition is the ledger-denominated identity of a delivered event: the
+// TOID fields (Ledger, Tx, Op) plus the event index within the operation, and
+// K, the within-ledger index over all stored events of the ledger. K is
+// portable across nodes only while ingest preserves the close meta's
+// within-ledger stream order and the store's inclusion rule stays fixed.
+type EventPosition struct {
+	Ledger, Tx, Op, Event, K uint32
```
tamirms: In EventPosition every field is named for what it is, Ledger, Tx, Op, Event, except K, which is notation, and it sits next to Event, the within-operation index, so the two sibling ordinals are distinguished only by the comment. What do you think about naming the field for what it is:

```go
// LedgerOrdinal is the within-ledger index over all stored events of the
// ledger. Portable across nodes only while ingest preserves the close
// meta's within-ledger stream order and the store's inclusion rule stays
// fixed.
LedgerOrdinal uint32
```

Independent of the rename, a #774 pointer on the EventPosition doc comment seems worth it: that comment holds the tuple rationale the pager will rely on, the arithmetic resume and the self-validation against the payload header at the claimed ordinal.

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// EventCursorQuery is the canonical (engine) form of the original query the
+// cursor pins: ledger bounds, direction, and the engine's filters.
+type EventCursorQuery struct {
+	MinLedger, MaxLedger uint32
```
tamirms: The proposal says an ascending query with `maxLedger` omitted follows the chain tip forever, and unlike descending, where the default is fixed at query start, that open upper bound persists across pages. So the envelope needs a way to say "no upper bound", and today it has none: 0 would trip the `min > max` decode check, and nothing else is defined. Since the v1 format freezes in this PR, could we pin the representation here rather than in the handler PR? I would argue for absence rather than a sentinel, `MaxLedger *uint32` with `"max"` omitted when unbounded, and the `min > max` check applying only when max is present: no legal ledger value gets stolen, and a zero-valued minting bug fails the presence rules instead of encoding an eternal tip-follower. A MaxUint32 sentinel also works with the current check if the pointer feels heavy, but either way the choice needs a doc line on the field and a format-test vector, since the unbounded ascending cursor will probably be the most common one in the wild.

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
```diff
@@ -0,0 +1,320 @@
...
+	ScannedLedger uint32
+}
+
+// The structs below define the envelope's JSON body. Their tags are the
+// version-1 format: cursors never expire, so once v1 ships the key names are
+// frozen and any change to them is a version bump. event.Filter carries no
+// JSON tags and stays serialization-free; these conversions own the body
+// shape instead. []byte fields serialize as standard base64 via
+// encoding/json.
+
```
tamirms: Before v1 freezes I want to raise the body encoding itself. Cursors never expire, so whatever format v1 accepts we have to keep decoding forever. Two things make JSON risky to freeze.

First, the definition of a valid v1 cursor does not fully live in this file; part of it lives in encoding/json's tolerance. The decoder quietly accepts many bodies we would never mint: keys in the wrong case, duplicate keys, unknown keys, null in place of a filter object. The tests pin some of this, but only the cases we thought to enumerate, and the stdlib's behavior is set to shift when json/v2 arrives. Freezing v1 on top of that means owning those quirks forever.

Second, hostile input can make json.Unmarshal do a lot of work before our checks run. The brace guard counts objects, but a filters array of nulls has no braces: a ~1 MiB cursor of `null,` entries allocates 114 MB inside json.Unmarshal before the filter cap rejects it, measured on this branch (a digit variant reaches 313 MB). Guards can patch each case, but every patch is one more rule to carry.

A simple binary body avoids both problems by construction rather than by guard. Concretely:

```
token: "gec1_" || base64url(body)
body:  u8 flags: dir | hasMax | hasPos | rest must be zero
       u32 min | u32 max if hasMax | 5 x u32 position if hasPos | u32 scanned |
       u16 filter count | per filter:
         u8 fflags: hasContract | 4 topic-presence bits | rest must be zero
         32B contract if hasContract | per present topic: u32 len | bytes   (u32: a topic can exceed 64 KiB)
```

What this buys, in plain terms:

- One encoding per cursor. Every optional field is a presence bit plus an omitted field, a set topic bit must carry a non-empty value, and decode must consume the body exactly. So decode-then-encode reproduces the input byte for byte, and the canonicalizationSlack escape hatch in the fuzz target disappears.
- No sentinels. "No maxLedger" is just a cleared bit, which also answers the open-upper-bound question from my other comment without stealing a legal ledger number.
- Allocation follows input. Decode is one pass of "check the remaining length, then read". There is nothing a hostile cursor can inflate.
- The guards shrink to a handful of length and count checks. The only subtle one: reject any payload character the base64 decoder would silently skip (it skips CR and LF), so that one token means one body.
- A readable version marker. The cleartext prefix ("gec" for getEvents cursor, then the format version, spelling up for grabs) makes version skew greppable in logs, and later changes are clean: the #904 filter fields take two of the reserved flag bits in a "gec2_", each version's decoder stays supported forever, and a bump ships the new decoder everywhere before anything mints it, so in-flight paginations never break during a deploy.

Cost-wise it is roughly a wash: about 250 lines of stdlib codec plus 350 of tests, against what is here once the leniency-pinning tests are counted, and the encoded size comes out comparable or a bit smaller. The real loss is `base64 -d | jq` on a pasted cursor; a ten-line dump helper covers that. The typed errors, the caps, the tripwire, and the golden-plus-fuzz test structure all carry over unchanged.

If a library feels safer than a hand-rolled codec, fxamacker/cbor/v2 in strict mode is workable; I would avoid protobuf here, since unknown fields round-trip by default and its decoder has the same unbounded allocation as encoding/json. The probe test behind the flood numbers is yours if useful. What do you think?

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
```diff
@@ -0,0 +1,320 @@
...
+	directionDesc = "desc"
+)
+
+// Encode serializes the envelope as base64url over (version byte ||
+// JSON). Encode fails on an invalid Direction, a wrong-length ContractID,
+// more than maxCursorFilters filters, or output beyond maxCursorBytes.
+//
+// Absent values must be nil: the wire format cannot tell nil from empty, so
+// empty non-nil values decode back as nil (no exact round-trip). Inverted
+// ledger bounds encode without error but fail the next decode.
```
tamirms: Encode fails minting bugs in place for direction, contract length, filter count, and size. The PR description frames that as the point of the double-sided checks, a bug fails on the page that has it, and the oversize test states the same rule as its comment: "Encode refuses output DecodeEventCursor would reject". Inverted bounds are the one exception to both statements, and they surface in the worst place, on the client's next resume. Was there a reason to leave this one out? Two lines in Encode would make it symmetric with decode. It also interacts with the maxLedger representation question: whichever form the open upper bound takes, both sides need to accept it.

tamirms (approved):

Issue #883, opened 2026-07-21
rpcv2 integration tests: run the rpcv1 suite unchanged against rpcv2, four CI legs


karthikiyer56:
Part of #986.

## Goal

Run the rpcv1 integration suite, assertions unchanged, against rpcv2. CI runs four legs instead of two.

| leg | daemon | protocol |
|---|---|---|
| 1 | rpcv1 | P27 |
| 2 | rpcv1 | P28 |
| 3 | rpcv2 | P27 |
| 4 | rpcv2 | P28 |

- The only thing that may differ between the two daemons is how the harness instantiates them.
- rpcv1's integration-test settings do not change.
- rpcv2 starts with the stock defaults from `cmd/stellar-rpc/rpcv2/rpc-v2-sample-config.toml`.
- Run model stays as today: one Core container per test, captive core as a subprocess, the daemon in-process.

This replaces the earlier in-process `paritytest` design. rpcv2 now serves every rpcv1 method natively (#908, #960), so the suite itself is the parity test.

## Layout

| path | holds |
|---|---|
| `cmd/stellar-rpc/internal/integrationtest/` | the common tests and `infrastructure/` (moved up from `rpcv1/`) |
| `cmd/stellar-rpc/internal/rpcv1/integrationtest/` | rpcv1-only tests |
| `cmd/stellar-rpc/internal/rpcv2/integrationtest/` | home for rpcv2-only tests (rpcv2 backfill, `getEventsV2`), empty except `doc.go` |

- The harness picks the daemon from `STELLAR_RPC_INTEGRATION_TESTS_DAEMON` (`rpcv1` default, `rpcv2`).
- Each CI leg runs the common package and its own daemon's package.

## Which tests run where

| test | placement | reason |
|---|---|---|
| `TestGetLedgersFromDatastore`, `backfill_test.go`, `migrate_test.go`, `ingest_loadtest_test.go` | rpcv1-only | mutate `rpcv1/config.Config`, read SQLite, or run released rpcv1 images |
| `TestArchiveUserAgent` | rpcv1-only | pins rpcv1's `User-Agent`; rpcv2 has none yet (follow-up below) |
| `TestHealth` | common, minus one line | `LedgerRetentionWindow == 17280` moves to a per-daemon test; rpcv2 measures retention in 10 000-ledger chunks |
| `TestUpgradeFrom20To21` | common | probe changes from `GetDaemon().GetDB()` to `getLatestLedger`; assertions unchanged |
[...]

PR #917, opened 2026-08-04
v2 ingestion: one walk per ledger — feed and rebuild the getFeeStats windows (#881, #888)

### LINKED ISSUE #881: v2 ingestion: one walk per ledger — migrate backfill + streaming to parts/products, and feed the getFeeStats windows
## Context — what changed

- Supersedes the 2026-07-27 "Part 1 (SDK) / Part 2 (rpc)" plan that previously lived on this issue.
- [tamirms' review on stellar/go-stellar-sdk#5966](https://github.com/stellar/go-stellar-sdk/pull/5966#issuecomment-5141047833) reshaped the SDK API: one `ExtractLedgerTxParts` walk + `EventsFromTxParts` / `FeesFromTxParts` products. The SDK side is now tracked in #912 (PR 5966 re-targeted there).
- This issue is now purely the stellar-rpc side, reframed:
  - **v2 ingestion walks each ledger exactly once — streaming and backfill — and every per-ledger product (txhash entries, event payloads, fee buckets) is a read over that one walk.**
  - Fee stats stops being a special second pass; it falls out of the walk like everything else.
- What this deletes from the old plan: the passphrase. Fee classification no longer touches envelopes, so the `resolveCore`-passphrase plumbing step is gone entirely.

## Work item 1 — migrate the two walk sites to parts + products

- Hot: `internal/rpcv2/stores/hotchunk/hotchunk.go` `IngestLedger` — currently calls `ExtractLedgerEvents`.
  - Switch to `ExtractLedgerTxParts` + `EventsFromTxParts`.
  - txhash entries build from `parts[i].Hash` / `parts[i].InnerHash` / `parts[i].FeeBump` (today they're mined out of the events bundle — the bundling forced that workaround).
  - The pre-extract-before-batch discipline is preserved: a walk or product failure still rejects the ledger before any batch opens.
- Cold: `internal/rpcv2/ingest/ingester.go` `coldChunk.ingest` — same switch.
  - **Fees are never computed on the cold path**, and that stays structural: no fee consumer → no `FeesFromTxParts` call. Backfill pays nothing for a product it won't serve.
- `internal/rpcv2/events` `PayloadsFromLedgerEvents`: the slimmed `TxEvents` no longer carries hashes, so the signature becomes `(parts []ingest.LedgerTxParts, events []ingest.TxEvents, ledgerSeq, ledgerClosedAt)` — index-aligned; payload `TxHash` reads `parts[i].Hash`.
- Tests that use `ExtractTxHashes` as an oracle (`rpcv2/ingest/ingest_test.go`, `rpcv2/daemon_test.go` helper) re-target `parts[i].Hash` — the SDK function is deleted.
- Metering: `PhaseExtract` / `ColdExtract` now time the walk + product reads together; the metered claim stays "one walk per ledger".

## Work item 2 — share the window (unchanged from the earlier plan)

- Drop the `db *sqlitedb.DB` field and constructor param from `FeeWindows`; methods return plain errors (removes v1's double-rollback).
- Move the package from `internal/rpcv1/feewindow` to shared `internal/feewindow`; the `AsMigration` adapter stays v1-side next to its only caller.
- `computeFeeDistribution`, `ledgerbucketwindow`, and `store.FeeStats` are untouched — the distribution math stays parity-by-construction between v1 and v2.
- v1 keeps feeding via its parsed-LCM `IngestFees`, behavior unchanged.

## Work item 3 — feed the windows from the live loop

- Hoist the walk one level: `ExtractLedgerTxParts` moves from inside `hotchunk.IngestLedger` up to `HotService.Ingest` (`internal/rpcv2/ingest/service.go`) — the one place every live ledger passes through.
  - `hotchunk.IngestLedger` takes the parts (it still needs the raw `lcm` for the ledgers-store write and the close time).
  - `HotService.Ingest` then computes `FeesFromTxParts(parts)` and appends both buckets via `AppendLedgerFees`.
- Ordering: append fees **only after** the store write succeeds — a rejected ledger must not contribute to the windows (mirrors v1's rollback-on-error).
- A fee classification error (negative fee, resource fee > FeeCharged) fails the ledger loudly, same as any other extract failure — tip-only trusted input.
- **Windows live at daemon level, injected into `HotService`** — the hot loop rebuilds `HotService` at every chunk boundary (`hotloop.go`), and a boundary rebuild must not reset the windows.
- Window sizes come from `[service.fee_stats]` (#882, landed in #897): `classic_fee_window_ledgers` / `soroban_inclusion_fee_window_ledgers`, defaults 10/50 matching v1, validated 1..1000 at startup.
- Expose the fed `FeeWindows` (it satisfies `store.FeeStats`) from the daemon so the handler wiring (#772/#889) consumes it.
- Flip the dormancy wording once this lands: `config.go`'s `ServiceConfig`/`FeeStatsConfig` comments, the sample-config header, and the design doc all say the fee windows "will be fed by live ingestion when #881 lands" — update all three to present tense.

### Why this shape — rationale for the implementer

- **Why the walk hoists to `HotService.Ingest` instead of feeding windows inside `hotchunk`.** `hotchunk` is a storage type: it knows how to write one chunk's data (ledgers, txhash, events) atomically, and nothing else. Fee windows are not storage — they're in-memory serving state for getFeeStats. If `hotchunk` fed them, a storage type would hold a reference to serving state, and every one of its rebuilds would need that reference re-plumbed. The walk output (`parts`) is needed by BOTH the storage write and the fee product, so the walk belongs to the one caller that has both in scope: `HotService.Ingest`.

- **Why the windows must live at daemon level, not inside `HotService`.** The hot loop throws `HotService` away and builds a fresh one at every chunk boundary (`hotloop.go`, roughly every 10,000 ledgers). If the windows lived inside it, every boundary would silently wipe up to 1,000 ledgers of fee history and getFeeStats would serve empty distributions until the windows refilled. The daemon owns the windows; each `HotService` borrows them.

- **Why fees append only AFTER the store write succeeds.** A ledger the store rejects gets retried by the hot loop. If its fees were appended before the failed write, the retry appends them again — the same ledger counted twice in the distribution, with nothing to detect it.

- **Why `PayloadsFromLedgerEvents` takes two slices.** The slimmed `TxEvents` deliberately carries no hash (see #912's rationale) — the hash lives on `parts` at the same index. The shaping needs both, index-aligned.

## Consequences for siblings

- #888 (restart warm-up) is rewritten against this shape: raw bytes → parts → `FeesFromTxParts` over the query router's read view — no LCM decode, no passphrase, and it drops its #885 dependency.
- #883 (parity harness): the getFeeStats case is now a **real** divergence test over the narrowed contract — v1 and v2 identical on organic current-protocol traffic. The documented deltas are never-ran transactions (`txINTERNAL_ERROR`, or a tx invalidated by an earlier tx in the same ledger — see stellar/go-stellar-sdk#5966's `FeesFromTxParts` godoc for the worked example), so a divergence means "go look at that ledger", not automatically "core had an incident".
- #885–#887 (read-path adapters, PR #908): unaffected — the reshape touches only ingestion call sites; the adapters keep `LedgerTransactionViewByHash` and its passphrase.
- The old plan's "optional follow-up: migrate v1's feeding to the same extractor" is **off the table** — the classifications now differ by design (see #912).

## Depends on

- #912 merged in `stellar/go-stellar-sdk` + a go.mod bump here.
- #876 (done), #882/#897 (done).

## Sizing

- Work item 1 is mechanical (2 call sites, 1 shaping signature, test re-targets).
- Work items 2–3: ~150–250 lines net (mostly moves), ~15 tests touched or added.

### LINKED ISSUE #888: getFeeStats: rebuild fee windows on restart
## What

- v2's fee windows are in-memory only, so a restart replays the last `max(classic window, soroban window)` ledgers (≤1000) to refill them before live ingestion resumes.
- Reframed after #912 / #881: the replay is the **fees-only consumer of the shared walk** — `ExtractLedgerTxParts` → `FeesFromTxParts`, nothing else computed. Exactly the third composition the SDK reshape was built for.

## How — raw bytes through the view path (changed from the earlier plan)

- The earlier plan was "`StreamLedgerRange` with v1's `IngestFees` as the `store.StreamLedgerFn` callback". It is dead. The ELI5 of why — do not resurrect it by routing the replay through `store.LedgerReader`, even though that looks like the blessed seam:
  - **The store interfaces exist for one audience: the shared JSON-RPC handlers (#772).** Their currency is parsed `xdr.LedgerCloseMeta` because handlers work on parsed data (`store.StreamLedgerFn` is `func(xdr.LedgerCloseMeta) error`). The replay is not a handler — it's ingestion refilling its own in-memory cache. Wrong seam, wrong audience.
  - **Going through that seam forces a full LCM decode per replayed ledger** — the only full decode anywhere on the v2 path, paid up to 1,000 times on every restart, just to walk data the view extractor reads for free off raw bytes.
  - **v1's `feewindow.IngestFees` lives under `internal/rpcv1/`, and v2 code may not import `rpcv1/`** (the depguard rules from #876). Reusing v1's ingestion here was never actually possible after the reorg.
  - **Bonus from the bypass: this ticket sheds its #885 dependency.** Reading the router's `ReadView` directly means the replay can be built as soon as #881 lands, without waiting for the `LedgerReader` adapter.
- New mechanism — the same code path as live ingestion's fee product:

```python
# pseudocode; real types are query.ReadView / xdr.LedgerCloseMetaView
view = registry.NewReadView()          # frozen range + handle set; defer view.Release()
latest = lastCommittedLedger           # the hot loop resumes at latest + 1
start  = max(latest - window + 1, view.OldestLedger())
if start > latest:
    return                             # fresh start / short history: no-op, not an error

for entry in view.ScanLedgers(start, latest):
    lcm   = LedgerCloseMetaView(entry.Bytes)
    parts = ingest.ExtractLedgerTxParts(lcm)   # the one walk
    fees  = ingest.FeesFromTxParts(parts)      # fees only; events never computed
    # seq + close time read off the view header; append to BOTH windows,
    # each trimming to its own retention
    windows.AppendLedgerFees(seq, closeTime, fees)
```

- `entry.Bytes` is **borrowed** (it aliases the reader's scratch buffer, overwritten on the next step) — fine here: fees are folded within the loop body and only plain `uint64` slices are retained.
- The replay range spanning a chunk boundary is the common case, not an edge case (any restart within the first `window` ledgers of a chunk). `ScanLedgers` does the split and per-chunk clip; ≤1000 ledgers means at most 2 chunks, so routing failures surface before any ledger is emitted.

## Ordering — the double-count hazard

- The replay covers `[start, lastCommitted]`; the hot loop starts feeding the same windows at `lastCommitted + 1`.
- The replay must **complete before** the hot loop starts, or the boundary ledger is counted twice.
- The windows live at daemon level and are injected into both the replay and `HotService` (see #881 work item 3).

## Scope notes (carried over, still true)

- Clamp to what exists: on a fresh start `OldestLedger()` can exceed `latest` by one, so a short history must be a no-op, not an error.
- No config-time window-vs-retention guard needed (v1 has one): windows are capped at 1,000 ledgers (`limits.MaxFeeStatsRetentionWindow`) and v2's smallest retention is one chunk = 10,000 ledgers, so the guard can never fire.

## Dependency change

- **No longer depends on #885.** The replay reads the query router's `ReadView` directly over raw bytes — it never touches `store.LedgerReader`. (The store seam exists for handlers; the replay is ingestion.)
- Depends on: #881 (shared windows + `FeesFromTxParts` wired), #865/#894 (query router — done).

## Sizing

- ~80–120 lines in 1 file, ~6 tests (unchanged estimate).

karthikiyer56:
## The story, in plain words

- `getFeeStats` answers "what did transactions pay recently?" from two small in-memory windows: one for classic per-op fees, one for soroban inclusion fees.
- Before this PR, v2 had config knobs for those windows but nothing ever filled them. This PR fills them (live, as each ledger is ingested) and refills them after a restart (they're memory-only, so a restart starts empty).
- Along the way, v2 ingestion switches to the new SDK API where each ledger is scanned **once**, and everything we need per ledger — tx hashes, events, fees — is just a cheap read over that single scan.

- 2026-08-04 update: merged the refreshed `feature/full-history` (which now carries main's Protocol 28 stack via #920) and resolved `go.mod` to `go-stellar-sdk v0.7.1` — this is what fixes the `dependency-sanity-checker` failure. Also addressed all four review comments (see the commit rows and replies on the threads).

## What changed, commit by commit

| Commit | ELI5 |
|---|---|
| 1. SDK bump + migration | • the SDK replaced its old per-ledger extractors with one `ExtractLedgerTxParts` scan whose outputs everything else reads<br>• v2's two scan sites (live ingestion, backfill) switch over — same behavior, same bytes written, tests re-checked against independent references |
| 2. v2 gets its own fee windows | • new `internal/rpcv2/feewindow`: append a ledger's fees, get a distribution back, reset, done<br>• deliberately NOT shared with v1 (see "why not shared" below)<br>• the percentile math lives once in `internal/store` as `ComputeFeeDistribution`, next to the `FeeDistribution` type it returns — both services call it (review feedback; it was initially copied) |
| 3. Feed the windows live | • every ledger the hot loop commits also drops its fees into the windows<br>• fees go in **only after** the ledger is durably written — a rejected ledger gets retried, and counting its fees early would count them twice<br>• the windows live at daemon level: the hot loop throws away and rebuilds its writer at every chunk boundary, and that rebuild must not wipe fee history |
| 4. Refill on restart | • on boot, re-read the last `max(classic, soroban)` ledgers (≤1000) as raw bytes, extract only fees, refill the windows — then start the live loop<br>• raw bytes on purpose: going through the parsed-ledger store interface would decode the full ledger up to 1000 times per restart just to read fee fields<br>• fresh start or short history → nothing to replay, not an error |
| 5. Move `ledgerbucketwindow` into rpcv1 | • v2's window is a ~25-line ring buffer of its own, so the generic container has exactly one consumer left: v1's feewindow<br>• pure move; the only edit is v1's import path |
| 6. Fold the events read into the shaping (review feedback) | • `PayloadsFromLedgerEvents` now derives the events product from `txParts` itself instead of taking two slices callers had to keep index-aligned<br>• mispairing the slices (events stamped with the wrong tx hash) becomes impossible by construction, and the runtime length check plus a chain of parameter plumbing deletes<br>• metrics note: on the backfill path the events read now counts in the events writer's own timing instead of the shared per-ledger extract timing — more honest attribution, since it only runs when the events writer is on |

## Why v1 and v2 don't share fee code

- They classify fees from different inputs: v1 reads parsed ledgers + envelopes (needs the network passphrase, a SQLite migration adapter), v2 reads the SDK's fee product off the one scan (needs none of that).
- Sharing one package meant v2 dragging around v1's baggage and v1's migration adapter getting exiled from its own type. Not worth it.
- What IS still shared, and why:
  - `store.FeeStats` / `FeeDistribution` — the interface the **shared getFeeStats handler** consumes; both trees implement it, the handler serves both daemons.
  - `store.ComputeFeeDistribution` — the pure percentile math both services' numbers come out of; it has none of v1's baggage, and it's the one place where silent drift would actually matter, so it lives beside the type it returns with its differential test. Cross-service agreement stays #883's parity harness's job.

## How we know it works

| Guarantee | Test |
|---|---|
| The one-scan switch changed no output byte | `TestWriteColdChunk_ByteIdentity_SharedWalk` |
| A rejected ledger's fees never land (no double-count on retry) | `TestHotService_RejectedLedgerFeedsNoFees` |
| A corrupt fee shape fails loudly instead of skewing stats | `TestHotService_FeeClassificationErrorFailsCommittedLedger` |
| Chunk-boundary writer rebuild keeps fee history | `TestRunIngestionLoop_FeeWindowsSurviveChunkBoundary` |
| Restart refill == computing from scratch, boundary ledger counted once | `TestFeeWindowReplay_SpansChunkBoundary` |
| Fresh start / short history handled quietly | `TestFeeWindowReplay_FreshStartIsNoOp`, `TestFeeWindowReplay_ShortHistoryClampsToOldest` |
| The shared percentile math matches an independent implementation | `TestComputeFeeDistributionAgainstAlternative` (in `internal/store`) |
| Backfill still never computes fees (nobody asks it to) | structural: no consumer → no `FeesFromTxParts` call in the cold path |

- `go test ./...`: green (38 packages, 0 failures). Lint on both trees: only findings already present on the base branch. depguard: passes.
- v1 behavior unchanged; its diff is two mechanical edits: the `ledgerbucketwindow` import path, and `feewindow` calling `store.ComputeFeeDistribution` (its distribution tests moved to `internal/store` with the function).

Closes #881, closes #888.


cmd/stellar-rpc/internal/rpcv2/ingest/service.go
```diff
@@ -2,37 +2,71 @@ package ingest
...
+// items — nothing landed durably); on success every phase is emitted. The walk
+// runs here (see the type doc) and its duration folds into PhaseExtract, keeping
+// that phase "the walk + product reads".
+//
+// Fees append ONLY after the store write succeeds: a rejected ledger gets
+// retried by the hot loop, and fees appended before a failed write would be
+// counted again on the retry. A fee classification error (negative fee,
+// resource fee above FeeCharged) fails the ledger loudly like any extract
+// failure — tip-only trusted input — even though the store write is already
+// durable; the restart replay refills the windows.
```
tamirms: This sentence describes a recovery that cannot happen for this failure class. `FeesFromTxParts` is deterministic over the committed bytes, and the failed ledger is the new last-committed frontier (`TestHotService_FeeClassificationErrorFailsCommittedLedger` asserts exactly that), so on the next startup `replayFeeWindows` re-runs the same classification over the same ledger and fails the same way. The restart never reaches the loop; `supervise` retries `run()` against the same startup failure.

I don't think the behavior itself needs to change: the trigger is corrupt tip input, and a corrupt ledger already wedges the daemon the same way at extract, pre-commit, so failing hard here is consistent. But I think this comment should say what actually happens (a classification failure is permanent until a code fix, like an extract failure, with the one difference that the ledger is already durable). Alternatively, if you would rather a restart genuinely heal fee stats, the replay could skip and log classification errors, which would make this sentence true; in that case a test for the commit-succeeds, classification-fails, restart path would pin it.

cmd/stellar-rpc/internal/rpcv2/feewindow/feewindow.go
```diff
@@ -0,0 +1,214 @@
...
+	w.dist = store.FeeDistribution{}
+}
+
+func (w *feeWindow) distribution() store.FeeDistribution {
+	w.lock.RLock()
+	defer w.lock.RUnlock()
+	return w.dist
+}
+
+// computeFeeDistribution is copied verbatim from internal/rpcv1/feewindow (see
```
tamirms: Would you consider moving `computeFeeDistribution` into `internal/store`, next to `FeeDistribution`, instead of copying it? I agree with the package doc that sharing the windows would drag v1's baggage into v2: the passphrase, the migration adapter, and the sqlite rollback joins all live in v1's classification layer, and v2 rightly wants none of that. But this particular function is pure, `([]uint64, uint32) -> store.FeeDistribution` over stdlib only, so the baggage argument does not apply to it. It can live beside the type it returns, the differential test and its montanaflynn dependency move with it once instead of being duplicated, and depguard is fine with it since it imports neither tree. The v1 side becomes a mechanical import change in a file this PR already touches for the ledgerbucketwindow move.

The ring staying separate seems right to me, and #883 would still cover end-to-end agreement. This is just about the one piece where silent drift would actually matter, since both services' getFeeStats numbers come out of this function.

cmd/stellar-rpc/internal/rpcv2/events/extract.go
```diff
@@ -34,23 +35,30 @@ import (
...
+// walking TxProcessing twice. The slimmed TxEvents carries no hash, so the
+// shaping reads each payload's TxHash off txParts at the same index — the two
+// slices MUST be index-aligned (txEvents = EventsFromTxParts(txParts)).
+// ledgerSeq and ledgerClosedAt are the view's header values (cheap reads, not
+// a walk).
 func PayloadsFromLedgerEvents(
-	txEvents []ingest.LedgerTransactionEvents, ledgerSeq uint32, ledgerClosedAt int64,
+	txParts []ingest.LedgerTxParts, txEvents []ingest.TxEvents, ledgerSeq uint32, ledgerClosedAt int64,
 ) ([]Payload, error) {
+	if len(txParts) != len(txEvents) {
```
tamirms: Everything that consumes `txEvents` is this function (I checked both tiers: hotchunk shapes immediately, and the cold events writer shapes inside `ingestSeq`), so would it be worth deriving it internally and dropping the paired-slice contract?

```go
func PayloadsFromLedgerEvents(
	txParts []ingest.LedgerTxParts, ledgerSeq uint32, ledgerClosedAt int64,
) ([]Payload, error) {
	txEvents, err := ingest.EventsFromTxParts(txParts)
	...
```

That makes the index alignment structural instead of runtime-checked, and a fair amount of plumbing deletes with it: this length check and its error path, the `txEvents` parameter threading through `coldChunk.ingest`, `eventsCold.write`, and `ingestSeq`, the conditional `EventsFromTxParts` block in `coldChunk.ingest`, hotchunk's separate `EventsFromTxParts` call, and `extractFor`'s third return value in the tests.

The one real consequence I see is metering: on the cold path the events-product read currently lands in the type-less ColdExtract signal, and with this change it would move into the events writer's ColdIngest total. I could argue that attribution is more correct, since it is events-specific work, but it does shift a documented boundary, so it is your call whether the trade is worth the deletion. The hot path's PhaseExtract attribution is unchanged either way, since the shaping already runs inside it.

cmd/stellar-rpc/internal/rpcv2/ingest/ingest_test.go
```diff
@@ -1139,6 +1144,128 @@ func hotTestLogger() *supportlog.Entry {
 	return l
 }
 
+// feeLCMBytes builds a one-tx V2 LCM whose result carries feeCharged and ONE
+// op result, so FeesFromTxParts observes exactly one classic per-op fee of
+// feeCharged. The shared successResult() fixture carries an EMPTY op-result
+// list, which the classification skips as a never-ran transaction — useless
+// for fee tests.
+func feeLCMBytes(t *testing.T, seq uint32, feeCharged int64) []byte {
```
tamirms: This is a line-for-line duplicate of `rpcv2test.FeeTxLCMBytes`: the construction is identical, and the passphrases are the same string, since `testPassphrase` in this file is the public-network passphrase that `FeeTxLCMBytes` uses via `network.PublicNetworkPassphrase`. The file already imports `rpcv2test`, so I think the local copy can just go.

cmd/stellar-rpc/internal/rpcv2/ingest/service.go
karthikiyer56: Done in fa42f32c.
Kept the fail-loud behavior (consistent with extract failures, and skipping in the replay would break the replay-equals-from-scratch invariant), and rewrote the comment to say the failure is permanent until a code fix, with the one difference that the ledger is already durable.

cmd/stellar-rpc/internal/rpcv2/feewindow/feewindow.go
karthikiyer56: done in 8e926029

cmd/stellar-rpc/internal/rpcv2/ingest/ingest_test.go
karthikiyer56: Done in fa42f32c.

cmd/stellar-rpc/internal/rpcv2/events/extract.go
karthikiyer56: Done in 2e38a04c — `PayloadsFromLedgerEvents` derives the events product itself now; the length check and the `txEvents` plumbing in both tiers deleted. Took the metering trade: the cold events read lands in the events writer's ColdIngest.

tamirms (approved):

PR #932, opened 2026-08-17
Key cold-index routing (events + txhash) against adversarial block-overflow

tamirms:
## Summary

Both cold stores build streamhash MPHFs, and streamhash routes each key to a block via a public, **unkeyed** function of the key bytes. An attacker who influences indexed keys — event topics / contract IDs they emit, or transaction hashes — can grind keys into one block, exceed its per-block cap, and abort the chunk's cold-index build (`ErrBlockOverflow`). Routing is deterministic, so retries fail identically: a permanent backfill stall.

Fix: **blind** every key with `SipHash-2-4-128` under a per-index secret before it reaches streamhash, at build and query. Without the secret an attacker can't predict which block a key lands in.

## Secret handling

- The deployment's cold-index secret (`catalogSecret`) is **minted once and persisted in the catalog** (`Catalog.Secret`, get-or-create at open) — no operator config, no TOML key.
- Consumers **source it at point of use** from the catalog they already hold — it is not threaded through configs.
- Each index's secret is derived: `indexSecret = HKDF(catalogSecret, domain, indexID)`, domain-separated per store (`"txhash"` / `"events"`).
- The transform is `BlindKey(indexSecret, key)` (shared helper in the `stores` package, KAT-tested).

## Per store

- **txhash** — keys are blinded at **ingest** (when the `.bin` is written), so the existing streaming `SortedBuilder` build is untouched and same-node rebuilds stay **byte-identical**. The `.bin` header records the secret its keys were blinded with, and the index build **adopts** that secret (erroring if inputs disagree), so an index can never be built under a secret that mismatches its keys. The derived secret is stored in the index metadata.
- **events** — term keys are blinded at build with the chunk's deterministic per-index secret (`domain = "events"`), stored in the index metadata. Deterministic, so rebuilds are byte-identical.

## Notes

- The secret is **per-node** (each deployment mints its own), so index files are not byte-identical across nodes. This is intentional — it preserves the streaming `SortedBuilder` build; cross-node reproducibility is not needed for serving (see the thread discussion) and is a possible future opt-in via a normalization pass.
- **Overflow is non-retryable** at the residual (~1e-12) rate: the secret is deterministic, so a rebuild can't clear a natural overflow; recovery is operator-level.
- Keying adds **no per-key heap allocation** on the ingest hot path (verified via escape analysis); secrets are derived once per writer/build.
- Fingerprints, per store: **events** keeps its 4-byte `index.pack` fingerprint (and post-filter) on the **original** term bytes. **txhash** has only streamhash's 1-byte fingerprint, which — since txhash feeds streamhash the blinded key — covers the **blinded** bytes (equivalent detection power, blinding being deterministic per key); the downstream cold-hit check resolves against the **original** tx hash in the ledger, so a wrong index can only miss, never return a wrong answer.
- No backward compatibility (unreleased): a single fixed metadata format, no versioning.

## Test

Unit tests pass for `catalog`, `stores`, `stores/txhash`, `stores/event`, `ingest`, `backfill`, `query`, `lifecycle`, and the daemon suite. Build / vet / gofmt / golangci-lint all clean.

### DIFF
--- a/cmd/stellar-rpc/internal/rpcv2/catalog/secret.go
+++ b/cmd/stellar-rpc/internal/rpcv2/catalog/secret.go
@@ -1,0 +1,42 @@
+package catalog
+
+import (
+	"crypto/rand"
+	"fmt"
+)
+
+// catalogSecretStoreKey holds the deployment's cold-index secret.
+const catalogSecretStoreKey = "meta/catalog-secret"
+
+// Secret returns a copy of the deployment's cold-index secret, minted once at
+// Open and cached. Per-index secrets are derived from it, so an attacker who
+// influences indexed keys cannot predict which block a key lands in. Returning
+// a fixed-size array (not the internal slice) states the length and prevents a
+// caller aliasing or mutating the cached value. Stable for the life of the
+// catalog.
+func (c *Catalog) Secret() [32]byte { return c.secret }
+
+// ensureSecret loads the persisted cold-index secret, minting and persisting a
+// fresh random one on first call. Open runs it single-threaded and caches the
+// result; nothing else should call it (get-or-create is not atomic).
+func (c *Catalog) ensureSecret() ([32]byte, error) {
+	var s [32]byte
+	v, found, err := c.get(catalogSecretStoreKey)
+	if err != nil {
+		return s, err
+	}
+	if found {
+		if len(v) != len(s) {
+			return s, fmt.Errorf("persisted cold-index secret is %d bytes, want %d", len(v), len(s))
+		}
+		copy(s[:], v)
+		return s, nil
+	}
+	if _, err := rand.Read(s[:]); err != nil {
+		return s, err
+	}
+	if err := c.put(catalogSecretStoreKey, string(s[:])); err != nil {
+		return s, err
+	}
+	return s, nil
+}
--- a/cmd/stellar-rpc/internal/rpcv2/backfill/execute.go
+++ b/cmd/stellar-rpc/internal/rpcv2/backfill/execute.go
@@ -82,7 +82,8 @@ func (cfg ExecConfig) validate() error {
 	return nil
 }
 
-// processConfig projects the shared Catalog/Logger into the ProcessConfig.
+// processConfig projects the shared Catalog/Logger into the ProcessConfig. The
+// cold-index secret is sourced from the Catalog at point of use.
 func (cfg ExecConfig) processConfig() ProcessConfig {
 	p := cfg.Process
 	p.Catalog = cfg.Catalog
--- a/cmd/stellar-rpc/internal/rpcv2/backfill/perf_test.go
+++ b/cmd/stellar-rpc/internal/rpcv2/backfill/perf_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/stellar/streamhash"
 
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/chunk"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/stores"
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/stores/txhash"
 )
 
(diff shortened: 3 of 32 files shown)


tamirms:
## Porting guidance: what this PR needs to merge cleanly into `tamirms/full-history-p99-campaign`

The mechanism (SipHash-2-4-128 blinding under a catalog-derived secret, at build and query) is right and the campaign branch wants it — it closes the one adversarial route to the reorder-heap's corrupt-index transient, so the count backstop's remaining trigger becomes hardware-grade corruption only. But this changeset targets `feature/full-history`'s builders, and the campaign branch (`24c6e613`, 12 commits) rewrote most of the files it touches: txhash hot ingest is now a packed-row engine (sorted 32B-hash rows + sealed runs; the cold `.bin` is produced by the freeze's k-way merge, not by ingest), and the events cold build is a streaming spill-and-merge feeding `streamhash.NewSortedBuilder` through a bounded slot-reorder heap. Applying this PR there is a port with two design changes, not a rebase:

### 1. Events: the per-build random secret cannot survive
The streaming builder's contract is **sorted input** — and with blinding, "sorted" must mean *sorted by blinded key*, because streamhash's non-decreasing-block routing (which bounds the reorder heap) operates on the bytes you feed it. The spill runs are sorted at ingest time, so a secret chosen at build time is too late. Port shape:
- Blind at **spill-write** (`runspill` entries carry the blinded TermKey), under the **deterministic** catalog-derived secret — the same `DeriveIndexSecret` scheme txhash already uses in this PR, domain-separated per chunk.
- The hot events tier consumes the same keys, so hot run/query probes blind first (~25ns each, measured on the bench box — noise).
- Keep exactly what this PR already does for the 4-byte fingerprint and the post-filter: **original TermKey bytes**, only routing is blinded.
- Consequence to document: events loses the rebuild-under-fresh-secret retry and becomes non-retryable on residual overflow, same as txhash. At ~1e-12 that is the right trade; update the Notes section.

### 2. txhash: blind at seal, not at ingest-into-rows
The hot window/rows should stay raw 32B — hot routing is bloom + binary search, no block-capped structure, so it needs no blinding, and the full-hash exact compare is worth keeping where it's free. The clean insertion point is **sealing**: sealed-run records become (blinded 16B key ‖ 4B seq) = 20B, down from today's 36B — seal write and freeze-merge read I/O drop ~44% as a side effect. The freeze merge then emits blinded-sorted runs, the cold `.bin` keeps its existing 20B layout (16B key + seq) with blinded keys, and the full-window `SortedBuilder` build is untouched — which preserves this PR's rebuilds-stay-byte-identical property, since the secret is deterministic. Hot run-file probes compare blinded 128-bit values; note the exactness-contract wording change in `hotindex.go`'s docs (keyed-collision-exact rather than raw-compare-exact).

### 3. Mechanical items
- **streamhash pin collision**: this PR needs `ErrBlockOverflow` (stellar/streamhash#14); the campaign pins `Index.MaxBlockKeys()` (stellar/streamhash#13). Merge both upstream and pin one version before porting — neither branch commit contains the other.
- **Byte-identity gates are the port's safety net, keep them green**: freeze-vs-walk identity must hold with both paths blinding through the same catalog secret; the walk builders source it from the catalog they already hold. All cold-index golden fixtures change bytes — regenerate, don't relax.
- The events `index.hash` user-metadata (version ‖ secret) must be produced identically by the in-memory (test-oracle) and streaming builders, or the identity gate fails on metadata before it ever compares payload.

### Expected perf on the campaign benchmarks (SipHash measured at 24.8ns/16B, 32ns/32B on the c6id.8xlarge bench box)
- **Hot ingest**: events spill blinding ~0.1–0.75ms/ledger at stress density → ≤1% on p50, noise at p99; txhash seal blinding is background — ~0.
- **Freeze**: events arm ~0% (keys arrive pre-blinded; merge and both index passes unchanged). txhash arm slightly **faster** (−44% sealed-run bytes through the merge).
- **Cold ingest (walk)**: ~2.3s of blinding across 93M keys on an ~8.5min build — well under 1%.

Happy to run the A/B cells on the bench harness once the port lands — the freeze/hot/cold rigs are warm and the acceptance bands above are the ones the campaign tracks.

https://claude.ai/code/session_013zyTXU8wkocJgN6mBafnou

cmd/stellar-rpc/internal/rpcv2/backfill/txindex.go
```diff
@@ -95,7 +95,7 @@ func buildTxhashIndex(ctx context.Context, w geometry.TxHashIndexID, lo, hi chun
 		return fmt.Errorf("buildTxhashIndex mkdir %s: %w", indexDir, mkErr)
 	}
 	if berr := txhash.BuildColdIndex(
-		ctx, inputs, idxPath, lo.FirstLedger(), hi.LastLedger(),
+		ctx, inputs, idxPath, lo.FirstLedger(), hi.LastLedger(), cat.Secret(), uint32(w),
```
karthikiyer56: The `.bin` writer calculates the index secret at ingest time. This line calculates the same secret again at build time. The two calculations are independent, and no code checks that they agree.

They do agree today.
My concern is that nothing enforces it, and the failure mode is quiet. 
If the two values ever differ, the build still succeeds: the blinded keys stay uniform and sorted, so streamhash makes a correct index over them and the `.idx` opens normally. 
Every `Get` in the window then returns not-found, which reads as a normal answer, so nothing is logged.

Can we make this more robust and double-check it?
One option, if it helps:
write a tag in the `.bin` header, calculated one-way from the index secret, for example the first 8 bytes of an HMAC. 
All the `.bin` files in one window share the same secret, so they all carry the same tag. 
`BuildColdIndex` can then confirm that every input carries the same tag, and that it matches the secret it is about to write into the index. 
It stops if they differ

cmd/stellar-rpc/internal/rpcv2/catalog/secret.go
```diff
@@ -0,0 +1,36 @@
...
+const catalogSecretStoreKey = "meta/catalog-secret"
+
+// catalogSecretLen is the byte length of the minted secret.
+const catalogSecretLen = 32
+
+// Secret returns the deployment's cold-index secret, minted once at Open and
+// cached. Per-index secrets are derived from it, so an attacker who influences
+// indexed keys cannot predict which block a key lands in. Stable for the life of
+// the catalog.
+func (c *Catalog) Secret() []byte { return c.secret }
```
karthikiyer56: nit:
This returns the slice itself, not a copy of it
The comment above says the value is stable, and I think the type should say it too.

can we return a [32]byte instead for this funciton?

karthikiyer56 (approved):

PR #933, opened 2026-08-18
Integrate changes made to stellar-rpc-blaster

### LINKED ISSUE #925: Update RPC release eval endpoint load test configuration with insights from Overcat's data
Prerequisites: #860, #924 

### What

We now have data and analysis of expected RPC traffic and the distributions/patterns followed by various endpoints along many dimensions: RPS, query patterns, etc. This issue provides for integrating the insights into both `stellar-rpc` and `stellar-rpc-blaster`. Though this should be driven by the conclusions of #924, preliminarily, this may look like:
- for `stellar-rpc-blaster`:
    - align the request bodies of parameter-backed endpoints crafted for use in endpoint load tests with what's seen in the data
    - create a benchmark endpoint load-test profile that follows the RPS distributions seen by the data
        - since this profile will likely not change, but must align with what's truly achievable under the minimum hardware requirements, there should probably be an optional `float scale` parameter one can specify in the profile. One way to accomplish this could be to normalize the profile such that the sum of all of the endpoints RPS values equals 100, and a benchmark load level `L` is established as some `scale=L`. Then, in the load test, each endpoint `i` is sent at a maximum of RPS of `RPS_i * L`. 
- for `stellar-rpc`:
    - integrate the new endpoint load test configuration changes into the endpoint load test
    - test robustly to establish reliable and consistent benchmarking expectations given the configurations

cjonas9:
### What

Integrates recent changes made to `stellar-rpc-blaster` based on Overcat's endpoint traffic data. These changes are aimed at making stellar-rpc-blaster more effective at representing real traffic (i.e. request bodies are now generated with probabilities based on real data rather than intuition about what production traffic is likely to look like). Among many things, this includes an overhaul to how getEvents traffic is generated; realistic RPS proportions among endpoints; and distributions around variable parameters (such as limit) fit to those seen in real traffic.

The stellar-rpc-blaster config file added in this PR is only intended to be used in a `--serial` endpoint test. This test evaluates the endpoints independent of one another; RPS values are based only on what an RPC on benchmark hardware can handle.

Please see [this PR](https://github.com/stellar/stellar-rpc-blaster/pull/42) for the complete blaster-side changes made.

Since this branch increased the RPS values each endpoint is tested at, a previously dormant test harness-side bug became active: under blast load, the daemon's request logging outran the rate at which the harness's pipe could drain at. This led to the capture mechanism itself throttling. The daemon in the harness was changed to now write to its own file, `var/log/stellar-rpc.log`.

### Why

Recently, a release evaluation system was added to Stellar RPC. Among four tests included in the suite is one that evaluates RPC's performance serving queries. The requests it serves, as well as the report that it receives, are generated in `stellar-rpc-blaster`. This PR and the work in `blaster` aims to make this test far better at representing/discovering regressions between releases that would happen in a production setting.

### Known Limitations

N/A


cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
```diff
@@ -117,27 +121,27 @@ func instantiate(ctx context.Context) error {
 	return nil
 }
 
-// fetchBlaster clones and builds stellar-rpc-blaster at dev HEAD
+// fetchBlaster clones and builds stellar-rpc-blaster at the pinned commit.
 func fetchBlaster(ctx context.Context, dir, repo string) (string, string, error) {
-	logger.Infof("fetching stellar-rpc-blaster (%s@dev)", repo)
+	logger.Infof("fetching stellar-rpc-blaster (%s@%s)", repo, blasterPin)
```
cjonas9, line 126: Flagging this, revert to `@dev` (or `@main` if we merge `main`<-`dev`) before merging (after commit merged to blaster of course)

cmd/stellar-rpc/internal/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
```diff
@@ -117,27 +121,27 @@ func instantiate(ctx context.Context) error {
...
-	if err := harness.RunStreaming(ctx, "", nil, 20, "git", "clone", "-q", "--depth", "1",
-		"--branch", "dev", "https://github.com/"+repo+".git", dir); err != nil {
-		return "", "", fmt.Errorf("git clone failed: %w", err)
-	}
-	out, err := exec.CommandContext(ctx, "git", "-C", dir, "rev-parse", "HEAD").Output()
-	if err != nil {
-		return "", "", fmt.Errorf("resolving blaster commit: %w", err)
+	for _, args := range [][]string{
+		{"init", "-q", dir},
+		{"-C", dir, "fetch", "-q", "--depth", "1", "https://github.com/" + repo + ".git", blasterPin},
```
cjonas9, line 132: as above, revert to simpler `git clone` after since we'll no longer need to get `blaster` by SHA

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
```diff
@@ -14,8 +14,8 @@ import (
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness"
 )
 
-// blasterCfg is the endpoint roster, shipped with the blaster checkout.
-const blasterCfg = "cmd/stellar-rpc-blaster/internal/config/config.example.toml"
+// blasterCfg is the realistic traffic-profile config, shipped with the blaster checkout.
+const blasterCfg = "cmd/stellar-rpc-blaster/internal/config/config.profile.toml"
```
Shaptic: This should live here rather than in blaster, I think. Obviously not a big deal, but it does make more sense as the test profile rather than the tool profile.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 )
 
+// expectedTrafficProfile is the traffic-model version blaster stamps on modeled
+// endpoints at dev HEAD; a mismatch breaks cross-run comparability.
+const expectedTrafficProfile = 3
```
Shaptic, line 14: Errr where is this from exactly?

Shaptic (approved):

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
cjonas9: This was one of my main outstanding questions. The blaster changes came first, so I'd temporarily put it in there, but then I began wondering if, because the profile could be argued to be general "representative profile" useful to other blaster ops, it was better to leave there. Could see it going either way, but I prefer it here in RPC too.
Happy to move it over, thanks for thinking about this!

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
cjonas9: This matches with what's stamped into the endpoint traffic profile results JSON since format isn't guaranteed to be consistent across these. [in stellar-rpc-blaster/.../run/parameters/endpoints.go:L133](https://github.com/stellar/stellar-rpc-blaster/blob/d446043db54e5489b10686cf7f7665c0f9c1df01/cmd/stellar-rpc-blaster/internal/run/parameters/endpoints.go#L133)

In this comment I'll note the file in blaster where the constant is defined for future reference

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
Shaptic: Yeah, I don't think either one is wrong, per se, and I do remember the open question and thinking "hmm," but after seeing it cross-referenced here it definitely felt more appropriate to keep the release eval profile near the release eval framework.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
Shaptic: Oh hm, I feel like this must've slipped through my original review of that part of the codebase. This is... odd. Could we use the SHA of the revision of the tool, or something, instead?

Shaptic: Not in this PR, obviously, but earmarked for the future

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
```diff
@@ -53,20 +72,57 @@ func summarize(data []byte) ([]endpointStats, error) {
...
 	fmt.Fprintf(&b, "### 🎯 Endpoint load test — `%s`\n\n", sha[:min(12, len(sha))])
-	fmt.Fprintf(&b, "Serial blast per endpoint (ramp-up %s, duration %s, blaster `%s`) against the backfilled RPC "+
-		"(ledgers `[%d, %d]`, handoff wait %ds).\n\n",
-		rampUp, duration, blasterSHA[:min(12, len(blasterSHA))], oldest, latest, handoffSecs)
+	fmt.Fprintf(&b, "Serial blast per endpoint (ramp-up %s, duration %s, error kill switch %s%%, blaster `%s`) "+
+		"against the backfilled RPC (ledgers `[%d, %d]`, handoff wait %ds).\n\n",
+		rampUp, duration, errorThreshold, blasterSHA[:min(12, len(blasterSHA))], oldest, latest, handoffSecs)
+	if aborted {
+		b.WriteString("> ⚠️ **Aborted early:** an endpoint's error rate crossed the kill switch, so blaster ended " +
+			"the run early.\n\n")
```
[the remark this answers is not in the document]
cjonas9: intended behavior -- rows are filled with 0s, including RPS, indicating that row didn't run

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
```diff
@@ -14,16 +14,18 @@ import (
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness"
 )
 
-// blasterCfg is the endpoint roster, shipped with the blaster checkout.
-const blasterCfg = "cmd/stellar-rpc-blaster/internal/config/config.example.toml"
+// legDir holds the leg's traffic-profile config; runners start w/ cwd = repo root.
+const legDir = "cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test"
```
Shaptic: could we just use a go `embed`?

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
```diff
@@ -53,20 +72,57 @@ func summarize(data []byte) ([]endpointStats, error) {
...
-		})
+		}, nil
+	}
+
+	rows := make([]endpointStats, 0, len(res.Endpoints))
+	var archRows []endpointStats
+	for name, ep := range res.Endpoints {
+		r, err := row(name, ep)
+		if err != nil {
+			return nil, nil, false, err
```
Shaptic: Wouldn't you always want to return `res.Aborted` here? Do we do anything with that value? Could it be an error sentinel, instead? Def some nuance here so I might be off base.

Shaptic (approved):

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
cjonas9: We check it in this file [here.](https://github.com/stellar/stellar-rpc/pull/933/changes#diff-d659e540fca1cb4a89867b4d34c1fd70c2d867b1d500dde9c8b65b6ae0e564c2R112)
I realized it only after the blaster-side change went in, but `res.Aborted` is a bool, so there isn't much to do with that value. It would've been far more clever to make it the name of the endpoint it aborted on, `""` if no abort, but this also isn't that tricky to deduce by just looking at the table.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/instantiate.go
cjonas9: totally, great call, will add in next batch of changes

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/endpoint-load-test/runner/results.go
Shaptic: No no, what I'm saying is that shouldn't we always return it from this function as opposed to hardcoding `false` in this `return nil, nil, false, err` case?

cjonas9: Ah, got it, yes I was misunderstanding what you meant here.

I don't think it changes much, and I returned false out of convention to return zero values on error paths, but I can see the value here in case a caller ever wants that. I think the only point of concern I have about this is re: returning it in L52 since I think that return value is technically undefined (i.e. returning `res.Aborted` in the unmarshal that populates `res.Aborted`), so I kept that one as `false`.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -43,7 +45,7 @@ type backfillEnv struct {
 	TargetSHA   string `env:"TARGET_SHA"`
 	RunID       string `env:"RUN_ID"       envDefault:"manual"`
 	// ~1 day by default for cheap test runs; the full week is 120960.
-	Retention int           `env:"HISTORY_RETENTION_WINDOW" envDefault:"17280"`
+	Retention int           `env:"HISTORY_RETENTION_WINDOW" envDefault:"120960"`
```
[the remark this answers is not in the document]
cjonas9: added to next/final batch of changes

.github/workflows/load-test-coordinator.yml
```diff
@@ -11,7 +11,8 @@ defaults:
 
 on:
   push:
-    branches: [release/**]
+    # use-blaster-update: temporary full-run trigger for the blaster-update PR
+    branches: [release/**, use-blaster-update]
```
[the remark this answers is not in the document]
cjonas9: added to next/final batch of changes, pending current CI run pass

PR #934, opened 2026-08-18
rpcv2: the getEvents v2 cross-chunk events pager

urvisavla:
## What

The first two pieces of the getEvents v2 read path (#774), on top of the cursor codec (#914) and rebased onto the #904 filter fields.

1. **Engine** (`stores/event`): `Matches` replaces the one-shot `Query` API. It streams one chunk's matching events in ordinal order, reversed when descending. Filters resolve to a single `LookupKeys` call, and the first fetch batch is sized to what the page still needs.
2. **Pager** (`query/event_page.go`): `QueryEvents(ctx, cursor, limit)` returns one page and the advanced cursor. The cursor holds the query scope and two bookmarks: `ScannedLedger`, the last fully covered ledger, and `Position`, present only while a ledger is partly served. The server keeps nothing between pages.

## Behavior

- A descending scope whose top is above the node's latest ledger waits: empty page, `ScanWaitingForLedgers`, cursor unchanged. Clamping it would skip those ledgers forever.
- A descending range below the retention floor returns an empty `ScanOldestReached` page, per the proposal's errors table. Ascending keeps the out-of-range error.
- A page scans at most 10,000 ledgers, so a filter that matches nothing cannot walk the whole retention in one call.
- Validation rejects every cursor shape the server never mints as `ErrCursorMalformed`. A non-positive limit is `ErrInvalidLimit`.
- `ScanOldestReached` is per node: a node with deeper retention resumes the same cursor.

Deferred to the handler PR (B2): the wire limit cap, `startLedger` normalization, and error-to-wire mapping. The `EventCursorQuery` to `EventScope` rename goes to the naming follow-up PR.

## Testing

Full `rpcv2` suite green. Every pager test round-trips its cursors through the codec. Coverage: the paging seam matrix in both directions, mid-ledger resumes, ingest between pages, a full-chunk seam fixture (skipped in -short), edge behavior per direction, and the validation table.


cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,412 @@
...
+		// Beyond latest: nothing to scan yet. Either an ascending
+		// resume moved past it, or a fresh scope is entirely above it.
+		return &EventPage{Next: cursor, Status: ScanWaitingForLedgers}, nil
+	}
+
+	parts, err := a.EventParts(cursor.Scope.Dir, lo, hi)
+	if err != nil {
+		return nil, err
+	}
+	walk, err := walkChunks(ctx, parts, cursor.Scope.Filters, reenter, desc, limit)
```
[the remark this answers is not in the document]
urvisavla: The eager open is fixed in a28633a4 (walkChunks resolves one chunk per step), and 2d403e1b adds the per-page scan window: each page scans at most 10,000 ledgers (at most two chunks), advances the watermark through the window, and returns hasMore. The value may become configuration.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
```diff
@@ -83,10 +72,16 @@ func (f *Filter) termKeys() []events.TermKey {
...
 // call for that request. Events ingested after the snapshot are
 // invisible to the in-flight request. Multi-page paginated requests
-// MUST share the same End across pages.
+// MUST share the same End across pages, either literally or in the
+// ledger-level form the cross-chunk pager uses: each page re-derives
+// its range from the same ledger bounds via IDRangeForLedgers, which
+// is identical over ledgers committed before the first page (the
+// store is append-only) and extends only over newly committed
+// ledgers. See design-docs/getevents-v2-implementation-plan.md,
+// "Reconciling the cursor with the store's End-pinning contract".
```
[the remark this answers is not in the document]
urvisavla: Removed the file reference in 2d403e1b; the contract text stays in the comment.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,412 @@
...
+		return fmt.Errorf("%w: %w", ErrCursorMalformed, err)
+	}
+	// The server mints scope and bookmarks together, always consistent.
+	// A bookmark outside the scope's own bounds can only come from a
+	// forged or corrupted cursor. Refuse it rather than guess.
+	if pos := cursor.Position; pos != nil {
+		if pos.Ledger < cursor.Scope.MinLedger ||
+			(cursor.Scope.MaxLedger != nil && pos.Ledger > *cursor.Scope.MaxLedger) {
+			return fmt.Errorf("%w: position ledger %d outside scope bounds",
+				ErrCursorMalformed, pos.Ledger)
```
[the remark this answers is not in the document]
urvisavla: Good catch — validation checked each bookmark against the scope but not against each other. e138e988 rejects pairs no walk mints: ascending, the watermark trails the position by at most one ledger (a zero watermark only pairs with a position in the scope's first ledger); descending mirrors it. The descending zero-watermark case stays unchecked since it pairs with the clamped top, which the cursor does not record.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+		return nil, err
+	}
+	desc := cursor.Scope.Dir == Descending
+
+	lo, hi, reenter := resumeBounds(&cursor)
+	if lo > hi {
+		// Resume moved past the scope's far bound: nothing is left to serve.
+		return &EventPage{Next: cursor, Status: ScanComplete}, nil
+	}
+	if desc && hi > a.LatestLedger() &&
```
tamirms: The wait guard fires only for resumed cursors, so a fresh descending scope whose MaxLedger is above this view's latest takes the other path: page 1 clamps to latest, the walk finishes, and terminalStatus reports ScanComplete because it only classifies the bottom edge. The ledgers between the clamp point and MaxLedger are then unreachable on this cursor forever: nothing signals they were skipped, and a client that treats COMPLETE as range-fully-served has a permanent gap. Reproduced by execution: scope [f, f+10] descending on a node whose latest is f+3 serves the six below-tip events and returns ScanComplete, while the entirely-above-latest variant of the same scope waits and eventually serves everything, so the two shapes are also inconsistent with each other.

The proposal pins the intended behavior: a descending maxLedger above the tip is legal and waits at WAITING_FOR_LEDGERS until the chain reaches it (stellar/stellar-protocol#1872, the Bounds bullets). Descending can never revisit a ledger, so waiting is the only shape that keeps the top of the range servable.

I think the fix is the same rule for fresh and resumed scopes:

```go
if desc && hi > a.LatestLedger() {
    // The scope's top edge is above this view's latest ledger. A
    // descending walk never goes back up, so clamping would skip
    // [latest+1, hi] forever. Wait until this view has those ledgers.
    return &EventPage{Next: cursor, Status: ScanWaitingForLedgers}, nil
}
```

This also answers the status question the PR description defers to B2: the proposal already names the status, and the engine cannot defer the fresh case anyway, because clamping on page 1 destroys the information B2 would need. The bookmark disjunction disappears, the three above-latest shapes (fresh, resumed, entirely above) become one rule, and terminalStatus's bottom-edge-only check becomes fully sound, since a descending walk that runs can no longer have had a silently clamped top. What do you think?

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+	// Clip to the offsets snapshot's coverage. In production this is a
+	// no-op: latest never exceeds a serving chunk's ingested range. It
+	// keeps test fixtures with partially ingested chunks from turning
+	// into IDRangeForLedgers errors.
+	end := ofs.EndLedger()
+	if end == ofs.StartLedger() {
+		return chunkResult{}, nil // nothing ingested yet
+	}
+	pLo := max(part.From, ofs.StartLedger())
+	pHi := min(part.To, end-1)
```
tamirms: The clip quietly narrows the scan when the offsets snapshot ends before the part's To, and the empty-chunk return does the same when nothing is ingested; walkChunks then records coveredThrough as the chunk's far ledger, so the page can claim coverage over ledgers nothing scanned, up to a ScanComplete with a permanent client-visible gap. The comment is right that production ordering makes this a no-op today, but that invariant is enforced by publish ordering two packages away, and the engine this path replaced failed loudly on the same state (validateMatchCall rejects End beyond EventCount as a caller bug). The clip also swallows the re-entry position when the first chunk's window clips empty, which skips resumeOrdinal's ErrPositionMismatch check.

What do you think about failing loudly instead, and giving the fixtures honest offsets where they currently lean on the clip:

```go
end := ofs.EndLedger()
if end == ofs.StartLedger() || part.To > end-1 {
    return chunkResult{}, fmt.Errorf(
        "query: chunk offsets cover through %d but the walk needs through %d", end-1, part.To)
}
```

borderFixtureEvents already stages the partially-ingested state, so the boundary stays covered, just asserting the loud path.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+		return nil, err
+	}
+	walk, err := a.walkChunks(ctx, chunks, lo, hi, cursor.Scope.Filters, reenter, desc, limit)
+	if err != nil {
+		return nil, err
+	}
+	return assemblePage(cursor, walk, lo, hi, a.OldestLedger(), a.LatestLedger(), desc, truncated), nil
+}
+
+func validateCursor(cursor *EventCursor, limit int) error {
```
tamirms, line 130: A scope with more filters than the codec's cap passes validateScope (event.ValidateFilters checks field shapes only), the page is computed at full cost, and then Encode on the advanced cursor fails its 256-filter cap, so the request surfaces as an internal error after the work was done instead of ErrCursorMalformed before it. Decoded cursors are covered by readFilters, but the handler will also mint fresh cursors straight from request params, and the wire-side count cap is deferred to B2, so this stays reachable through the front door until then:

```go
if len(scope.Filters) > maxCursorFilters {
    return fmt.Errorf("%w: %d filters exceeds the %d-filter cap",
        ErrCursorMalformed, len(scope.Filters), maxCursorFilters)
}
```

And the non-positive limit rejection is the one client-input rejection without a sentinel, so the handler has no errors.Is target to map it to invalid-params and it would land in the store-failure bucket; an ErrInvalidLimit alongside the other sentinels keeps the property the validation table pins, an exact sentinel per rejection.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
```diff
@@ -381,64 +403,96 @@ func Query(ctx context.Context, r Reader, filters []Filter, opts QueryOptions) (
...
-	}
+// streamUnion walks the union bitmap in internal batches: collect up
+// to matchBatchSize candidate ordinals, fetch, post-filter, yield the
+// survivors. Drops advance the walk with no yield.
+//
+// FetchEvents requires ascending ids, so a descending batch is
+// collected highest-first and flipped before the fetch, then the
+// fetched matches are flipped back. Stepping one id at a time is fine
+// here: the fetch I/O dominates a 512-step loop.
+func streamUnion(
```
tamirms, line 458: streamUnion always fetches matchBatchSize candidates with no knowledge of the consumer's remaining room, and the deleted path applied the cap before fetching, so filtered pages regressed: a default-limit page over a chunk whose union holds 512 or more candidates fetches 512 scattered ids to deliver 100, roughly 400 extra record reads and decompressions per page on cold and about 5x the point reads on hot, plus the wasted clones and post-filter parses. The descending match-all arm similarly clones a full block to deliver limit plus one.

A first-batch size hint through Matches would remove the waste without changing the contract: the consumer passes its room, the first batch is sized to it, and later batches, which only run after post-filter drops, stay at matchBatchSize. While touching that plumbing, two tiny things in the same functions: the forward Iterator is allocated and immediately discarded on the descending path, and a positivity clamp at streamRange's top would keep a future table-driven test that sets the seam to 0 from spinning forever (step 0 never advances the loop).

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	if desc && hi > a.LatestLedger() &&
+		(cursor.Position != nil || cursor.ScannedLedger > 0) {
+		// A bookmark above this view's latest ledger proves more
+		// ledgers exist; another node already served them. A descending
+		// walk never goes back up, so serving now would skip those
+		// ledgers forever. Wait until this view has them.
+		return &EventPage{Next: cursor, Status: ScanHasMore}, nil
+	}
+	lo, hi, err := a.ClampRange(cursor.Scope.Dir, lo, hi)
```
tamirms: Two small things about the clamp. ClampRange's doc in bounds.go still says a descending hi is truncated to latest "since nothing exists there to drop", and the guard above this line exists precisely because that claim is false for resumed cursors. The next pager built on this infrastructure (scan.go's header names getTransactions as the next cursor-driven scan) could follow the documented contract and reintroduce the skip one endpoint over, so I would either scope the bounds.go rationale to fresh queries or move the resume-aware rule into bounds.go so the policy has one home. Separately, the range is clamped twice, here and again inside ChunksForRange, whose error return is dead code on this path; one clamp with one owner seems better.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+// maxScanLedgers bounds the ledgers one page may scan, so a filter
+// that matches nothing cannot walk the node's whole retention in one
+// call: the page stops at the window's edge and returns ScanHasMore
+// with the watermark advanced through it. At 10,000 ledgers a page
+// touches at most two chunks. A var, not a const, so tests can shrink
+// it to force window seams; production never writes it. The value may
+// become configuration.
+//
+//nolint:gochecknoglobals // test seam; production never writes it
+var maxScanLedgers = uint32(10_000)
```
tamirms: maxScanLedgers and matchBatchSize are the first mutable package globals in rpcv2, and this one is semantic: it changes page shape and status, and its comment already anticipates becoming configuration, which a process global cannot serve per instance. The tests in this package already run two registries side by side, and any test that writes the global forecloses t.Parallel for the whole package. The codebase's existing pattern for exactly this is an unexported struct field with a zero-value default (StartConfig.runBackfill, ExecConfig.retryBackoff), and a Registry field flowing into ReadView would match how geometry.Retention already threads through NewRegistry. Worth making that move before B2 builds on it?

cmd/stellar-rpc/internal/rpcv2/query/scan.go
```diff
@@ -135,35 +134,11 @@ func (w *ledgerWalk) closeAll() error {
 }
 
 // EventPart is one chunk's slice of a clamped range: the reader plus the
-// intersected ledger bounds — the input shape the events query engine consumes.
+// intersected ledger bounds, the input shape the events query engine
+// consumes. The pager builds one part at a time in walkChunks, so a
+// reader opens only when the walk reaches its chunk.
 type EventPart struct {
```
tamirms: EventPart now has exactly one producer and one consumer, both in this package, and Chunk is written in walkChunks and a test but read nowhere. Worth deleting the field, or folding the two remaining fields next to their call pair? If the offsets invariant check from my other comment lands and wants the chunk id in its error, keeping Chunk with a real reader is fine too.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,494 @@
...
+		return 0, fmt.Errorf("%w: ledger %d index %d holds (tx %d, op %d, event %d)",
+			ErrPositionMismatch, pos.Ledger, pos.LedgerOrdinal, p.TxIdx, p.OpIdx, p.EventIdx)
+	}
+	return ord, nil
+}
+
+// assemblePage is the one place cursor outputs are derived from the
+// walk's facts: the advanced bookmarks never regress and the watermark
+// never leaves the clamped window [clo, chi].
+func assemblePage(
```
tamirms: Four consecutive uint32s plus a desc that is derivable from the cursor it already receives: a transposed call (oldest and latest swapped with clo and chi, or a desc inconsistent with cursor.Scope.Dir) compiles silently and mints a wrong watermark or terminal status, the exact bug class this one-place-derives-outputs function exists to prevent. Making it a ReadView method (oldest and latest from the receiver) and deriving desc from cursor.Scope.Dir shrinks it to (cursor, walk, clo, chi, truncated); watermark and terminalStatus could drop their desc the same way.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
```diff
@@ -1,33 +1,22 @@
 package event
 
-// query.go is the events-side coordinator that turns a {filters,
-// Range, MaxEvents, Descending} spec into matching events for one
+// query.go is the events-side read surface: Matches turns a
+// {filters, window} spec into a stream of verified matches for one
```
tamirms, line 4: Three references didn't survive the rework: the cold-parity coverage map in query_test.go (the comments near lines 624 and 1158) still names fetchAllInRange and selectEventIDs, events/concurrent_bitmaps.go:101 still cites event.Query as the caller of the safe-methods contract, and design-docs/query-routing-design.md (lines 349 and 405) still specifies EventParts with eager up-front reader opens, which the lazy one-chunk-per-step walk deliberately replaced and TestQueryEvents_OpensOnlyScannedChunks pins. The design doc one matters most, since it now contradicts the implementation it governs. reader.go's parallel comment was updated, so these three look like misses rather than policy.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
```diff
@@ -456,10 +510,9 @@ func validateFilters(filters []Filter) error {
 		}
 		// Above MaxTopicCount the index cannot answer a count exactly: every
```
tamirms: This comment's mechanism didn't survive the move to Matches: post-filter drops are now invisible to the consumer, the stream keeps fetching batches until the window ends, so a superset-returning term can no longer leave a page short. That failure mode belonged to the old Query, where the MaxEvents cap applied before the post-filter. The cap itself still seems right, for the surviving reasons: no getEvents filter shape can ask for a count above MaxTopicCount, and the cursor codec carries the count in one byte. Worth trimming the comment to those, so the next reader doesn't go hunting for a page-shortening path that no longer exists.

cmd/stellar-rpc/internal/rpcv2/query/event_page_test.go
```diff
@@ -0,0 +1,888 @@
...
+	assert.Empty(t, page.Events)
+	assert.Equal(t, ScanHasMore, page.Status)
+	assert.Equal(t, f+6, page.Next.ScannedLedger)
+}
+
+// TestQueryEvents_WatermarkStaysInsideWindow pins the watermark
+// derivation: a mid-ledger stop never reports a value outside the
+// clamped window (an earlier version minted latest+1), and a stop
+// whose ledger was fully examined claims exactly that ledger.
+func TestQueryEvents_WatermarkStaysInsideWindow(t *testing.T) {
```
tamirms, line 738: The ascending clamp-echo branch of watermark (v below clo keeps the incoming mark) is never asserted: this test's ascending sub-case cannot fire the clamp at limit 3 and its assertion duplicates FilteredMidLedgerResume, so a change that breaks the echo (returning clo-1, say, which resume validation would then reject) passes the suite. Repointing the ascending sub-case at limit 1 with assert.Zero on ScannedLedger would make it the true mirror of the descending sub-case above it. And no pager-level test exercises a descending filtered mid-ledger resume; the descending rows are all match-all. Two smaller harness thoughts while there: pageDriver could Encode/Decode round-trip every minted cursor so the seam matrix also enforces the codec contract, and only limit 0, never a negative limit, exercises the limit guard.

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
```diff
@@ -100,7 +105,7 @@ type EventCursorQuery struct {
 // delivered event, and the scanned-ledger watermark. A nil Position means no
 // event has been delivered yet, so resume starts from the watermark alone.
 type EventCursor struct {
-	Query         EventCursorQuery
+	Scope         EventCursorQuery
```
tamirms, line 110: The field rename reads well in the pager, but the type is still EventCursorQuery, so the struct reads Scope EventCursorQuery and validateScope takes an *EventCursorQuery it calls scope. Two names for one concept become permanent once B2 imports the type; finishing the rename now (EventScope?) costs one package.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
tamirms: The unified rule also enables one more tightening, in validateBookmarkPair. With no clamped tops possible, the zero-watermark pair is exactly characterizable, and the updated comment states it: a zero watermark only pairs with a position in the scope's own MaxLedger. The branch does not enforce that yet, while the ascending mirror enforces its MinLedger equivalent, so a pair no walk mints, a zero watermark with a position below MaxLedger, still passes validation, resumes at the position, and silently never serves the ledgers between it and MaxLedger, which is the unaccounted-ledgers class this function exists to refuse. The symmetric check:

```go
if mark == 0 && pos.Ledger != *cursor.Scope.MaxLedger {
    return fmt.Errorf("%w: position ledger %d with no scanned ledger",
        ErrCursorMalformed, pos.Ledger)
}
```

validateScope has already refused a descending scope without a MaxLedger by this point, so the dereference is safe. What do you think?

urvisavla: Done in b5d83fbe: a descending scope whose top is above latest now always waits and returns ScanWaitingForLedgers, fresh and resumed alike. Added TestQueryEvents_DescendingFreshScopeAboveLatestWaits. Moved the rule into ClampRange (f69ed019).

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: Done in 91b44392: the clip is deleted; scanChunk now refuses a window if the chunk's offsets don't cover it.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: Done in 468fa38e: validateScope now rejects scopes exceeding the codec's filter cap and non-positive limits trigger a new error, ErrInvalidLimit.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
urvisavla: Done in 9f371270: streamUnion now fetches candidates based on the page's remaining room. The discarded forward iterator is removed, and batch sizes are clamped to be positive.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: Done in f69ed019: ClampRange now owns the wait rule — a descending top above latest returns an empty range instead of being truncated — and the pager's gate is removed. ChunksForRange is deleted; the pager clamps once.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: maxScanLedgers moved onto the Registry in 55891bf2, copied to each ReadView at acquisition. matchBatchSize stays a package var: the engine has no per-instance object to hold it, production never writes it, and since 9f371270 every use goes through batchSizes, which clamps it.

cmd/stellar-rpc/internal/rpcv2/query/scan.go
urvisavla: Done in f1d640dd: The struct was moved next to walkChunks and scanChunk as unexported eventPart. Chunk remains, as it's needed for the offsets-coverage error from 91b44392.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: Done in eff4d1cd: assemblePage is a ReadView method now, reads the serving range off the view, and derives desc from the cursor.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
urvisavla: Done in b43c3c4d: the test comments name streamRange and streamUnion now, concurrent_bitmaps cites event.Matches, and the design doc describes the lazy per-chunk resolution.

cmd/stellar-rpc/internal/rpcv2/stores/event/query.go
urvisavla: Trimmed in d83bd05f. The comment claimed the cap prevents pages coming back short of the limit; that could only happen in the old Query API, so the claim is gone. What stays: counts above MaxTopicCount share one index bucket, so the index cannot answer them exactly; no getEvents filter can ask for such a count; and the codec stores the count in one byte.

cmd/stellar-rpc/internal/rpcv2/query/event_page_test.go
urvisavla: Done in 47b0cb10: the ascending sub-case now stops inside the first ledger and asserts the zero watermark, mirroring the descending one; a descending filtered mid-ledger resume test is added; and pageDriver now encodes and decodes every minted cursor, so every pager test exercises the codec. The negative-limit row landed with 468fa38e.

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
urvisavla: Agreed on the name — the rename was already planned. We deferred it to the naming/cleanup follow-up PR to keep this diff small; it lands there with the other queued renames. Ok with that?

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
urvisavla: Done in 75143c70, with a validation-table row. I'd held the same tightening as an option after the wait-rule change. The paging tests still pass, so the check rejects nothing the server itself produces.

cmd/stellar-rpc/internal/rpcv2/query/event_cursor.go
tamirms: Works for me.

cmd/stellar-rpc/internal/rpcv2/query/bounds.go
```diff
@@ -67,6 +71,9 @@ func (a *ReadView) ClampRange(dir Direction, lo, hi uint32) (uint32, uint32, err
 	}
 
 	if hi > latest {
+		if dir == Descending {
+			return latest + 1, latest, nil // top above latest: wait
```
tamirms: The floor edge has the same divergence the top edge had, in mirror. ClampRange rejects a below-oldest leading edge in both directions, but the proposal is deliberately asymmetric at the floor: the ledger_out_of_range row enumerates three producing conditions, none of them a descending scan below the window, and closes with "Descending scans never get this error: they stop with OLDEST_REACHED instead" (stellar/stellar-protocol#1872, the errors table). The Bounds section also makes a descending minLedger below oldestLedger legal: "the node serves the ledgers it holds".

Three descending shapes hit it today, all returning *RangeError where the model says an empty OLDEST_REACHED page with the cursor unchanged:

1. A fresh scope entirely below the floor (explicit maxLedger below oldestLedger). The model makes this legal and empty rather than an error.
2. A re-pull of the cursor the floor page itself minted: its watermark is the floor, so the resume edge is floor minus one. Under the model a client that keeps re-pulling OLDEST_REACHED polls futilely and quietly; here the second pull throws.
3. A descending resume after pruning moved the floor past the remaining range. This is one of the warts the status model was redesigned to remove: a descending drain on a pruned node ending in a thrown error.

The walk-runs case is already right (a finished walk at the clamped floor reports OldestReached with its events), and the precedence is already right (the range-used-up check runs first, so a finished scope stays Complete forever, even below a pruned floor). What's missing is the no-walk case: descending, with the remaining top below the view's oldest, return the empty OldestReached page instead of the error.

On where it lives: the top edge's wait signal is ClampRange's empty range, but the floor case needs a different status, so the empty range cannot carry both meanings. Either ClampRange returns a typed outcome (serve, wait above, floor below) so both edge policies keep one home, or the descending floor check sits in the pager and ClampRange's leading-edge rejection becomes ascending-only with its doc saying why. I lean to the typed outcome for symmetry with f69ed019, but either keeps the policy in one place. Tests would mirror the top-edge pair: a fresh entirely-below case, and an OLDEST_REACHED re-pull that stays OLDEST_REACHED.

cmd/stellar-rpc/internal/rpcv2/query/event_page.go
```diff
@@ -0,0 +1,513 @@
...
+// assemblePage is the one place cursor outputs are derived from the
+// walk's facts: the advanced bookmarks never regress and the watermark
+// never leaves the clamped window [clo, chi].
+func (a *ReadView) assemblePage(
+	cursor EventCursor, walk walkResult, clo, chi uint32, truncated bool,
+) *EventPage {
+	desc := cursor.Scope.Dir == Descending
+	next := cursor
+	if walk.last != nil {
+		next.Position = walk.last
```
tamirms, line 463: One more tightening became available once the zero-watermark anchors were enforced: the cursor still carries Positions the watermark has passed, and those are now the only reason validateBookmarkPair needs inequalities instead of equalities. A Position is only ever read while it leads the watermark (resumeBounds takes the further-along bookmark, and resumeOrdinal only verifies on mid-ledger re-entry), so a Position at or behind the mark is dead weight: twenty bytes in every such cursor forever, never read and never verified, on any node.

The change is one minting rule plus the matching validation. assemblePage drops the Position once the watermark passes it (ascending pos.Ledger <= mark, descending mark != 0 && pos.Ledger >= mark), and validateBookmarkPair tightens to exact equalities: a present Position's ledger is mark+1 ascending, mark-1 descending, or the zero-watermark anchors already enforced. Every reachable cursor is then either a bare watermark or a watermark plus the one live mid-ledger re-entry point, and resumeBounds loses the which-bookmark-is-further merge: Position present means re-enter, absent means resume from the mark.

No wire change: Position is already presence-flagged in the envelope, and the rules compare cursor fields to cursor fields, so validity stays node-independent across a mixed fleet. The reason to decide it now rather than later: the tightened validation rejects shapes today's server mints, which is free while no cursors exist and impossible after release, when every stale-Position cursor ever handed out must stay accepted forever. What do you think?

tamirms (approved):

PR #936, opened 2026-08-19
bench(perf-eval): add a windowed relay poller for long campaigns

### LINKED ISSUE #1010: Wait for a benchmark campaign across several GitHub Actions jobs
## TL;DR

Campaigns run paced ingestion: one 10,000-ledger chunk at one ledger per 2 s is about 5.6 h of replay per profile, and a full suite outlives the six-hour Actions job limit. `Gather` waits inside one job budget and cannot hand a wait to the next job. Add a `Relay` command that polls the campaign's S3 result object for one bounded window and reports `ok`, `fail`, or `running` so a chain of jobs can wait under one campaign deadline.

## Scope

- **`perf-eval/relay` command** — polls `s3://$BUCKET/$RESULT_KEY` until the earlier of window end and campaign deadline; writes `state=ok|fail|running` to `$GITHUB_OUTPUT`; the caller owns the job failure gate. At the deadline it makes one final fetch before reporting failure.
- **Shared `resultPoller`** — one polling loop for `Gather` and `Relay`: fetch, classify, log a debug tail over SSM every N polls, sleep, stop at window end. `Gather` is rewritten onto it with unchanged outputs.
- **Fetch classification** — an absent key (404) and a result from another run attempt are healthy waits; `ErrInvalidResult` and 4xx (except S3 `RequestTimeout`) fail now; transport failures and 5xx keep retrying inside the window.
- **Bounded remote calls** — 30 s per result fetch, 60 s per SSM diagnostic including its send retries; sleeps honour context cancellation.
- **`Result.Validate`** — schema version 1, non-empty `runId`, verdict `ok` or `fail`. No `pending` verdict: an absent key is the wait state, and the workflow role needs `s3:ListBucket` so absence is 404, not 403 (stellar/terraform#6902).
- **Coordinator identity check** — `reportableLeg`: a leg result counts only when its `RunID` belongs to this run and its `TargetSHA` matches, so a partial re-run keeps earlier passing legs and never reports another run's result as its own.
- **Tests** — poller states, retries, handoff, deadline, cancellation, and both commands' output contracts, on fake AWS transports and virtual time.

The multi-job campaign workflow is separate (#1014).

## Done when

`Relay` exposes `ok` / `fail` / `running` and the final deadline fetch; `Gather` runs on the shared poller with unchanged outputs; the coordinator reports only this run's leg results; the poller paths above are covered by tests.

## Related

- #1006 — follow-up that shares one poller configuration between the two commands (#1011).
- #868 — the first campaign automation this relay replaces the waiting logic of.

marwen-abid:
## Purpose

The existing apply-load benchmark measures ingestion throughput and per-ledger processing time under fast replay. The upcoming RPCv2 campaigns add a complementary measurement: whether hot ingestion sustains a set ledger arrival rate over time, and how far commits fall behind that schedule.

Paced ingestion can exceed GitHub Actions' six-hour job limit by design. One RPCv2 chunk contains 10,000 ledgers; at one ledger every five seconds, replay alone takes about 14 hours.

This PR adds **Relay**, a bounded result poller that lets successive GitHub Actions jobs wait for the same campaign while its EC2 benchmark continues running. Relay shares Gather's polling loop, result validation, and diagnostics. The multi-job campaign workflow is separate work.

> [!NOTE]
> Example 19h run for Bench Campaign https://github.com/stellar/stellar-rpc/actions/runs/34421921550

## Result Flow

Arrows show result flow. The existing coordinator continues to use Gather; Relay is a separate caller of the shared poller.

```mermaid
flowchart TB
    subgraph before["Before"]
        B["EC2 benchmark"] --> S[("S3 final result")]
        S --> G["Gather: polling + reporting"]
        G --> O["found / passed<br/>Job gate + instance cleanup"]
        S --> C["Coordinator<br/>PR report"]
    end
```

```mermaid
flowchart TB
    subgraph after["With this PR"]
        B2["EC2 benchmark"] --> S2[("S3 final result")]
        S2 --> P["Shared result poller<br/>waits through 404, stale attempt, transport, 5xx<br/>fails now on 4xx fault or invalid object"]
        P --> G2["Gather<br/>found / passed"]
        G2 --> O2["Existing job gate<br/>+ instance cleanup"]
        P --> R["New Relay<br/>ok / fail / running"]
        R -. "running: caller schedules another window" .-> R
        S2 --> C2["Coordinator<br/>checks run + target identity, then reports"]
    end
```

## Relay Contract

Relay returns `ok` or `fail` for a final result, and `running` when its window ends with campaign time remaining. Polling faults also produce `fail`. At the campaign deadline, it makes one final fetch before reporting failure if no final result is available.

These states are step outputs; the caller owns the job failure gate and schedules further windows. Configuration, cancellation, and local output errors return nonzero. Result fetches have a 30-second limit and diagnostics a 60-second limit; the final fetch and diagnostics can extend beyond the campaign deadline.

## What's Next

- Launch configurable benchmark campaigns on EC2 from GitHub Actions.
- Use Relay to wait across multiple jobs under one campaign deadline.
- Add scheduled cleanup of expired instances, Slack summaries, and results-site publication.
- Follow-up: one `pollerConfig` with a `newResultPoller` constructor shared by Gather and Relay.

## Validation

Validated that existing Benchmarks don't break. https://github.com/stellar/stellar-rpc/actions/runs/34518652855

Closes #1010


cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/poller.go
```diff
@@ -0,0 +1,113 @@
...
+	// their gatherer leaves this false and 404s stay normal.
+	keySeeded       bool
+	debugLogLines   int
+	debugEveryPolls int
+}
+
+// poll polls until `until`. It returns (res, nil) when a final verdict for
+// this run appears, (nil, nil) when the window closes without one, and an
+// error after maxConsecutiveFetchErrors consecutive failed fetches.
+func (p *resultPoller) poll(ctx context.Context, until time.Time) (*Result, error) {
```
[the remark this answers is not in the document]
marwen-abid: Added deterministic shared-poller tests in 0bce1941, with boundary coverage in b4ebf3a6.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/coordinator-runner.go
```diff
@@ -99,6 +99,12 @@ func run(ctx context.Context) error {
 			logger.Warnf("no result at s3://%s/%s: %v", l.Bucket, l.Key, err)
 			continue
 		}
+		// A campaign that never overwrote its seed marker has no outcome to
+		// report; it renders through the no-result fallback below.
+		if res.Verdict == harness.VerdictPending {
+			logger.Warnf("result at s3://%s/%s is still pending", l.Bucket, l.Key)
+			continue
```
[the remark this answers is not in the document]
marwen-abid: Fixed in d57db3a2.

.github/workflows/ec2-leg.yml
```diff
@@ -155,6 +155,19 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
           role-duration-seconds: ${{ steps.deadlines.outputs.role_seconds }}
 
+      # Only the shared poller understands pending. Older target refs use their
+      # original Gather and must wait on an unseeded key.
+      - name: Seed the pending result marker
```
tamirms: Was granting `s3:ListBucket` on the `runs/*` prefix to the gha-rpc-ci role considered instead of seeding? Without ListBucket, GetObject on a missing key comes back as 403, which is what makes an absent key indistinguishable from a permissions fault and what the pending marker is working around. With ListBucket, a missing key is a 404 that the poller treats as a healthy wait, a 403 stays a fault, and the fail-fast keeps the same coverage. That would let this PR drop the `pending` verdict, the seed step here and in bench-campaign.yml, the `hashFiles` guard for older target refs, `TestLegSeedingContract`, and the pending branch in `reportableLeg`. The one thing seeding buys that IAM does not is overwriting a predecessor's result at launch, and the RunID check already covers that case. If the IAM change is slow to land, seeding is fine as a bridge, but I would rather not bake a third verdict value into the result protocol for it.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/poller.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+// Ten failed polls in a row, about 5 min at the 30 s interval.
+const maxConsecutiveFetchErrors = 10
```
tamirms: Ten consecutive fetch errors is about five minutes at the 30 s interval, against budgets measured in hours, and for a leg a false give-up is expensive: the `Fail leg` step exits 1 and the `Terminate instance` step then kills a box that may be hours into its run and would have published fine. Each `GetObject` already goes through the SDK's three retry attempts, so this only fires when S3 or the runner's network is unhealthy for the full five minutes, but those blips happen and the box is unaffected by them because it writes once at the end. Would it make sense to fail fast only on faults that cannot heal (AccessDenied, an invalid object) and keep transport and 5xx errors on the budget clock, or at least raise the threshold to something like 30 minutes for those? The previous Gather waited out the budget on any error, so this is a behavior change for the existing legs as well as the campaigns.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/relay.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+
+	awsCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(cfg.region))
+	if err != nil {
+		return err
+	}
+	s3Client := s3.NewFromConfig(awsCfg)
+	runner := &ssmRunner{client: ssm.NewFromConfig(awsCfg), instanceID: cfg.instanceID}
+	poller := &resultPoller{
```
tamirms, line 37: The poller loop is shared, but the two commands still duplicate everything around it: the same six `RequireEnv` keys, the overlapping int keys and their positivity checks, `LoadDefaultConfig`, the S3 and SSM clients, and this identical `resultPoller` literal (gather.go:50). Relay carries that in `relayConfig` while Gather keeps loose locals and passes eight positional arguments to `reportGather`, so the two halves read as different designs. One `pollerConfig` loaded from env with a `newResultPoller(ctx, cfg)` constructor would leave each command as load config, build poller, poll, report, with Gather adding `RESULTS_TIMEOUT` and Relay adding `WINDOW_SECONDS` and `DEADLINE_EPOCH`. The `relay` struct could then hold the poller plus those two fields instead of a config whose fields mirror the poller's, and `reportFault` would not need to reach into `r.poller.runner`. If the config became a tagged struct through `caarlos0/env`, which `Leg` already uses on the box side, `RequireEnvInts`, `requirePositive`, and `requireSeconds` would go with it; the overflow guard in `requireSeconds` in particular protects against a 292-year window that nobody will set. This is a reshape rather than a bug, so I am fine with it as a follow-up if you would rather keep this PR to the campaign feature.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/workflow_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// These checks read workflow text only. They never execute infrastructure steps.
+func TestLegSeedingContract(t *testing.T) {
```
tamirms: I would drop this test, or keep only the step-order loop. It pins the YAML by substring, including gates that have nothing to do with seeding (`KEEP_INSTANCE`, `ADOPT_RUN_LABEL`, the `found` and `passed` checks), so any cosmetic edit to ec2-leg.yml breaks a Go test while a real regression in the step's shell passes as long as the quoted fragments survive. The invariant that matters, seed after the OIDC step and before launch, is the one thing the loop checks, and the comment on the step already states it.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/poller.go
```diff
@@ -0,0 +1,123 @@
...
+// checkOnce fetches the key once and classifies what it finds. It returns a
+// result only for a final verdict from this run. A current pending marker is
+// a healthy wait; stale results and fetch failures count toward the error limit.
+func (p *resultPoller) checkOnce(ctx context.Context) (*Result, error) {
+	ctx, cancel := context.WithTimeout(ctx, resultFetchTimeout)
+	defer cancel()
+	res, err := FetchResult(ctx, p.s3Client, p.bucket, p.key)
+	switch {
+	case errors.Is(err, ErrResultNotReady):
+		logger.Infof("still waiting for s3://%s/%s", p.bucket, p.key)
```
tamirms, line 75: A few messages did not follow the semantics change. An absent key now logs `still waiting` at info and then `result fetch failed (n/10); retrying` at warn on every poll, which reads as healthy and faulty at once; since absence is a fault now, the `still waiting` line can go. Line 92 says `campaign still running` for legs too (it shows up in the go-bench leg log of run 34518652855); `result still pending` fits both callers. The two `giveUpErr` branches differ in style (one starts with ❌ and a capital, the other lowercase) and both land as the step-summary headline. The package doc in harness.go lists the package's parts and does not mention Relay or the shared poller yet.

.github/workflows/ec2-leg.yml
marwen-abid: I agree it's better to add this now. Here is the terraform https://github.com/stellar/terraform/pull/6902
Also made the related changes in 6cf5bba7f2b4834fefa803cf7644c36820db4d42

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/poller.go
marwen-abid: addressed in 6cf5bba7f2b4834fefa803cf7644c36820db4d42

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/relay.go
marwen-abid: Agreed on the shape, and I'd rather keep this PR to the campaign feature. I'll add this as a follow up PR to the stack.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/workflow_test.go
marwen-abid: addressed in 6cf5bba7f2b4834fefa803cf7644c36820db4d42

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/poller.go
marwen-abid: addressed in 6cf5bba7f2b4834fefa803cf7644c36820db4d42

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/s3.go
```diff
@@ -30,6 +29,28 @@ type Result struct {
...
+		return fmt.Errorf("unsupported schemaVersion %d", r.SchemaVersion)
+	}
+	if r.RunID == "" {
+		return errors.New("runId is required")
+	}
+	switch r.Verdict {
+	case VerdictOK, VerdictFail:
+		return nil
+	default:
+		return fmt.Errorf("unknown verdict %q", r.Verdict)
```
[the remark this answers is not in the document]
marwen-abid: Fixed

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/harness/relay.go
marwen-abid: Opened this PR https://github.com/stellar/stellar-rpc/pull/1006

tamirms (approved):

tamirms (approved):
Approving. Every item from the first round is addressed here or deferred to #1006, and the rebase carried the series unchanged. One heads-up for the rebase of `bench-campaign` onto this: that branch still seeds a `pending` marker from its own older harness copy, and `Validate` here rejects `pending` as an invalid object, which the poller treats as a permanent fault, so the seed step there should go in the same commit.

Issue #935, opened 2026-08-19
Auto-refresh release evaluation test assets


cjonas9:
### What

In epic #706, an automated release evaluation framework is established. Two of the four tests use a number of assets stored in S3: 
- from apply-load live ingestion test:
    - golden DB: `current/golden.sqlite.zst`
    - `stellar-core`: `core/stellar-core.zst`
    - 3 ledger bundles: `ledgers/load-test-ledgers-*.xdr.zst`
- from backfill ingestion test: 
    - `stellar-core`: `core/stellar-core.zst` (same as above)

On a monthly cadence, the DB in `current/` should be replaced by a freshly-backfilled DB; the current should be moved into `prev1/`, the DB in `prev1/` should be moved into `prev2/`, and the DB in `prev2/` should be evicted. Alternatively, whenever the backfill test is run and successful and the golden DB is older than ~2 weeks, this sequence should initiate. The `prev*/` folders will probably need to be established, along with some logic to defer back to these DBs if the DB in `current/` is corrupted.

On that same cadence, one should likely have the 3 ledger bundles be regenerated using `apply-load` machinery in `stellar-rpc`, and have the old replaced with the new.

Then, on some cadence (likely matching `stellar-core` releases), the prebuilt binary should be updated. Alternatively, rather than using a prebuilt binary, one could set this up to pull the latest from `stellar/stellar-core.`

### Why

These assets are made semi-constant to avoid constant drift between assets. With that said, they should still be periodically refreshed to ensure the test stays up-to-date.

roebee:
Three questions:

1. Cadence: a fixed monthly schedule, on backfill success with a 2-week staleness check, or both?
2. On a refresh failure: does the framework fall back to `prev1`/`prev2`, or fail the evaluation run?
3. The stellar-core binary: pin the release build per core release, or pull the latest from stellar/stellar-core?

PR #938, opened 2026-08-21
perf-eval: fix per-eval for full-history

### LINKED ISSUE #944: Fix load-test coordinator on feature/full-history
The load-test coordinator does not complete successfully on `feature/full-history`.

The go-bench leg attempts to build rpcv2 packages that require native libraries unavailable on the benchmark box. The endpoint-load leg also encounters keep-alive connection resets, and serve-phase logs are lost when the box is terminated.

Update the perf-eval workflow to:

- benchmark only packages supported by the stock box;
- preserve logs throughout the serve phase;
- avoid premature idle connection closure.

The full load-test coordinator should pass on `feature/full-history`.

marwen-abid:
## Summary

- Exclude rpcv2 and native-library packages from go-bench on stock boxes.
- Upload backfill logs while serving and report early daemon exits.
- Set the rpcv1 HTTP idle timeout to 2 minutes to prevent keep-alive reuse failures.

## Validation
- [x] CI
- [x] Copilot Review  

Full coordinator runs: 
- [x]  [32521300483](https://github.com/stellar/stellar-rpc/actions/runs/32521300483)
- [x] [32538909114](https://github.com/stellar/stellar-rpc/actions/runs/32538909114)
- [x] [32763039473](https://github.com/stellar/stellar-rpc/actions/runs/32763039473)

## Note for Reviewer
* I added a comment for each changeset to mark the changes related to `feature/full-history` merge and the ones that were discovered during testing. 

Closes #944

### DIFF
--- a/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
@@ -40,6 +40,7 @@ import (
 
 const (
 	defaultShutdownGracePeriod = 10 * time.Second
+	defaultIdleTimeout         = 2 * time.Minute
 
 	// Since our default retention window will be 7 days (7*17,280 ledgers),
 	// choose a random 5-digit prime to have irregular logging intervals at each
@@ -402,6 +403,7 @@ func (d *Daemon) setupHTTPServers(cfg *config.Config) {
 	d.server = &http.Server{
 		Handler:     createHTTPHandler(d.logger, d.jsonRPCHandler),
 		ReadTimeout: jsonrpc.DefaultHTTPReadTimeout,
+		IdleTimeout: defaultIdleTimeout,
 	}
 
 	if cfg.AdminEndpoint != "" {
--- a/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
@@ -40,7 +40,6 @@ import (
 
 const (
 	defaultShutdownGracePeriod = 10 * time.Second
-	defaultIdleTimeout         = 2 * time.Minute
 
 	// Since our default retention window will be 7 days (7*17,280 ledgers),
 	// choose a random 5-digit prime to have irregular logging intervals at each
@@ -403,7 +402,7 @@ func (d *Daemon) setupHTTPServers(cfg *config.Config) {
 	d.server = &http.Server{
 		Handler:     createHTTPHandler(d.logger, d.jsonRPCHandler),
 		ReadTimeout: jsonrpc.DefaultHTTPReadTimeout,
-		IdleTimeout: defaultIdleTimeout,
+		IdleTimeout: jsonrpc.DefaultHTTPIdleTimeout,
 	}
 
 	if cfg.AdminEndpoint != "" {
@@ -425,7 +424,11 @@ func (d *Daemon) setupAdminServer(cfg *config.Config) {
 	if err != nil {
 		d.logger.WithError(err).WithField("endpoint", cfg.AdminEndpoint).Fatal("cannot listen on admin endpoint")
 	}
-	d.adminServer = &http.Server{Handler: adminMux} //nolint:gosec
+	d.adminServer = &http.Server{
+		Handler:     adminMux,
+		ReadTimeout: jsonrpc.DefaultHTTPReadTimeout,
+		IdleTimeout: jsonrpc.DefaultHTTPIdleTimeout,
+	}
 }
 
 // mustInitializeStorage initializes the storage using what was on the DB
--- a/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
+++ b/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
@@ -258,7 +258,7 @@ func runBackfill(
 			fmt.Fprintln(os.Stderr, scanner.Text())
 		}
 		pr.Close()
-		_ = cmd.Wait() // reap; a kill from cancel surfaces here and is expected
+		logger.Infof("daemon exited: %v", cmd.Wait())
 	}()
 	daemon := &daemonHandle{cancel: cancel, done: done}
 	if !keepAlive {
--- a/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
+++ b/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
@@ -4,11 +4,13 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/config"
@@ -112,7 +114,7 @@ func instantiate(ctx context.Context) error {
 
 	if daemon != nil {
 		// hand the serving box off to the chained blaster leg
-		servePhase(ctx, daemon)
+		servePhase(ctx, daemon, cfg.Bucket, cfg.ResultKey)
 	}
 	return nil
 }
@@ -181,12 +183,15 @@ func renderConfig(repoRoot, workDir, coreCfg, retention, endpoint string) (strin
 
 // daemonHandle controls a daemon left running past its backfill phase.
 type daemonHandle struct {
-	cancel context.CancelFunc
-	done   chan struct{} // closed once the daemon is reaped
+	cancel   context.CancelFunc
+	done     chan struct{} // closed once the daemon is reaped
+	err      error         // cmd.Wait's result; valid once done is closed
+	stopping atomic.Bool   // set by Stop, so the reaper can tell a requested kill from a crash
 }
 
 // Stop kills the daemon and waits (bounded) for it to be reaped.
 func (d *daemonHandle) Stop() {
+	d.stopping.Store(true)
 	d.cancel()
 	select {
 	case <-d.done:
@@ -249,23 +254,35 @@ func runBackfill(
 		return 0, 0, 0, nil, fmt.Errorf("daemon exited or hit the %s deadline before backfill completed", deadline)
 	}
 
-	// keep draining the pipe (the daemon blocks on it once full) and teeing
-	// its catchup/ingestion output to the box log until it dies
-	done := make(chan struct{})
-	go func() {
-		defer close(done)
-		for scanner.Scan() {
-			fmt.Fprintln(os.Stderr, scanner.Text())
-		}
-		pr.Close()
-		logger.Infof("daemon exited: %v", cmd.Wait())
-	}()
-	daemon := &daemonHandle{cancel: cancel, done: done}
+	daemon := &daemonHandle{cancel: cancel, done: make(chan struct{})}
+	go daemon.reap(scanner, pr, cmd)
 	if !keepAlive {
 		daemon.Stop() // stop the daemon before it starts live ingestion
 		return elapsed, lo, hi, nil, nil
 	}
 	return elapsed, lo, hi, daemon, nil
+}
+
+// reap keeps draining the pipe (the daemon blocks on it once full), teeing its
+// catchup/ingestion output to the box log until it dies, then records how.
+func (d *daemonHandle) reap(scanner *bufio.Scanner, pr *os.File, cmd *exec.Cmd) {
+	defer close(d.done)
+	for scanner.Scan() {
+		fmt.Fprintln(os.Stderr, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		// a line over the scanner's cap must not stop the drain: closing the
+		// read end would SIGPIPE-kill a healthy daemon on its next write
+		logger.Warnf("reading daemon output: %v; draining the rest unbuffered", err)
+		_, _ = io.Copy(os.Stderr, pr)
+	}
+	pr.Close()
+	d.err = cmd.Wait()
+	if d.stopping.Load() {
+		logger.Infof("daemon stopped on request: %v", d.err)
+	} else {
+		logger.Warnf("daemon exited on its own: %v", d.err)
+	}
 }
 
 func renderMarkdown(sha, retention string, lo, hi, ingested int, elapsed time.Duration) string {
--- a/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
+++ b/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"math"
 	"os/exec"
 	"path"
 	"strings"
@@ -14,28 +15,39 @@ import (
 const (
 	rpcPort               = "8000"
 	serveSnapshotInterval = 2 * time.Minute
+	// snapshotTimeout bounds one box-log upload. Under the tick, so uploads
+	// never overlap and a stalled S3 path cannot keep the serve loop from
+	// noticing a daemon exit or the ceiling.
+	snapshotTimeout = time.Minute
+	// ceilingBackstopSlackMinutes keeps the OS poweroff behind the in-process
+	// ceiling, so the teardown defers (the final snapshot included) run first.
+	ceilingBackstopSlackMinutes = 2
 )
 
 type serveEnv struct {
-	Ceiling   time.Duration `env:"SERVE_CEILING" envDefault:"6h"`
-	Bucket    string        `env:"BUCKET"        envDefault:"stellar-rpc-ci-load-test"`
-	ResultKey string        `env:"RESULT_KEY"`
+	Ceiling time.Duration `env:"SERVE_CEILING" envDefault:"6h"`
 }
 
-// servePhase keeps the daemon available to the blaster leg.
-func servePhase(ctx context.Context, daemon *daemonHandle) {
+// servePhase keeps the daemon available to the blaster leg, snapshotting the
+// box log next to the result object until the box is terminated.
+func servePhase(ctx context.Context, daemon *daemonHandle, bucket, resultKey string) {
 	cfg, err := env.ParseAs[serveEnv]()
 	if err != nil {
 		logger.Warnf("parsing serve env: %v", err)
 		cfg.Ceiling = 6 * time.Hour
 	}
 
-	defer rescheduleShutdown(ctx, 1, "serve phase over")
-	defer snapshotBoxLog(context.WithoutCancel(ctx), cfg.Bucket, cfg.ResultKey)
+	// teardown outlives ctx: the final snapshot and the poweroff must run
+	// however the loop below exits
+	teardown := context.WithoutCancel(ctx)
+	defer rescheduleShutdown(teardown, 1, "serve phase over")
+	defer snapshotBoxLog(teardown, bucket, resultKey)
 	defer daemon.Stop()
 
-	rescheduleShutdown(ctx, int(cfg.Ceiling.Minutes()), "serve ceiling")
-	snapshotBoxLog(ctx, cfg.Bucket, cfg.ResultKey)
+	// the OS poweroff is only the safety net; the in-process ceiling fires first
+	backstop := int(math.Ceil(cfg.Ceiling.Minutes())) + ceilingBackstopSlackMinutes
+	rescheduleShutdown(ctx, backstop, "serve ceiling backstop")
+	snapshotBoxLog(ctx, bucket, resultKey)
 
 	logger.Infof("serving :%s until external termination (ceiling %s)", rpcPort, cfg.Ceiling)
 	ceiling := time.After(cfg.Ceiling)
@@ -49,11 +61,11 @@ func servePhase(ctx context.Context, daemon *daemonHandle) {
 			logger.Warnf("serve ceiling passed without external termination")
 			return
 		case <-daemon.done:
-			logger.Warnf("serving daemon exited")
-			dumpDmesgTail(ctx)
+			logger.Warnf("serving daemon exited: %v", daemon.err)
+			dumpDmesgTail(teardown)
 			return
 		case <-ticker.C:
-			snapshotBoxLog(ctx, cfg.Bucket, cfg.ResultKey)
+			snapshotBoxLog(ctx, bucket, resultKey)
 		}
 	}
 }
@@ -76,6 +88,8 @@ func snapshotBoxLog(ctx context.Context, bucket, key string) {
 	if bucket == "" || key == "" {
 		return
 	}
+	ctx, cancel := context.WithTimeout(ctx, snapshotTimeout)
+	defer cancel()
 	dst := fmt.Sprintf("s3://%s/%s/user-data-serving.log", bucket, path.Dir(key))
 	cmd := exec.CommandContext(ctx, "aws", "s3", "cp", "/var/log/user-data.log", dst)
 	if out, err := cmd.CombinedOutput(); err != nil {
--- a/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
+++ b/cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
@@ -5,46 +5,70 @@ import (
 	"fmt"
 	"os/exec"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/caarlos0/env/v11"
 )
 
-const rpcPort = "8000" // rpcPort is where the daemon serves, crosses VPC
+const (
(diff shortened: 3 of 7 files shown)


cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
```diff
@@ -44,6 +44,7 @@ import (
 const (
 	defaultReadTimeout         = 5 * time.Second
 	defaultShutdownGracePeriod = 10 * time.Second
+	defaultIdleTimeout         = 2 * time.Minute
```
marwen-abid, line 43: _Issue pre-existing on main and discovered during testing._

Added a 2-minute HTTP `IdleTimeout`. Without it, Go reused the 5-second `ReadTimeout` for idle keep-alive connections, causing pooled clients to receive EOF/reset errors after pause.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -258,7 +258,7 @@ func runBackfill(
 			fmt.Fprintln(os.Stderr, scanner.Text())
 		}
 		pr.Close()
-		_ = cmd.Wait() // reap; a kill from cancel surfaces here and is expected
+		logger.Infof("daemon exited: %v", cmd.Wait())
```
marwen-abid, line 261: _Pre-existing observability gap on main._ 

The daemon reaper now logs `cmd.Wait()` instead of discarding it. This distinguishes normal shutdowns from crashes or kills.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid, line 1: _Pre-existing observability gap on main._

The serving phase now:
- uploads the box log immediately and every two minutes;
- performs a final upload on normal exit;
- watches `daemon.done` and exits promptly if the daemon dies;
- appends the last 100 `dmesg` lines for crash/OOM diagnosis.

This is necessary because the blaster hard-terminates the box, bypassing the shell EXIT trap that previously uploaded logs.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -30,6 +30,16 @@ var benchDenylist = []string{
 	"BenchmarkTransactionFetch", // is broken in current baseline release
 }
 
+var nativeLibRoots = []string{
+	"github.com/linxGnu/grocksdb",
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/zstd",
+}
+
+var v2Prefixes = []string{
```
marwen-abid, line 38: *👋 Issue caused by full-history* 

Filter RPC v2 packages. The branch added rpcv2 and native-library dependencies that the stock box cannot compile.

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
cjonas9: It's definitely not clear to me what this affected. In what situation did this show up?

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
cjonas9: good catch!

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
cjonas9: hm, this logging behavior intuitively feels quite excessive to me -- the logs, in the first place, are kept in-place from debugging this and realizing they're a nice-to-have going forward. with that in mind, had I designed this more robustly, I'd still have avoided logging at this cadence. Of course, it's not about the storage space kept in S3, but rather the cluttering that would accumulate.

I'm unsure of what the ideal compromise is -- perhaps clearing stale logs could be added to #935 and/or rather than an "upload every N minutes," the logs could be streamed somewhere to be kept in memory and uploaded once the box is killed under all circumstances.

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
marwen-abid: When `IdleTimeout` is 0, `net/http` uses `ReadTimeout` (5s here) as the keep-alive idle timeout, so the server closed any connection that stayed idle for more than 5s.

The blaster keeps pooled connections for 45s, and its seeder pauses 5-6.5s between seeding phases before it reuses a pooled connection. That pause lands right on the server's 5s close boundary: the client sends on a connection it thinks is open but the server has just closed it and the request fails with `EOF` or `connection reset by peer`. This is a concurrency issue so the failures are intermittent and the error message would be different depending on where the request is when the connection closes.  This race condition killed the endpoint-load-test leg twice : 
- https://github.com/stellar/stellar-rpc/actions/runs/32506853991/job/96860136155
- https://github.com/stellar/stellar-rpc/actions/runs/32324377300/job/96308045998

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid: Yeah it's a tricky one. In this case, each snapshot overwrites the same `user-data-serving.log` key, so the cadence does not create additional visible objects; each run still leaves only one serving log. Also the issue is that a final (in-process) upload cannot be guaranteed because the box is hard-terminated. 

Another way to solve that more cleanly would be streaming to CloudWatch but it's a larger change. 

I'm open to reducing cadence or we can address this in another PR/issue.

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
```diff
@@ -40,6 +40,7 @@ import (
 
 const (
 	defaultShutdownGracePeriod = 10 * time.Second
+	defaultIdleTimeout         = 2 * time.Minute
```
tamirms, line 43: Should the same fix cover the v2 servers? `internal/rpcv2/serve.go` builds both of its servers with only `ReadTimeout: jsonrpc.DefaultHTTPReadTimeout` (lines 58 and 99), so they still have the 5s idle-close this PR diagnoses, and the v2 read server is the path the full-history daemon actually serves on.

One option: hoist this constant next to `jsonrpc.DefaultHTTPReadTimeout` (whose doc comment already describes it as the timeout both daemons use) as a `DefaultHTTPIdleTimeout`, and set the pair at all four `http.Server` construction sites. That would also pick up the v1 admin server at line 428, which currently sets no timeouts at all (the `//nolint:gosec` there is silencing the missing-timeouts lint), so it has no slowloris bound and an infinite idle window.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
```diff
@@ -258,7 +258,7 @@ func runBackfill(
 			fmt.Fprintln(os.Stderr, scanner.Text())
 		}
 		pr.Close()
```
tamirms, line 260: Could we log `scanner.Err()` here before `pr.Close()`, the way the `elapsed == 0` path above does? If the daemon ever emits a line over the 1MB cap set at line 229, `scanner.Scan` returns false with `bufio.ErrTooLong`, this goroutine silently stops draining and closes the read end, and the still-healthy daemon gets SIGPIPE-killed on its next write, which the serve loop then reports as a spontaneous daemon crash.

I reproduced the chain with this exact scanner config: the child stays healthy until the drain stops, then dies with `signal: broken pipe` immediately after the close, and the box log's only breadcrumb blames the daemon rather than the runner. Captive-core log lines pass through this pipe and nothing upstream guarantees they stay under the cap, so a single pathological line during a 6h serve would take down the blast run.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -193,6 +218,75 @@ func runSuite(ctx context.Context, dir, outFile string, count int) []string {
 	return nil
 }
 
+// benchPackages excludes rpcv2 and packages that need native libraries absent
+// from the benchmark box.
+func benchPackages(ctx context.Context, dir string) ([]string, error) {
+	var stdout, stderr bytes.Buffer
+	cmd := exec.CommandContext(ctx, "go", "list", "-test", "-json", "./...")
```
tamirms, line 225: Would it be worth adding `-e` here and skipping entries with a non-nil `Error`/`DepsErrors` (or `Incomplete: true`)? Without it, one package that fails to load (an unresolvable test import, a syntax error) in either tree makes `go list` exit 1 and fails the whole leg before any benchmark runs, even if the broken package is one `excludeFromBench` would have dropped anyway. The old `go test ./...` flow degraded to a per-package FAIL row and still produced benchstat and the report.

The sharp case is the baseline: it is pinned and immutable, and `benchDenylist` already documents baseline breakage happening before, so a load-broken baseline would brick the leg until the runner itself is patched. I checked the behavior difference on a toy module: `go list -test -json ./...` exits 1 on one broken `_test.go` import while `go test -bench . ./...` still benches every other package.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
```diff
@@ -5,46 +5,70 @@ import (
...
 	}
 
-	// the boot-time self-terminate (budget+15m) can predate the blast's end
-	rescheduleShutdown(ctx, int(cfg.Ceiling.Minutes()), "serve ceiling")
+	defer rescheduleShutdown(ctx, 1, "serve phase over")
+	defer snapshotBoxLog(context.WithoutCancel(ctx), cfg.Bucket, cfg.ResultKey)
+	defer daemon.Stop()
 
-	// the happy-path hard kill skips the EXIT-trap upload, persist the log up to now
+	rescheduleShutdown(ctx, int(cfg.Ceiling.Minutes()), "serve ceiling")
```
tamirms, line 37: I think the backstop wants a round-up and some slack. `int(cfg.Ceiling.Minutes())` truncates toward zero, and this OS shutdown is scheduled before the in-process ceiling timer gets armed (which only happens after the synchronous first upload below), so on the ceiling path the poweroff SIGTERMs the runner first and none of the defers run, including the final `snapshotBoxLog` this PR adds for exactly these orphaned runs.

The degenerate cases: `SERVE_CEILING=30s` becomes `shutdown -P +0` (immediate poweroff), and 90s becomes a 60s shutdown that beats the 90s timer. Something like

```go
rescheduleShutdown(ctx, int(math.Ceil(cfg.Ceiling.Minutes()))+2, "serve ceiling backstop")
```

keeps the in-process ceiling as the one that fires first, with the OS shutdown purely as the safety net.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
```diff
@@ -5,46 +5,70 @@ import (
...
+			return
+		case <-ceiling:
+			logger.Warnf("serve ceiling passed without external termination")
+			return
+		case <-daemon.done:
+			logger.Warnf("serving daemon exited")
+			dumpDmesgTail(ctx)
+			return
+		case <-ticker.C:
+			snapshotBoxLog(ctx, cfg.Bucket, cfg.ResultKey)
```
tamirms, line 56: Since `ctx` here is `context.Background` all the way from `harness.Run`, the `CommandContext` inside `snapshotBoxLog` has no bound, and the upload runs synchronously in this select arm, so while a slow or retrying `aws s3 cp` runs, the loop cannot observe `daemon.done` or the ceiling. The aws CLI's own connect/read timeouts keep it from hanging forever, but a stalled S3 path can still block each tick for minutes and delay noticing a daemon exit.

What do you think about giving the snapshot its own deadline?

```go
func snapshotBoxLog(ctx context.Context, bucket, key string) {
	if bucket == "" || key == "" {
		return
	}
	ctx, cancel := context.WithTimeout(ctx, time.Minute)
	defer cancel()
	...
```

A cap under the 2-minute tick also guarantees uploads never overlap, and it bounds the deferred final snapshot so a bad S3 day cannot hold up the `rescheduleShutdown` defer that runs after it.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -30,6 +30,16 @@ var benchDenylist = []string{
 	"BenchmarkTransactionFetch", // is broken in current baseline release
 }
 
+var nativeLibRoots = []string{
```
tamirms, line 33: Nothing verifies that the kept set actually builds on the box, so if a package outside this roster grows a native-library dependency later (say a cgo wrapper under shared `internal/`), the failure only surfaces after both full legs as a candidate bench failure, which reads as a perf regression in the PR under test rather than a roster gap. `go list` does not catch it: it exits 0 for a cgo package whose native library is missing, so `excludeFromBench` keeps it (I checked).

A cheap guard would be a `go build` preflight over the kept packages in `prepareSuites` that bails immediately with a message pointing at this list.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
tamirms, line 261: Building on this: since the reason still only lives in this goroutine's log line, `servePhase`'s `serving daemon exited` warning cannot say why, and a normal teardown (`Stop`'s context cancel delivers `os.Kill`) logs `daemon exited: signal: killed`, which is byte-for-byte what a kernel OOM kill looks like in the box log.

What do you think about storing `cmd.Wait()`'s error in a `daemonHandle` field before `close(done)`? Then the serve loop can log the reason when it notices the exit, and `Stop`, which knows it initiated the kill, can label the expected case so backfill-only runs do not end on an alarming kill line.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
```diff
@@ -5,46 +5,70 @@ import (
...
 	cfg, err := env.ParseAs[serveEnv]()
 	if err != nil {
 		logger.Warnf("parsing serve env: %v", err)
 		cfg.Ceiling = 6 * time.Hour
 	}
 
-	// the boot-time self-terminate (budget+15m) can predate the blast's end
-	rescheduleShutdown(ctx, int(cfg.Ceiling.Minutes()), "serve ceiling")
+	defer rescheduleShutdown(ctx, 1, "serve phase over")
+	defer snapshotBoxLog(context.WithoutCancel(ctx), cfg.Bucket, cfg.ResultKey)
```
tamirms, line 34: The teardown paths treat `ctx` inconsistently: this defer guards with `WithoutCancel`, but the sibling `rescheduleShutdown` defer and `dumpDmesgTail` keep the cancelable `ctx`. Today they are equivalent because `harness.Run` passes `context.Background` (which also makes the `ctx.Done` arm below dead code), but if signal handling ever lands, the unguarded pair breaks silently: `servePhase` returns via `ctx.Done`, the `shutdown -c` / `shutdown -P +1` execs fail with `context.Canceled`, and the box lingers until the stale ceiling shutdown instead of powering off a minute later.

I would apply `WithoutCancel` to all three teardown paths, or to none and drop the `ctx.Done` arm, so the model is consistent either way.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
```diff
@@ -5,46 +5,70 @@ import (
...
+const (
+	rpcPort               = "8000"
+	serveSnapshotInterval = 2 * time.Minute
+)
 
-// serveEnv is the serve phase's env-derived config.
 type serveEnv struct {
 	Ceiling   time.Duration `env:"SERVE_CEILING" envDefault:"6h"`
-	Bucket    string        `env:"BUCKET"`
+	Bucket    string        `env:"BUCKET"        envDefault:"stellar-rpc-ci-load-test"`
```
tamirms, line 21: `servePhase` re-parses `BUCKET`/`RESULT_KEY` that this same process already parsed into `backfillEnv`, and this `envDefault` adds the fifth copy of the `stellar-rpc-ci-load-test` literal in the tree (bootstrap-common.sh, harness.go, leg.go, backfill instantiate.go, and here). Since `cfg` is in scope at the `servePhase` call site, would it be simpler to pass `Bucket`/`ResultKey` in and shrink `serveEnv` to just `SERVE_CEILING`?

The failure mode this avoids is quiet: a future bucket or key change updated in `backfillEnv` but not here sends the serving-phase snapshots to the wrong place, and since `snapshotBoxLog` only warns, the run stays green while the artifact this PR exists to preserve stops uploading.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
```diff
@@ -193,6 +218,75 @@ func runSuite(ctx context.Context, dir, outFile string, count int) []string {
...
+	cmd.Stdout, cmd.Stderr = &stdout, &stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("go list -test in %s: %w (%s)", dir, err, strings.TrimSpace(stderr.String()))
+	}
+
+	all, excluded := map[string]bool{}, map[string]bool{}
+	decoder := json.NewDecoder(&stdout)
+	for {
+		var listed struct {
+			ImportPath string   `json:"importPath"`
```
tamirms, line 236: These tags are lowerCamel while `go list -json` emits `ImportPath`/`ForTest`/`Deps`, so the decode only works through encoding/json's case-insensitive fallback. Dropping the tags does the right thing (the exported field names already match the keys exactly) and keeps this from silently breaking under case-sensitive matching, for example an encoding/json/v2 migration.

tamirms (commented):
Some notes from a pass over the serve-phase runner and the go-bench changes.

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
cjonas9: Wow, this ran far deeper than I'd known. Thanks so much for the explanation, this was an amazing catch!!

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
cjonas9: On second thought, considering what you're saying, I think this is ok. With that said, considering this is a "fix perf eval" PR, I do think a tiny bit more discussion is appropriate. But I will preface that I don't have any substantial objections to leaving this as you've written it.

We're trying to fix the case that exists because the box is hard-killed and logs go missing, but we still are open to a missing log tail, which is inherent to the design here. Increasing/decreasing the cadence has this tradeoff in either direction. But we could get around this by:

GHA side:
- stop leading by hard-killing the serving box with `terminate-instances`
- have GHA send a graceful reap over SSM (e.g. creating a sentinel file on the box, or `send-command` using the `ssmRunner`) + wait a few minutes for the box to shut itself down
    - hard terminate as fallback

On the box:
- runner handles the reap signal (e.g. `SIGTERM` via `signal.NotifyContext`)
    - stop the daemon, upload the complete serve-phase log to S3, and return so the normal `shutdown -P` exit path runs

This is more deterministic and replaces the 2-minute log flush, and because the reap lives in the same step that does the kill, there's no race between uploading and getting terminated by GHA. I think this design ultimately "solves" the problem and could be pretty compact without involving CloudWatch, so I wanted to offer it, but if you'd like to call all this good enough as-is, LGTM.

marwen-abid: Just to be clear, this is a `fix perf eval **on feature/full-history**` PR.  Most of these issues were surfaced during testing of the port to feature/full-history.  

I like your suggested design (a lot). A graceful reap over SSM with terminate-instances as the fallback is the deterministic version of what the 2-minute snapshot tries to approximate and removes the missing-tail tradoff. 

The scope of this PR is already getting bigger - so I'd rather address this in another issue. I created https://github.com/stellar/stellar-rpc/issues/956 for that. Feel free to edit it and I'll tackle that after I'm out of the critical path.

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
marwen-abid: addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

Also added a change to keep draining the pipe with `io.Copy` until the daemon exits, to prevent the read from getting closed under a healthy daemon.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
marwen-abid: I don't think a `go build` preflight catches what we want because itskips `_test.go`, and that is where the benches and their imports live. The check that actually catches it is compiling + linking the test binaries, which is the first thing `go test` does in `runSuite`

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/instantiate.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

One `teardown := context.WithoutCancel(ctx)` now feeds all three defers.

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/backfill-test/runner/serve.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/perf-eval/go-bench/runner/instantiate.go
marwen-abid: Addressed in a003f3bb3df8c257378e337090532ff5814dfea9

tamirms (approved):

PR #939, opened 2026-08-21
rpcv2: the getEvents term budget knob, and the EventScope rename

urvisavla:
## What

The naming follow-up #934 deferred, plus the first half of the getEvents v2 term budget (#774).

1. **Rename**: `EventCursorQuery` becomes `EventScope`, the name the `Scope` field already implies. Stale `PR-3c` and "query coordinator" comments now name the events pager and `Matches`.
2. **Term budget inputs**: `term_budget` on `[service.methods.getEvents]`, default 15 per the proposal, and `event.CountDistinctTerms`.


tamirms (approved):
LGTM. One forward note for the enforcement half: since cursors are decodable by any node and are not authenticated, the budget check will need to run on cursor resumes too, not just fresh requests, otherwise a hand-built cursor bypasses the budget entirely. The proposal's split guidance already covers a client that gets rejected mid-pagination after a budget drop, so rejecting there is fine.

PR #941, opened 2026-08-21
Use GetLedgerRaw in getLatestLedger endpoint

### LINKED ISSUE #731: `getLatestLedger`: apply the same raw-bytes optimization that `getLedgers` already uses
## Summary

`getLedgers` (via `BatchGetLedgers` in `cmd/stellar-rpc/internal/db/ledger.go`) avoids unmarshaling the full `LedgerCloseMeta` from the SQLite BLOB. It selects directly into `[][]byte`, keeps the raw bytes for base64 encoding, and parses only the small header prefix (V discriminator + optional `LedgerCloseMetaExt` + `LedgerHeaderHistoryEntry`). This bounds per-ledger XDR work to ~200 bytes regardless of the LCM size.

`getLatestLedger` doesn't follow this pattern. After #554 added the full `LedgerMetadata` field to its response, the handler now:

1. Calls `GetLedger` → SQL scan into `[]xdr.LedgerCloseMeta`, which triggers a full XDR Unmarshal of the entire BLOB (~2.8 MB on pubnet) into a deep struct tree.
2. Calls `latestLedger.MarshalBinary()` → re-serializes that same struct tree back to bytes.
3. Base64-encodes the result.

Steps 1 and 2 are wasted work — XDR encoding is canonical, so the bytes from the DB are byte-identical to the bytes we send in the response. The header fields we actually need (`Hash`, `ProtocolVersion`, `LedgerCloseTime`, the marshaled `LedgerHeader`) all live in the first ~200 bytes of the serialization.

## Proposed fix

Apply the same raw-bytes pattern. Likely shape:

- Add a new `GetLedgerRaw(ctx, sequence) ([]byte, xdr.LedgerHeaderHistoryEntry, bool, error)` method to `LedgerReader` — returns the raw BLOB plus the parsed header.
- Factor the header-parsing logic out of `BatchGetLedgers` into a shared helper (e.g., `parseLedgerHeaderFromMeta`) so both `BatchGetLedgers` and `GetLedgerRaw` use the same parsing.
- Update the `getLatestLedger` handler to call `GetLedgerRaw`, derive the response fields from the parsed header, and base64-encode the raw bytes directly without re-marshaling.

cjonas9:
### What

Uses the new raw-bytes patterning used in `getLedger` in `getLatestLedger`:
- Add `GetLedgerRaw(ctx, sequence) ([]byte, xdr.LedgerHeaderHistoryEntry, bool, error)` to `LedgerReader`
    - Returns raw bytes of ledger meta, the header, and whether it was found/if the request error'd
- Updates the `getLatestLedger` handler to call `GetLedgerRaw`
- Updates mocks/tests with `GetLedgerRaw`

### Why

`getLatestLedger` currently wastes time calling GetLedgers to get a single `LedgerCloseMeta`, which causes an `Unmarshal` to XDR. The handler then marshals this XDR to binary. On each call, we pay for a ~2.8Mb `Unmarshal` and `Marshal` that can be completely bypassed.

### Known limitations

N/A

### DIFF
--- a/cmd/stellar-rpc/internal/db/ledger.go
+++ b/cmd/stellar-rpc/internal/db/ledger.go
@@ -91,27 +91,35 @@ func (l ledgerReaderTx) BatchGetLedgers(
 
 	batch := make([]LedgerMetadataChunk, len(results))
 	for i, meta := range results {
-		batch[i] = LedgerMetadataChunk{Lcm: meta}
-
-		var v xdr.Int32
-		rd := bytes.NewReader(meta)
-		if _, err := xdr.Unmarshal(rd, &v); err != nil {
+		header, err := l.parseLedgerHeaderFromMeta(meta)
+		if err != nil {
 			return nil, err
 		}
-
-		if v > 0 { // V0 has no extension
-			var ext xdr.LedgerCloseMetaExt
-			if _, err := xdr.Unmarshal(rd, &ext); err != nil { // skipped
-				return nil, err
-			}
-		}
-
-		if _, err := xdr.Unmarshal(rd, &batch[i].Header); err != nil {
-			return nil, err
-		}
+		batch[i] = LedgerMetadataChunk{Header: header, Lcm: meta}
 	}
 
 	return batch, nil
+}
+
+func (l ledgerReaderTx) parseLedgerHeaderFromMeta(meta []byte) (xdr.LedgerHeaderHistoryEntry, error) {
+	var v xdr.Int32
+	rd := bytes.NewReader(meta)
+	if _, err := xdr.Unmarshal(rd, &v); err != nil {
+		return xdr.LedgerHeaderHistoryEntry{}, err
+	}
+
+	if v > 0 { // V0 has no extension
+		var ext xdr.LedgerCloseMetaExt
+		if _, err := xdr.Unmarshal(rd, &ext); err != nil { // skipped
+			return xdr.LedgerHeaderHistoryEntry{}, err
+		}
+	}
+
+	var header xdr.LedgerHeaderHistoryEntry
+	if _, err := xdr.Unmarshal(rd, &header); err != nil {
+		return xdr.LedgerHeaderHistoryEntry{}, err
+	}
+	return header, nil
 }
 
 // GetLedger fetches a single ledger from the db using a transaction.
--- a/cmd/stellar-rpc/internal/db/ledger.go
+++ b/cmd/stellar-rpc/internal/db/ledger.go
@@ -45,7 +45,7 @@ type LedgerWriter interface {
 }
 
 type readDB interface {
-	Select(ctx context.Context, dest interface{}, query sq.Sqlizer) error
+	Select(ctx context.Context, dest any, query sq.Sqlizer) error
 }
 
 type ledgerReader struct {
@@ -397,24 +397,23 @@ func getLedgerRawFromDB(ctx context.Context, db readDB, sequence uint32) ([]byte
 
 // Parses a raw ledger close meta blob into a LedgerHeaderHistoryEntry.
 func parseLedgerHeaderFromMeta(meta []byte) (xdr.LedgerHeaderHistoryEntry, error) {
-	var v xdr.Int32
-	rd := bytes.NewReader(meta)
-	if _, err := xdr.Unmarshal(rd, &v); err != nil {
-		return xdr.LedgerHeaderHistoryEntry{}, err
-	}
-
-	if v > 0 { // V0 has no extension
-		var ext xdr.LedgerCloseMetaExt
-		if _, err := xdr.Unmarshal(rd, &ext); err != nil { // skipped
-			return xdr.LedgerHeaderHistoryEntry{}, err
+	raw, err := xdr.Try(func() []byte {
+		m := xdr.LedgerCloseMetaView(meta)
+		switch m.MustV() {
+		case 0:
+			return m.MustV0().MustLedgerHeader().MustRaw()
+		case 1:
+			return m.MustV1().MustLedgerHeader().MustRaw()
+		default:
+			return m.MustV2().MustLedgerHeader().MustRaw()
 		}
-	}
-
+	})
 	var header xdr.LedgerHeaderHistoryEntry
-	if _, err := xdr.Unmarshal(rd, &header); err != nil {
-		return xdr.LedgerHeaderHistoryEntry{}, err
+	if err != nil {
+		return header, err
 	}
-	return header, nil
+	err = header.UnmarshalBinary(raw)
+	return header, err
 }
 
 // InsertLedger inserts a ledger in the db.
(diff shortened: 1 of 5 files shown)


cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -362,20 +366,59 @@ func (l ledgerWriter) trimLedgers(latestLedgerSeq uint32, retentionWindow uint32
...
-	sql := sq.Select("meta").From(ledgerCloseMetaTableName).Where(sq.Eq{"sequence": sequence})
-	var results []xdr.LedgerCloseMeta
-	if err := db.Select(ctx, &results, sql); err != nil {
+	raw, found, err := getLedgerRawFromDB(ctx, db, sequence)
+	if err != nil {
+		return xdr.LedgerCloseMeta{}, false, err
+	}
+	if !found {
+		return xdr.LedgerCloseMeta{}, false, nil
+	}
```
Shaptic:
```suggestion
	if err != nil || !found {
		return xdr.LedgerCloseMeta{}, false, err
	}
```
:wink:

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -362,20 +366,59 @@ func (l ledgerWriter) trimLedgers(latestLedgerSeq uint32, retentionWindow uint32
...
+
+	var lcm []xdr.LedgerCloseMeta
+	if _, err := xdr.Unmarshal(bytes.NewReader(raw), &lcm); err != nil {
```
Shaptic, line 390: Since we're not using `Select` anymore, we can just use the normal way of decoding LCM rather than this length-one check.

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -362,20 +366,59 @@ func (l ledgerWriter) trimLedgers(latestLedgerSeq uint32, retentionWindow uint32
...
+
+// getLedgerRawFromDB is a helper function that encapsulates the common logic
+// for fetching a single ledger's bytes from the database
+func getLedgerRawFromDB(ctx context.Context, db readDB, sequence uint32) ([]byte, bool, error) {
+	sql := sq.Select("meta").From(ledgerCloseMetaTableName).Where(sq.Eq{"sequence": sequence})
+	var result []byte
+	if err := db.Select(ctx, &result, sql); err != nil || len(result) == 0 {
+		return nil, false, err
+	}
+	return result, true, nil
```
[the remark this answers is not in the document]
Shaptic: This sounds right actually

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -362,20 +366,59 @@ func (l ledgerWriter) trimLedgers(latestLedgerSeq uint32, retentionWindow uint32
...
+}
+
+// Parses a raw ledger close meta blob into a LedgerHeaderHistoryEntry.
+func parseLedgerHeaderFromMeta(meta []byte) (xdr.LedgerHeaderHistoryEntry, error) {
+	var v xdr.Int32
```
Shaptic: Sir this is where you're supposed to use views

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -24,6 +24,7 @@ type StreamLedgerFn func(xdr.LedgerCloseMeta) error
 
 type LedgerReader interface {
 	GetLedger(ctx context.Context, sequence uint32) (xdr.LedgerCloseMeta, bool, error)
+	GetLedgerRaw(ctx context.Context, sequence uint32) ([]byte, xdr.LedgerHeaderHistoryEntry, bool, error)
```
Shaptic: I'm confused.. this doesn't seem to be.. used anywhere?

Shaptic (approved):

cmd/stellar-rpc/internal/db/ledger.go
cjonas9: `internal/methods/get_latest_ledger.go:L27`, no?

cmd/stellar-rpc/internal/db/ledger.go
cjonas9: No totally, I botched this on my first try. We need the length check after the `db.Select`, and we can't `db.Select` into a `[]byte`, it's gotta be into a `[][]byte`

cmd/stellar-rpc/internal/db/ledger.go
cjonas9: I was stubborn about using Claude for this PR since it's a little one, and it shows. 😔

cmd/stellar-rpc/internal/db/ledger.go
Shaptic: 🤦 yep, missed that diff. Only noticed the implementation and was like 🤔

Shaptic: Wait I feel like extracting the header shouldn't be part of this. The return signature should be the same as `GetLedger`, except `byte`ified. If someone wants the header they can use `parseLedgerHeaderFromMeta` like you already do.

cjonas9: Yes, good thinking. Was thinking about this a bit too hard and still wanted your thoughts on something that came up: this function _barely_ belongs as a method on `LedgerReader`, but still feels like it belongs as a method on an object. I think that if we wanted to engineer this further, we could even add a defined slice type for []byte that provides `ParseLedgerHeaderFromMeta` as a method. i.e., 
```
type RawLedger []byte
func (r RawLedger) ParseLedgerHeader...
// whatever other methods end up being handy
```

Thoughts? I'm certainly overthinking it, but think this deserves a little thought

UPDATE: If you'd like to see how this looks vs. not using a `RawLedger` name over `[]byte`, compare commit `fe59a393` with commit `78cd042d`. It's not much work, so I'm happy to go either way!

cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -253,6 +256,27 @@ func (r ledgerReader) GetLatestLedgerSequence(ctx context.Context) (uint32, erro
 	return getLatestLedgerSequence(ctx, r, r.db.cache)
 }
 
+// Parses a raw ledger close meta blob into a LedgerHeaderHistoryEntry.
+func (r RawLedger) ParseLedgerHeaderFromMeta() (xdr.LedgerHeaderHistoryEntry, error) {
```
Shaptic, line 260: This should for sure go into the SDK at some point, but nice workaround here to add it as a method 👍

Shaptic (approved):

cmd/stellar-rpc/internal/db/ledger.go
Shaptic: I like it, picasso

cmd/stellar-rpc/internal/db/ledger.go
Shaptic: Oh wait we wouldn't able to do that because it's a `[]byte` underneath :facepalm:

cjonas9: Yeah, if this wasn't so skinny, I'd totally agree. I think it'll probably wind up being convenient again to have an alias to receive methods that would be otherwise awkward to place. Plus it's pretty cool how Go doesn't complain about using it interchangeably with a `[]byte`, I was not expecting that.

Shaptic: We'll definitely add to these in the future for other PRs so it's fine to use here tbh

PR #943, opened 2026-08-24
xdr2json: add the JSON-to-XDR direction

### LINKED ISSUE #940: xdr2json: add the JSON-to-XDR direction for getEvents v2 xdrInputFormat
The getEvents v2 request carries `xdrInputFormat`. The `json` value selects the JSON encoding for topic values inside filters. We cannot serve it today.

The term index keys each topic term on the canonical XDR byte encoding of the `ScVal`, so a JSON topic value must convert to those exact bytes. The repository has no converter for that direction. `internal/xdr2json` is cgo over a Rust staticlib that exposes only `xdr_to_json`, and the Go SDK's `xdr.ScVal` has no `UnmarshalJSON`.

Requirement: a JSON-to-XDR converter for `ScVal`, so getEvents v2 accepts `xdrInputFormat: "json"`.

Parent issue: #774
Spec: https://github.com/orgs/stellar/discussions/1872

tamirms:
getEvents v2's `xdrInputFormat: "json"` needs topic ScVals converted from their JSON encoding to canonical XDR bytes, and the repository had no converter for that direction. Closes #940.

The pinned stellar-xdr crate already ships the reverse parser, `Type::from_json`, behind its `serde_json` feature. `Type` is serde(untagged), so the JSON that `xdr_to_json` emits is exactly the bare inner-value JSON that `Type::from_json` parses: the emitter and the parser stay in lockstep by construction, which is the round-trip property users need when they paste values from json-formatted responses back into filters.

- enable the stellar-xdr `serde_json` feature
- add a `json_to_xdr` FFI entry point mirroring `xdr_to_json`; its failure paths return errors instead of panicking, since it parses user-supplied values
- add private `vec_to_c_xdr`/`free_c_xdr` byte-buffer helpers (mirroring preflight's, which stay untouched), and generalize the panic catcher over both directions
- expose `ConvertJSON` from the Go package, symmetric to `ConvertBytes`, with round-trip coverage across the ScVal variants, a pinned-dialect golden test, and error cases

Two input bounds are documented and pinned by tests: inputs over 32 MiB are rejected before parsing, and serde_json's recursion limit caps container nesting at 63 levels (the read direction allows 500), so the deepest protocol-legal values only convert from their base64 form; a too-deep filter value fails with a clear error for the handler to map to `invalid_params`.

### DIFF
--- a/cmd/stellar-rpc/internal/xdr2json/conversion.go
+++ b/cmd/stellar-rpc/internal/xdr2json/conversion.go
@@ -58,6 +58,10 @@ func ConvertInterface(xdr encoding.BinaryMarshaler) (json.RawMessage, error) {
 	return convertAnyBytes(xdrTypeName, data)
 }
 
+// maxJSONInputLen mirrors DEFAULT_XDR_RW_LIMITS.len in
+// lib/xdr2json/src/lib.rs, which stays authoritative for direct FFI callers.
+const maxJSONInputLen = 32 * 1024 * 1024
+
 // ConvertJSON is the inverse of ConvertBytes: it takes an XDR object (`xdr`,
 // used only to determine the name of the structure, as in ConvertBytes) and
 // the JSON serialization of a value of that type, in the encoding
@@ -71,6 +75,13 @@ func ConvertJSON(xdr any, js json.RawMessage) ([]byte, error) {
 	// zero-length allocation, whose pointer may be null.
 	if len(js) == 0 {
 		return nil, errors.New("JSON input is empty")
+	}
+
+	// Reject oversized input before C.CBytes copies it into C memory, so an
+	// over-limit value cannot force an unbounded native allocation.
+	if len(js) > maxJSONInputLen {
+		return nil, errors.Errorf(
+			"JSON input is %d bytes, over the %d-byte limit", len(js), maxJSONInputLen)
 	}
 
 	xdrTypeName := reflect.TypeOf(xdr).Name()
--- a/cmd/stellar-rpc/internal/xdr2json/conversion.go
+++ b/cmd/stellar-rpc/internal/xdr2json/conversion.go
@@ -58,6 +58,39 @@ func ConvertInterface(xdr encoding.BinaryMarshaler) (json.RawMessage, error) {
 	return convertAnyBytes(xdrTypeName, data)
 }
 
+// ConvertJSON is the inverse of ConvertBytes: it takes an XDR object (`xdr`,
+// used only to determine the name of the structure, as in ConvertBytes) and
+// the JSON serialization of a value of that type, in the encoding
+// ConvertBytes produces, and returns the value's XDR byte encoding. Unlike
+// ConvertBytes, empty input is an error: it is not valid JSON. Inputs over
+// 32 MiB are rejected, and serde_json's recursion limit caps container
+// nesting at 63 levels, below the 500 the read direction allows, so the
+// deepest protocol-legal values only convert from their base64 form.
+func ConvertJSON(xdr any, js json.RawMessage) ([]byte, error) {
+	// Rejecting empty input here also keeps C.CBytes below from making a
+	// zero-length allocation, whose pointer may be null.
+	if len(js) == 0 {
+		return nil, errors.New("JSON input is empty")
+	}
+
+	xdrTypeName := reflect.TypeOf(xdr).Name()
+
+	goRawJSON := CXDR(js)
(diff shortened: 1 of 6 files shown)


cmd/stellar-rpc/internal/xdr2json/conversion.go
```diff
@@ -58,6 +58,50 @@ func ConvertInterface(xdr encoding.BinaryMarshaler) (json.RawMessage, error) {
 	return convertAnyBytes(xdrTypeName, data)
 }
 
+// maxJSONInputLen mirrors DEFAULT_XDR_RW_LIMITS.len in
+// lib/xdr2json/src/lib.rs, which stays authoritative for direct FFI callers.
+const maxJSONInputLen = 32 * 1024 * 1024
```
Shaptic, line 63: I get that this mirrors a library constant, but do we really want to process 32MB HTTP requests? Or maybe an earlier limit will be hit far before this and this is just defensive in which case that’s fine.

cmd/stellar-rpc/internal/xdr2json/conversion.go
```diff
@@ -58,6 +58,50 @@ func ConvertInterface(xdr encoding.BinaryMarshaler) (json.RawMessage, error) {
...
+	defer C.free(unsafe.Pointer(b))
+
+	result := C.json_to_xdr(b, goRawJSON)
+	defer C.free_json_to_xdr_result(result)
+
+	if errStr := C.GoString(result.error); errStr != "" {
+		return nil, errors.New(errStr)
+	}
+
+	return C.GoBytes(unsafe.Pointer(result.xdr.xdr), C.int(result.xdr.len)), nil
```
Shaptic, line 102: We defer a free to result, but xdr is still valid, right? It’s weird to see access to a “will be freed” structure here, maybe it makes sense to reassign it first before returning to make the separation explicit?

cmd/stellar-rpc/lib/xdr2json/src/lib.rs
```diff
@@ -108,34 +112,135 @@ pub unsafe extern "C" fn free_conversion_result(ptr: *mut ConversionResult) {
...
-                error: "xdr_to_json() failed: unknown cause".to_string(),
-            },
-        },
+        Err(panic) => {
+            // Payloads are String from format-style panics and &str from
+            // literal ones (e.g. assert! in dependencies).
+            let msg = panic
+                .downcast_ref::<String>()
+                .map(String::as_str)
+                .or_else(|| panic.downcast_ref::<&'static str>().copied())
```
Shaptic, line 239: Is this just defensive? I’m confused why it wasn’t required previously.

Shaptic (commented):
I’m generally skeptical of the idea that we should allow such complex user-supplied inputs so I need to do another pass on this but here’s a few comments in the meantime.

cmd/stellar-rpc/internal/xdr2json/conversion.go
tamirms: we already cap whole request bodies at 512 kb via http.MaxBytesHandler so this limit should never be exercised via an rpc request. it's more of a defensive sanity check

cmd/stellar-rpc/internal/xdr2json/conversion.go
tamirms: C.GoBytes allocates a go managed byte slice and copies the c bytes into it so that's why it is safe. the pattern is similar to convertAnyBytes() below:

```go
	defer C.free_conversion_result(result)

	jsonStr = C.GoString(result.json)
	errStr = C.GoString(result.error)
```

cmd/stellar-rpc/lib/xdr2json/src/lib.rs
tamirms: It was a gap before too: the old catcher only downcast String payloads, so a dependency panicking with a string literal (an assert! without a message, for example) surfaced as "unknown cause". Our own panic sites all use format-style panics, which carry String, so this arm only matters for panics inside stellar-xdr or serde internals.

cmd/stellar-rpc/internal/xdr2json/conversion.go
Shaptic: Ah yeah, that makes sense - I didn't realize it was a copy.

Shaptic (approved):
Just gonna call out once more that I do think we should discuss whether or not this feature is worth significantly broadening the security surface.

Issue #732, opened 2026-05-06
Epic: Adopt zero-copy XDR views and GetLedgerRaw for performance


tamirms:
## Summary

The new XDR view types (`xdr.LedgerCloseMetaView`, `xdr.LedgerEntryView`, `xdr.TransactionEnvelopeView`, etc.) and the `LedgerBackend.GetLedgerRaw` method coming from [go-stellar-sdk](https://github.com/stellar/go-stellar-sdk) (PR [#5941](https://github.com/stellar/go-stellar-sdk/pull/5941)) enable zero-copy partial decoding of XDR data. stellar-rpc has multiple hot paths that currently:

1. Decode entire XDR structures via SQL `Scanner` auto-decode, then read only a few fields
2. Decode XDR structures and immediately re-marshal sub-fields back to bytes for response/storage
3. Walk full transaction batches when only metadata or a single transaction is needed

This issue tracks the work to migrate those paths to view-based access.

### Constraint

The public JSON-RPC API contract is preserved exactly. All changes are internal Go interfaces, types, and implementations.

### Prerequisite

This work depends on go-stellar-sdk PR [#5941](https://github.com/stellar/go-stellar-sdk/pull/5941) merging. That PR generates view types for every XDR struct (the wrapper types are `[]byte`-based with field accessors). Convenience helpers on those views (e.g., `LedgerCloseMetaView.LedgerCloseTime()`, `LedgerEntryView.LedgerKey()`) will be added per-task as needed.

For ingest-pipeline work (Task 7), an additional SDK addition is needed: a view-based replacement for `ingest.NewLedgerTransactionReaderFromLedgerCloseMeta` that operates on raw bytes / a `LedgerCloseMetaView` and yields per-transaction views without full decode.

### Measured baseline

In the BSB throughput benchmark from the SDK PR (1000 ledgers from GCS, NumWorkers=100):
- `GetLedger` (full decode): 167 lps
- `GetLedgerRaw` (decode skipped): 198 lps
- **~19% throughput improvement** when LCM decode is eliminated

This is the unit cost of the LCM decode (~1ms per typical pubnet ledger). Per-call savings on stellar-rpc paths range from ~1ms (single-LCM read) to ~5ms (`getTransactionByHash` with re-encode round trip).

---

### Task 1: Migrate DB ledger read path to bytes / view

[...]

PR #945, opened 2026-08-24
Migrate DB ledger read path to raw ledger bytes/view

### LINKED ISSUE #947: Adopt XDR views: Migrate DB ledger read path to bytes / view
Parent: #732 

### What

Migrate RPC's DB ledger path to read bytes rather than decoding LCM by default. Most of this work will occur in `cmd/stellar-rpc/internal/db/ledger.go`, though callers may also need to be adjusted accordingly. This includes, but is not limited to, using `LedgerCloseMetaView`s in: 
- `getLedgerRange` / `getLedgerRangeWithCache` / `getLedgerRangeWithoutCache` 
- `SteamAllLedgers` / `StreamLedgerRange`

### Why

Every `get_latest_ledger` request, every `get_health` poll, every `get_fee_stats` call, every range query, and every backfill row pays a full LCM decode for a few-field read. `get_latest_ledger` additionally pays two re-encodes on top. This is all because every `db.Select(... []xdr.LedgerCloseMeta ...)` triggers `xdr.LedgerCloseMeta.Scan` in the SQL driver layer despite only a subset of the data decoded being needed.

cjonas9:
### What

This PR covers some lingering ledger read path migrations to use ledger bytes/views where applicable. This has the immediate affect of optimizing `getHealth` and `getFeeStats`, though no changes to these handlers was required. Here are the functions modified:
- `getLedgerRange`/`getLedgerRangeWithCache`/`getLedgerRangeWithoutCache`

An adjacent optimization to the `getLedgerRange*` functions was also made alongside these changes. In particular, these functions return only sequence numbers, but fetch entire LCM blobs (>1Mb). This is on the hot-path. To mitigate this, now, this behavior is a fallback -- we fetch the first 1Kb _prefix_ of an LCM and use a view over it to fetch our sequences. This saves us some heavier fetching work and is relevant here because we no longer need to decode anything but the data that's relevant to the function. :-)

`getLatestLedger` is also migrated. Instead of decoding the entire LCM and re-marshaling its pieces, the handler reads the header fields and base64 payloads straight off the raw bytes through view accessors. The raw read is exposed as a loan on the serving interface with the following:

```go
WithLedgerRaw(ctx, sequence, fn) (found bool, err error)   // on store.LedgerReader
```

`fn` borrows the ledger's marshaled LCM for the duration of the call: the v1 (sqlite) backend lends the row blob, and the v2 backend lends the routed point read's buffer via `query.ReadView.WithLedger` (#961) with no copy. 

### Why

See #732. This is a component of adopting ledger views in RPC for query-path performance purposes.

### Known limitations

N/A

### DIFF
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/hex"
-	"fmt"
 
 	"github.com/creachadair/jrpc2"
 
@@ -36,7 +35,7 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
 		if err != nil {
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: fmt.Sprintf("could not read latest ledger meta: %v", err),
+				Message: "could not parse latest ledger header",
 			}
 		}
 		return response, nil
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -25,21 +25,19 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
 			}
 		}
 		var response protocol.GetLatestLedgerResponse
+		var parseErr error
 		found, err := ledgerReader.WithLedgerRaw(ctx, latestSequence, func(raw []byte) error {
-			var lerr error
-			response, lerr = latestLedgerResponse(xdr.LedgerCloseMetaView(raw), latestSequence)
-			return lerr
+			response, parseErr = latestLedgerResponse(xdr.LedgerCloseMetaView(raw), latestSequence)
+			return parseErr
 		})
-		if err != nil && found {
-			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: "could not parse latest ledger header",
-			}
-		}
 		if err != nil || !found {
+			msg := "could not get latest ledger"
+			if parseErr != nil {
+				msg = "could not parse latest ledger header"
+			}
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: "could not get latest ledger",
+				Message: msg,
 			}
 		}
 		return response, nil
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/hex"
+	"fmt"
 
 	"github.com/creachadair/jrpc2"
 
@@ -31,9 +32,14 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
 			return parseErr
 		})
 		if err != nil || !found {
-			msg := "could not get latest ledger"
-			if parseErr != nil {
-				msg = "could not parse latest ledger header"
+			var msg string
+			switch {
+			case parseErr != nil:
+				msg = fmt.Sprintf("could not parse latest ledger header: %v", parseErr)
+			case err != nil:
+				msg = fmt.Sprintf("could not get latest ledger: %v", err)
+			default: // clean miss: no underlying error to report
+				msg = "could not get latest ledger"
 			}
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -50,8 +50,6 @@ func latestLedgerResponse(view xdr.LedgerCloseMetaView, sequence uint32,
 	if err != nil {
 		return protocol.GetLatestLedgerResponse{}, err
 	}
-	// Must* accessors panic with *ViewError on malformed fields and Try
-	// recovers it, so the extraction needs a single error check.
 	return xdr.Try(func() protocol.GetLatestLedgerResponse {
 		header := headerEntry.MustHeader()
 		return protocol.GetLatestLedgerResponse{
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -3,11 +3,13 @@ package methods
 import (
 	"context"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/creachadair/jrpc2"
 
 	protocol "github.com/stellar/go-stellar-sdk/protocols/rpc"
+	"github.com/stellar/go-stellar-sdk/xdr"
 
 	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/store"
 )
@@ -23,39 +25,43 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
 				Message: "could not get latest ledger sequence",
 			}
 		}
-		latestLedger, found, err := ledgerReader.GetLedger(ctx, latestSequence)
+		latestLedgerView, found, err := ledgerReader.GetLedgerView(ctx, latestSequence)
 		if (err != nil) || (!found) {
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
 				Message: "could not get latest ledger",
 			}
 		}
-		header := latestLedger.LedgerHeaderHistoryEntry().Header
-		headerBytes, err := header.MarshalBinary()
+		response, err := latestLedgerResponse(latestLedgerView, latestSequence)
 		if err != nil {
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: fmt.Sprintf("could not marshal latest ledger header: %v", err),
+				Message: fmt.Sprintf("could not read latest ledger meta: %v", err),
 			}
 		}
-		response := protocol.GetLatestLedgerResponse{
-			Hash:            latestLedger.LedgerHash().HexString(),
-			ProtocolVersion: latestLedger.ProtocolVersion(),
-			Sequence:        latestSequence,
-			LedgerCloseTime: latestLedger.LedgerCloseTime(),
-			LedgerHeader:    base64.StdEncoding.EncodeToString(headerBytes),
-		}
-
-		raw, err := latestLedger.MarshalBinary()
-		if err != nil {
-			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: fmt.Sprintf("could not marshal latest ledger metadata: %v", err),
-			}
-		}
-		response.LedgerMetadata = base64.StdEncoding.EncodeToString(raw)
-
 		return response, nil
 	}
 	return NewHandler(coreHandler)
+}
+
+// latestLedgerResponse extracts the response fields from a ledger close meta view.
+func latestLedgerResponse(view xdr.LedgerCloseMetaView, sequence uint32,
+) (protocol.GetLatestLedgerResponse, error) {
+	headerEntry, err := view.LedgerHeader()
+	if err != nil {
+		return protocol.GetLatestLedgerResponse{}, err
+	}
+	// Must* accessors panic with *ViewError on malformed fields and Try
+	// recovers it, so the extraction needs a single error check.
+	return xdr.Try(func() protocol.GetLatestLedgerResponse {
+		header := headerEntry.MustHeader()
+		return protocol.GetLatestLedgerResponse{
+			Hash:            hex.EncodeToString(headerEntry.MustHash().MustRaw()),
+			ProtocolVersion: header.MustLedgerVersion().MustValue(),
+			Sequence:        sequence,
+			LedgerCloseTime: int64(header.MustScpValue().MustCloseTime().MustValue()), //nolint:gosec // safe for ~292B years
+			LedgerHeader:    base64.StdEncoding.EncodeToString(header.MustRaw()),
+			LedgerMetadata:  base64.StdEncoding.EncodeToString(view),
+		}
+	})
 }
--- a/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
+++ b/cmd/stellar-rpc/internal/methods/get_latest_ledger.go
@@ -24,18 +24,22 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
 				Message: "could not get latest ledger sequence",
 			}
 		}
-		latestLedgerView, found, err := ledgerReader.GetLedgerView(ctx, latestSequence)
-		if (err != nil) || (!found) {
-			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: "could not get latest ledger",
-			}
-		}
-		response, err := latestLedgerResponse(latestLedgerView, latestSequence)
-		if err != nil {
+		var response protocol.GetLatestLedgerResponse
+		found, err := ledgerReader.WithLedgerRaw(ctx, latestSequence, func(raw []byte) error {
+			var lerr error
+			response, lerr = latestLedgerResponse(xdr.LedgerCloseMetaView(raw), latestSequence)
+			return lerr
+		})
+		if err != nil && found {
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
 				Message: "could not parse latest ledger header",
+			}
+		}
+		if err != nil || !found {
+			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
+				Code:    jrpc2.InternalError,
+				Message: "could not get latest ledger",
 			}
 		}
 		return response, nil
--- a/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
+++ b/cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
@@ -472,8 +472,11 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 		readTxMetaCtx,
 		ledgerSeqRange.First,
 		ledgerSeqRange.Last,
-		func(txMeta xdr.LedgerCloseMeta) error {
-			currentSeq = txMeta.LedgerSequence()
+		func(txMeta xdr.LedgerCloseMetaView) error {
+			currentSeq, err := txMeta.LedgerSequence()
+			if err != nil {
+				return err
+			}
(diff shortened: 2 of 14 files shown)


cmd/stellar-rpc/internal/db/ledger.go
```diff
@@ -256,51 +262,61 @@ func (r ledgerReader) GetLatestLedgerSequence(ctx context.Context) (uint32, erro
...
-		case 1:
-			return m.MustV1().MustLedgerHeader().MustRaw()
-		default:
-			return m.MustV2().MustLedgerHeader().MustRaw()
-		}
-	})
-	var header xdr.LedgerHeaderHistoryEntry
+// ledgerCloseTimePrefixBytes is the fast-path meta prefix fetched for range
+// endpoints. Parsing falls back to the full blob if the header extends past it.
+const ledgerCloseTimePrefixBytes = 1024
```
Shaptic, line 267: This seems like a micro-optimization that isn't worth the add'l double-fetch complexity. The bulk of the time should be on encodes and decodes rather than the DB fetches themselves, which we mitigate via views already. Can you justify that this is worthwhile via some numbers?

cjonas9: Yeah, good catch, I'll try to get some profiling on this later. 

I'd intended to essentially eliminate the double-fetch path and just left that in for safety. I had figured that it was some wasted effort to fetch the entire DB when almost assuredly, only a small number of prefix bytes of the LCM were needed. But that can be discovered with numbers, too.

Shaptic (commented):
This LGTM pending the Go benchmark side-by-side for the cool factor!

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
```diff
@@ -256,51 +262,61 @@ func (r ledgerReader) GetLatestLedgerSequence(ctx context.Context) (uint32, erro
...
-		case 1:
-			return m.MustV1().MustLedgerHeader().MustRaw()
-		default:
-			return m.MustV2().MustLedgerHeader().MustRaw()
-		}
-	})
-	var header xdr.LedgerHeaderHistoryEntry
+// ledgerCloseTimePrefixBytes is the fast-path meta prefix fetched for range
+// endpoints. Parsing falls back to the full blob if the header extends past it.
+const ledgerCloseTimePrefixBytes = 1024
```
cjonas9: Alright, happy to say I got some results. I committed the bench test I used for this. I'm uncertain of whether or not this is exactly what you had in mind, but I hope it suffices. It checks this for the min/avg/max ledger sizes (512Kb/2Mb/4Mb), and it's neat as you can see how the results scale with ledger sizes.
Tested on a c5.2xlarge on this branch. I added a human-friendly summary at the end of the benchmark:
```
go test ./cmd/stellar-rpc/internal/rpcv1/sqlitedb/ -run '^$' -bench BenchmarkOldestLedgerRangeLookup
goos: linux
goarch: amd64
pkg: github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv1/sqlitedb
cpu: Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
BenchmarkOldestLedgerRangeLookup/512KiB/prefix-8                   12426             95669 ns/op            7793 B/op         93 allocs/op
BenchmarkOldestLedgerRangeLookup/512KiB/fullBlob-8                  4064            295293 ns/op         1070335 B/op         87 allocs/op
BenchmarkOldestLedgerRangeLookup/2MiB/prefix-8                      1495            782332 ns/op            7788 B/op         94 allocs/op
BenchmarkOldestLedgerRangeLookup/2MiB/fullBlob-8                     505           2245511 ns/op         4216505 B/op         88 allocs/op
BenchmarkOldestLedgerRangeLookup/4MiB/prefix-8                       780           1502594 ns/op            7790 B/op         94 allocs/op
BenchmarkOldestLedgerRangeLookup/4MiB/fullBlob-8                     276           4249209 ns/op         8410865 B/op         88 allocs/op

prefix / fullBlob:
  512KiB:  0.32x ns/op | 0.007x B/op | 1.07x allocs/op
  2MiB:    0.35x ns/op | 0.002x B/op | 1.07x allocs/op
  4MiB:    0.35x ns/op | 0.001x B/op | 1.07x allocs/op
PASS
ok      github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv1/sqlitedb  7.332s
```

Shaptic: Oh neat approach to do the side-by-side! In the future, it'd be sufficient to just post the before-and-after text as a GH comment by just running the benchmark on `main` rather than over-complicating the reporting and re-implementing the old approach in the suite itself. Feel free to just do that and/or leave it as-is, your call!

Also you might need to add this to the ignores for the release eval suite since there's no "before"?

Shaptic (approved):

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -55,6 +55,21 @@ func (r *LedgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledg
 	return lcm, found, err
 }
 
+// GetLedgerView is GetLedger without the decode: the raw bytes come from the
+// store as a fresh, caller-owned buffer (see getLedgerRaw), so the view may
+// outlive the call.
+func (r *LedgerReader) GetLedgerView(ctx context.Context, sequence uint32) (xdr.LedgerCloseMetaView, bool, error) {
+	view, err := viewFrom(ctx)
```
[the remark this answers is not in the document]
Shaptic: based on CI I think bro is right

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
cjonas9: I was actually feeling strongly that this was a temp test, at least as I'd written it (especially because my release eval job does exactly this before/after automagically). But it did make for a theatric result :-). Absolutely I'll remove the comparison element and leave the rest, thanks for the look!

cmd/stellar-rpc/internal/rpcv1/daemon/daemon.go
```diff
@@ -482,7 +486,7 @@ func (d *Daemon) mustInitializeStorage(cfg *config.Config) *feewindow.FeeWindows
 					Debug("Still initializing in-memory store")
 			}
 
-			if err := dataMigrations.Apply(readTxMetaCtx, txMeta); err != nil {
+			if err := dataMigrations.Apply(readTxMetaCtx, currentSeq, txMeta); err != nil {
```
Shaptic: On second look, I actually think we should just pass the full LCM here since the migration needs the whole thing, anyway. Reduces the diff and makes it clear in the function signature that a full decode is required regardless.

cjonas9: Great catch, agreed completely, implemented, and the diff is way cleaner now

tamirms (approved):

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
```diff
@@ -23,39 +24,45 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
...
 		}
-		latestLedger, found, err := ledgerReader.GetLedger(ctx, latestSequence)
-		if (err != nil) || (!found) {
+		var response protocol.GetLatestLedgerResponse
+		found, err := ledgerReader.WithLedgerRaw(ctx, latestSequence, func(raw []byte) error {
+			var lerr error
+			response, lerr = latestLedgerResponse(xdr.LedgerCloseMetaView(raw), latestSequence)
+			return lerr
+		})
+		if err != nil && found {
```
karthikiyer56: `err != nil && found` does not prove a parse failure.
On rpcv2, `ColdReader.WithLedger` can run `fn` with success and then return a packfile error.
the adapter in `ledger_reader.go` already set `found = true` inside `fn`
This handler then reports a header parse failure for what really could be a disk-read failure.
The latest ledger lives in the hot store, so the risk today is low.
but the mapping breaks for the next caller of `WithLedgerRaw`

Suggested Fix: record the parse error in its own variable. Branch on that variable. This also collapses the two `if` blocks into one.

```go
var response protocol.GetLatestLedgerResponse
var parseErr error
found, err := ledgerReader.WithLedgerRaw(ctx, latestSequence, func(raw []byte) error {
	response, parseErr = latestLedgerResponse(xdr.LedgerCloseMetaView(raw), latestSequence)
	return parseErr
})
if err != nil || !found {
	msg := "could not get latest ledger"
	if parseErr != nil {
		msg = "could not parse latest ledger header"
	}
	return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
		Code:    jrpc2.InternalError,
		Message: msg,
	}
}
return response, nil
```

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
```diff
@@ -23,39 +24,45 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
...
+				Message: "could not parse latest ledger header",
 			}
 		}
-		header := latestLedger.LedgerHeaderHistoryEntry().Header
-		headerBytes, err := header.MarshalBinary()
-		if err != nil {
+		if err != nil || !found {
 			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: fmt.Sprintf("could not marshal latest ledger header: %v", err),
```
karthikiyer56, line 38: The old messages included the underlying error (%v). The new messages are static.  Please include the error text in the message, or log it. A 500 with no cause is hard to debug.

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
```diff
@@ -23,39 +24,45 @@ func NewGetLatestLedgerHandler(ledgerReader store.LedgerReader) jrpc2.Handler {
...
 }
+
+// latestLedgerResponse extracts the response fields from a ledger close meta view.
+func latestLedgerResponse(view xdr.LedgerCloseMetaView, sequence uint32,
+) (protocol.GetLatestLedgerResponse, error) {
+	headerEntry, err := view.LedgerHeader()
+	if err != nil {
+		return protocol.GetLatestLedgerResponse{}, err
+	}
+	return xdr.Try(func() protocol.GetLatestLedgerResponse {
```
karthikiyer56, line 61: The `xdr.Try` block and the `Must*` chains re-implement helpers the view already has. `LedgerCloseMetaView` has `LedgerHash()` and `LedgerCloseTime()`, and `parseLedgerInfo` already uses them with plain error checks:
https://github.com/stellar/stellar-rpc/blob/0bc19d80828d4b046f9a4a1eff13ddab36df7773/cmd/stellar-rpc/internal/methods/get_ledgers.go#L283-L300

Use the same style here. This removes the panic/recover control flow, the local `//nolint:gosec` (the SDK helper carries that conversion), and this call to `xdr.Try` (the SDK function itself stays).

```go
func latestLedgerResponse(view xdr.LedgerCloseMetaView, sequence uint32,
) (protocol.GetLatestLedgerResponse, error) {
	hash, err := view.LedgerHash()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	closeTime, err := view.LedgerCloseTime()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	headerEntry, err := view.LedgerHeader()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	header, err := headerEntry.Header()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	headerRaw, err := header.Raw()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	versionView, err := header.LedgerVersion()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	version, err := versionView.Value()
	if err != nil {
		return protocol.GetLatestLedgerResponse{}, err
	}
	return protocol.GetLatestLedgerResponse{
		Hash:            hex.EncodeToString(hash),
		ProtocolVersion: version,
		Sequence:        sequence,
		LedgerCloseTime: closeTime,
		LedgerHeader:    base64.StdEncoding.EncodeToString(headerRaw),
		LedgerMetadata:  base64.StdEncoding.EncodeToString(view),
	}, nil
}
```

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
```diff
@@ -223,30 +218,61 @@ func (r ledgerReader) GetLatestLedgerSequence(ctx context.Context) (uint32, erro
...
+// ledgerRangeRow is one endpoint of the stored ledger range.
+type ledgerRangeRow struct {
+	Sequence   uint32 `db:"sequence"`
+	MetaPrefix []byte `db:"meta_prefix"`
+}
+
+// ledgerInfoFromRow reads the close time out of the row's meta prefix,
+// refetching the full blob for rare metas whose close time lies beyond it.
+func ledgerInfoFromRow(ctx context.Context, db readDB, row ledgerRangeRow) (store.LedgerInfo, error) {
+	closeTime, err := xdr.LedgerCloseMetaView(row.MetaPrefix).LedgerCloseTime()
```
karthikiyer56, line 234: error optimization nit:


Lines 241 and 245 build the identical error wrap twice. Keep the prefix-parse error when the refetch finds nothing, and fall through to one wrap at the end.

suggest to do the following:
```go
func ledgerInfoFromRow(ctx context.Context, db readDB, row ledgerRangeRow) (store.LedgerInfo, error) {
	closeTime, err := xdr.LedgerCloseMetaView(row.MetaPrefix).LedgerCloseTime()
	if err != nil {
		meta, found, dbErr := getLedgerRawFromDB(ctx, db, row.Sequence)
		if dbErr != nil {
			return store.LedgerInfo{}, dbErr
		}
		if found {
			closeTime, err = xdr.LedgerCloseMetaView(meta).LedgerCloseTime()
		}
		if err != nil {
			return store.LedgerInfo{}, fmt.Errorf("couldn't get ledger %d close time: %w", row.Sequence, err)
		}
	}
	return store.LedgerInfo{Sequence: row.Sequence, CloseTime: closeTime}, nil
}
```

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
```diff
@@ -163,6 +149,15 @@ func (r ledgerReader) GetLedger(ctx context.Context, sequence uint32) (xdr.Ledge
 	return getLedgerFromDB(ctx, r.db, sequence)
 }
 
+// WithLedgerRaw lends the ledger's stored meta blob without decoding it.
+func (r ledgerReader) WithLedgerRaw(ctx context.Context, sequence uint32, fn store.WithLedgerRawFn) (bool, error) {
```
karthikiyer56, line 153: The adapters got three `WithLedgerRaw` tests. 
The sqlite backend got none. 

May I suggest you add a found case and a not-found case to `ledger_test.go`.

```go
func TestWithLedgerRaw(t *testing.T) {
	db := NewTestDB(t)
	tx, err := NewReadWriter(logger, db, host.MakeNoOpDaemon(), 15, passphrase).NewTx(t.Context())
	require.NoError(t, err)
	lcm := createLedger(42)
	require.NoError(t, tx.LedgerWriter().InsertLedger(lcm))
	require.NoError(t, tx.Commit(lcm, nil))
	want, err := lcm.MarshalBinary()
	require.NoError(t, err)

	reader := NewLedgerReader(db)
	var got []byte
	found, err := reader.WithLedgerRaw(t.Context(), 42, func(raw []byte) error {
		got = bytes.Clone(raw)
		return nil
	})
	require.NoError(t, err)
	assert.True(t, found)
	assert.Equal(t, want, got)

	ran := false
	found, err = reader.WithLedgerRaw(t.Context(), 43, func([]byte) error {
		ran = true
		return nil
	})
	assert.NoError(t, err)
	assert.False(t, found)
	assert.False(t, ran)
}
```

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
cjonas9: Imo, I much prefer the `xdr.Try()`-style implementation as I find the other style in this case to be excessively verbose, to the point where I feel like the readability of this degrades. Also, this isn't one we'd expect to panic almost ever, no? The rare cases where we see the slower exception-catch of the `Try` bind would likely mean something much more problematic is going on, so the off-happy-path case isn't too urgent perhaps?

Definitely just good for me to know going forward, but do we truly prefer the `if err` , or are there cases where `xdr.Try` is ok too?

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
cjonas9: good catch!!

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
karthikiyer56: I believe this comment is still unanswered?
> Also you might need to add this to the ignores for the release eval suite since there's no "before"?

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
karthikiyer56: actually, the sdk itself builds `LedgerCloseTime()` on xdr.Try [here](https://github.com/stellar/go-stellar-sdk/blob/81dd8f0aec3f/xdr/ledger_close_meta_view.go#L68-L79) 🤦 

I was more coming from the angle of using existing functions in the views for `closeTime` and `hash` rather than the clunky `hex.EncodeToString(headerEntry.MustHash().MustRaw()),`

I dont have a strong answer to when to use Try vs not.
The verbosity reason seems valid enuf 
(go error verbosity doesnt bother me anymore 🙈 )

tl;dr - no need ot chagne

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
cjonas9: done, good catch!

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
cjonas9: ok, awesome! thanks.

I totally remember feeling this way some time ago when Tamir first proposed the two ideas, with the accessors being convenience overlays over `xdr.Try`. Seems I'm still going to need some extra time before I stop wincing at go error verbosity :cry-laugh:

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
cjonas9: agreed. thanks for providing me with this, I've included it in the test suite

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
cjonas9: This was a little tricky, but I am pretty sure it is ok. Without getting into too much detail, the perf eval `go-bench` suite handles this without an ignore -- if a test isn't there for a build, it's just not run for that build. benchstat handles the asymmetry fine (candidate-only benchmarks show up as rows with an empty base column and no delta, and it still exits 0), so the leg doesn't fail. This test will still be there for the candidate build, which is useful information to see reported on.

karthikiyer56 (approved):

cmd/stellar-rpc/internal/methods/get_latest_ledger.go
cjonas9: Great catch. I'd applied it as suggested, then extended this to carry the underlying error per the follow-up comment. This forced the miss/error paths apart into a switch, but I think that's fine because each arm of the switch distinguishes the actual type of error it seeks to identify (of which there's a parse error, store read error, and `!found` but `err == nil` case).

PR #953, opened 2026-08-26
937 follow-ups: getTransactions cursor guards, getHealth first-commit gate, snapshot gauge, method-table completeness test

### LINKED ISSUE #745: getTransactions returns `cursor: "0"` for empty-result responses; round-tripping decodes to invalid ledger 0
## What's happening

`getTransactions` returns `cursor: "0"` whenever the response processed zero transactions — typically when a polling client has caught up to the head. If the client follows the documented guidance and sends that cursor back in the next request, the server interprets `"0"` as a TOID pointing at `ledger=0` and replies with:

```
{"code": -32602, "message": "database does not contain metadata for ledger: 0"}
```

Every subsequent poll with the same cursor returns the same error. The client loops on it.

## Reproduction

1. Call `getTransactions` with `startLedger` at or past the current head, OR in a steady-state poll loop that has paged forward to the head.
2. Response has `transactions: []` and `cursor: "0"`.
3. Send `"0"` back as `pagination.cursor` (with `startLedger` omitted per the mutex rule).
4. You get the `-32602` error above. Forever, until something clears the cursor.

## Where this comes from in the server

`cmd/stellar-rpc/internal/methods/get_transactions.go` initializes the cursor as `toid.New(0, 0, 0)` before the ledger loop:

```go
cursor := toid.New(0, 0, 0)
for ledgerSeq := start.LedgerSequence; ledgerSeq <= int32(ledgerRange.LastLedger.Sequence); ledgerSeq++ {
    ...
    cursor, done, err = h.processTransactionsInLedger(ledger, start, &txns, limit, request.Format)
    ...
}
return protocol.GetTransactionsResponse{
    ...
    Cursor: cursor.String(),
}, nil
```

If the loop body doesn't run (request is at/past head), or if `processTransactionsInLedger` never updates the cursor (no transactions in any iterated ledger), the cursor stays at its initial zero value. `toid.New(0,0,0).String()` returns the literal `"0"` because `ToInt64()` packs to `0` and `String()` is `fmt.Sprintf("%d", id.ToInt64())`.

## Possible fixes

Setting `cursor = ""` in the response when no transactions were processed would match typical pagination conventions (empty cursor = no more pages) and save clients from special-casing `"0"`. `omitempty` on the field is functionally similar.

If changing the response shape isn't desirable, just documenting the `"0"` behavior on the `getTransactions` docs page would be enough. Clients can then detect it and skip the round-trip.

## Related

Issue #398 (fixed in v23.0.0) changed `transactions: null` → `transactions: []` for the same kind of empty-response consistency. The example response in that thread already shows `cursor: "0"` next to the empty array, but the cursor side wasn't flagged at the time.

Happy to send a PR if you have a preferred direction.

### LINKED ISSUE #937: Post-#908 follow-ups: store-boundary error observability, markErr enforcement, boot-window tip reads
> [!IMPORTANT]
> **This issue is a snapshot** of items deferred during PR #908's review. Contents can change as the branch refines and as #774 / #889 land. **Re-triage every item before picking it up.**

## Open items

| Item | Area | Severity | One line |
|---|---|---|---|
| H2 | cold tx-index read amplification | LOW | frozen `.idx` files reopen per request; no reader reuse across requests |

## H2 — cold tx-index readers reopen per request

**What is wrong:**

- A `getTransaction` that misses the hot indexes probes the frozen window `.idx` files. Each request's read view opens them lazily and closes them at `Release` (`lazyColdTxIndex` in `cmd/stellar-rpc/internal/rpcv2/query/txlookup.go`).
- Back-to-back cold lookups each pay their own open + header read of the same immutable file.

**What to do:**

- Add a process-wide reader cache keyed by coverage generation. A rebuilt index gets a new coverage, so a stale reader is never returned.
- The hard part is lifetime, not the diff: define when a cached reader is evicted, and what happens when a prune deletes the file under it. Getting this wrong is a use-after-close on the read path. Keep the current view-owned design until that answer is written down.

---

## Closed — no longer tracked here

| Item | Outcome |
|---|---|
| B2 (boot-window tip point-reads) | Resolved on the PR #908 branch: `adapters.SeedCloseTimes` stamps both window edges at startup |
| D2 (permanent data loss answered "retry" with no signal) | Resolved: `unavailable_chunk_resolves_total` + a chunk-id log at the routing decision; the `-32002` remap no longer exists |
| D3 (cold half of "outlived the deletion grace" invisible) | Resolved: `missing_cold_pack_opens_total` at the cold reader; `store_ops_after_deferred_close_total` at the store guard |
| E2 (`markErr` convention unenforced) | Dissolved: the error side channel was deleted; nothing is left to enforce |
| G1 (snapshot-leak gauge) | Resolved: `open_snapshots` GaugeFunc beside the serving-invariant counters; a process-wide tally in `rocksdb` mirrors the `deferredCloseOps` pattern |
| H1 (four hand-synced method tables) | Resolved with the minimum viable option: a reflection completeness test ties `MethodsConfig` ↔ `limitsByMethod` ↔ `validateService`; `Apply` already ties limits ↔ served specs at startup |
| F1 + F2 (getTransactions cursor contract) | Resolved together: the returned cursor is floored at the request's cursor, so a caught-up poller gets its own cursor echoed back instead of `"0"` (F1, reported in #745) or a regressed cursor that re-delivers a consumed ledger (F2). Shared-handler wire change, called out loudly in the CHANGELOG |
| I1 (getHealth after a restart) | Resolved for v2: the registry records the boot-seeded latest ledger and getHealth fails until ingestion advances past it. v1 keeps the shared behavior. Cost: after every restart, getHealth reports unhealthy until the first commit (about one ledger interval) |

## Related but tracked elsewhere

| What | Home |
|---|---|
| getEvents / getEventsV2 serving, including classifying the v2 pager's error kinds (`ErrCursorMalformed`, `ErrInvalidLimit`, `ErrPositionMismatch`, the offsets-shortfall internal error) — the wrapper classifies nothing, so the v2 events adapter must map them itself | #774 |

karthikiyer56:
Executes the remaining #937 items against the current branch tip. Four items land here; H2 stays deferred in #937.

## What changed

| Item | Change | Files |
|---|---|---|
| G1 | `open_snapshots` gauge: a process-wide tally of unreleased RocksDB snapshots, exported beside the serving-invariant counters. A leaked read view is visible on a dashboard, not only in a teardown log | `rpcv2/rocksdb/rocksdb.go`, `rpcv2/observability/observability.go` |
| H1 | Completeness test ties the per-method tables together: every `MethodsConfig` method must appear in `limitsByMethod` and be checked by `validateService`. `Apply` already ties limits to the served specs | `rpcv2/methods_completeness_test.go` (test-only) |
| F1 + F2 | `getTransactions` never returns a cursor below the request's cursor. A caught-up poller gets its cursor echoed back instead of `"0"` (#745) or a regressed cursor that re-delivers a consumed ledger | `methods/get_transactions.go`, `CHANGELOG.md` |
| Review follow-up | `getLedgers` gets the same caught-up contract: a cursor at or past the tip returns an empty page with the cursor echoed, instead of `-32602`. The cursor parse keeps only the lower-bound rejection; a max-uint32 cursor wraps to start 0 and stays rejected | `methods/get_ledgers.go`, `CHANGELOG.md` |
| I1 | v2 `getHealth` fails until this run's ingestion commits its first ledger. The registry records the boot-seeded latest ledger; the gate compares against it | `rpcv2/query/registry.go`, `rpcv2/jsonrpc.go` |

## Wire-visible changes

| Method | Before | After |
|---|---|---|
| `getTransactions` (v1 AND v2 — shared handler) | Cursor at/past the tip returned `"0"`; resending it failed every poll with `-32602`. A consumed ledger's cursor regressed and the next poll re-delivered its transactions | The request's cursor is echoed back on an empty caught-up page. The returned cursor always fetches what comes next. Fixes #745. CHANGELOG carries a loud entry |
| `getLedgers` (v1 AND v2 — shared handler) | Resending the last page's cursor at the tip failed with `-32602` — the same code a malformed cursor gets — until the next ledger closed | Empty page with the cursor echoed back. Below-oldest cursors and explicit above-tip `startLedger` still error. CHANGELOG carries a loud entry |
| `getHealth` (v2 only) | Healthy for up to `max_healthy_ledger_latency` after a restart, off the previous run's close time — even in a crash loop | Fails with `-32603` until the first commit of this run lands (about one ledger interval after boot). v1 unchanged |

## Not in this PR

| Item | Why |
|---|---|
| H2 (cold `.idx` reader cache) | Deferred in #937: needs a written reader-lifetime design first; a wrong answer is a use-after-close on the read path |

## Notes

- Each item is one commit, in the order reviewed.
- `open_snapshots` mirrors the `deferredCloseOps` pattern: a package atomic read by a `GaugeFunc`.
- The H1 test is reflection over `MethodsConfig`: adding a method field extends the checks with no test edits.


cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -280,6 +286,14 @@ func (h transactionsRPCHandler) getTransactionsByLedgerSequence(ctx context.Cont
 		}
 	}
 
+	// A caught-up poller's cursor points at or past the tip. The walk then
+	// produces nothing and leaves the cursor below the request's own — at the
+	// zero value, or at the consumed ledger's start. Echo the request's cursor
+	// instead: the returned token must always fetch what comes next (#745).
+	if requestCursor != nil && cursor.ToInt64() < requestCursor.ToInt64() {
```
tamirms, line 293: While verifying the cursor guard here I checked how the other two paginated methods behave for the same caught-up poll. getEvents is fine: an empty page returns a max cursor at the last scanned ledger, which stays valid when resent at the tip. getLedgers still breaks the polling loop, and it fails with an error instead of an empty page:

1. A poller reaches the tip ledger T. getLedgers returns its last page with `cursor: "T"`.
2. The poller resends `cursor: "T"`. `parseCursor` computes `start = T + 1` and rejects it, because `IsLedgerWithinRange` requires `start <= T` (get_ledgers.go, line 154).
3. The poller gets `-32602`: "cursor ('T') must be between the oldest ledger: X and the latest ledger: T". Every poll fails until the next ledger closes.

This self-heals within one ledger interval, so nobody gets permanently stuck the way #745 clients did. But it is the steady state for anyone polling faster than the close interval, and the code is the same `-32602` a malformed cursor gets, so a client cannot tell "wait and retry" apart from "my cursor is garbage" without parsing the message text. The handler already contains the intended behavior: the empty-result branch echoes `request.Pagination.Cursor` (line 106), but validation rejects the caught-up request before that branch can run, so it is unreachable for exactly the case it was written for.

Since the changelog entry in this PR states the contract generally ("the returned cursor now always does what the docs promise"), I think we should bring getLedgers in line, either as one more commit here or as an immediate follow-up, ideally landing in the same release so the caught-up contract changes once for both methods. The shape I have in mind mirrors the getTransactions fix: `parseCursor` keeps only the lower-bound rejection (a cursor below the oldest ledger is real data loss and should stay an error), and the handler returns early for a cursor at or past the tip, before `end` is computed or `fetchLedgers` runs:

```go
// A caught-up poller's cursor points at or past the tip. Echo it back on an
// empty page instead of rejecting the server's own token.
if request.Pagination != nil && request.Pagination.Cursor != "" &&
	start > availableLedgerRange.LastLedger {
	return protocol.GetLedgersResponse{
		Ledgers:               []protocol.LedgerInfo{},
		LatestLedger:          ledgerRange.LastLedger.Sequence,
		LatestLedgerCloseTime: ledgerRange.LastLedger.CloseTime,
		OldestLedger:          ledgerRange.FirstLedger.Sequence,
		OldestLedgerCloseTime: ledgerRange.FirstLedger.CloseTime,
		Cursor:                request.Pagination.Cursor,
	}, nil
}
```

Two details worth pinning in tests: an explicit `startLedger` above the tip should stay an error, the same asymmetry getTransactions has (only the server-issued token gets the echo, an explicit out-of-range start is a client mistake), and the cursor `"4294967295"` makes `start` wrap to 0 in `parseCursor`, which the retained lower-bound check needs to catch. Placing the early return before `end := start + uint32(limit) - 1` also keeps that addition from ever running with an above-tip start.

If you'd rather defer this, a row on #937 plus scoping the changelog sentence to getTransactions would keep the docs honest in the meantime.

karthikiyer56: Fair callout — done in 37d84243, making getLedgers toe the same cursor-semantics line as getTransactions.

tamirms (approved):

PR #954, opened 2026-08-26
rpcv2: serve getEventsV2

urvisavla:
Serves `getEventsV2`. It was a not-implemented stub before this.

The handler takes a range request or a cursor. A range request starts paging, and a cursor continues it. Both check the term budget before reading anything, because anyone can build a cursor by hand.

Notes:

- `minLedger` below genesis is raised to genesis instead of rejected. The retention floor then decides what happens: `ledger_out_of_range` going up, `OLDEST_REACHED` going down. `maxLedger` is never raised, because that would add genesis to a range that excluded it. A range wholly below genesis returns an empty `COMPLETE` page.
- An explicit `limit` is checked against `max_items_per_response` first, so the error names this node's limit and not the spec's 1000. Config validation keeps that setting at 1000 or below.
- Unknown request fields are rejected. The handler decodes its own params, so a wrong-typed field is reported by name instead of by the decoder's message.
- Errors carry the typed `error.data`. Anything unexpected returns an internal error, the same as v1.
- A hand-built cursor could carry filter shapes that no v2 request can send. The dangerous one was an empty filter clause, which scanned everything and counted as zero terms against the budget. All such shapes are now rejected as `cursor_malformed`.
- `limit` and `xdrFormat` apply to each request and are not stored in the cursor. The spec defines them that way.
- `xdrInputFormat: "json"` is rejected for now. Support is deferred to #940.

`WithView` and `ViewFrom` moved from `adapters` into `query`. They carry the read view through the request context, which is the query layer's concern, and the handler is not an adapter.

End-to-end tests come in a separate PR.


cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,385 @@
...
+	ctx context.Context, view *query.ReadView, limits Limits,
+	req *protocol.GetEventsV2Request, oldest, latest uint32,
+) (protocol.GetEventsV2Response, error) {
+	cursor, limit, err := requestCursor(limits, req, latest)
+	if err != nil {
+		return protocol.GetEventsV2Response{}, err
+	}
+	// Checked on the cursor path too: a cursor carries the filters and is
+	// not signed, so a hand-built one could ask for any number of lookups.
+	if err := checkTermBudget(cursor.Scope.Filters, limits.TermBudget); err != nil {
```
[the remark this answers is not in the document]
urvisavla: fixed

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	eventType, err := responseEventType(ev.Type)
+	if err != nil {
+		return protocol.EventInfoV2{}, err
+	}
+
+	info := protocol.EventInfoV2{
+		EventType:      eventType,
+		Ledger:         int32(p.LedgerSequence), //nolint:gosec // ledger sequences fit int32 by protocol
```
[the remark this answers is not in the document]
urvisavla: fixed

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,396 @@
...
+
+// response carries a cursor on every page but the last: an absent cursor is
+// what tells a client the query is finished.
+func response(
+	page *query.EventPage, format string, oldest, latest uint32,
+) (protocol.GetEventsV2Response, error) {
+	resp := protocol.GetEventsV2Response{
+		Events:        make([]protocol.EventInfoV2, 0, len(page.Events)),
+		ScanStatus:    responseScanStatus(page.Status),
+		ScannedLedger: page.Next.ScannedLedger,
```
tamirms: The proposal defines the no-progress value of `scannedLedger` as the ledger just before the range: `minLedger - 1` ascending, `maxLedger + 1` descending. Here the pager's internal "nothing scanned yet" sentinel (0) goes out on the wire instead. It shows up on any page that completes no ledger: a fresh query whose page fills midway through its first ledger, and every `WAITING_FOR_LEDGERS` page before the first scan.

Ascending it is harmless in practice, since 0 is still below the range. Descending it inverts the field's meaning: `scannedLedger` promises that everything between the query's starting bound and this ledger has been delivered, and for a descending query the starting bound is `maxLedger`, so 0 claims the whole range is covered on a page that delivered part of one ledger.

The cursor itself has to keep 0, since the codec rejects nonzero marks outside the scope bounds. So I think this is a translation at the response edge only:

```go
scanned := page.Next.ScannedLedger
if scanned == 0 {
	if page.Next.Scope.Dir == query.Descending {
		scanned = *page.Next.Scope.MaxLedger + 1
	} else {
		scanned = page.Next.Scope.MinLedger - 1
	}
}
```

The dereference is safe because a descending scope always carries `MaxLedger`, and ascending `MinLedger` is at least genesis, so there is no underflow. The one theoretical edge is a descending `maxLedger` of `MaxUint32` wrapping to 0, which no real ledger number will approach; a clamp is cheap if you want it airtight.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,396 @@
...
+	// A below-genesis minLedger is raised, not rejected. The floor rules
+	// then decide the outcome: ledger_out_of_range ascending,
+	// OLDEST_REACHED descending. Only a forged cursor now trips the
+	// pager's own below-genesis check.
+	scope := query.EventScope{MinLedger: max(req.MinLedger, chunk.FirstLedgerSeq)}
+	if req.Order == protocol.OrderDescending {
+		scope.Dir = query.Descending
+		maxLedger := req.MaxLedger
+		if maxLedger == 0 {
+			maxLedger = latest
```
tamirms, line 364: When `maxLedger` is omitted and `minLedger` is above the tip, pinning the default first turns the request into an inverted scope, and the client gets `invalid_params` with the pager's message, for example "query: inverted range (lo > hi): [1005, 1000]", naming a bound the request never contained. The proposal's errors table makes this exact case `ledger_out_of_range` with `missingLedger` = `minLedger`, and its retry column says retrying succeeds once the node holds that ledger. The difference matters to a poller: the condition is transient (the same request serves once the chain reaches `minLedger`), and it is also what separates this shape from the same request with an explicit `maxLedger` above the tip, which is legal and waits at `WAITING_FOR_LEDGERS`.

The handler is the only layer that knows the max was defaulted, so I think the check belongs here, inside the descending branch before the pin:

```go
if req.MaxLedger == 0 && req.MinLedger > latest {
	return query.EventScope{}, &query.RangeError{
		Requested: req.MinLedger, Oldest: oldest, Latest: latest,
	}
}
```

with `oldest` threaded in; `responseError` already maps `RangeError` to the right wire error. Fallout: `TestEventScopeDescendingMinAboveLatest` asserts the current behavior and would flip to asserting this one, and the `ErrInvertedRange` entry in `responseError` becomes unreachable from client input, so it can be dropped or kept as defense.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,396 @@
...
+type Limits struct {
+	TermBudget uint32
+	// MaxLimit is protocol.MaxLimitV2 unless an operator lowers it.
+	MaxLimit uint
+	// DefaultLimit applies when the request sets no limit.
+	DefaultLimit uint
+}
+
+// NewHandler builds the getEventsV2 handler.
+func NewHandler(limits Limits) jrpc2.Handler {
```
tamirms, line 41: The proposal requires unknown request fields to be rejected (the invalid_params table has "unknown field: contractID"), with the rationale that omission is a wildcard here, so a silently ignored typo silently widens the results. jrpc2 only enforces that when the wrapper opts in, and `methods.NewHandler` doesn't, so today `{"minLedger":100,"maxLedgor":200}` becomes an unbounded tip-following query, and a filter with `"topicc1"` quietly drops its topic1 constraint.

I don't think we should flip `methods.NewHandler` itself, since that changes wire behavior for every v1 method. Two per-method options:

1. A strict sibling in `internal/methods` (`handler.Check` plus `AllowArray(false)` plus `SetStrict(true)`), used only by this handler. Unknown fields then fail, including inside the nested filter objects, but the error carries no `data.reason`.
2. Have this handler take `*jrpc2.Request`, unmarshal with `jrpc2.StrictFields`, and shape the failure through `jrpcError`, so `data.reason` is present like every other error this endpoint returns.

I'd lean toward 2 since the plumbing already exists, but 1 already closes the hole where a typo silently widens results.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,396 @@
...
+// An absent ascending maxLedger stays nil, the open bound. A descending one
+// is pinned to latest, so every page of the session shares one top edge.
+func eventScope(
+	req *protocol.GetEventsV2Request, latest uint32,
+) (query.EventScope, error) {
+	// A below-genesis minLedger is raised, not rejected. The floor rules
+	// then decide the outcome: ledger_out_of_range ascending,
+	// OLDEST_REACHED descending. Only a forged cursor now trips the
+	// pager's own below-genesis check.
+	scope := query.EventScope{MinLedger: max(req.MinLedger, chunk.FirstLedgerSeq)}
```
tamirms, line 352: Raising a below-genesis `minLedger` instead of rejecting it seems right to me: ledger 1 never exists to serve, so an error claiming the node is missing it would send the client hunting for history no node can have. Two consequences worth pinning down: on a node whose floor is above genesis, `minLedger: 1` now reports `missingLedger: 2` rather than the requested 1, and on a node that holds genesis it serves instead of erroring, while the proposal's errors table says an ascending `minLedger` below `oldestLedger` errors with `missingLedger` = `minLedger`. I'll update the proposal text to record the raise so the doc and the wire agree.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,396 @@
...
+		}
+		resp.Cursor = token
+	}
+	return resp, nil
+}
+
+// responseError maps the pager's and the SDK's errors onto the three
+// machine-readable reasons. Anything else becomes an internal error; jrpc2
+// would otherwise code a plain error as SystemError.
+func responseError(err error, oldest, latest uint32) error {
```
tamirms, line 294: Minor: the messages that reach clients on these paths are the internal error strings, so a bad cursor reads "query: malformed cursor: base64: illegal base64 data at input byte 4", and the json-input rejection carries an "rpcv2:" prefix. The reasons and data payloads are typed, so nothing is broken, but the proposal's message table is clean text, and these leak layering the client has no use for. A small pass on the client-visible strings, or stripping the package prefixes here, would close the gap. Fine as a follow-up.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2_test.go
```diff
@@ -0,0 +1,295 @@
...
+		Reason string `json:"reason"`
+	}
+	require.NoError(t, json.Unmarshal(jerr.Data, &reason))
+	assert.Equal(t, wantReason, reason.Reason)
+	if into != nil {
+		require.NoError(t, json.Unmarshal(jerr.Data, into))
+	}
+}
+
+func TestGetEventsV2_AscendingPagesToComplete(t *testing.T) {
```
tamirms: Nit: the name says the query pages to COMPLETE, but the range has no `maxLedger`, so the second page correctly asserts `WAITING_FOR_LEDGERS`. Maybe `TestGetEventsV2_AscendingPagesToTheTip` or similar, so the name and the final assertion agree.

cmd/stellar-rpc/internal/rpcv2/adapters/sharedview.go
```diff
@@ -24,8 +24,10 @@ func WithView(ctx context.Context, view *query.ReadView) context.Context {
 	return context.WithValue(ctx, viewKey{}, view)
 }
 
-// viewFrom returns the view WithView installed, or errNoView.
-func viewFrom(ctx context.Context) (*query.ReadView, error) {
+// ViewFrom returns the view WithView installed, or errNoView. It is exported
+// for the getEventsV2 handler. That handler is not an adapter, but it takes
+// its view from the request context the same way.
+func ViewFrom(ctx context.Context) (*query.ReadView, error) {
```
tamirms: The export works; what caught my eye is that the comment has to explain it. The reason it needs explaining is that the function now sits in the wrong package for its consumers. WithView and ViewFrom are ReadView context plumbing, which is the query layer's subject matter; this package's subject is satisfying the v1 method interfaces. getEventsV2 is the first handler that is not an adapter, and the v1 getEvents shim will be the second, so this package is about to be a dependency of every native handler for no reason beyond this pair of functions.

What would you think about moving WithView and ViewFrom (together with viewKey and errNoView) into a new query/view_context.go and deleting this file, either here or as a small follow-up PR? jsonrpc.go and the two adapter readers switch to query.WithView / query.ViewFrom, and eventsapi then does not import the adapters package at all. It comes to about ten files and roughly a forty-line net diff, with no new exports and no cycle risk, since query imports neither adapters nor eventsapi. The comment on ViewFrom then shrinks to one line, because nothing needs excusing anymore.

One optional companion if this becomes a follow-up PR: renaming query/scan.go to ledger_scan.go and txlookup.go to tx_lookup.go so query's per-domain files read as one family next to event_page.go.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Added
+* The full-history service serves `getEventsV2`, the cursor-paged events method ([#774](https://github.com/stellar/stellar-rpc/issues/774)).
```
tamirms: I think the changelog should be limited to changes that affect rpc v1. we'll probably want a separate changelog for rpc v2 once we do the first release of that service

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,455 @@
...
+		// With no max from the client, a low edge past the tip is out of
+		// range.
+		if req.MaxLedger == 0 && scope.MinLedger > latest {
+			return query.EventScope{}, &query.RangeError{
+				Requested: scope.MinLedger, Oldest: oldest, Latest: latest,
+			}
+		}
+		maxLedger := latest
+		if req.MaxLedger != 0 {
+			maxLedger = max(req.MaxLedger, chunk.FirstLedgerSeq)
```
tamirms: Raising a below-genesis maxLedger expands the range instead of preserving it. Raising minLedger is safe because no ledgers exist below genesis, so `[1, X]` and `[2, X]` contain the same existing ledgers. Raising maxLedger adds a ledger the request excluded: `{minLedger: 1, maxLedger: 1}` passes Valid (1 <= 1), both bounds get raised, and the scope becomes `[2, 2]`. I ran it against a genesis chunk holding one event at ledger 2: the response carries that event with COMPLETE and scannedLedger 2, for a request that asked only about ledger 1. `{order: "desc", maxLedger: 1}` answers the same way. Before this change the same requests got invalid_params through the inverted scope, which was the wrong class but returned no wrong data.

Since a range whose intersection with existing ledgers is empty is just an exhausted range, I think the honest answer is the empty COMPLETE page: when an explicit max lands below the raised min, return the response with no events, COMPLETE, and scannedLedger reporting no progress, instead of raising the max. That keeps the "raising both keeps min <= max" property this change wanted, without serving data from outside the requested bounds.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 5ccddda7

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 5ccddda7

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 3e51c744

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 2e695173

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2_test.go
urvisavla: fixed in 291e8fd6

cmd/stellar-rpc/internal/rpcv2/adapters/sharedview.go
urvisavla: fixed in 6abaf6eb

CHANGELOG.md
urvisavla: fixed in 291e8fd6

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 9192f75e

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,478 @@
...
+
+func serve(
+	ctx context.Context, view *query.ReadView, limits Limits,
+	req *protocol.GetEventsV2Request, oldest, latest uint32,
+) (protocol.GetEventsV2Response, error) {
+	cursor, limit, err := requestCursor(limits, req, oldest, latest)
+	if err != nil {
+		return protocol.GetEventsV2Response{}, err
+	}
+	if req.Cursor == "" && exhausted(req) {
```
tamirms: The shortcut runs before checkTermBudget, so an over-budget query gets an empty COMPLETE page whenever its maxLedger sits below genesis. I probed it: {order: "desc", maxLedger: 1} with three distinct topic0 values under a budget of 2 answers COMPLETE, while the same filters with maxLedger: 2 get the invalid_params budget error, so whether validation runs depends on the range's emptiness. The proposal's Limits section makes the budget row unconditional, and the errors section says request validation runs before any scanning, so I think checkTermBudget belongs above this return, and a below-genesis case in TestGetEventsV2_TermBudgetRejectsBothRequestShapes would pin the precedence.

While here, one possible simplification: the built scope already states the condition (an explicit max the raised min has passed), and the finished page for it is exactly what response() produces for a ScanComplete page with no events. I checked that response(&query.EventPage{Next: cursor, Status: query.ScanComplete}, req.Format, oldest, latest) returns a byte-identical response for both orders, so the shortcut could be `if cursor.Scope.MaxLedger != nil && cursor.Scope.MinLedger > *cursor.Scope.MaxLedger` reusing response(), and exhausted() plus exhaustedResponse() both go away.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,478 @@
...
+	// raised, because that would add genesis to a range that excluded it;
+	// serve handles that shape as an exhausted range.
+	scope := query.EventScope{MinLedger: max(req.MinLedger, chunk.FirstLedgerSeq)}
+	if req.Order == protocol.OrderDescending {
+		scope.Dir = query.Descending
+		// With no max from the client, a low edge past the tip is out of
+		// range.
+		if req.MaxLedger == 0 && scope.MinLedger > latest {
+			return query.EventScope{}, &query.RangeError{
+				Requested: scope.MinLedger, Oldest: oldest, Latest: latest,
```
tamirms, line 359: Reusing query.RangeError here sends the below-floor message for the above-tip condition. I probed it: {order: "desc", minLedger: 50102} on a node serving [50002, 50004] answers "ledger 50102 is below the retention floor; available range is [50002, 50004]", telling the client the ledger was pruned when it does not exist yet, which inverts the recovery action the proposal's retry column describes for this case (retry succeeds once the node holds that ledger). The error data is right, so only the text misleads. RangeError.Error() could pick its wording by comparing Requested to Latest ("above the latest ledger" when it is), with the doc comments updated to match; ClampRange only ever mints the below-floor direction, so the existing wording survives everywhere it appears today. A message assertion in TestEventScopeDescendingMinAboveLatest would keep the two directions from swapping again.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,478 @@
...
+		if err := r.UnmarshalParams(jrpc2.StrictFields(&req)); err != nil {
+			return nil, decodeError(err)
+		}
+		return getEventsV2(ctx, limits, &req)
+	}
+}
+
+// decodeError re-shapes a params failure to carry a reason. jrpc2 puts the
+// decoder's message in error.data, which leaves no room for one.
+func decodeError(err error) error {
```
tamirms: The strict decode fires before the range checks, so a negative limit never reaches the "limit must be between 1 and %d" path: {"limit": -1} answers "cannot unmarshal number -1 into Go struct field GetEventsV2Request.limit of type uint", and {"minLedger": "abc"} leaks the same Go struct naming the unknown-fields test says is not the client's business. Since NewHandler already holds the raw params, decoding them directly with json.NewDecoder + DisallowUnknownFields would let a *json.UnmarshalTypeError map to a field-named message (the spec's limit wording for limit, "minLedger must be a number" and the like elsewhere) instead of round-tripping through the decoder's prose.

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
```diff
@@ -0,0 +1,478 @@
...
+			MissingLedger: outOfRange.Requested,
+			OldestLedger:  outOfRange.Oldest,
+			LatestLedger:  outOfRange.Latest,
+		})
+	}
+	switch {
+	case errors.Is(err, query.ErrCursorMalformed),
+		errors.Is(err, query.ErrCursorUnknownVersion),
+		errors.Is(err, query.ErrPositionMismatch),
+		errors.Is(err, query.ErrInvertedRange):
```
tamirms: After the exhausted-range change, I could not find a wire input that reaches this arm: a forged cursor with min > max dies inside DecodeEventCursor (its own cursor_malformed, from the codec's min/max check), and a range request cannot build an inverted scope anymore. So the only way ErrInvertedRange surfaces here is a scope this server minted wrong, and cursor_malformed would then blame the client for a server bug. Letting it fall through to the internal-error default seems more honest, and the arm's test goes with it.

tamirms (approved):
I left comments for a few more issues but I think the PR should be good to merge after they are addressed

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in 63a04469

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in a296ad9d

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in cac33bb3

cmd/stellar-rpc/internal/rpcv2/eventsapi/get_events_v2.go
urvisavla: fixed in cb14babd

PR #962, opened 2026-08-27
Migrate transaction read path to use views

### LINKED ISSUE #948: Adopt XDR views: migrate transaction read path to utilize XDR views
Parent: #732 

### Scope

 `getTransactionByHash` currently walks decoded LCMs within a rows[] struct. Then, when it finds the target `tx`, it reencodes it. This work could be trimmed using Views. From the parent:

getTransactionByHash (line 200) reads:
```
var rows []struct {
    TxIndex int                 `db:"application_order"`
    Lcm     xdr.LedgerCloseMeta `db:"meta"`  // Scanner auto-decodes the entire LCM
}
```

Then, `NewLedgerTransactionReaderFromLedgerCloseMeta` walks the just-decoded structure to find the `tx` at `application_order.` `ParseTransaction` calls `MarshalBinary` on `Result` (line 258), Meta (line 261), and Envelope (line 264) to repackage as bytes for the response. The transactions table only stores (hash, ledger_sequence, application_order) lookups, so every GetTransaction joins back to ledger_close_meta.

Instead, we may:
- Read raw bytes from the JOIN (`Lcm []byte`).
- Wrap as `LedgerCloseMetaView,` view-walk to `application_order,` slice `tx` envelope/result/meta as raw bytes via view accessors.
- `ParseTransaction` accepts the view + index; constructs `Transaction` struct with raw byte slices directly.

This requires `LedgerCloseMetaView` accessors that yield per-transaction byte slices or transaction-level views (envelope/result/meta) within the SDK.

### Acceptance criteria

Simply restated, `getTransactionByHash` no longer decodes and reencodes LCMs and instead uses a view-walk.

cjonas9:
### What

This PR migrates the transaction read paths to ledger bytes/views, following #945. `getTransaction` (by hash) and `getTransactions` no longer decode an entire multi-MB LCM to read a handful of per-transaction fields back out of it. Here is what's modified:

- `getTransactionByHash`: the row's meta blob is scanned into a view and exactly one transaction is extracted via `ingest.LedgerTransactionViewRange`
    - removes `NewLedgerTransactionReaderFromLedgerCloseMeta` + `Seek` over a fully decoded LCM.
- `getTransactions`: each ledger in the page walk is borrowed as raw bytes (`readLedgerPage`) and walked through views (`processTransactionsInLedger`)
    - only the page's worth of transactions is materialized
    - page-entry rendering is extracted into `transactionInfo`, shared with the differential suite below.
- `store.ParseTransactionView` takes an `ingest.LedgerTransactionView` and just reshapes it
    - old re-marshal loop is deleted 🎉
- `store.LedgerReaderTx` gains loan-shaped `WithLedgerRaw`: fn borrows the marshaled LCM, valid only inside the call.
    - for v1: it lends the scanned blob (database/sql clones BLOB scans)
    - for v2: it advances the same walk cursor as `GetLedger` and lends the step's scratch bytes w/o copy.

Two related correctness fixes were made: 
- bumps SDK to stellar/go-stellar-sdk#5997
    - the view extractor disagreed with the decode path on one shape — (TransactionMetaV3 + Soroban envelope + absent SorobanMeta)

Note that an exhaustive test suite relating to this work exists as a fast-follow PR, #976. Both should be merged around the same time.

#### Minor behavior changes to note

- Legacy TransactionMeta V0 is now tolerated as event-free by the view path. This is pinned by the new test `TestTransactionsWithLegacyV0Meta`
- `getTransactions` now returns `jrpc2.InternalError` instead of `jrpc2.InvalidParams` when it is unable to read a transaction out of a ledger

### Why

See #732. This is a component of adopting ledger views in RPC for query-path performance purposes.

### Known Limitations

N/A

### Attribution

Substantial parts of this PR are @tamirms's work, cherry-picked from
`tamirms/gettransactions-view-walk`:
- the loan-shaped `WithLedgerRaw` accessor on `LedgerReaderTx` and its contract
- the raw-bytes page walk in `getTransactions` and the shared `transactionInfo` renderer
- the Tx-level `WithLedgerRaw` tests
- the discovery of the V3/absent-`SorobanMeta` event-arity wire bug, + its upstream fix (`stellar/go-stellar-sdk#5997`)


cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
Shaptic, line 1: Do we need to change V2 code?

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
cjonas9, line 1: This is @tamirms work that I can't take credit for. It's the differential suite of 

> ~410 cases pinning byte-identical responses old-vs-new across meta versions, fee-bumps, event shapes, cursors, and page boundaries

cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -74,149 +72,153 @@ func (h transactionsRPCHandler) initializePagination(
...
 }
 
-// fetchLedgerData calls the meta table to fetch the corresponding ledger data.
-func (h transactionsRPCHandler) fetchLedgerData(ctx context.Context, ledgerSeq uint32,
-	readTx store.LedgerReaderTx,
-) (xdr.LedgerCloseMeta, error) {
-	ledger, found, err := readTx.GetLedger(ctx, ledgerSeq)
-	if err != nil {
-		return ledger, &jrpc2.Error{
+// readLedgerPage borrows ledgerSeq's raw LedgerCloseMeta and runs the page
```
cjonas9, line 75: attribution: @tamirms loan-shaped pager design from `tamirms/gettransactions-view-walk`

cmd/stellar-rpc/internal/rpcv1/sqlitedb/transaction.go
```diff
@@ -175,11 +164,11 @@ func (txn *transactionHandler) GetTransaction(ctx context.Context, hash xdr.Hash
...
 // Note: Caller must do input sanitization on the hash.
 func (txn *transactionHandler) getTransactionByHash(ctx context.Context, hash xdr.Hash) (
-	xdr.LedgerCloseMeta, ingest.LedgerTransaction, error,
+	store.Transaction, error,
 ) {
 	var rows []struct {
-		TxIndex int                 `db:"application_order"`
-		Lcm     xdr.LedgerCloseMeta `db:"meta"`
+		TxIndex int                     `db:"application_order"`
+		Lcm     xdr.LedgerCloseMetaView `db:"meta"`
```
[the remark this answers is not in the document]
cjonas9: Untrue comment, disregard. Named `[]byte` types scan via `convertAssign`'s reflect fallback (`convert.go`: `AssignableTo` -> `bytes.Clone` for `[]byte` sources). `Scanner` is only one of the paths, not the last one. `TestTransactionFound` covers this exact scan against real SQLite.

cmd/stellar-rpc/internal/rpcv1/sqlitedb/transaction.go
```diff
@@ -189,27 +178,26 @@ func (txn *transactionHandler) getTransactionByHash(ctx context.Context, hash xd
...
 	txIndex, lcm := rows[0].TxIndex, rows[0].Lcm
-	reader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(txn.passphrase, lcm)
+	ledgerSeq, err := lcm.LedgerSequence()
 	if err != nil {
-		return lcm, ingest.LedgerTransaction{},
-			fmt.Errorf("failed to create ledger reader: %w", err)
+		return store.Transaction{}, fmt.Errorf("failed to get ledger sequence: %w", err)
 	}
-	err = reader.Seek(txIndex - 1)
+	txnViewRange, err := ingest.LedgerTransactionViewRange(lcm, txIndex-1, 1, txn.passphrase)
```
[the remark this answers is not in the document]
cjonas9: The change is intentional and required by full-history. "retain the previous outcome" would 500 on pre-V1-meta history. SDK enforces this in `TestTransactionEventsFromMeta_LegacyV0`

cmd/stellar-rpc/internal/rpcv1/sqlitedb/transaction.go
```diff
@@ -203,27 +192,26 @@ func (txn *transactionHandler) getTransactionByHash(ctx context.Context, hash xd
...
-		return lcm, ingest.LedgerTransaction{},
-			fmt.Errorf("failed to index to tx %d in ledger %d (txhash=%s): %w",
-				txIndex, lcm.LedgerSequence(), hash, err)
+		return store.Transaction{}, fmt.Errorf("failed to index to tx %d in ledger %d (txhash=%s): %w",
+			txIndex, ledgerSeq, hex.EncodeToString(hash[:]), err)
 	}
-
-	ledgerTx, err := reader.Read()
-	return lcm, ledgerTx, err
+	if len(txnViewRange) == 0 {
```
karthikiyer56, line 210: - The old code read the transaction at `txIndex`, then returned an internal error if the read failed.
- The new code returns `store.ErrNoTransaction` when `txnViewRange` comes back empty (lines 210-212).
- `store.ErrNoTransaction` means one thing everywhere else in this codebase: "this hash does not exist." An empty result here means something else: the row already found at lines 187-198 says this hash's `application_order` sits inside this ledger. If the view extraction at that exact position finds nothing, the `transactions` table and the ledger meta table disagree. That is database corruption, not a missing transaction.
- This repo already has a convention for that distinction, in this same PR:
  - `rpcv2/adapters/transaction_reader.go:41-49` — https://github.com/stellar/stellar-rpc/pull/962/files#diff-62717ea5e6bb51839fa71cb0a142f72ad81fdb735c527a703ddbfa8e018f6755R41-R49 — an index disagreeing with the ledger store returns the underlying error (counted as `txhash.ErrInconsistent`), not `store.ErrNoTransaction`.
  - `rpcv2/adapters/ledger_reader.go:253-269` — https://github.com/stellar/stellar-rpc/pull/962/files#diff-64106cf72c1990237c454dbf154a37911ec98a9d69a8e4f46c4ba3c2a656c91fR253-R269 — a broken walk contract returns a plain `fmt.Errorf`, with the comment "Fail loudly rather than serve a wrong-position read."
  - Both flow up through the same shared handler, `methods/get_transaction.go:71-81` (unchanged by this PR) — https://github.com/stellar/stellar-rpc/blob/e4a6f8d982cdae94054e11a8cff0aabceb6f0732/cmd/stellar-rpc/internal/methods/get_transaction.go#L71-L81 — only `store.ErrNoTransaction` becomes `TransactionStatusNotFound`; any other error is logged and returned to the client as `jrpc2.InternalError`, for free.
- Return a wrapped error here instead of `store.ErrNoTransaction`, so this case gets the same treatment the v2 backend already gives the same kind of problem.

Code suggestion:
```go
if len(txnViewRange) == 0 {
    return store.Transaction{}, fmt.Errorf(
        "application_order %d does not resolve to a transaction in ledger %d (txhash=%s): index/meta mismatch",
        txIndex, ledgerSeq, hex.EncodeToString(hash[:]))
}
```

cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -74,149 +72,153 @@ func (h transactionsRPCHandler) initializePagination(
...
-			diagEvents, convErr := jsonifySlice(xdr.DiagnosticEvent{}, tx.Events)
-			if convErr != nil {
-				return nil, false, &jrpc2.Error{
-					Code:    jrpc2.InternalError,
-					Message: convErr.Error(),
-				}
-			}
+// transactionInfo renders one extracted transaction as a page entry. Pure
+// formatting.
+func transactionInfo(tx store.Transaction, format string) (protocol.TransactionInfo, error) {
```
karthikiyer56, line 175: consolidation suggestion
- This PR extracts `transactionInfo` out of `processTransactionsInLedger`, so the paginated handler renders one transaction in one place.
- `GetTransaction` in `get_transaction.go` still carries the same format switch inline, field for field.
- The two blocks will drift the next time someone adds a response field.
- Call `transactionInfo` from `GetTransaction`, then remove the inline block in `get_transaction.go`.

cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -74,149 +72,153 @@ func (h transactionsRPCHandler) initializePagination(
...
-			return nil, false, &jrpc2.Error{
-				Code:    jrpc2.InvalidParams,
-				Message: err.Error(),
-			}
+	// One walk of this ledger's TxProcessing, materializing only the page's
+	// worth of transactions. Apply indices are 0-based here; application order
+	// (and the cursor) is 1-based. A startTxIdx past the last transaction
+	// yields nothing, which is how a spent cursor lands on an empty page.
+	txViews, err := ingest.LedgerTransactionViewRange(lcmView, startTxIdx-1, remaining, h.networkPassphrase)
+	if err != nil {
```
karthikiyer56, line 145: This one is a very very minor semantic change

- The old code returned `jrpc2.InvalidParams` when it could not read a transaction out of a ledger.
- The new code returns `jrpc2.InternalError` for the same case.
- This actually matches the sibling handler's own convention, `get_ledgers.go:212-215` (unchanged by this PR) — https://github.com/stellar/stellar-rpc/blob/e4a6f8d982cdae94054e11a8cff0aabceb6f0732/cmd/stellar-rpc/internal/methods/get_ledgers.go#L212-L215 — a failure to parse our own stored ledger data is `InternalError` there too. The new code is more consistent with the rest of the package, not less.
- Still, a client that branched on the old `InvalidParams` code for this case sees a different code now. Worth one line in the PR description, since it is client-visible. and potentially in unreleased Changelog

cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -74,149 +72,153 @@ func (h transactionsRPCHandler) initializePagination(
...
-			}
-		}
 	}
-
-	txCount := ledger.CountTransactions()
 	cursor := toid.New(ledgerSeqInt32, 0, 1)
-	for i := startTxIdx; i <= txCount; i++ {
-		cursor.TransactionOrder = int32(i)
+	remaining := limitInt - len(*txns)
+	if remaining <= 0 {
```
karthikiyer56, line 136: - `if remaining <= 0 { return cursor, true, nil }` returns a cursor with `TransactionOrder` at `0`.
- No caller reaches this today: `processTransactionsInLedger` has exactly one call site, `readLedgerPage` at `cmd/stellar-rpc/internal/methods/get_transactions.go:86` — https://github.com/stellar/stellar-rpc/pull/962/files#diff-c2641387f86faddacb07386103234ef8ec7d8f64acda9943070bf54c52b62600R86 — and the outer loop stops as soon as `done` comes back `true`.
- A later change to the call order could hand a client a cursor at the wrong transaction.
- Set `cursor.TransactionOrder` before this early return, so the value stays correct even if a future caller re-enters this branch.

Code suggestion:
```go
cursor := toid.New(ledgerSeqInt32, int32(startTxIdx-1), 1)
remaining := limitInt - len(*txns)
if remaining <= 0 {
    return cursor, true, nil
}
```

cmd/stellar-rpc/internal/rpcv1/sqlitedb/transaction.go
cjonas9: good catch!

cmd/stellar-rpc/internal/methods/get_transactions.go
cjonas9: agreed that this should be noted. I recall this being an explicit choice I made, and I'd not remembered to note it. thanks for the catch!

cmd/stellar-rpc/internal/methods/get_transactions.go
cjonas9: implemented, thanks!

cmd/stellar-rpc/internal/methods/get_transactions.go
cjonas9: This was an outstanding suggestion. It went into GetTransaction pretty smoothly except for preserving the case of the hash in GetTransaction -- I'm sure this was wholly inconsequential, but just for the sake of not touching any more wire-bytes than this PR already has, I ensured the hash is copied from the response exactly as it was before this consolidation.

karthikiyer56 (approved):

cmd/stellar-rpc/internal/store/ledger.go
```diff
@@ -50,14 +50,20 @@ type WithLedgerRawFn func(raw []byte) error
 // LedgerReaderTx is a read-only snapshot of the ledger store. Call Done to
 // release it.
 //
-// GetLedger is a walk, not a free-form point read. Call it with ascending,
-// contiguous sequences, starting from the first call's sequence. Read at most
-// methods.LedgerScanLimit ledgers per Tx. This is getTransactions' access
-// pattern. The v1 (SQL) backend accepts any pattern; the v2 backend serves
-// the walk from a forward iterator primed on the first call and fails loudly
-// on anything else.
+// GetLedger and WithLedgerRaw share one walk, not free-form point reads. Calls
```
tamirms: After this PR nothing in production calls `LedgerReaderTx.GetLedger`: getTransactions moved to `WithLedgerRaw`, and the two remaining `GetLedger` callers in methods (simulate_transaction.go, util.go) go through `store.LedgerReader`, not the Tx. The only callers left are fourteen sites in the adapter tests, so this paragraph now explains how a method nobody calls shares a cursor with the one that is called. I would drop `GetLedger` from the Tx here: the v1 and v2 methods and the mock go, the adapter tests that drive the walk through it convert through a small helper that decodes inside the loan, and `TestTxWithLedgerRaw_SharesTheWalkWithGetLedger` goes with them. Then this paragraph describes one accessor. Five files, no production line added.

The larger question this paragraph raises, why the Tx has point-read accessors over what v2 serves as a scan, I would keep out of this PR. I have written it up as a follow-up: one `ScanLedgers(ctx, start, end) iter.Seq2[RawLedger, error]` on the Tx replaces both `WithLedgerRaw` and `BatchGetLedgers`, which deletes the walk, the span cap and the getLedgers per-page clone on v2 without changing a wire byte. Landing it after this PR and #976 lets the differential suite check the reshape instead of being rebased across it. It is written up as #1004; would you want to take it as the next PR once these two are in?

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
```diff
@@ -367,9 +379,10 @@ func getLedgerFromDB(ctx context.Context, db readDB, sequence uint32) (xdr.Ledge
 	return lcm, true, nil
 }
 
-// getLedgerRawFromDB is a helper function that encapsulates the common logic
-// for fetching a single ledger's bytes from the database.
-func getLedgerRawFromDB(ctx context.Context, db readDB, sequence uint32) ([]byte, bool, error) {
+// withLedgerRawFromDB is a helper function that encapsulates the common logic
+// for fetching a single ledger's bytes from the database. The bytes returned
+// are lent and shouldn't be reused by a later statement.
```
tamirms: The bytes this returns are owned, not lent: database/sql clones every BLOB it scans into a `*[]byte`, which is exactly what the `WithLedgerRaw` doc at line 95 relies on. So this comment contradicts that one, and the `with` prefix suggests a callback shape the function does not have. I would keep the old `getLedgerRawFromDB` name and say the bytes are owned, so `WithLedgerRaw` stays the only place in this file that talks about a loan.

cmd/stellar-rpc/internal/rpcv1/sqlitedb/transaction.go
tamirms: With the mismatch error in place, one step further would close the remaining gap: the positional read trusts `application_order`, so a value that is wrong but still in range serves another transaction's envelope, result and meta under the requested hash, and only a hash comparison catches that. `ingest.LedgerTransactionViewByHash(lcm, hash, txn.passphrase)` does the comparison for free and turns `!found` into this same mismatch error.

```go
txView, found, err := ingest.LedgerTransactionViewByHash(lcm, hash, txn.passphrase)
if err != nil {
	return store.Transaction{}, fmt.Errorf("failed to extract tx %s from ledger %d: %w",
		hex.EncodeToString(hash[:]), ledgerSeq, err)
}
if !found {
	return store.Transaction{}, fmt.Errorf("transactions table maps %s to ledger %d, which does not contain it",
		hex.EncodeToString(hash[:]), ledgerSeq)
}
return store.ParseTransactionView(txView), nil
```

The cost is the same: both extractors walk TxProcessing sequentially up to the target and both hash TxSet envelopes only until the wanted one turns up; ByHash adds a 32-byte read per entry on the way. It also covers the inner-hash rows `InsertTransactions` writes for fee bumps, since the extractor matches either hash. And it is the extractor the v2 read path and this PR's own `MockTransactionHandler` already use, so the three lookups share one code path. `application_order` would simply stop being read here. Fine as a follow-up if you would rather not touch this again now.

cmd/stellar-rpc/internal/store/ledger.go
cjonas9: > I would drop GetLedger from the Tx here: the v1 and v2 methods and the mock go, the adapter tests that drive the walk through it convert through a small helper that decodes inside the loan, and TestTxWithLedgerRaw_SharesTheWalkWithGetLedger goes with them. Then this paragraph describes one accessor. Five files, no production line added.

This is a good catch, I'd missed how few callers I'd whittled this down to.

> would you want to take it as the next PR once these two are in?

sure! happy to cut back on this PR's blast radius and defer this question for after this + the differential tests are merged in a separate PR.

tamirms (approved):

PR #967, opened 2026-08-29
rpcv2: version the on-disk format before the first release freezes it

tamirms:
The v2 store writes immutable cold artifacts, so whatever the first release ships can never be retrofitted: a file written without self-description stays that way forever. And today nothing stops a binary from misreading — or destroying — data written by a different version: the resolver treats any catalog state it doesn't recognize as missing work, so rolling back to an older binary after a format change would overwrite the newer binary's artifacts.

This PR adds the minimum that must exist before the format freezes:

**Version handshake.** `catalog.Open` checks every catalog key and value against the exact vocabulary this binary writes (`catalog/census.go`) and refuses to start — before writing anything — when it meets an entry it doesn't know, with a message distinguishing "a newer binary wrote this" from corruption. Upgrades stay seamless; rollback across a format change refuses instead of destroying data. Format-touching releases are roll-forward-only, by policy.

**Self-describing files.** `.bin` gains a magic + version prelude (it was the one cold file with none); txhash `.idx` metadata and ledger-pack app data gain leading version bytes; `index.pack` gains a build stamp (term-schema version + indexed-field bitmask), so "index lacks this term family" is distinguishable from "no matches". All three pack writers enable the packfile content hash, computed over pre-compression bytes so it survives zstd encoder changes.

**Pins against accidental format changes.** Every RocksDB column family pins its block-based-table `format_version` (a grocksdb upgrade could otherwise silently change on-disk bytes), and byte-level golden fixtures pin the term-key derivation, which is de facto on-disk format.

**The policy, in design docs:** what forces a format bump (a new indexed term family bumps the hot format too; dependency bumps that change RocksDB defaults count; write-path switches must flip freeze and walk together) and the per-tier upgrade shapes. The rest of the versioning design — `state@id` grammar, per-format read sets, capability gating — deliberately waits for the first release that actually changes a format.

Verification: the full `internal/rpcv2` tree is green and `golangci-lint --new-from-rev=feature/full-history` is clean at every commit. New tests pin the acceptance/refusal matrices (including first-start crash residues and secret redaction), each decoder's newer-version refusal, stamp round-trips, and per-pack hash verification.

Interactions: rebased onto #910; the build-stamp gate runs after #910's record-checksum gate in `OpenColdReader`. #902's zero-decompression freeze will need a decompress extract or an ingest-time hash handoff for the ledgers content hash when it rebases (recorded in the doc section).


cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
```diff
@@ -56,8 +57,46 @@ func OpenSnapshots() int64 { return openSnapshots.Load() }
...
+	"Column families not opened",
+	"Unknown Footer version",
+	"future feature not supported",
+}
+
+// wrapIfEngineTooNew re-labels an open or read failure whose error text
+// matches a known newer-RocksDB signature, so a rollback across a RocksDB
+// upgrade reads as "deploy the newer binary" instead of corruption. Any other
+// error (nil included) passes through unchanged.
+func wrapIfEngineTooNew(err error) error {
```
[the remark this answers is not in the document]
tamirms: Applied in b5d319b0: BatchMultiGet and LastKey now route through wrapIfEngineTooNew as well.

cmd/stellar-rpc/internal/rpcv2/stores/txhash/cold_merge.go
```diff
@@ -255,6 +255,17 @@ func newFileReader(path string, bufBytes int) (*fileReader, error) {
...
+	// prelude itself rather than trusting that scanAndValidate ran upstream.
+	if magic := string(buf[:4]); magic != coldBinMagic {
+		_ = f.Close()
+		return nil, fmt.Errorf("txhash: %s is not a cold txhash .bin (magic %q, want %q)", path, magic, coldBinMagic)
+	}
+	if buf[4] != coldBinVersion {
+		_ = f.Close()
+		return nil, fmt.Errorf("txhash: %s has .bin version %d unsupported by this binary "+
+			"(written by a newer stellar-rpc?)", path, buf[4])
+	}
```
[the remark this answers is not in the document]
tamirms: Applied in b5d319b0: the merge reader now enforces the reserved bytes as zero, matching scanBinHeader.

cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
```diff
@@ -56,8 +57,46 @@ func OpenSnapshots() int64 { return openSnapshots.Load() }
...
+	pinnedTableFormatVersion = 6
 )
 
+// engineTooNewSignatures are the RocksDB open-failure strings produced when a
+// database was written by a newer RocksDB than the linked one. They arrive
+// corruption-classed from the engine, so wrapIfEngineTooNew re-labels them.
+//
+//nolint:gochecknoglobals // immutable signature list
+var engineTooNewSignatures = []string{
+	"unsupported format_version",
```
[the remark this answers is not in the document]
tamirms: Checked against the linked librocksdb (strings over the dylib): the rollback path, an older library reading a newer-format SST footer, emits "Corrupt or unsupported format_version: ", which this entry matches verbatim and case-sensitively. "Unsupported BlockBasedTable format_version" is a different failure: options validation when a binary requests a format_version its own linked library does not support. That is a build misconfiguration rather than version skew, so the deploy-newer advice would be misdirected there and it is deliberately not matched.

cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
```diff
@@ -854,10 +861,10 @@ func (s *Store) applySharedTableOptions(cfNames []string, cfOpts []*grocksdb.Opt
 	}
 	for i, o := range cfOpts {
 		override := s.cfg.PerCFOptions[cfNames[i]]
-		if s.cache == nil && override.BloomFilterBitsPerKey == 0 && override.BlockSize == 0 {
-			continue
-		}
+		// Every CF gets an explicit BBTO so the on-disk table format is pinned
+		// (see pinnedTableFormatVersion) instead of riding grocksdb's default.
 		bbto := grocksdb.NewDefaultBlockBasedTableOptions()
+		bbto.SetFormatVersion(pinnedTableFormatVersion)
```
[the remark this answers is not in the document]
tamirms: The relabeling existed in an earlier revision and was removed deliberately (fd8c93f6): the pin prevents the drift, and the refusal scenario requires the pin doctrine to have been violated first. For that case RocksDB's own messages already carry the hypothesis ("Unknown Footer version. Maybe this file was created with newer version of RocksDB?"), and matching engine error text across library versions proved to be a fragility treadmill. The PR body no longer promises relabeling.

cmd/stellar-rpc/internal/rpcv2/catalog/census.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		if !geometry.IsKnownState(geometry.State(value)) {
+			return fmt.Sprintf("unknown artifact state %q", value), false
+		}
+	case strings.HasPrefix(key, geometry.TxHashIndexPrefix):
+		if _, ok := geometry.ParseTxHashIndexKey(key); !ok {
+			return "malformed index coverage key", false
+		}
+		if !geometry.IsKnownState(geometry.State(value)) {
+			return fmt.Sprintf("unknown artifact state %q", value), false
```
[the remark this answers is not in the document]
karthikiyer56: Fixed on the current head (ebfc7f0c): both coverage endpoints must map to the key's own index id, and TestCensus_RefusesForeignEntries has the "cross-window index coverage" case. Resolving.

cmd/stellar-rpc/internal/rpcv2/stores/event/index.go
```diff
@@ -26,6 +26,39 @@ const (
...
+	FieldContractID, FieldTopic0, FieldTopic1, FieldTopic2, FieldTopic3,
+	FieldEventType, FieldTopicCount,
+}
+
+// IndexedFieldMask is the set of indexed fields as a bitmask (bit i set means
+// Field i is indexed), derived from allFields; the build stamp records it per
+// artifact and the term-key golden tests iterate the same registry.
+//
+//nolint:gochecknoglobals // derived from the immutable field registry
+var IndexedFieldMask = func() uint64 {
```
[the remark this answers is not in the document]
karthikiyer56: Agree: the mask is on-disk identity and should not be reassignable from outside the package.

Rationale for the fix shape. The suggestion above is an explicit const. A const cannot be computed from a slice in Go, so it would be a second copy of the field list next to `allFields`, and the two could drift. An accessor keeps one list and still removes the mutability: unexport the var, keep the loop, add

```go
func IndexedFieldMask() uint64 { return indexedFieldMask }
```

The golden test already fails when a Field is added without a pin, so no second list and no extra test are needed.

cmd/stellar-rpc/internal/rpcv2/catalog/catalog.go
```diff
@@ -46,6 +46,14 @@ func Open(
 		return nil, err
 	}
 	c := &Catalog{store: store, logger: logger, layout: layout, txhashIndex: txhashIndex}
+	// Census before anything writes: a catalog holding entries outside this
+	// binary's vocabulary (a newer binary's formats, or corruption) is refused
+	// here, ahead of the secret mint below — Open must never write into a
+	// catalog it is about to refuse.
+	if err := c.census(); err != nil {
```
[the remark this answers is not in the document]
karthikiyer56: Agreed, "write-free" overstates it. Will reword the comment and the doc to "writes no catalog entry of its own". Not adding a read-only pre-open: it doubles every startup open for a path that is already safe by content.

cmd/stellar-rpc/internal/rpcv2/stores/event/index.go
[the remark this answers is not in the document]
karthikiyer56: Resolved in 934b6fe9.

cmd/stellar-rpc/internal/rpcv2/catalog/catalog.go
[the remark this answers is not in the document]
karthikiyer56: Resolved in 934b6fe9.

cmd/stellar-rpc/internal/rpcv2/stores/event/cold_reader.go
```diff
@@ -194,6 +196,20 @@ func OpenColdReader(chunkID chunk.ID, bucketDir string, opts ColdReaderOptions)
 			return fmt.Errorf("%w: %s: built without a record checksum (stale build)",
 				stores.ErrCorrupt, indexPackPath)
 		}
+		ad, aerr := c.index.AppData()
```
karthikiyer56: The stamp gate is three checks with their own error wrapping, sitting inside an open closure that already carries the format, checksum, and pair cross-checks. Should probably move into one `checkIndexBuildStamp(indexPackPath, r)` helper next to `decodeIndexBuildStamp`, so the open path reads as a list of gates and the stamp rules live beside their decoder. Same strings, same order.

cmd/stellar-rpc/internal/rpcv2/catalog/census_test.go
```diff
@@ -0,0 +1,133 @@
...
+// TestCensus_RefusesForeignEntries walks the refusal matrix: every entry shape
+// a newer binary (or corruption) could leave must fail reopen with
+// ErrForeignCatalog and an actionable message.
+func TestCensus_RefusesForeignEntries(t *testing.T) {
+	cases := []struct {
+		name, key, value string
+	}{
+		{"novel prefix", "format:events", "2"},
+		{"novel meta key", "meta/other", "WOULD-BE-SECRET-BYTES"},
+		{"unknown kind under chunk prefix", "chunk:00000001:bogus", "frozen"},
```
karthikiyer56: These four rows spell "frozen" as a literal while the rows above them use the geometry constants. Should probably be `string(geometry.StateFrozen)` too, so the refusal table is tied to the token it is testing against and a renamed state cannot leave the fixture behind.

cmd/stellar-rpc/internal/rpcv2/stores/txhash/cold_merge.go
```diff
@@ -255,6 +255,22 @@ func newFileReader(path string, bufBytes int) (*fileReader, error) {
 		_ = f.Close()
 		return nil, fmt.Errorf("txhash: %s too short for header (%d bytes)", path, n)
 	}
+	// Self-defending: the merge consumes raw entry bytes, so it verifies the
+	// prelude itself rather than trusting that scanAndValidate ran upstream.
+	if magic := string(buf[:4]); magic != coldBinMagic {
```
karthikiyer56: Should probably be one `checkBinPrelude(path, hdr []byte) error` in `cold_bin.go`, called from here and from `scanBinHeader`.

- The three checks and their three error strings are typed out twice, here and in `cold_index.go`.
- This reader is the one that must not drift: it consumes raw entry bytes and, per its own comment, does not trust the upstream scan.
- `TestColdBin_ScanRejectsForeignHeader` says "BOTH .bin consumers" but only the reserved-byte case runs against `newFileReader`. A table of the three mutations asserted on both readers covers it in fewer lines.

cmd/stellar-rpc/internal/rpcv2/stores/ledger/cold_reader_test.go
```diff
@@ -182,8 +182,9 @@ func TestColdReader_RejectsWrongAppDataSize(t *testing.T) {
 	})
 	require.NoError(t, err)
 	require.NoError(t, pw.AppendItem([]byte("v")))
-	// 7-byte payload — appDataSize is 4.
-	require.NoError(t, pw.Finish([]byte("seven-b")))
+	// Valid version byte, 7-byte payload; appDataSize is 5, so the size
+	// check (which runs after the version check) is what fires.
+	require.NoError(t, pw.Finish([]byte{coldAppDataVersion, 's', 'e', 'v', 'e', 'n', 'b'}))
```
karthikiyer56: Should probably add a second case here: a 1-byte blob `{coldAppDataVersion}`.

- Test-only change. `loadHeader` already handles it.
- Today the only wrong-size case is too long (7 bytes). Nothing covers 1 to 4 bytes with a valid version byte.
- That is the one case where `binary.BigEndian.Uint32(ad[1:])` depends on the `len(ad) != appDataSize` guard. If the length check ever folds into `CheckBlobVersion`, which only rejects empty, a 1-byte blob panics inside `LastSeq`.

cmd/stellar-rpc/internal/rpcv2/geometry/keys.go
```diff
@@ -62,6 +62,28 @@ var allKinds = []Kind{KindLedgers, KindEvents, KindTxHash}
...
+)
+
+// AllStates returns the artifact lifecycle states.
+func AllStates() []State { return append([]State(nil), allStates...) }
+
+// AllHotStates returns the hot-DB lifecycle states.
+func AllHotStates() []HotState { return append([]HotState(nil), allHotStates...) }
+
+// IsKnownState reports whether s is a registered artifact lifecycle state.
+func IsKnownState(s State) bool { return slices.Contains(allStates, s) }
```
karthikiyer56: Nit: `IsKnownState` and `IsKnownHotState` should probably be a `switch` over the named constants rather than `slices.Contains` over the registry.

- The comment above says a token added to the registry is automatically accepted. The failure mode is the other way round: a constant added to the const block but not to `allStates` is written by the daemon, then refused by the census on the next restart as foreign.
- `TestCensus_AcceptsOwnVocabulary` cannot catch that, since it iterates the same slice.
- A `switch` names each constant at the point of the check, so the omission is visible where the decision is made. `allStates` stays for `AllStates()` and the tests.

cmd/stellar-rpc/internal/rpcv2/stores/event/cold_reader.go
karthikiyer56: Resolved in cf4510d1.

cmd/stellar-rpc/internal/rpcv2/catalog/census_test.go
karthikiyer56: Resolved in cf4510d1.

cmd/stellar-rpc/internal/rpcv2/stores/txhash/cold_merge.go
karthikiyer56: Resolved in cf4510d1.

cmd/stellar-rpc/internal/rpcv2/stores/ledger/cold_reader_test.go
karthikiyer56: Resolved in cf4510d1.

cmd/stellar-rpc/internal/rpcv2/geometry/keys.go
karthikiyer56: Resolved in cf4510d1.

cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
```diff
@@ -56,6 +56,13 @@ func OpenSnapshots() int64 { return openSnapshots.Load() }
 const (
 	dirPerm       os.FileMode = 0o700
 	defaultCFName             = "default"
+
+	// pinnedTableFormatVersion pins the block-based table format written to
+	// disk, so a grocksdb/librocksdb upgrade cannot silently change the on-disk
+	// format. Raising it is a format-touching change: an older binary cannot
+	// open the newer tables, so it must ship as a declared storage-format bump,
+	// never as a side effect of a dependency bump.
+	pinnedTableFormatVersion = 6
```
karthikiyer56, line 69: Nit, comment trims. Should probably land as one small commit:

- This constant: add that 6 is the default in librocksdb 10.9.1, the version `scripts/install-rocksdb.sh` builds, so the pin changes no byte today, and that we pin on purpose against RocksDB's "leave it at the default" advice, because an unpinned default bump would strand a rollback binary.
- The "every CF gets an explicit BBTO so the format stays pinned" sentence appears four times in this file and its test. Keep this one and the one-line trap warning on `applySharedTableOptions`; drop the inline loop copy and the test copy.
- `stores/event/index.go`, `TermSchemaVersion`: "events format id", "per-id read sets", and "the format-id grammar" do not exist in code yet. Trim to the bump rule and point to the design doc.
- `stores/ledger/cold_reader.go`: `appDataSize = 1 + 4` with field names, like the other blob widths in this PR..

cmd/stellar-rpc/internal/rpcv2/stores/txhash/cold_format.go
```diff
@@ -55,28 +60,34 @@ func ColdIndexSecret(catalogSecret []byte, indexID uint32) [stores.SecretLen]byt
...
 	return buf
 }
 
 // ParseColdMetadata recovers [minLedger, maxLedger, secret] from the metadata
-// blob, rejecting a wrong size or maxLedger < minLedger with ErrInvalidMetadata.
+// blob, rejecting a wrong size or maxLedger < minLedger with
+// ErrInvalidMetadata and an unknown version byte with its own message.
 func ParseColdMetadata(metadata []byte) (uint32, uint32, [stores.SecretLen]byte, error) {
 	var secret [stores.SecretLen]byte
+	if err := stores.CheckBlobVersion(metadata, coldMetadataVersion); err != nil {
```
karthikiyer56, line 79: Nit: the doc above says a wrong size returns `ErrInvalidMetadata`, but size zero now returns the bare version error. Same for `ErrShortLedgerOffsets` in `event.DecodeLedgerOffsets`.

- Should probably wrap: `fmt.Errorf("%w: %w", ErrInvalidMetadata, err)`. The "written by a newer stellar-rpc?" text stays, `errors.Is` works again.
- Then the three tests can go back from `ErrorContains("empty")` to `ErrorIs`, which is what they asserted before.

design-docs/full-history-streaming-workflow.md
```diff
@@ -1043,8 +1043,25 @@ A storage walk against the invariants is enough to find these without inspecting
...
+
+**Roll-forward-only.** A release that changes any storage format, adds a catalog key family, or changes a value form cannot be rolled back past: the older binary refuses the tree, and the remedy is redeploying the newer binary. Release notes for a format-touching release must say so.
+
+**Where format identity lives.** The catalog value that makes an artifact visible also names its format: the planned grammar widens `"frozen"` to `"frozen@id"`, with a bare value meaning format 1 forever. Release 1 writes only bare values, so the grammar ships with the first bump. The grammar applies to the three state families only, never to `config:*` or `meta/*` values (the secret's random bytes can contain `@`). The files themselves stay self-describing as witnesses: the packfile trailer carries a container version and a per-store `Format` id, the streamhash files carry their own header version, and every app-data blob leads with its own version byte.
```
karthikiyer56: Three sentences should probably be added to this paragraph:

- Which blobs may grow without a version bump. The `index.pack` stamp decoder ignores trailing bytes; the ledger app data and the txhash `.idx` metadata require the exact length. Both rules are fine, but the next extension should not pick by accident.
- Who verifies the content hash. Nothing on the serving path calls `Verify`; it exists for offline tooling and repair.
- A `TermSchemaVersion` or field-mask change must ship as a new events format id. Nothing ties the stamp to `indexPackFormat` today, and a stamp-only bump leaves every chunk `frozen` in the catalog with an index that fails on every filtered query.

cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
karthikiyer56: Fixed in 615c65e5.

cmd/stellar-rpc/internal/rpcv2/stores/txhash/cold_format.go
karthikiyer56: Fixed in 615c65e5.

design-docs/full-history-streaming-workflow.md
karthikiyer56: Fixed in 615c65e5.

cmd/stellar-rpc/internal/rpcv2/rocksdb/rocksdb.go
```diff
@@ -56,6 +56,16 @@ func OpenSnapshots() int64 { return openSnapshots.Load() }
 const (
 	dirPerm       os.FileMode = 0o700
 	defaultCFName             = "default"
+
+	// pinnedTableFormatVersion pins the block-based table format written to
+	// disk, so a grocksdb/librocksdb upgrade cannot silently change the on-disk
+	// format. 6 is librocksdb 10.10.1's default, so the pin changes no byte
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in e2793963. The 10.9.1 header also defaults format_version to 6, so the pin still matches the linked default.

PR #971, opened 2026-09-02
Optimize integration test run times

karthikiyer56:
## What this PR does

Makes the integration-test job faster and stops it failing at random. Three changes, all in `cmd/stellar-rpc/internal/rpcv1` and the CI workflow. No change to rpcv2.

| # | change | effect |
|---|---|---|
| 1 | Close captive core before waiting for the ingestion worker | integration package 22m 56s → 11m 01s (P27), 18m 05s → 11m 30s (P28) |
| 2 | Print per-test timings in the CI job summary | shows which tests are slow, on every run |
| 3 | Remove three causes of random test failures | about 5 in 32 jobs red → about 1 in 32 |

## Change 1: the daemon took minutes to shut down

**Which test showed it.** `TestArchiveUserAgent` points the daemon at a history archive that answers HTTP 500 to everything, so the daemon never leaves catch-up and every shutdown hits the stall. Every other test hits the same stall for a shorter time. That test now takes 0.09s (P27) and 0.15s (P28).

| step | old order | new order |
|---|---|---|
| 1 | wait for the ingestion worker to finish | cancel ingestion's context, which does not block |
| 2 | stop stellar-core | stop stellar-core, which frees the worker |
| 3 | | wait for the worker, which now exits at once |

- The worker sits inside a blocking `stellar-core catchup` call. Only stopping stellar-core can interrupt it. The old order waited on the worker before doing the one thing that could free it, so shutdown lasted until stellar-core exited on its own. On a busy 4-CPU runner: minutes, in 45 tests.
- The new order is the one you would pick without knowing the bug: stop the thing the worker is waiting on, then wait for the worker. A worker cannot finish while the subprocess it is blocked on is alive.
- Cancelling first matters. Stopping stellar-core while ingestion is live makes `GetLedger` error, the retry loop exhausts, and `logger.Fatal` kills the process.
- This is a production shutdown bug: any node restarted mid-catch-up stalls the same way. rpcv2 does not have it; its iterator tears down its own core process.

| | P27 | P28 |
|---|---|---|
| before this PR, [sample 1](https://github.com/stellar/stellar-rpc/actions/runs/33665308366) | [22m 55.7s](https://github.com/stellar/stellar-rpc/actions/runs/33665308366/job/100365471562) | [18m 05.1s](https://github.com/stellar/stellar-rpc/actions/runs/33665308366/job/100365471551) |
| before this PR, [sample 2](https://github.com/stellar/stellar-rpc/actions/runs/33532435250) | [20m 35.1s](https://github.com/stellar/stellar-rpc/actions/runs/33532435250/job/99938589444) | [23m 35.9s](https://github.com/stellar/stellar-rpc/actions/runs/33532435250/job/99938590149) |
| after this PR, [sample 1](https://github.com/stellar/stellar-rpc/actions/runs/33677405497) | [**11m 00.5s**](https://github.com/stellar/stellar-rpc/actions/runs/33677405497/job/100405446617) | [**11m 29.6s**](https://github.com/stellar/stellar-rpc/actions/runs/33677405497/job/100405446712) |
| after this PR, [sample 2](https://github.com/stellar/stellar-rpc/actions/runs/33802748148) | [**11m 00.4s**](https://github.com/stellar/stellar-rpc/actions/runs/33802748148/job/100805895431) | [**11m 34.6s**](https://github.com/stellar/stellar-rpc/actions/runs/33802748148/job/100805895739) |
| after this PR, [sample 3](https://github.com/stellar/stellar-rpc/actions/runs/33812529055) | [**11m 00.6s**](https://github.com/stellar/stellar-rpc/actions/runs/33812529055/job/100837406804) | [**12m 07.6s**](https://github.com/stellar/stellar-rpc/actions/runs/33812529055/job/100837407008) |

Before: 8 samples, 18m 05s to 23m 36s. After: 8 samples, 11m 00s to 12m 08s, all about 40% below the lowest "before" value.

## Change 2: per-test timings in the job summary

The log printed one package total. Now `go test -v` runs, and a step writes the 20 slowest tests and the package total to the job summary on every run.

## Change 3: the random failures

Sampled at **job** level, because the workflow sets `continue-on-error: true` and `gh run list` says success while the job failed: 16 runs, 32 jobs, 5 red (P27 1, P28 4), 16 failed tests, none failing consistently. Every failure sits in environment setup or transaction submission, never in an assertion about the behaviour under test.

| cause | failed tests | fixed here |
|---|---|---|
| Core answered `TRY_AGAIN_LATER` | 13 | yes |
| the protocol upgrade was dropped | 1 | yes |
| a port collision killed captive core | 1 | yes |
| simulation asked Core for a snapshot it did not have yet | 1 | no |

Wiring, for the rows below: the **validator** runs in a container and closes ledgers; **captive core** runs inside the RPC daemon, follows the validator, and feeds RPC; tests submit through RPC to the validator.

### Cause 1: `TRY_AGAIN_LATER`, 13 of 16, fixed

| | |
|---|---|
| what the harness does | Submits four "raise the Soroban limits" transactions from the root account, one after the other, waiting for RPC to say "applied" before each next one. |
| the gap | "Applied" to RPC means captive core applied the ledger. On a busy runner the validator applies it a few hundred milliseconds later, and until then its queue still holds the previous transaction. Core allows one pending transaction per account and answers `TRY_AGAIN_LATER` for a second: "nothing was queued, submit it again". |
| what the harness did | Treated that as a hard failure. |
| fix | New helper `infrastructure.SendTransaction` resubmits the same signed transaction every 500ms for up to 30s and logs each retry. On a healthy run the first attempt succeeds. |

[Job 100365471551](https://github.com/stellar/stellar-rpc/actions/runs/33665308366/job/100365471551): tx 1 lands in ledger 23 at 14.137s, RPC reports it applied at 14.475s, tx 2 is sent at 14.478s and gets `TRY_AGAIN_LATER`.

### Cause 2: dropped protocol upgrade, 1 of 16, fixed

| | |
|---|---|
| what went wrong | The harness polls `/info` until the validator answers, then arms the upgrade. In [job 100490569892](https://github.com/stellar/stellar-rpc/actions/runs/33704478634/job/100490569892), `/info` answered at ledger 1, before Core had finished booting. Core accepted the request and lost it. The network ran 28 ledgers on protocol 0, so a longer wait would not have helped. |
| fix | `UpgradeProtocol` re-sends the request on every poll. Sending it twice is harmless. |

### Cause 3: port collision, 1 of 16, fixed

| | |
|---|---|
| what went wrong | Captive core needs fixed ports in its config. The harness took two free ports from the kernel's ephemeral range, the pool every outgoing connection uses. In [job 100494963490](https://github.com/stellar/stellar-rpc/actions/runs/33705948339/job/100494963490) a client socket took the port before Core bound it. Captive core died with `bind: Address already in use` and `TestBackfillEmptyDB` waited 60s for ledgers that never came. |
| fix | Ports come from 20000-29999, outside the ephemeral and Docker ranges, with a process-wide counter so tests in one run cannot collide. |

### Cause 4: missing query snapshot, 1 of 16, not fixed

| | |
|---|---|
| what went wrong | In [job 100422271810](https://github.com/stellar/stellar-rpc/actions/runs/33682426774/job/100422271810), RPC had ingested ledger 32 before captive core had published its query snapshot for 32. Simulation asked for entries at 32, Core answered 404: `setting ContractComputeV0 is not present in the snapshot`. |
| why not here | A race inside the RPC daemon (`ledgerentries.coreLedgerEntryGetter`), not in the harness. Needs its own issue. |

### Failure rate

| | integration jobs | red |
|---|---|---|
| before this PR, last 16 runs | 32 | 5 |
| after this PR, 2 runs so far ([run 1](https://github.com/stellar/stellar-rpc/actions/runs/33802748148), [run 2](https://github.com/stellar/stellar-rpc/actions/runs/33812529055)) | 4 | 0 |

- The "after" row grows with each run on this branch. The one cause not fixed here (cause 4) hit 1 of the 32 jobs before.
- In [run 2, P28](https://github.com/stellar/stellar-rpc/actions/runs/33812529055/job/100837407008) the retry fired once, in `TestSimulateTransactionUnmarshalError`: Core answered `TRY_AGAIN_LATER`, the second attempt was accepted, the test passed. Before this PR that job would have been red.

## Testing

| what | result |
|---|---|
| `go test -race ./cmd/stellar-rpc/internal/rpcv1/ingest/... ./cmd/stellar-rpc/internal/util/...` | pass |
| integration jobs on this branch, 4 runs, 8 jobs | 7 green; the 1 red was cause 1, before the fix |
| local integration suite after the fix, 2 runs | 41 pass each |


cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/client.go
```diff
@@ -53,6 +53,29 @@ func logTransactionResult(t testing.TB, response protocol.GetTransactionResponse
...
+// soon as captive core has applied it, which on a loaded CI runner can be a few
+// hundred milliseconds before the validator has. Nothing was queued, so the
+// same transaction is submitted again after a short wait.
+func SendTransaction(t testing.TB, client *client.Client, b64 string) protocol.SendTransactionResponse {
+	const maxAttempts = 60
+	request := protocol.SendTransactionRequest{Transaction: b64}
+	var result protocol.SendTransactionResponse
+	for attempt := 1; ; attempt++ {
+		var err error
+		result, err = client.SendTransaction(context.Background(), request)
```
Shaptic:
```suggestion
		result, err = client.SendTransaction(t.Context(), request)
```

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/client.go
```diff
@@ -53,6 +53,29 @@ func logTransactionResult(t testing.TB, response protocol.GetTransactionResponse
...
+// Core answers TRY_AGAIN_LATER when the source account's previous transaction
+// is still in its queue. RPC reports that previous transaction as applied as
+// soon as captive core has applied it, which on a loaded CI runner can be a few
+// hundred milliseconds before the validator has. Nothing was queued, so the
+// same transaction is submitted again after a short wait.
+func SendTransaction(t testing.TB, client *client.Client, b64 string) protocol.SendTransactionResponse {
+	const maxAttempts = 60
+	request := protocol.SendTransactionRequest{Transaction: b64}
+	var result protocol.SendTransactionResponse
+	for attempt := 1; ; attempt++ {
```
Shaptic: Never gives up?

Shaptic: Er wait it does but it doesn't express it in the normal `for` loop way

.github/workflows/integration-tests.yml
```diff
@@ -153,5 +153,29 @@ jobs:
...
+      # single package total, which is not enough to tell a slow suite from one
+      # slow test.
       - name: Run Stellar RPC Integration Tests
-        run: go test -race -timeout 30m ./cmd/stellar-rpc/internal/rpcv1/integrationtest/...
+        run: |
+          set -o pipefail
+          go test -race -v -timeout 30m ./cmd/stellar-rpc/internal/rpcv1/integrationtest/... 2>&1 \
+            | tee "${RUNNER_TEMP}/integration-tests.log"
+
+      - name: Report the slowest tests
```
Shaptic, line 165: Is this necessary?

Shaptic (approved):

.github/workflows/integration-tests.yml
karthikiyer56: I am trying to expedite a few things.
things run fast in local docker container but slower in the 4-cpu github runners.
I have more work slated for "making things faster" and for that I need to know whats the holdup in the CI runs.
and for that this is needed 🙈

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/client.go
karthikiyer56: done

cmd/stellar-rpc/internal/rpcv1/integrationtest/infrastructure/client.go
karthikiyer56: fixed

karthikiyer56: thanks for catching
one of the several instances where AI seems to think its cool to do wierd things instead of a simple `for`

.github/workflows/integration-tests.yml
Shaptic: fair nuf!

PR #976, opened 2026-09-04
Add getTransactions differential test suite

### LINKED ISSUE #987: Adopt XDR views: add differential test suites for ensuring byte-equal JSON responses
### What

Each XDR-view migration swaps the decode-based path behind a v1 method for a view-walk over raw ledger bytes. The wire contract must not move and a client cannot tell the two backends apart. This issue tracks the differential suites that prove it, one per migrated method, by freezing the pre-view path test-locally and asserting it and the view path produce byte-identical response JSON over a corpus that sweeps the shapes the migration touches. 

This should ensure _byte_ equality, not semantic equality, so if a field appears, vanishes or reorders, the differential test suite fails.

- `getTransactions` (#948 / #962): #976
- `getEvents` (#950 / #978): #982

### Acceptance criteria

- [x] #976 merged: `getTransactions` suite plus the shared machinery in `differential_test.go` (generic comparator, cursor-chain driver, corpus builders, sqlite seeding).
- [x] #982 merged: `getEvents` suite built on that machinery, with the SDK's `GetEventsRequest.Matches` frozen locally rather than called.
    - [ ] deprecate/remove unused legacy helpers in the SDK after
- [x] Each suite's reference path is frozen in test code and no production code depends on it.
- [x] Each suite pins its own corpus and filter table as non-vacuous, so a silently empty or one-shaped fixture cannot pass.
- [x] Both suites pass and lint clean on `feature/full-history` once the stack lands.

cjonas9:
### What

This PR adds the `getTransactions` differential suite out of #962 into its own PR, stacked on the branch `getTxByHash-views`. It's split off from #962 for organization purposes, and all work relating to actually using views in the transaction read path is left over there.

`get_transactions_differential_test.go` reconstructs the pre-view decode path test-locally and asserts that it and the view-walk page loop produce byte-identical `GetTransactionsResponse` JSON over a corpus that sweeps LCM V0/V1/V2, `TransactionMeta` V1/V2/V3/V4, `classic`/`Soroban`/`fee-bump` envelopes, event shapes, empty ledgers, page boundaries and cursor round-trips (~400 subtests).

`differential_test.go` holds the method-agnostic machinery factored out of the suite: a generic two-sided comparator and cursor-chain driver, the corpus builders, and sqlite seeding. The upcoming `getEvents` differential suite (in PR #982) reuses it unchanged.

The suite is @tamirms's work from `tamirms/gettransactions-view-walk`, but the shared-helper refactor is new.

Retarget to `feature/full-history` once #962 merges.

### Why

The suite validates #962's migration exhaustively, and ensures no regression of the arity fixes in stellar/go-stellar-sdk#5997. At ~900 lines, this suite obscured the few hundred load-bearing lines within that PR. Reviewing it separately keeps #962 focused on the read-path change, testing those changes more narrowly, and lets the corpus be evaluated on its own.

### Known Limitations
N/A


cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,648 @@
...
+	cursor := toid.New(0, 0, 0)
+	endLedger := min(int64(ledgerRange.LastLedger.Sequence), int64(start.LedgerSequence)+LedgerScanLimit-1)
+	for ledgerSeq := start.LedgerSequence; int64(ledgerSeq) <= endLedger; ledgerSeq++ {
+		if ledgerSeq < 0 {
+			return protocol.GetTransactionsResponse{}, &jrpc2.Error{
+				Code:    jrpc2.InvalidParams,
+				Message: "cursor ledger sequence cannot be negative",
+			}
+		}
+		ledger, found, gerr := readTx.GetLedger(ctx, uint32(ledgerSeq))
```
tamirms: `LedgerReaderTx.GetLedger` was removed in #962 at 735e19b8, and this branch is one commit behind that, so the green CI here ran against a base that no longer exists: after the rebase this file does not compile. The reference wants a decoded LCM, so decode inside the loan and it stays byte for byte the old read:

```go
var ledger xdr.LedgerCloseMeta
found, gerr := readTx.WithLedgerRaw(ctx, uint32(ledgerSeq), ledger.UnmarshalBinary)
```

The doc on lines 38 to 40 names `LedgerReaderTx.GetLedger` as well.

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,648 @@
...
+// ---- the corpus ----
+//
+
+// transactionsCorpus is the contiguous ledger run [transactionsCorpusFirst,
+// transactionsCorpusLast] the differential sweeps over.
+// Every ledger is a deliberate shape; the comment on each says which axis it
+// covers. Account sequence numbers are unique across the corpus so no two
+// transactions share an envelope hash (the TxSet is paired to TxProcessing by
+// hash, so a collision would hide a mispairing).
+func transactionsCorpus(t *testing.T) []xdr.LedgerCloseMeta {
```
tamirms, line 254: Every cell here is LCM V1 or V2 with a generalized set of one V0 phase, meta V1, V3 or V4, and `TX` or `TX_FEE_BUMP` envelopes. That leaves out both ends of what full history serves. Before protocol 20, every ledger is LCM V0 with a plain `TransactionSet`, which the view walks through a separate path (`v0TxSetEnvelopes`); protocol 13 to 19 meta is V2; and pre-13 transactions are `TX_V0` envelopes, which both hashers special-case. At the other end, protocol 23 puts Soroban transactions in a `TransactionPhase` V1 with stages and clusters, and the view has its own `case 1` for it that no cell reaches. The legacy reader handles all four already (`LedgerCloseMeta.TransactionEnvelopes`), so they slot straight in: a `version == 0` branch in `diffLCM`, a `diffMetaV2()`, a `diffV0Envelope(acctSeq)`, and one V2 ledger whose set carries both phases:

```go
Phases: []xdr.TransactionPhase{
	{V: 0, V0Components: &classicComponents},
	{V: 1, ParallelTxsComponent: &xdr.ParallelTxsComponent{
		ExecutionStages: []xdr.ParallelTxExecutionStage{
			{xdr.DependentTxCluster{sorobanA}},
			{xdr.DependentTxCluster{sorobanB, sorobanC}},
		},
	}},
},
```

Two smaller things in the same spirit. Meta V0 is the one intentional divergence of this migration and belongs here as a pinned divergence cell (reference errors, new path serves) rather than only in the handler test. And the multi-event cells use `{ev, ev}`, so a reversed event order inside an operation is invisible; Copilot's open comment has that one. For the record, the suite on my branch had the same gaps, so this is an ask, not a regression of the port.

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,648 @@
...
+// from an explicit cursor, including cursors that land mid-ledger, on a
+// ledger's last transaction, past a ledger's last transaction, and past the
+// tip (where the request's own cursor has to be echoed back).
+func TestGetTransactions_ViewWalkMatchesParsedPath_Cursors(t *testing.T) {
+	diff := seededTransactionsDifferential(t)
+
+	for ledger := transactionsCorpusFirst; ledger <= transactionsCorpusLast; ledger++ {
+		for txOrder := range 7 {
+			for _, limit := range []uint{1, 3, 10} {
+				cursor := toid.New(int32(ledger), int32(txOrder), 1).String()
```
tamirms: Every cursor in the sweep has operation order 1, and that is the one axis on which the two sides now differ. #962 (14814e4e) initializes the per-ledger cursor at `(L, startTxIdx-1, 1)` where the reference keeps `(L, 0, 1)`. For a client-built cursor `(L, N, 0)` with N above 0 whose ledger is the tip and holds at most N transactions, the reference echoes `(L, N, 0)` and the new path answers `(L, N, 1)`: same paging semantics, different bytes. Adding op 0 and op 2 to this sweep lets the suite see it, and then it needs a deliberate decision: pin it here as a divergence cell, or narrow #962's init back to the early-return branch so the bytes match. I lean to the second, since it is one line and keeps the changelog honest, but either works as long as the arbiter records it.

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,648 @@
...
+func seededTransactionsDifferential(t *testing.T) transactionsDifferential {
+	t.Helper()
+	corpus := transactionsCorpus(t)
+	require.Len(t, corpus, transactionsCorpusLast-transactionsCorpusFirst+1)
+	return newTransactionsDifferential(seedDifferentialDB(t, corpus))
+}
+
+// TestGetTransactions_ViewWalkErrorParity pins that both paths reject an
+// invalid request identically: same message, same jrpc2 code.
+func TestGetTransactions_ViewWalkErrorParity(t *testing.T) {
```
tamirms, line 422: The two cells cover the start-ledger bounds; the cursor side has none, and one of its cases is the pin the follow-up will need. A cursor below the oldest ledger (a poller resuming after retention moved past it) reaches the walk because `ValidatePagination` does not range-check cursors, and both sides must answer InvalidParams "database does not contain metadata for ledger: N". Worth a cell here, alongside a malformed cursor string, a limit above `maxLimit` and an invalid format, all cheap with `assertSameError`.

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,648 @@
...
+	protocol "github.com/stellar/go-stellar-sdk/protocols/rpc"
+	"github.com/stellar/go-stellar-sdk/toid"
+	"github.com/stellar/go-stellar-sdk/xdr"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv1/sqlitedb"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/store"
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/xdr2json"
+)
+
+// This file is the correctness harness for the getTransactions page loop's
```
tamirms, line 28: One line for this header: both sides read sqlite, so what the suite pins is the shared handler's extraction, not the v2 adapter's walk, which has its own tests under rpcv2/adapters. Worth stating, since the description reads as validating the migration on both backends.

cmd/stellar-rpc/internal/methods/get_transactions.go
```diff
@@ -131,9 +131,10 @@ func (h transactionsRPCHandler) processTransactionsInLedger(
 	if ledgerSeqInt32 == start.LedgerSequence {
 		startTxIdx = int(start.TransactionOrder)
 	}
-	cursor := toid.New(ledgerSeqInt32, int32(startTxIdx-1), 1)
+	cursor := toid.New(ledgerSeqInt32, 0, 1)
 	remaining := limitInt - len(*txns)
 	if remaining <= 0 {
+		cursor.TransactionOrder = int32(startTxIdx - 1)
 		return cursor, true, nil
```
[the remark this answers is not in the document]
cjonas9: intended behavior, hardening

tamirms (approved):
looks like there are some lint failures? but apart from that everything looks good!

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,791 @@
...
+// phases, envelope shapes, event shapes, page boundaries, empty ledgers and
+// cursor round-trips.
+// Shared machinery lives in differential_test.go.
+
+// legacyGetTransactionsByLedgerSequence is the pre-view-walk pagination loop:
+// same handler, same cursor math, but reading each ledger's bytes through
+// LedgerReaderTx.WithLedgerRaw and decoding them into an xdr.LedgerCloseMeta.
+func legacyGetTransactionsByLedgerSequence(
+	ctx context.Context, h transactionsRPCHandler, request protocol.GetTransactionsRequest,
+) (protocol.GetTransactionsResponse, error) {
```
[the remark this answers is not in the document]
cjonas9: this is untrue. the linter currently passes.

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,791 @@
...
+// ---- the corpus ----
+//
+
+// transactionsCorpus is the contiguous ledger run [transactionsCorpusFirst,
+// transactionsCorpusLast] the differential sweeps over.
+// Every ledger is a deliberate shape; the comment on each says which axis it
+// covers. Account sequence numbers are unique across the corpus so no two
+// transactions share an envelope hash (the TxSet is paired to TxProcessing by
+// hash, so a collision would hide a mispairing).
+func transactionsCorpus(t *testing.T) []xdr.LedgerCloseMeta {
```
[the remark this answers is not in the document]
cjonas9: untrue

cmd/stellar-rpc/internal/methods/get_transactions_differential_test.go
```diff
@@ -0,0 +1,767 @@
...
+	}
+	if ledgerSeqInt32 == start.LedgerSequence {
+		startTxIdx = int(start.TransactionOrder)
+		if ierr := reader.Seek(startTxIdx - 1); ierr != nil && !errors.Is(ierr, io.EOF) {
+			return nil, false, &jrpc2.Error{Code: jrpc2.InternalError, Message: ierr.Error()}
+		}
+	}
+
+	txCount := ledger.CountTransactions()
+	cursor := toid.New(ledgerSeqInt32, 0, 1)
```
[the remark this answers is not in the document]
cjonas9: This change incorrectly breaks the differential. `cursor.TransactionOrder = startTxIdx - 1` sits inside the remaining `<= 0 `guard in `get_transactions.go:139`. That branch means the page was already full before this ledger was touched. It is unreachable from the outer loop, since the previous ledger returns done as soon as the page fills, and a zero limit is replaced by the default. In the case Copilot describes, a cursor past the last transaction of the starting ledger, production falls through to the view range walk, which yields nothing, and returns transaction order 0. The reference does exactly the same. The two paths already agree.

PR #978, opened 2026-09-04
RPCv1 XDR views: migrate getEvents

### LINKED ISSUE #950: Adopt XDR Views: View-based event filtering + response
## Summary

The event scan loop Unmarshals on every row regardless of whether or not the `scanner` callback filters the events. This decoding work can be avoided.

### Scope

Files: `cmd/stellar-rpc/internal/db/event.go`, `cmd/stellar-rpc/internal/methods/get_events.go`

- `db/event.go:428` event scan loop: event.UnmarshalBinary(row.eventData) per row. 
    - We should wrap `row.eventData` as `DiagnosticEventView` and run filter checks against view accessors (contract ID, type, topic patterns)
    - Only call `event.UnmarshalBinary` when filter passes AND the scanner callback genuinely needs the decoded form
- `get_events.go:296`: `xdr.MarshalBase64(segment)` for each topic in non-JSON response output + `get_events.go:304`: `xdr.MarshalBase64(v0.Data)` for event data in non-JSON response output
    - We should slice topic and data bytes from the source via view rather than re-encoding
- Possibly change `ScanFunction` signature to accept view (or both).

### Acceptance criteria

The decoding work done in getEvents query handling no longer decodes in filtering and no longer encodes its responses beyond what it must.

cjonas9:
### What

Moves RPCv1 `getEvents` onto XDR views. The filter loop and the response renderer both paid for `Unmarshal`s of every scanned `xdr.DiagnosticEvent`; now nothing in the query path decodes an event.

**Handler-local matcher** (`methods/event_filter.go`): a view port of the SDK's `GetEventsRequest.Matches`. Filters compile once per request into wire-form comparison values (decoded contract ids, marshaled topic segments, a trailing `**` as an arity flag) and match against the raw bytes the view hands back. rpcv2's `stores/event` and its `v1Filters` shim are untouched.

**RPCv1 query path:**
- `store.ScanFunction` -> `store.ViewScanFunction`: takes `xdr.DiagnosticEventView`, returns `(bool, error)`.
- `sqlitedb.GetEvents` reads rows as `sql.RawBytes` and wraps them as views; scanner errors abort the scan.
- `methods/get_events.go` locates each event's fields once via `Fields()`, matches, and renders matches inside the scan callback. Topics are read only after type and contract id pass. The callback's own error is captured and returned unchanged, as getTransactions does, so a render failure keeps its system error code.

**Shared renderer:** `methods.EventInfoFromView` replaces both `eventInfoForEvent` and the rpcv2 `eventInfoV2` body. `ev.Fields()` and `v0.Fields()` locate type, contract id, topics and data in one pass each and return trimmed views, so their bytes are used directly with no `Raw()` re-walks. The type name comes from a switch instead of the SDK's per-call map. rpcv2 wraps it and rejects non contract/system types on the returned name. Wire output is byte-identical.

### Why

See epic #732.


cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -194,38 +185,53 @@ func (h eventsRPCHandler) getEvents(ctx context.Context, request protocol.GetEve
 
 	eventTypes := combineEventTypes(request.Filters)
 
+	filters, err := store.CompileV1EventFilters(request.Filters) // nil matches every event
```
tamirms, line 188: Could the v1 match stay beside this handler as a view port of `request.Matches`, instead of borrowing rpcv2's clause compiler and matcher through `store`? With the comparison values precomputed once per request it stays small:

```go
// compiledFilter is one validated protocol.EventFilter with its comparison
// values in wire form, matched the way EventFilter.Matches matches a decoded event.
type compiledFilter struct {
	types       []xdr.ContractEventType // empty matches every type
	contractIDs [][]byte                // empty matches every contract
	topics      []compiledTopic         // empty matches every topic list
}

// compiledTopic is one protocol.TopicFilter: a nil value is "*", and atLeast
// is a trailing "**".
type compiledTopic struct {
	values  [][]byte
	atLeast bool
}

func matchesAny(filters []compiledFilter, typ xdr.ContractEventType, cid []byte, topics [][]byte) bool {
	if len(filters) == 0 {
		return true
	}
	for _, f := range filters {
		if len(f.types) > 0 && !slices.Contains(f.types, typ) {
			continue
		}
		if len(f.contractIDs) > 0 &&
			!slices.ContainsFunc(f.contractIDs, func(id []byte) bool { return bytes.Equal(id, cid) }) {
			continue
		}
		if len(f.topics) == 0 ||
			slices.ContainsFunc(f.topics, func(t compiledTopic) bool { return t.matches(topics) }) {
			return true
		}
	}
	return false
}

func (t compiledTopic) matches(topics [][]byte) bool {
	if len(topics) < len(t.values) || (!t.atLeast && len(topics) != len(t.values)) {
		return false
	}
	for i, want := range t.values {
		if want != nil && !bytes.Equal(topics[i], want) {
			return false
		}
	}
	return true
}
```

The type, contract id and raw topics come off the view with the same `Must` chain `EventInfoFromView` already runs, and compiling is the contract id decode and segment marshaling `topicShapeOf` does, without the cross product. The rest of the PR stands as it is: the view scan, `sql.RawBytes`, rendering in the callback and the shared renderer.

What moves with it:
- `store/event_match.go` and its test go away, and `store/event.go` keeps the `ViewScanFunction` change.
- `stores/event/match.go`, `match_test.go`, `v1Filters` in `eventsapi/get_events_v1.go` and `v1_conversions_test.go` return to the base.
- The renderer changes in `get_events_v2.go`, `eventInfoV1` and `conversions_test.go` stay, as does the scan signature in `extract_test.go`.
- #982 drives the handler end to end against the decoded path, so it covers this matcher unchanged.

cmd/stellar-rpc/internal/store/event_match.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/stellar/go-stellar-sdk/strkey"
+	"github.com/stellar/go-stellar-sdk/xdr"
+)
+
+// EventFilter is one clause in the union of an events query. Within a
+// clause every constrained field is AND-ed against the event; zero values
+// are wildcards. Shared by the sqlite (rpcv1) and chunk (rpcv2) backends.
+//
+// Topics[i] constrains topic position i.
+type EventFilter struct {
```
tamirms, line 18: No interface in `store` consumes this type. `EventReader` still takes the SQL prefilter vocabulary, and the only things here that use `EventFilter` are the compiler and matcher that moved with it. The package does already hold cross-backend event policy, `TopicCondition` and `StageSentinels` next door, so it is not an arbitrary home.

What the move costs is the shape of the rpcv2 side. `stores/event` has to alias the type back, and `termKeys`, `valueTermKeys`, `termGroups` and `impliesTopicCount` stop being methods and become `topicCountTermKeys`, `filterValueTermKeys`, `filterTermGroups` and `filterImpliesTopicCount`, since Go does not allow methods on a type declared in another package. That demotion follows from sharing at all rather than from picking this package. With the v1 match beside its handler (the comment on `get_events.go`), the type and its methods stay in `stores/event` next to the engine that plans with them.

cmd/stellar-rpc/internal/store/event_match.go
```diff
@@ -0,0 +1,410 @@
...
+		if matchAll {
+			return nil, nil
+		}
+		out = append(out, expanded...)
+	}
+	return out, nil
+}
+
+func expandV1Filter(f *protocol.EventFilter) ([]EventFilter, bool, error) {
+	// One clause per named type + an empty set is the wildcard. Valid admits
```
tamirms, line 77: The `{contract, system}` fold from #960 still holds for sqlite, so I don't think it needs to go. sqlite has ingested only `GetTransactionEvents` output since v23.0.0 (`sqlitedb/event.go:108`), with a week's retention by default, and rpcv2 ingests the same transaction and operation event lists (`stores/event/extract.go:73-84`). Core puts only contract and system events in those lists, which is what Urvi pointed out when Copilot raised the same thing on [#960](https://github.com/stellar/stellar-rpc/pull/960#discussion_r3930112326). The diagnostic-typed operation events that make the fold differ from `EventTypeSet.matches` exist only in #982's corpus. So the fold can stay either way: with the match beside the handler, `v1Filters` keeps it untouched, and if the compiler stays shared, #982's corpus can drop the diagnostic-typed operation events, since neither ingest writes that shape.

cmd/stellar-rpc/internal/store/event_match.go
```diff
@@ -0,0 +1,410 @@
...
+	if f.Exact {
+		return n == f.Count
+	}
+	return n >= f.Count
+}
+
+// CompileV1EventFilters expands a validated v1 filter list into clauses. The
+// OR dimensions within one v1 filter (contract ids, topic filters) multiply
+// out, one clause per combination: at most 5 filters x 5 contract ids
+// x 5 topics x 2 filter types = 250. A combination with no constraints matches
```
tamirms, line 59: Dropping the fold also changes how rpcv2 plans these requests. A v1 request that names both types and nothing else used to compile to no clauses, which the engine serves by streaming the window (the match-all path from `unionForFilters` into `streamRange`). It now compiles to two type clauses, so the engine looks up both type terms, unions them and fetches the candidates, for a constraint every stored event satisfies. A request that adds contract ids or topics doubles its clauses the same way, and the worst case moves from 125 to 250 against the pager's 256-filter cap (`maxCursorFilters`, checked in `validateScope`). I haven't measured the difference, and keeping the fold makes it moot.

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -194,38 +185,53 @@ func (h eventsRPCHandler) getEvents(ctx context.Context, request protocol.GetEve
...
+		if matched {
+			info, err := EventInfoFromView(
+				event,
+				cursor,
+				time.Unix(ledgerCloseTimestamp, 0).UTC().Format(time.RFC3339),
+				txHash.HexString(),
+				request.Format,
+			)
+			if err != nil {
+				return false, errors.Wrap(err, "could not parse event")
```
tamirms, line 221: This error now leaves the callback, so the sqlite loop joins it with "failed to scan event" and the post-scan branch codes every `GetEvents` failure as `jrpc2.InvalidRequest`. On base the same failure returned as a plain error from the handler, which jrpc2 codes as a system error. So the code moves from -32098 to -32600 and the message gains a second line.

The direction is backwards. The reachable case is a request with `xdrFormat: "json"` whose value the converter rejects, which is a server-side failure now reported as a bad request. An embedded newline in a JSON-RPC `message` is also new.

getTransactions hit this shape in #962 and kept the callback's own error in a captured variable, returning it from the handler unchanged so the wrap at the call site never sees it. The same trick works here, or the callback can return a `*jrpc2.Error` that the call site passes through.

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -250,83 +256,86 @@ func (h eventsRPCHandler) getEvents(ctx context.Context, request protocol.GetEve
...
+		dataRaw []byte
+		cidRaw  []byte
+	)
+	err := xdr.TryVoid(func() {
+		xdrType = ev.MustType().MustValue()
+		v0 := ev.MustBody().MustV0() // panics on a non-V0 body, replacing "unknown event version"
+		views := v0.MustTopics().MustAll()
+		topics = make([][]byte, len(views))
+		for i, t := range views {
+			topics[i] = t.MustRaw()
```
tamirms, line 277: `All()` already returns each topic trimmed to its exact extent, so `MustRaw()` on an element walks it a second time only to size it again. The rpcv2 events index has this same loop with a comment saying exactly that: the elements are already the topic's raw XDR, so use them directly rather than calling `Raw()`, which re-walks size.

`Data()` is worse. It walks the whole topics array to find its own offset and returns an untrimmed slice, so `MustData().MustRaw()` walks the topics a second time and then the value.

`ev.Fields()` and `v0.Fields()` locate these in one pass and hand back trimmed views, which removes both re-walks. Worth doing here because this runs once per event in the page, and not paying for work the view already did is the point of the PR.

cjonas9: Great catch, delegated this work to a new helper `eventRawFields` that does this.

cmd/stellar-rpc/internal/methods/get_events.go
cjonas9: Done as you laid out here. `methods/event_filter.go` compiles each `protocol.EventFilter` once into wire-form values and matches the way `Matches` does with a couple of smaller changes: the filters are a type with methods rather than a free function, and there is a `matchHeader` pre-check on type and contract id so a rejected event never sizes its topics. Fields come off the view once via `ev.Fields()` and `v0.Fields()` and feed both the match and the render. 
Your checklist is all in, so `store/event_match.go` and its test are gone, `store/event.go` keeps `ViewScanFunction,` and `stores/event/match.go,` `match_test.go,` `v1Filters` and `v1_conversions_test.go` are identical to base. The renderer changes, `eventInfoV1` and the `extract_test.go` scan signature stay. The parity harness caught a bug in my first cut where one filter's header and another filter's topics could each pass, so there is now a unit test pinning that shape too.

cmd/stellar-rpc/internal/store/event_match.go
cjonas9: Addressed/made moot by the changes in my reply above, the file is gone.

cmd/stellar-rpc/internal/store/event.go
```diff
@@ -24,12 +24,15 @@ type TopicFilter []TopicCondition
 // matches a row, the DB event is considered a candidate for further filtering.
 type TopicFilters []TopicFilter
 
-type ScanFunction func(
-	event xdr.DiagnosticEvent,
+// ViewScanFunction visits one event. eventView and txHash are on loan for the
+// call's duration, so copy anything that must persist. Returning false stops
+// the scan and an error aborts it.
+type ViewScanFunction func(
```
tamirms, line 30: The awkwardness in the handler's error path traces back to this shape rather than to the wrapping around it. A callback cannot return to the function that set it up, so everything it produces escapes through captured state: `results`, and now `procErr` as well, which is why the error travels out twice and the handler discards the copy the reader hands back. This supersedes the captured-variable fix I pointed at on the error-code thread, since that pattern is what makes the error travel twice.

An iterator removes the cause, and it is the shape the rest of the tree already uses for this job, including the rpcv2 events reader, `ScanLedgers` and the catalog scans.

```go
// ScannedEvent is one row of an event scan. Event aliases the backend's row
// buffer and is valid only inside the loop body that received it, so copy
// whatever outlives that body. TxHash is already a copy and is free to keep.
type ScannedEvent struct {
	Event           xdr.DiagnosticEventView
	Cursor          protocol.Cursor
	LedgerCloseTime int64
	TxHash          xdr.Hash
}

type EventReader interface {
	// GetEvents yields the events in cursorRange that pass the coarse
	// filters, in ascending cursor order. Break to stop early. A non-nil
	// error ends the stream and the ScannedEvent beside it is zero.
	GetEvents(
		ctx context.Context,
		cursorRange protocol.CursorRange,
		contractIDs [][]byte,
		topics TopicFilters,
		eventTypes []int,
	) iter.Seq2[ScannedEvent, error]
}
```

The handler becomes an ordinary loop over ordinary locals:

```go
for ev, err := range h.dbReader.GetEvents(ctx, cursorRange, contractIDs, topics, eventTypes) {
	if err != nil {
		return protocol.GetEventsResponse{}, &jrpc2.Error{Code: jrpc2.InvalidRequest, Message: err.Error()}
	}
	event, err := ev.Event.Event()
	if err != nil {
		return protocol.GetEventsResponse{}, errors.Wrap(err, "could not parse event")
	}
	head, err := eventHeader(event)
	if err != nil {
		return protocol.GetEventsResponse{}, errors.Wrap(err, "could not parse event")
	}
	if !filters.matchHeader(head) {
		continue
	}
	body, err := eventBody(head.v0)
	if err != nil {
		return protocol.GetEventsResponse{}, errors.Wrap(err, "could not parse event")
	}
	if !filters.match(head, body.topics) {
		continue
	}
	info, err := eventInfo(head, body, ev.Cursor,
		time.Unix(ev.LedgerCloseTime, 0).UTC().Format(time.RFC3339), ev.TxHash.HexString(), request.Format)
	if err != nil {
		return protocol.GetEventsResponse{}, errors.Wrap(err, "could not parse event")
	}
	results = append(results, info)
	if uint(len(results)) >= limit {
		break
	}
}
```

A render failure returns plainly, which is the code base gave it, so the join in the sqlite loop has nothing left to route around and `procErr` goes away.

On the sqlite side the body of the existing `rows.Next()` loop becomes a `yield`, the query and scan failures become yields of their own, and `rows.Err()` becomes the trailing one. `defer rows.Close()` still runs when the consumer breaks, and the borrow is unchanged, since the loop body runs inside `yield` before the next `Next()`.

It also lets the loan sentence say something true. `txHash` is an `xdr.Hash` array conversion, so it is already a copy and safe to retain; only the event view aliases the row buffer.

Scope is one interface method, one implementation, one production caller and six test files.

The shape change is mechanism neutral: range-over-func compiles the loop body into a closure the iterator calls, so it is the same one indirect call per row that the callback is today. `BenchmarkGetEvents` and `BenchmarkGetEventsTopicFilters` are in this package if you want before and after numbers.

#1004 takes the ledger reads the same way and deletes `readLedgerPage` with its own `procErr`, so this is one direction rather than two separate asks.

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -250,83 +265,128 @@ func (h eventsRPCHandler) getEvents(ctx context.Context, request protocol.GetEve
...
-	if event.Event.ContractId != nil {
-		info.ContractID = strkey.MustEncode(
-			strkey.VersionByteContract,
-			(*event.Event.ContractId)[:])
-	}
 	return info, nil
 }
 
+// eventTypeName is protocol.GetEventTypeFromEventTypeXDR without the per-call map; "" for an unknown type.
+func eventTypeName(t xdr.ContractEventType) string {
```
tamirms: `ContractEventTypeView.Value()` rejects any discriminant outside the three enum members, and `eventHeader` reads the type through `Fields()` and `Value()`, so an unknown type cannot reach here and the `""` is unreachable.

Returning `(string, bool)` and erroring on `!ok` would say that outright. It would also let `checkResponseEventType` in eventsapi stop re-deriving an enum check from a rendered name: today it reports `stored event has type ""` for a case that cannot happen, when its real job is refusing the diagnostic type, which is a perfectly valid name.

cmd/stellar-rpc/internal/methods/get_events.go
```diff
@@ -194,36 +185,60 @@ func (h eventsRPCHandler) getEvents(ctx context.Context, request protocol.GetEve
...
-	results := make([]protocol.EventInfo, 0, len(found))
-	for _, entry := range found {
-		info, err := eventInfoForEvent(
-			entry.event,
-			entry.cursor,
-			time.Unix(entry.ledgerCloseTimestamp, 0).UTC().Format(time.RFC3339),
-			entry.txHash.HexString(),
-			request.Format,
-		)
+	results := []protocol.EventInfo{}
```
tamirms, line 195: Base sized this to the page, first `make([]entry, 0, limit)` and then `make([]protocol.EventInfo, 0, len(found))`, so a full page no longer regrows from empty. `make([]protocol.EventInfo, 0, limit)` restores that and is still non-nil, so an empty page still serializes as `[]`.

tamirms (approved):
Approving. All six earlier comments are addressed in the code rather than only in replies: the matcher is beside the handler, the shared file is gone and rpcv2 is back to base, the type fold is restored, the error code is fixed with the captured-error pattern, and the renderer reads through the located field bundles. I checked that those bundles trim every field the new code takes bytes from, so the rendering is still byte-identical.

The three comments below change no behavior, so they do not need another review round. Merge once they are addressed or consciously skipped. The first is the largest of them and can equally land here or fold in with #1004, which takes the ledger reads the same way.

cmd/stellar-rpc/internal/methods/get_events.go
cjonas9: This change does intentionally make the handler allocate lazily as opposed to eager, but that's for a reason. But first...

> Base sized this to the page, first make([]entry, 0, limit) and then make([]protocol.EventInfo, 0, len(found))

Just to say it explicitly, this shape no longer works within our constraints given the loan doc. Each `entry` holds a view that aliases the row buffer. Since the `RawBytes` change, that buffer is reused on the next row, so anything retained past the callback reads bytes that have already been overwritten. Trying to keep this shape leads to wasteful behavior pretty quickly (e.g. copying `RawBytes` where we don't need to).

That then leaves us with two options -- initialize `results` as an empty list, or size `results` to `limit`. The "right choice" here clearly depends on the time it takes an empty slice to grow to dynamically contain what it needs, or the time wasted by sizing these lists so optimistically at `limit`. I took the former based on two things:
- overwhelmingly, real `getEvents` requests usually don't land any results (in Overcat's data, ~93% didn't)
- benchmarking either case tells us how each impl handles each case, and not optimistically sizing the slice doesn't perform _that_ poorly when it needs a large slice. Some scrappy throwaway Claude benchmarks on this:

| matches / limit | empty literal | presize to limit |
| --- | --- | --- |
| 0 / 100 | 2 ns, 0 B, 0 allocs | 1.4 µs, 18 KB, 1 alloc |
| 3 / 100 | 0.2 µs, 1.3 KB, 3 allocs | 1.4 µs, 18 KB, 1 alloc |
| 100 / 100 | 4.9 µs, 49 KB, 8 allocs | 1.7 µs, 18 KB, 1 alloc |
| 0 / 10000 | 2 ns, 0 B, 0 allocs | 17 to 125 µs, 1.8 MB, 1 alloc |
| 1000 / 10000 | 50 µs, 606 KB, 12 allocs | 46 µs, 1.8 MB, 1 alloc |
| 10000 / 10000 | 707 µs, 9.3 MB, 20 allocs | 154 µs, 1.8 MB, 1 alloc |

cmd/stellar-rpc/internal/store/event.go
cjonas9: Agreed on the diagnosis. The error does travel twice: each failure sets `procErr`
and is also returned from the callback, where the reader joins it with "failed to
scan event" before the handler's switch throws that copy away. And the callback
shape is what forces it, since captured state is the only way out of a callback.

I'd like to take the iterator as a follow-up rather than here:

- This PR's correctness argument is the differential suite pinning view-migration
  behaviour. Reshaping the reader interface is a different axis, touching the store
  interface, the sqlite reader and its test callers, which dilutes the diff that's
  actually under review.
- Nothing is broken today. `procErr` produces the right codes and matches the
  existing `readLedgerPage` precedent, and your own point that the change is
  mechanism-neutral means there's no correctness or perf pressure to land it now.
- The sibling work already exists for ledger reads, so doing events as the next PR
  in that series is the single direction you're asking for, and it gets reviewed
  alongside the shape it mirrors.

I'm going to fix the loan sentence, since this PR wrote it and it's wrong.
The sqlite reader converts to an `xdr.Hash` array inside the loop, so the hash is a
per-row copy and safe to retain. Only the event view aliases the row buffer.

Two things worth settling when we do the follow-up:

- You cite `ScanLedgers` as precedent, but it returns `(iter.Seq2[...], error)`, so
  setup failures surface before the loop, whereas `ScannedEvent` folds them into the
  first yield. Worth picking one deliberately rather than by accident.
- Since the struct is new, it's the moment to carry `xdr.ContractEventView` directly
  instead of the diagnostic wrapper, so consumers stop doing `ev.Event.Event()`.

Minor: I think #1004 is the issue and #1005 is the PR.

cmd/stellar-rpc/internal/methods/get_events.go
cjonas9: good catch, implemented as recommended

Issue #986, opened 2026-09-08
Integration-test infrastructure for rpcv1 and rpcv2


karthikiyer56:
## Goal

One integration suite, identical assertions, run against rpcv1 and rpcv2 on every supported protocol. rpcv2-specific behaviour gets its own tests in its own home.

## Legs

| daemon | protocols |
|---|---|
| rpcv1 | P27, P28 |
| rpcv2 | P27, P28 |

- The protocol list moves with releases: `[28, 29]`, `[29, 30]`, and so on.

## Children

| order | issue | what |
|---|---|---|
| 0 | #883 | run the rpcv1 suite unchanged against rpcv2; four CI legs |
| 1 | #983 | rpcv2 sends a versioned `User-Agent`, so `TestArchiveUserAgent` can go common |
| 2 | #984 | rpcv2 backfill integration test |
| 3 | #985 | SPIKE: one shared network for the identical environments, after #883 |

## Done when

- Every child is closed.
- Four legs are green on `main`.

PR #1001, opened 2026-09-10
rpcv2 integration tests: run the rpcv1 suite unchanged against rpcv2, four CI legs

### LINKED ISSUE #883: rpcv2 integration tests: run the rpcv1 suite unchanged against rpcv2, four CI legs

karthikiyer56:
Closes #883 (task 0 of #986).

## What this PR does

- Runs the shared integration suite against either daemon. `STELLAR_RPC_INTEGRATION_TESTS_DAEMON` picks `rpcv1` (default) or `rpcv2`.
- Adds two CI legs: `stellar-rpc.yml` now runs a `protocol x daemon` matrix (P27, P28) x (rpcv1, rpcv2).
- Keeps every assertion identical for both daemons.

## Layout

| package | contents | runs on |
|---|---|---|
| `cmd/stellar-rpc/internal/integrationtest/` | the harness (`infrastructure/`) and 13 daemon-neutral test files | both daemons |
| `cmd/stellar-rpc/internal/rpcv1/integrationtest/` | `archive_test.go`, `backfill_test.go`, `ingest_loadtest_test.go`, `migrate_test.go`, `get_ledgers_datastore_test.go`, `health_retention_test.go`, `TestMain` with the fake GCS server | rpcv1 legs |
| `cmd/stellar-rpc/internal/rpcv2/integrationtest/` | `health_retention_test.go`; the home for rpcv2-only tests | rpcv2 legs |

## Where the changes are

| area | files | kind of change |
|---|---|---|
| Moves (`git mv`) | 18 files identical, 14 with the import path and `infrastructure.SkipLimitsUpgrade()` only; `infrastructure/docker/` byte-identical | mechanical |
| Harness, new | `infrastructure/daemon.go`, `daemon_rpcv1.go`, `daemon_rpcv2.go`, `docker/rpcv2-integration-tests.toml`, `rpcv2_config_test.go`, `skip_limits.go` | - the harness drives the daemon through a four-method interface<br>- `daemon_rpcv1.go` is today's boot code moved behind it<br>- `daemon_rpcv2.go` boots rpcv2 in-process: three captive-core ports from `getFreeTCPPorts`, the daemon's own listeners on port 0, the sample TOML minus the datastore tables, per-test values as flag overrides<br>- `rpcv2_config_test.go` decodes the test TOML and the sample and requires them equal outside the three deliberate differences (no datastore tables, `enable_debug`, log level) |
| Harness, `test.go` | `infrastructure/test.go` | - `waitForRPC` returns when the daemon is healthy **and** caught up with the Core container<br>- the limits upgrade also waits until the daemon has ingested the ledger that applied it, since preflight reads state from captive core, not the container<br>- every address the harness dials is `127.0.0.1`, never `localhost`<br>- a daemon that exits before it is healthy fails the test at once<br>- under rpcv2, every rpcv1-only `TestConfig` setting fails the test at startup instead of being ignored |
| rpcv2 | `cmd/stellar-rpc/internal/rpcv2/daemon.go`, `serve.go`, `startup.go`, `metrics_test.go` | - `Options` and `RunDaemonWithOptions`: an exported entry point taking a logger, flag overrides, and an `OnListen` callback<br>- `OnListen` reports the bound read and admin addresses once, so a caller may configure port 0<br>- the process metrics (log-line counters, `soroban_rpc_build_info`, Go and process collectors) come from `host.RegisterProcessMetrics`, so `TestMetrics` passes unchanged<br>- nothing else under `rpcv2/` changes |
| shared | `cmd/stellar-rpc/internal/host/metrics.go` | - `RegisterProcessMetrics`: the one copy of the process-metrics registration both daemons call, so the label set cannot drift between them |
| rpcv1 | `cmd/stellar-rpc/internal/rpcv1/daemon/metrics.go` | - `registerMetrics` calls `host.RegisterProcessMetrics`; same families, same labels<br>- the rpcv1-only tests change only their import path and `GetDaemon()` -> `infrastructure.RPCv1Daemon(t, test)`<br>- in the harness, rpcv1 now submits through its own captive core (`STELLAR_CAPTIVE_CORE_HTTP_PORT` set, `STELLAR_CORE_URL` pointed at it), as it does by default in production |
| Tests, five edits and one comment | `health_test.go`, `upgrade_test.go`, `metrics_test.go`, `get_ledgers_test.go` (x2 files) | - `TestHealth`: retention-window line moves to a per-daemon test<br>- `TestUpgradeFrom20To21`: latest ledger via `getLatestLedger`, not SQLite<br>- `TestMetrics`: `soroban_rpc_log_error_total` read from `/metrics`, not the registry<br>- `TestGetLedgersFromDatastore`: split into the rpcv1 package<br>- `TestGetLedgers`: beyond-latest request asks for `LatestLedger + 1000`<br>- `TestGetLedgers`: a comment explains why `OldestLedger - 4` wraps under rpcv2 and why the assertion still holds |
| CI and tooling | `integration-tests.yml`, `stellar-rpc.yml`, `scripts/check-dependencies.bash`, `.golangci.yml`, `README.md` | - `daemon` input; each leg runs the shared package plus that daemon's package<br>- the dependency checker reads the protocol matrix<br>- depguard lets the shared harness import both trees |

## Latent problems fixed along the way

| problem | where it showed | fix |
|---|---|---|
| `healthy` does not mean caught up. A daemon that starts at genesis reports healthy as soon as its first ledger commits, because with 1 s test ledgers every ledger it still has to replay is inside the 30 s latency window. A test then submitted into a captive core still replaying, or read a protocol-0 ledger. | rpcv2 (`TestGetTransactions`, `TestGetVersionInfoSucceeds`, `TestSendTransactionFailedInLedger`) | `waitForRPC` also waits until the daemon's latest ledger reaches the Core container's ledger. Tests that delay the daemon on purpose (`DelayDaemonForLedgerN`) are exempt. |
| The limits upgrade was checked on the Core container's `/sorobaninfo`, but preflight reads state from the daemon's captive core, which closes the same ledger a moment later. A `simulateTransaction` sent in that gap ran with the old limits (`Error(Budget, ExceededLimit)`). | both daemons (`TestGetLedgerEntriesSucceeds`) | after the `/sorobaninfo` check, wait until the daemon has ingested the upgrade ledger. |
| Every harness client dialed `localhost`. Go tries the IPv6 side first, and on Docker Desktop for macOS a forwarder can hold the IPv6 side of a port whose IPv4 side the daemon or a container bound; the client connected to the forwarder and was reset for the whole 180 s health wait. | both daemons, local bench | every address the harness dials is `127.0.0.1`. |
| A daemon that died before it was healthy failed the test only after the 180 s health wait. A CI leg runs two test binaries that could draw the same port range (`pid % 10`), so one test's RPC port could be another's captive-core peer port. | both daemons | `waitForRPC` watches the daemon's exit and fails the test at once with the exit error. Each test process claims one of 25 port ranges with an exclusive `flock` on a temp-dir lock file for its lifetime, so two binaries never share a range. rpcv2's own listeners bind port 0 and report their ports back, so only the three captive-core ports are chosen in advance. rpcv1's exit hook reports through the same exit channel. |
| `TestGetLedgers` asked for `LatestLedger + 1` two calls after learning the latest ledger. On a 1 s ledger network that ledger sometimes existed by then, so the expected error did not come. | both daemons | the beyond-latest request asks for `LatestLedger + 1000`. Same assertion. |
| rpcv1 in the harness submitted transactions to the Core container, while rpcv1's own default (and rpcv2's only path) is the daemon's captive core. | rpcv1 harness | rpcv1 gets a captive-core admin HTTP port and `STELLAR_CORE_URL` points at it. |
| `scripts/check-dependencies.bash` read protocol and Core version pairs out of `stellar-rpc.yml` with an awk that only understood one literal `protocol_version` per job. | CI, dependency sanity checker | the awk also reads a `protocol: [...]` matrix list; the old and new workflow produce the same pairs. |
| The harness waited for Core's `/info` to pass the checkpoint ledger, but the container publishes that checkpoint to the history archive a few seconds later; until then the archive's root state says ledger 0 (`new-hist` writes it that way), which rpcv2 treats as a permanent startup failure. | rpcv2 (`TestSendTransactionFailedInLedger`, `TestSimulateInvokeContractTransactionUseUpgradedAuth`) | `waitForArchive` polls `/.well-known/stellar-history.json` until it reports the first checkpoint before any daemon starts. |

| The rpcv1 harness ended a failed daemon start with a bare `runtime.Goexit` on the test goroutine. Go's test runner treats that as a panic and aborts the whole test binary. | rpcv1 harness, only when the daemon fails to build | the exit hook calls `t.Fatal` while the daemon is being built and `t.Error` plus `Goexit` from daemon goroutines. |


.github/workflows/integration-tests.yml
```diff
@@ -154,7 +159,9 @@ jobs:
       - name: Run Stellar RPC Integration Tests
         run: |
           set -o pipefail
-          go test -race -v -timeout 30m ./cmd/stellar-rpc/internal/rpcv1/integrationtest/... 2>&1 \
+          go test -race -v -timeout 30m \
+            ./cmd/stellar-rpc/internal/integrationtest/... \
+            ./cmd/stellar-rpc/internal/${{ inputs.daemon }}/integrationtest/... 2>&1 \
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 340e255e. Each test process now claims one of 25 port ranges by holding an exclusive flock on a lock file in the temp dir for its lifetime, so two binaries in one leg can never share a range; the kernel releases the lock when the process ends. On top of that, an rpcv2 daemon that hits a bind collision (in its own listeners or in captive core) restarts on fresh ports, up to three attempts. With disjoint ranges rpcv1 needs no restart path.

cmd/stellar-rpc/internal/integrationtest/infrastructure/daemon_rpcv1.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if i.ingestLoadTest.Enabled() {
+		cfg.IngestLoadTest = i.ingestLoadTest
+	}
+
+	d.log = supportlog.New()
+	d.log.SetOutput(newTestLogWriter(i.t, `rpc="daemon" `))
+	d.log.SetExitFunc(func(code int) {
+		i.t.Fatalf("Exited with code %d", code)
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 340e255e. The exit hook now uses t.Error (safe from any goroutine) and sends the error on a buffered channel that exited() returns, so waitForRPC fails on the test goroutine at once instead of polling out its deadline.

scripts/check-dependencies.bash
```diff
@@ -105,12 +105,19 @@ fi
...
-# core_version; source jobs have core_git_ref instead and are skipped.
+# stellar-rpc.yml. A job either names one protocol (protocol_version: '27') or runs a matrix
+# (protocol: ['27', '28'] with protocol_version: ${{ matrix.protocol }}); in the matrix form
+# every listed protocol pairs with the job's core_deb_version. Source jobs have core_git_ref
+# instead and are skipped.
 PROTO_VERSION_PAIRS=$(awk "
   /^[[:space:]]*#/ { next }
-  /protocol_version:/ { p = \$NF; gsub(/'/, \"\", p) }
-  /core_deb_version:/     { c = \$NF; gsub(/'/, \"\", c); if (c != \"\") print p, c }
+  /^[[:space:]]*protocol:[[:space:]]*\\[/ { m = \$0; sub(/.*\\[/, \"\", m); sub(/\\].*/, \"\", m); gsub(/['\" ]/, \"\", m); matrix = m }
```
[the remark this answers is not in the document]
karthikiyer56: Not a bug. That double quote is escaped in the script (`gsub(/['\\" ]/, ...)` reaches awk as `['" ]`), the file passes `bash -n`, and the awk assignment run on its own prints the two expected pairs. The Dependency sanity checker job passed on heads 63a74388, a1edf048 and 60905ad7 with this exact code.

cmd/stellar-rpc/internal/integrationtest/infrastructure/daemon_rpcv1.go
```diff
@@ -0,0 +1,100 @@
...
+	// channel lets waitForRPC stop at once instead of polling out its deadline.
+	d.done = make(chan error, 1)
+	d.log.SetExitFunc(func(code int) {
+		err := fmt.Errorf("rpcv1 daemon exited with code %d", code)
+		i.t.Error(err)
+		select {
+		case d.done <- err:
+		default:
+		}
+	})
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in 07f0a789. After t.Error and the channel send, the hook calls runtime.Goexit(), so the goroutine that called Fatal ends the way it did before: on the test goroutine during MustNew that is what FailNow does, and a daemon goroutine just stops. No FailNow from daemon-owned goroutines.

cmd/stellar-rpc/internal/integrationtest/infrastructure/daemon_rpcv2.go
```diff
@@ -0,0 +1,162 @@
...
+	values := map[string]string{
+		"storage.default_data_dir":                 filepath.Join(i.t.TempDir(), "rpcv2"),
+		"service.endpoint":                         fmt.Sprintf("127.0.0.1:%d", i.testPorts.RPCPort),
+		"service.admin_endpoint":                   fmt.Sprintf("127.0.0.1:%d", i.testPorts.RPCAdminPort),
+		"service.methods.getNetwork.friendbot_url": FriendbotURL,
+		"ingestion.captive_core_config":            filepath.Join(i.rpcConfigFilesDir, captiveCoreConfigFilename),
+		"ingestion.history_archive_urls":           "http://" + i.testPorts.CoreArchiveHostPort,
+		"ingestion.stellar_core_binary_path":       findCoreBinary(i.t),
+		"ingestion.captive_core_storage_path":      i.captiveCoreStoragePath,
+		"ingestion.core_http_port":                 strconv.Itoa(int(d.captiveCoreHTTPPort)),
```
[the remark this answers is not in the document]
karthikiyer56: Fixed in d4ae951c: ingestion.core_url is now set to http://127.0.0.1:<core_http_port> explicitly, which validateCoreURL accepts because the port matches.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -528,23 +584,64 @@ func (vars rpcConfig) toMap() map[string]string {
...
+	// A nil channel never fires, which covers the released-RPC-container path
+	// (TestMigrate): there is no in-process daemon to watch.
+	var exited <-chan error
+	if i.daemon != nil {
+		exited = i.daemon.exited()
+	}
+	for {
+		select {
+		case err := <-exited:
+			if isBindError(err) && attempts < maxDaemonStartAttempts {
```
tamirms, line 602: With 340e255e giving each test process its own flock'd range, and every port probed before it is handed out, I think this restart path has nothing left to catch. The two collision sources were two processes walking the same range and a client socket taking a probed port back, and both are gone now. Your reply on the workflow thread makes the same argument for rpcv1, and it applies to rpcv2 too.

What the path still does is treat every "stellar core exited unexpectedly" as a port race, through isBindError. A captive core that dies at startup for a real reason, say a bad line in the generated config, gets rebooted three times with a full replay each time, and until the third attempt the log says "daemon lost a port race". I grepped the job logs of all four legs on d4ae951c: no port races, no core exits, no "address already in use".

I would keep the early exit, so a daemon that dies before it is healthy still fails the test at once with its error, and drop the restart: maxDaemonStartAttempts, isBindError, this branch, and the per-goroutine channel copies in rpcv2Daemon.start that exist only so a restarted daemon reports on its own channels.

cmd/stellar-rpc/internal/integrationtest/infrastructure/rpcv2_config_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/stellar-rpc/cmd/stellar-rpc/internal/rpcv2/config"
+)
+
+func TestRPCv2ConfigParsesStrict(t *testing.T) {
+	data, err := os.ReadFile(filepath.Join("docker", rpcv2ConfigFilename))
+	require.NoError(t, err)
+
+	cfg, err := config.ParseConfig(data)
```
tamirms, line 18: This pins the test config to the schema, but not to the sample it was copied from. If a default in rpc-v2-sample-config.toml changes and the copy is not updated, the legs keep running with the old value and nothing fails, and "rpcv2 starts with the stock defaults from the sample" in #883 stops being true without anyone noticing.

The way I would close that is to not have a copy at all: give rpcv2 an entry point that takes a decoded config, and have the harness decode the sample itself (three levels up, at cmd/stellar-rpc/rpcv2/rpc-v2-sample-config.toml) and set the per-test values on the struct.

```go
// rpcv2: RunDaemon keeps loading from a path for the CLI and calls this.
func RunDaemonWithConfig(ctx context.Context, cfg config.Config, opts Options) error

// harness, in place of flags()
data, err := os.ReadFile(sampleConfigPath)
require.NoError(i.t, err)
cfg, err := config.DecodeConfig(data)
require.NoError(i.t, err)
cfg.Backfill.DataStore = config.DataStoreConfig{}
cfg.Storage.DefaultDataDir = filepath.Join(i.t.TempDir(), "rpcv2")
cfg.Service.Endpoint = fmt.Sprintf("127.0.0.1:%d", i.testPorts.RPCPort)
cfg.Service.AdminEndpoint = fmt.Sprintf("127.0.0.1:%d", i.testPorts.RPCAdminPort)
cfg.Service.Methods.GetNetwork.FriendbotURL = FriendbotURL
cfg.Service.Preflight.EnableDebug = new(true)
cfg.Ingestion.CaptiveCoreConfig = filepath.Join(i.rpcConfigFilesDir, captiveCoreConfigFilename)
cfg.Ingestion.HistoryArchiveURLs = []string{"http://" + i.testPorts.CoreArchiveHostPort}
cfg.Ingestion.StellarCoreBinaryPath = findCoreBinary(i.t)
cfg.Ingestion.CaptiveCoreStoragePath = i.captiveCoreStoragePath
cfg.Ingestion.CoreHTTPPort = new(uint(d.captiveCoreHTTPPort))
cfg.Ingestion.CoreURL = fmt.Sprintf("http://127.0.0.1:%d", d.captiveCoreHTTPPort)
cfg.Ingestion.CoreHTTPQueryPort = new(uint(i.testPorts.captiveCoreHTTPQueryPort))
cfg.Logging.Level = "debug"
err = rpcv2.RunDaemonWithConfig(ctx, cfg.WithDefaults(), rpcv2.Options{Logger: log})
```

That deletes rpcv2-integration-tests.toml, this test, and the fs.Set block in flags(), and every per-test value becomes a typed assignment instead of a string. What it gives up is the harness going through LoadConfigWithFlags, which the config unit tests and sample_config_test.go already cover.

If you would rather keep the copy, the smaller fix is to make this test decode both files, apply the three deliberate differences to the sample side (no datastore tables, enable_debug true, level debug), and require the two structs to be equal.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -277,6 +286,18 @@ func NewTest(t testing.TB, cfg *TestConfig) *Test {
 	return i
 }
 
+// rejectRPCv1OnlySettings fails a test that asks for a setting only the rpcv1
+// daemon has while another daemon is selected.
+func (i *Test) rejectRPCv1OnlySettings() {
+	if selectedDaemon(i.t) == daemonRPCv1 {
+		return
+	}
+	if i.datastoreConfigFunc != nil || i.ingestLoadTest.Enabled() || i.historyRetentionWindow != 0 {
```
tamirms, line 295: SQLitePath, IgnoreLedgerCloseTimes and UseReleasedRPCVersion are rpcv1 settings too, and none of them reaches this check. The one a test can hit today is SQLitePath: BenchmarkGetLedgerEntries in the shared package passes it when STELLAR_RPC_BENCH_BASE_STORAGE_DIR is set, so that the daemon's storage lands on the disk under test. Under rpcv2 the path is ignored and the daemon uses a temp dir, so the benchmark measures the wrong disk without saying so. UseReleasedRPCVersion is a latent one: it bypasses daemon selection entirely, so a test that set it would run against the released rpcv1 image whatever the selector says. Only TestMigrate uses it and that lives in the rpcv1 package, so nothing trips it today.

Adding the three here is one line. For SQLitePath, the alternative that keeps the benchmark meaningful under rpcv2 is to honor it: when a test set it explicitly, put storage.default_data_dir next to it.

cmd/stellar-rpc/internal/rpcv2/daemon.go
```diff
@@ -388,6 +412,33 @@ func newPreflightPool(
 	})
 }
 
+// registerProcessMetrics registers the process-level families rpcv1 exposes on
+// /metrics, so one dashboard reads both daemons: per-level log-line counters,
+// the build-info gauge, and the Go runtime and process collectors.
+func registerProcessMetrics(registry *prometheus.Registry, logger *supportlog.Entry) {
```
tamirms, line 418: Optional. This is registerMetrics from cmd/stellar-rpc/internal/rpcv1/daemon/metrics.go line for line: the same logmetrics hook, the same build_info gauge with the same five labels, the same two collectors. Since the reason for adding it is that one dashboard reads both daemons, the label set is a contract between the two trees, and a change on one side would go unnoticed until a panel broke. One function both trees call, for example host.RegisterProcessMetrics(registry, logger) next to host.PrometheusNamespace, keeps that from drifting. rpcv1 would keep its versionLabel and commitLabel constants for the startup log fields.

cmd/stellar-rpc/internal/integrationtest/infrastructure/daemon_rpcv2.go
```diff
@@ -0,0 +1,163 @@
...
+	stopped chan struct{}
+}
+
+func (d *rpcv2Daemon) start() {
+	i := d.test
+	ports := getFreeTCPPorts(i.t, 5)
+	i.testPorts.captiveCorePeerPort = ports[0]
+	i.testPorts.captiveCoreHTTPQueryPort = ports[1]
+	d.captiveCoreHTTPPort = ports[2]
+	i.testPorts.RPCPort = ports[3]
```
tamirms, line 53: Optional, and fine as a follow-up. These two do not have to be pre-allocated. rpcv2 could bind 127.0.0.1:0 for its read and admin listeners and report the bound addresses back, the way rpcv1 does through GetEndpointAddrs, for example with an OnListen func(rpc, admin net.Addr) field on Options that the daemon calls once both listeners are bound. The harness would then pre-allocate the same three captive-core ports for both daemons and nothing else. Captive core's ports have to stay pre-allocated because stellar-core reads fixed port numbers from its config file.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
karthikiyer56: Done in 03c1c266.

cmd/stellar-rpc/internal/integrationtest/infrastructure/rpcv2_config_test.go
karthikiyer56: Partly accepted in 03c1c266: the smaller fix. A typed-config entry point would have this harness as its only caller, so we kept the path-plus-flags one.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
karthikiyer56: Done in 03c1c266. Rejected all three.

cmd/stellar-rpc/internal/rpcv2/daemon.go
karthikiyer56: Done in c439ea71.

cmd/stellar-rpc/internal/integrationtest/infrastructure/daemon_rpcv2.go
karthikiyer56: Done in c439ea71 and 03c1c266.

tamirms (approved):

PR #1003, opened 2026-09-11
[DRAFT] Optimize query-path caching and bypass jrpc2 marshal + unmarshal

cjonas9:
[TODO] break this PR up into several separate PRs
### What

Draft. This branch explores three optimizations greatly improving the performance of several endpoints. All of them live in the shared handlers, so both daemons get them through the same constructors. This PR is meant to be the capstone for the optimizations in #732. 

You can compare the optimized performance seen in coordinator runs here to the base perf measured in #1018.

**1. getLatestLedger memoizes its rendered response.** The fully rendered JSON is cached keyed by ledger sequence. A new ledger invalidates it by moving the key, so there is no ingestion hook. Misses are serialized so a burst at a ledger boundary renders once, a request on an older read view cannot evict a newer ledger's render, and errors are never memoized. The mechanism is a small generic, `latestMemo[T]` in `methods/latest_memo.go`, an atomic pointer to a `{seq, value}` entry plus a miss mutex.

**2. getNetwork and getVersionInfo stop decoding the whole ledger for one field.** `getProtocolVersion` went through `store.GetLedger`, a full `UnmarshalBinary` of the latest LedgerCloseMeta (about 2 MB on pubnet, and on v2 a zstd decompress first) to read a 4-byte protocol version. It now reads the header off the raw bytes through `LedgerCloseMetaView`, and memoizes the result by sequence with the same `latestMemo`. Both handlers hold a `protocolVersionCache`.

**3. The response path no longer round-trips through jrpc2's in-memory client.** Every response used to go through jrpc2's jhttp bridge, which routes the call through a `server.Local` client/server pair. For a 2.8 MB pubnet getLatestLedger response that meant the result was marshaled once, re-parsed twice and compacted again before hitting the socket: about 35 ms and 33 MB of allocation per call, with the handler itself under 6% of that. An earlier revision of this branch fixed that with a local `dispatch.go`; it has been replaced by a fork of jrpc2 (`github.com/stellar-experimental/jrpc2`, branch `raw-message-passthrough`, four commits on upstream `main`):

- `json.RawMessage` results pass through `invoke` verbatim.
- `Server.ServeRequests` dispatches parsed requests straight to the handler map.
- The jhttp bridge serves requests directly instead of through the in-memory client.
- `Response.WriteTo` writes the envelope without copying the result.
- A negative `ServerOptions.Concurrency` removes jrpc2's handler semaphore. We set `-1` because the backlog queue and request-duration limiters already gate concurrency per method.

The integration tests still use upstream `creachadair/jrpc2` as their client.

**4. rpcv1 requests no longer stall while a ledger is being ingested.** The ingestion commit used to hold the ledger-range cache lock across the SQLite commit, so every request waited out the whole ledger write (tens to hundreds of milliseconds, once per ledger). The commit now runs outside the lock, and the oldest retained ledger is looked up on the ingestion path instead of by the first request after each trim.

### Measurements

Handler-level, from the perf-eval go-bench leg on this branch (pubnet-sized ledger):

| | time | B/op | allocs/op |
|---|---|---|---|
| GetLatestLedger, uncached | 8.8 ms | 11.2 MiB | 40 |
| GetLatestLedger, cached | 1.06 µs | 164 B | 5 |

getNetwork over the pubnet sqlite fixture (`BenchmarkLedgerReads` harness; v1 sqlite / v2 hot). getVersionInfo tracks it within noise:

| | time | B/op | allocs/op |
|---|---|---|---|
| before | 4.2 ms / 4.4 ms | 10.5 MB / 9.2 MB | 103,000 |
| header view read | 175 µs / 264 µs | 1.5 MB / 8.7 KB | ~200 |
| view + memo | 10 µs / 13 µs | 5 KB / 6 KB | ~120 |

The memo row is getHealth cost. The 103K allocations per call were what made these endpoints fold under concurrency.

### Tests

- getLatestLedger: serves the memo until the ledger advances, an older view does not evict a newer render, 32 concurrent misses render once, errors are not memoized, and the rendered bytes are byte-equal to marshaling the response struct.
- Protocol version: same memo behaviour, plus getNetwork doing one raw read across two requests.
- `BenchmarkGetLatestLedger` gained cached and uncached rows.

## Why

getLatestLedger, getNetwork and getVersionInfo are what SDKs and wallets poll continuously, and they were the first endpoints to buckle under load even though their answers change once per ledger. Two things made them expensive out of proportion to what they return. Every request touched the 2 MB latest ledger, a full decode for one field or a base64 render of the whole thing. And the jhttp bridge added a client/server round trip that re-parsed multi-megabyte results, about 65% of end-to-end time. Anything that only changes when the ledger sequence changes can be memoized by sequence with no invalidation plumbing, and one memo shape covers every "fact about the latest ledger". The response-path fix belongs inside jrpc2 rather than in a parallel dispatcher, hence the fork.

## Known limitations / additional notes

- **Depends on a jrpc2 fork.** `stellar-experimental/jrpc2@raw-message-passthrough` is four commits ahead of upstream. Upstream, vendor or keep the fork is an open decision.
- **Supercedes #969** (own the response envelope, retire the bridge). Both remove the bridge round trip and only one should land. They perform the same on struct results; they differ on `json.RawMessage` results and in how the handler context is built.
- **simulateTransaction still has the old pattern.** `getLatestLedgerPreflightInfo` full-decodes the latest ledger for three header fields. The same view read applies; left for a follow-up.
- **Per-handler memos, not one shared cache.** getLatestLedger, getNetwork and getVersionInfo each hold their own. A shared one would need wiring through `specs.go`, and a getNetwork miss would then either trigger the 2 MB render or need per-field laziness.
- **Testing scaffolding on the branch, revert before merge:** the coordinator's push trigger lists `cache-getLatestLedger`, and the plan job drops the apply-load and go-bench legs from the roster (marked `# temporary, for testing`).
- **Perf-eval leg failures are not from this PR.** Apply-load has failed on this branch since #1001 made the shared integration infrastructure import rpcv2, so the v1 integration test binary now needs RocksDB and zstd headers the perf boxes do not have. The fix (ship the runner-built libs to the box via S3, move the box AMI to 24.04) is going into a separate PR. The blaster leg's "no route to host" to the backfill box is a separate failure, not yet investigated.
- **The go-bench `JSONTransactions` delta is inherited.** The leg reports +87% / +215% against v28.0.1. This PR's only change to `GetTransaction` is the import path; the delta comes from the base branch (#1005: `get_transaction.go`, the sqlite transaction store, the SDK bump) and is flat on the pre-reshape branch. Worth a look on #1005.


cmd/stellar-rpc/internal/methods/get_latest_ledger.go
```diff
@@ -14,41 +17,86 @@ import (
...
-			return protocol.GetLatestLedgerResponse{}, &jrpc2.Error{
-				Code:    jrpc2.InternalError,
-				Message: "could not get latest ledger sequence",
-			}
+	c := &latestLedgerCache{ledgerReader: ledgerReader}
+	return NewHandler(c.handle)
+}
+
+func (c *latestLedgerCache) handle(ctx context.Context, _ protocol.GetLatestLedgerRequest) (json.RawMessage, error) {
+	latestSequence, err := c.ledgerReader.GetLatestLedgerSequence(ctx)
```
Shaptic, line 43: Instead of adding an additional complex caching layer which in turn is relying on an existing caching layer (`GetLatestLedgerSequence` uses the ledger range cache), we should consider extending the latter layer, instead. Namely, we could do something like extend `store.LedgerRange`'s `LedgerInfo`:

```diff
// LedgerInfo identifies one ledger: its sequence number and close time.
type LedgerInfo struct {
	Sequence  uint32
	CloseTime int64
+	Response json.Message // can be nil
}
```

where `Response` gets filled by `getLedgerRangeWithCache` where you can selectively decide whether `substr(meta, 1, %d) AS meta_prefix` is included in the DB query or if the entire meta gets returned. That way the default stays as-is, where you get faster lookups if you don't care about the entire cached meta value, but in the opt-in case, we cache the meta into `LedgerInfo` and propagate `Response` directly to the JSON marshaller.

**Pros:** one cache mechanism as a source of truth, DRY.
**Cons:** violation of abstraction barriers since the DB caching layer now speaks method JSON; the optimization propagates through the entire stack

The con can be alleviated somewhat by caching _only_ the meta rather than the entire `GetLatestLedgerResponse` and computing the other fields on the fly, but then it's not as good of an optimization. But in either case, we remove the entire hairy secondary caching layer.

What do you think?

cjonas9: Hmm, it's a good question, and I'm not certain I have an off-the-cuff answer. 

My first instinct got a bit ahead of myself -- I was hoping that the jrpc2 `Bridge` bypass would have covered enough slack to justify throwing out the caching optimization. Ran a quick, non-benchmark A/B test locally and discovered that is sadly not the case:

getLatestLedger (through dispatcher) | Per request | Allocated
-- | -- | --
Uncached, render per request | 3.8 ms | 15.6 MB
Cached RawMessage memo (what is committed) | 0.34 ms | 13 KB

Thus it doesn't seem I can dodge your question here. Also, I checked only quickly, but that speedup above is dominated by skipping the rendering -- saving only the raw fields doesn't help us much.

I like the idea of DRYing this, though I do see the con about this violating our abstraction layers. And more than that, this adds a struct that's only used by getLatestLedger rides along on a struct that shows up in `getHealth`, `getLedgers`, `getTransactions`, and `getEvents`. 

I had a hard time finding a middle ground here. To me, the alteration my Claude came up with seems like perhaps a marginal improvement, but largely in the spirit of the original optimization:

> Pull it out as a ~20-line generic in `methods,` `ledgerMemo[T],` keyed by sequence: load, compare, render on miss, store with a `compare-and-swap` that keeps the higher sequence. That covers the older-read-view case in one line. The only remaining design choice is the miss mutex: keep it if you want one render per ledger boundary under load, drop it if 3.8 ms times a few concurrent misses every five seconds is acceptable. Either way it is a self-contained, backend-agnostic piece with a two-line contract, and the store's range cache stays what it is today, a cache of scalars with its own trim and reset rules that the memo never touches.

I think I've always found memoization to be particularly tricky, and it may just be a bit late in the evening for me to try to reason about improvements to a memoization design. That is to say, I don't have strong feelings about which potential design is best here. Either way, happy to hear your thoughts!

PR #1005, opened 2026-09-11
store/LedgerReaderTx: replace per-ledger reads on LedgerReader and LedgerReaderTx with ScanLedgers iterator

### LINKED ISSUE #1004: Replace the Tx-level point reads on store.LedgerReaderTx with one ScanLedgers iterator
`store.LedgerReaderTx` exposes a point-read accessor, `WithLedgerRaw(ctx, seq, fn)`, that the v2 adapter serves from a single `ScanLedgers` iterator primed on the first call. The point-read shape is what forces the walk protocol: callers must ask for ascending, contiguous sequences, the adapter checks that on every step, fails loudly on a gap and past a span cap (`walkSpanCap`), and the interface doc has to explain the rule. `BatchGetLedgers` on the same Tx clones every ledger of a getLedgers page on v2 (up to 200) only for the handler to base64- or JSON-encode it. And a v2 in-window gap answers InternalError from the walk's guard where v1 answers InvalidParams for the same missing ledger.

The fix is to let the Tx expose the scan it already runs, so the sequential access pattern is the contract:

```go
// RawLedger is one ledger as ScanLedgers yields it. Raw is the marshaled
// LedgerCloseMeta on loan: read-only, valid only inside the loop body that
// received it, overwritten by the next step. Copy whatever outlives the body.
type RawLedger struct {
	Sequence uint32
	Raw      []byte
}

// LedgerReaderTx is a read-only snapshot of the ledger store: GetLedgerRange
// and every ScanLedgers on it answer from the same committed state. Call Done
// to release it, however the loops ended.
type LedgerReaderTx interface {
	// ScanLedgers yields the stored ledgers in [start, end], ascending, as one
	// forward read that ends when the loop body breaks. Ledgers the snapshot
	// does not hold, below its oldest, above its latest, or missing inside the
	// range, are not yielded; a caller that needs every ledger checks the
	// sequences it receives. start > end yields nothing. A non-nil error,
	// including ctx's, ends the stream and the RawLedger beside it is zero.
	ScanLedgers(ctx context.Context, start, end uint32) iter.Seq2[RawLedger, error]
	// GetLedgerRange is the snapshot's oldest and latest ledger.
	GetLedgerRange(ctx context.Context) (LedgerRange, error)
	Done() error
}
```

`store.LedgerReader` (the non-Tx point reads getLatestLedger and simulateTransaction use, and `StreamLedgerRange`), `store.TransactionReader` and `processTransactionsInLedger` are untouched.

v2: `ledgerReaderTx` becomes `struct{ view *query.ReadView }` and `Done` returns nil. `ScanLedgers` raises `start` to `view.OldestLedger()` (a start below the floor would otherwise surface `ClampRange`'s `*RangeError` as InternalError, where today's `inWindow` gate yields InvalidParams), returns an empty stream when `start > end`, calls `view.ScanLedgers(start, end)` once, checks `ctx.Err()` per step, and yields `RawLedger{entry.Seq, entry.Bytes}` with no copy. The I/O is byte for byte today's `BatchGetLedgers` walk. Gone: `walk`, `next`/`stop`, `iter.Pull2`, `walkSpanCap`, the exhausted-span and non-sequential errors, `BatchGetLedgers`, `rawHeaderFromLCMBytes`, and the `bytes`, `iter`, `chunk`, `stores/ledger` imports.

v1: `ScanLedgers` runs `SELECT sequence, meta ... ORDER BY sequence asc` through `l.tx.Query` inside the read-only transaction `NewTx` began, one row at a time, `defer rows.Close()` covering an early break. Each scanned BLOB is database/sql's own copy, so the loan is a formality there. Gone: `BatchGetLedgers`, the Tx `WithLedgerRaw`.

getTransactions, replacing `readLedgerPage` and the per-ledger loop:

```go
first := uint32(start.LedgerSequence)
last := uint32(min(int64(ledgerRange.LastLedger.Sequence), int64(first)+LedgerScanLimit-1))
next := first
for entry, err := range readTx.ScanLedgers(ctx, first, last) {
	if err != nil {
		return protocol.GetTransactionsResponse{}, &jrpc2.Error{Code: jrpc2.InternalError, Message: err.Error()}
	}
	if entry.Sequence != next {
		return protocol.GetTransactionsResponse{}, missingLedger(next)
	}
	cursor, done, err = h.processTransactionsInLedger(entry.Raw, start, &txns, limit, request.Format)
	if err != nil {
		return protocol.GetTransactionsResponse{}, err
	}
	if done {
		break
	}
	next++
}
if !done && next <= last {
	return protocol.GetTransactionsResponse{}, missingLedger(next)
}
```

`missingLedger` mints today's InvalidParams text. A gap inside the range is the jump branch (TestGetTransactions_LedgerNotFound: ledgers 1 and 3 present, names 2); a stale cursor below the floor yields from the floor on v2 and from the oldest row on v1, both mismatch `next` and name the cursor's ledger, as today; a cursor above the tip has `first > last`, yields nothing, and the existing echo of the request cursor fires. `LedgerScanLimit` is the scan's `last`.

getLedgers ranges over `readTx.ScanLedgers(ctx, start, end)`, breaks at the page limit, and calls `parseLedgerInfo(raw, format)`, which slices the header out of `raw` itself; the datastore fallback marshals each decoded LCM once and calls the same function; `metaToChunk` and `store.LedgerMetadataChunk` are deleted. Missing ledgers stay silent for getLedgers, as today.

Wire behavior is unchanged on every pinned path. One case converges: a v2 in-window hot-store gap (skipped ingestion or corruption) moves from InternalError "non-sequential ledger walk" to the InvalidParams v1 already gives.

Tests to add: the dry-tail branch through a mock Tx that yields ledgers 1 and 2 of a requested [1, 3]; a below-floor cursor on both backends asserting InvalidParams for the cursor's ledger; a per-backend ordering test (ascending, duplicate-free, within [start, end]), since the gap check rests on it; `TestTxBatchGetLedgers_HeaderMatchesFullDecode` moved to methods as a `parseLedgerInfo` test; the LedgerScanLimit pairing test retargeted to `chunk.LedgersPerChunk` (eventsapi still relies on the one-window property).

Size: about 12 files, 760 lines out and 440 in, 12 tests deleted, 12 changed, 5 added. Sequencing: after #962 and #976 land, so the differential suite in #976 checks this reshape rather than being rebased across it.

Loan and lock discipline, stated in the ScanLedgers doc and here: on v2 the loop body runs while the hot store's RocksDB iterator is open and its lifecycle read lock is held (rocksdb.Store.IterateRange takes RLock for the whole iteration), exactly as GetPinned's fn does today. A body must only encode: no blocking, no call back into a store. getTransactions already runs this way on the PR head (its whole page runs under that lock through the Pull2-suspended walk); F1 extends the same rule to getLedgers. Close() is shutdown-only; CloseIfIdle() (deferred chunk deletion) uses TryLock and retries, so a long body delays a teardown, never blocks a reader. Cold readers hold no lock.

Acceptance, measured not argued: before/after on v2 over a real pubnet chunk (bench pkg, --source=bsb), p50 and allocations for getTransactions limit 50 (xdr and json), getLedgers limit 200 (xdr and json), and getLatestLedger; the reshape must be flat or better on every row. The one expected behavior shift to watch: getLedgers on v2 now holds the scan's readers (RocksDB iterator plus scratch, or a cold pack reader) across the page's encode instead of only the clone; getTransactions already does this today.

Step 2, same follow-up, second commit: make the contract congruent. `store.LedgerReader` gets the same `ScanLedgers` (no snapshot; v2 through `ViewFrom(ctx)`, v1 on the pooled connection). `LedgerReader.GetLedger` and `LedgerReader.WithLedgerRaw` become `store` helpers over a scan of one, `store.GetLedger(ctx, scanner, seq)` and `store.WithLedgerRaw(ctx, scanner, seq, fn)`, so the contract has one loan idiom and the callback survives only as sugar for the single-ledger case. The v2 adapter routes `start == end` to `ReadView.WithLedger`, keeping the window check on that branch, so getLatestLedger's I/O stays a pinned point read. `GetLatestLedgerSequence` and `StreamLedgerRange` stay as they are. Callers: getLatestLedger, util.go, simulate_transaction.go; about ten more files, one-line call sites and mocks.

cjonas9:
### What

`store.LedgerReaderTx` and `store.LedgerReader` now expose one read idiom: `ScanLedgers(ctx, start, end) iter.Seq2[RawLedger, error]`. Each `RawLedger` lends its bytes for one loop body. The point reads become `store` helpers over a scan of one (`store.GetLedger`, `store.WithLedgerRaw`), and both backends implement the scan once:

- **sqlite**: one ordered range query inside the reader's transaction, or on the pooled connection for the non-Tx reader. A scan of one is a primary-key equality query. The blob is scanned as `sql.RawBytes`, so database/sql lends the driver's buffer instead of cloning it; that is the loan the contract already promises.
- **v2 adapter**: the view's `ScanLedgers`, with `start` raised to the view's floor so a below-floor request yields nothing instead of a `*RangeError`. A scan of one routes to the view's pinned point read behind the window gate, so getLatestLedger's I/O is unchanged.

This replaces: `WithLedgerRaw` on the Tx, `BatchGetLedgers`, `LedgerMetadataChunk`, the adapter's `walk`, `walkSpanCap` and `rawHeaderFromLCMBytes`, and the handlers' `readLedgerPage` and `procErr`.

- **getTransactions** is one scan over `[start, min(latest, start+LedgerScanLimit-1)]`. A yielded sequence that is not the one expected, or a range that runs dry, returns the same InvalidParams naming the missing ledger as before.
- **getLedgers** collects its page (at most 200 ledgers) out of the scan and renders it after the scan's readers close, slicing the header off the view in `parseLedgerInfo`. The datastore fallback marshals each ledger into the same function. See the benchmark section for why the page is collected first.

### Why

#1004. The point-read shape forced a walk protocol the adapter had to police, cloned every ledger of a getLedgers page only to encode it, and made a v2 in-window gap answer `InternalError` where v1 answers `InvalidParams`. The scan makes sequential access the contract.

#### Behavior

Wire output is unchanged on every pinned path. The benchmark's response-byte metric is identical on all 21 rows. One case converges: a v2 in-window hot-store gap now returns v1's `InvalidParams`. Missing ledgers stay silent for getLedgers, as this is the behavior that we have today.

#### Tests

The `getTransactions` differential suite passes unchanged, including error parity. Added: scan-gap and dry-tail cells for `getTransactions`; ordering, gap, below-floor, empty-range, cancellation, early-break and chunk-border tests for the adapter; an ordering and gap test for sqlite; a `parseLedgerInfo` header test over all three LCM wire versions; a getLedgers test that a scan error yielded after a full page is reported rather than dropped; a `ScanLedgersFrom` test that an inclusive range ending at `MaxUint32` terminates. Removed: the fourteen tests on the retired walk and batch surface.

The getEvents differential suite (from #982) now validates the scan as well. Its frozen reference no longer reads the events table, and instead re-derives each request window's events from the ledgers themselves, one ScanLedgers pass over the window. This new scan approach and the old events table, its SQL prefilters and the view path are each other's oracles.

#### One tangential optimization

In `xdr2json/conversion.go`, `convertAnyBytes` copied the JSON out of C twice: `C.GoString` allocates a Go string and memcpys the C bytes into it, then `json.RawMessage(jsonStr)` converts string to byte slice, which must allocate and copy again because strings are immutable. A single `C.GoBytes` does one copy. It sits on the hot path of every JSON response, and the eight-line change is worth 8 to 11% of bytes/op on the JSON rows across all three backends.

#### Benchmark

`BenchmarkLedgerReads` (rpcv2) serves `getTransactions`, `getLedgers` and `getLatestLedger` over 865 real pubnet ledgers on v2-hot, v2-cold and v1-sqlite through an in-memory JSON-RPC pipe. Interleaved builds, `-count 6`, M4 Pro, branch base vs head:

| | time | bytes/op | allocs/op |
|---|---|---|---|
| v2-hot / v2-cold getLedgers, xdr | flat | flat | -3% to -11% |
| v2-hot / v2-cold getLedgers, json | flat | -8% to -9% | -13% to -31% |
| v1 getLedgers, xdr | flat | flat | +6% to +7% |
| v1 getLedgers-20/json | flat | -8% | flat |
| v1 getLedgers-200/json | flat | -9% | -5% |
| getTransactions-50, xdr | flat | flat (v2), -26% (v1) | flat |
| getTransactions-50, json | flat | -10% to -23% | -18% to -24% |
| getLatestLedger, all backends | flat | flat (v2), -7% (v1) | +5 to +6 allocs |

An earlier head of this PR regressed v2-hot getLedgers-200/json by 8 to 12% (p<0.01, n=6) while its cold twin stayed flat. The cause was encoding the page inside the open RocksDB iterator, the shift #1004 named as the one to watch. `getLedgers` therefore collects its page and renders after the scan closes; the `sql.RawBytes` change keeps that from costing v1 a second copy. That row is now flat (p=0.59).

The full-suite run showed v1 getLedgers-200/json at +8% (p=0.04). It is the last of 21 rows in each invocation, so it carries the most drift; an eight-round rerun of that row alone measured 15.59s vs 15.54s (p=0.57), so it is truly actually flat.

The remaining allocs increases (v1 xdr `getLedgers`, `getLatestLedger`) are the two-column scan and the scan-of-one closure, each a handful of allocations per request.

### Known Limitations

N/A


cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -265,27 +116,59 @@ func inWindow(view *query.ReadView, seq uint32) bool {
...
+			return
+		}
+		if start == end {
+			if !inWindow(view, start) {
+				return
+			}
+			err := view.WithLedger(start, func(raw []byte) error {
+				yield(store.RawLedger{Sequence: start, Raw: raw}, nil)
+				return nil
+			})
```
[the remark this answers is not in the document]
cjonas9: The multi-ledger path checks per step because it loops. Here the window between the top-of-iterator check and the callback is one RocksDB get. An abandoned handler decodes at most one ledger. Not worth the lines.

cmd/stellar-rpc/internal/store/ledger.go
```diff
@@ -47,27 +49,80 @@ type LedgerReader interface {
...
+// comes back verbatim with found true.
+func WithLedgerRaw(ctx context.Context, s LedgerScanner, seq uint32, fn WithLedgerRawFn) (bool, error) {
+	for l, err := range s.ScanLedgers(ctx, seq, seq) {
+		if err != nil {
+			return false, err
+		}
+		return true, fn(l.Raw)
+	}
+	return false, nil
+}
```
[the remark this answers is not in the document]
cjonas9: A scan of one is a primary-key equality on SQLite and a pinned `WithLedger` point read on v2. Neither has a stream tail after the row arrives. Once the row is delivered, found and fn's result are already determined. The loan contract also forces fn to run inside the loop body, so this "drain after the callback" restructure adds code for a path that does not exist.

cmd/stellar-rpc/internal/methods/get_events_differential_test.go
```diff
@@ -222,9 +228,8 @@ type legacyEventEntry struct {
 	txHash               *xdr.Hash
 }
 
-// legacyGetEvents is the pre-view handler: each event decoded, then the frozen matcher and renderer.
-//
-//nolint:cyclop,funlen // frozen reference; it mirrors the production handler's shape by design
+// legacyGetEvents is the pre-view handler over ledger-derived events: the
+// frozen matcher and renderer on what legacyScanEvents finds.
```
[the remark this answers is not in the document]
cjonas9: the lint job does not fail

cmd/stellar-rpc/internal/store/ledger.go
```diff
@@ -35,8 +36,9 @@ func (lr LedgerRange) ToLedgerSeqRange() protocol.LedgerSeqRange {
 // LedgerReader is the serving-side read contract every storage backend
 // implements. Handlers depend on this interface, never on a concrete backend.
 type LedgerReader interface {
-	GetLedger(ctx context.Context, sequence uint32) (xdr.LedgerCloseMeta, bool, error)
-	WithLedgerRaw(ctx context.Context, sequence uint32, fn WithLedgerRawFn) (found bool, err error)
+	// ScanLedgers reads whatever the request already sees (the live store on
+	// v1 and the request's read view on v2), without its own snapshot.
+	LedgerScanner
 	GetLedgerRange(ctx context.Context) (LedgerRange, error)
 	StreamLedgerRange(ctx context.Context, startLedger uint32, endLedger uint32, f StreamLedgerFn) error
```
tamirms, line 43: `StreamLedgerRange` is the last read that is still a method per backend, so the contract is one idiom plus this one.

It has a single production caller, the migration replay at `rpcv1/daemon/daemon.go:486`, and that call holds the concrete `sqlitedb` reader rather than this interface. Nothing calls the v2 implementation at all: `rpcv2/feereplay.go:57` goes straight to `view.ScanLedgers`. So `adapters/ledger_reader.go:57-85` exists only to satisfy this line, and four fakes carry a stub for the same reason.

Dropping it from the interface lets both implementations and all four stubs go, and the one caller becomes a loop over the scan it already has:

```go
for l, err := range sqlitedb.NewLedgerReader(d.db).ScanLedgers(ctx, first, last) {
	if err != nil {
		return err
	}
	var lcm xdr.LedgerCloseMeta
	if err := lcm.UnmarshalBinary(l.Raw); err != nil {
		return err
	}
	// the existing body
}
```

Roughly seventy lines out across seven files. One behavior note: the v2 version answers a below-floor start with a `*RangeError`, pinned by `TestStreamLedgerRange_BelowFloorIsRangeError`, where the scan raises the start to the floor instead. Nothing depends on that, since v2 has no caller.

Fine as a follow-up if you would rather not grow this diff.

cmd/stellar-rpc/internal/rpcv1/sqlitedb/ledger.go
```diff
@@ -151,18 +109,49 @@ func (r ledgerReader) StreamLedgerRange(
...
+		// Runs on an early break too, which is how the consumer ends the scan.
+		defer rows.Close()
+
+		for rows.Next() {
+			if err := ctx.Err(); err != nil {
+				yield(store.RawLedger{}, err)
+				return
+			}
+			var seq uint32
+			var meta sql.RawBytes // database/sql's per-Rows buffer, reused by the next Scan: RawLedger's loan
```
tamirms, line 142: A blob does not go through the shared per-Rows buffer. `database/sql` uses `rows.rawbuf()` only for `string` and `time.Time` sources; for a `[]byte` source the `*RawBytes` case is a plain `*d = s` (`$GOROOT/src/database/sql/convert.go:274-279`), aliasing whatever the driver left in the row. With mattn/go-sqlite3 that is a fresh `C.GoBytes` copy per row (`sqlite3.go:2228`), so nothing is reused here and the next `Scan` overwrites nothing.

The loan itself is right and should stay. The standard library only promises the slice until the next `Next`, `Scan` or `Close`, another driver may well hand back a reused buffer, and on v2 the reuse is real.

This PR already had the accurate wording, in 93326c0c, before e76193a7 rewrote it:

```go
var meta sql.RawBytes // the driver's buffer, valid until the next Next: RawLedger's loan
```

cmd/stellar-rpc/internal/store/ledger.go
```diff
@@ -47,27 +49,82 @@ type LedgerReader interface {
...
-// terms. found=false means fn never ran; fn's own error comes back verbatim
-// with found=true.
+// LedgerScanner is the contract's one read idiom, embedded by LedgerReader and LedgerReaderTx.
+type LedgerScanner interface {
+	// ScanLedgers yields the stored ledgers in [start, end], ascending, as one
+	// forward read that ends when the loop body breaks. Ledgers the source
+	// does not hold, below its oldest, above its latest, or missing inside the
+	// range, are not yielded, so callers should check the sequences they receive.
+	// A non-nil error ends the stream and the RawLedger beside it is zero.
+	ScanLedgers(ctx context.Context, start, end uint32) iter.Seq2[RawLedger, error]
```
tamirms, line 59: Worth saying here that the loop body runs inside the source's read. The callback shape made that obvious; a `for range` hides it.

On v2 a yielded ledger arrives from inside the tier's read: the hot store's pinned lookup for a scan of one (`stores/ledger/hot_store.go:97-107`) or the RocksDB iterator for a range, both holding the store's lifecycle read lock, and the cold reader's pooled record (`packfile/reader.go:428-444`). `rocksdb.Store.GetPinned` spells out the consequence at `rocksdb/rocksdb.go:230-232`: fn must not block on anything that could close the store, nor call back into a `Store` method that takes the lock, because a nested read lock behind a waiting `Close` deadlocks both. Nothing on this interface carries that up to where a handler author reads.

Every consumer today is pure over the bytes, so this is about keeping it that way. A sentence beside the loan would do it: the body runs inside the source's read, so keep it to encoding and copying and never call back into the same store.

cmd/stellar-rpc/internal/rpcv2/adapters/ledger_reader.go
```diff
@@ -265,27 +116,61 @@ func inWindow(view *query.ReadView, seq uint32) bool {
...
+func scanView(ctx context.Context, view *query.ReadView, start, end uint32) iter.Seq2[store.RawLedger, error] {
+	return func(yield func(store.RawLedger, error) bool) {
+		// The request duration limiter answers the client at the deadline but
+		// only abandons the handler goroutine; without these checks an
+		// abandoned scan would keep decoding while holding its read view.
+		if err := ctx.Err(); err != nil {
+			yield(store.RawLedger{}, err)
+			return
+		}
+		if start == end {
```
tamirms, line 132: The two branches decide "is this ledger servable" in different ways. The point branch gates on `inWindow`; the range branch raises `start` to the floor and lets `ClampRange` truncate the top. Same answers, two mechanisms.

Hoisting the clamp above the split leaves one:

```go
start = max(start, view.OldestLedger())
end = min(end, view.LatestLedger())
if start > end {
	return
}
```

After that the point branch is just the `WithLedger` call and its `ErrNotFound` mapping, and `inWindow` has no caller left. Its own doc calls it "the one gate every point read must apply", which is already only half true.

Small and optional; worth folding in if you touch the file again.

cmd/stellar-rpc/internal/methods/get_ledgers.go
```diff
@@ -278,10 +280,32 @@ func (h ledgersHandler) fetchLedgers(
 	return result, err
 }
 
+// collectLedgerPage copies up to room ledgers of [start, end] out of the scan, so
+// rendering runs after the scan's readers close (encoding a page under an open
+// hot-store iterator measurably slows it).
```
tamirms, line 285: I tried to reproduce this and could not, so I think the copy can go.

Three measurements over 320 real pubnet ledgers pulled from the lake, averaging about 2 MB each, on an M1 Max:

| measurement | result |
|---|---|
| interleaved in-process A/B, 6 rounds each | rendering in the scan is equal or faster on every row; the only significant hot result is it being faster, on xdr-200 at p=0.004 |
| your protocol replicated: alternating processes, benchstat | hot getLedgers-200/json flat, 22.93s vs 22.73s, p=0.49 |
| the iterator isolated | 0.6 ms per ledger, about 125 ms for the whole page, so 3% of the handler and 0.5% of the op |

Holding the iterator across the conversion does cost something, 43 to 84 µs per ledger, but it costs the same on cold, so it reads as ordinary cache pollution between steps rather than anything iterator-specific. Over a page that is under a tenth of a percent.

What I think produced the original signal: one op on this row takes about 23 seconds, which is past the default benchtime, so `b.Loop` runs exactly one iteration and each `-count` contributes a single sample of an operation that allocates roughly 10 GB and triggers several collections. Within one shape I measured a 6.5 to 17.7 percent spread, and one round landed 1.5 seconds faster purely because a pooled encoder buffer survived a GC. With six samples a side, p below 0.01 needs near-total separation of the twelve values, and drift of that size produces it on its own.

Caveats: a different machine from yours, 320 ledgers rather than 865, and my hot chunk never flushed, so I added a flushed variant to cover the SST path your fixture would have hit. Flat there too.

If that holds up, `collectLedgerPage` can go back to calling `appendLedger` inside the scan loop. That drops about 384 MB of clones per 200-ledger page and stops holding the whole page live. The `sql.RawBytes` change is worth keeping either way, and worth more without the clone: the render consumes the loaned bytes before the next `Next`, so v1 pays no copy at all rather than one.

I did consider one non-performance reason to keep it, that it releases the scan's readers before a multi-second render, and I do not think it survives. The store's read lock is shared, deferred chunk deletion sets its closed flag before trying the lock so a long reader costs it a retry rather than starving it, and the hot ledgers column family turns over slowly enough that a four-second hold overlaps almost no flushing. Against that, the copy is 400 MB per in-flight request.

If you would rather keep the copy, could the comment give the real reason and note the per-ledger cost instead of claiming a speedup? Happy to share the harness either way.

PR #1027, opened 2026-09-17
rpcv2: getEventsV2 integration tests

### LINKED ISSUE #1021: rpcv2: getEventsV2 integration test
## What
Integration coverage for `getEventsV2`, run against a live rpcv2 daemon in the same environments as the rest of the suite. #883 runs the rpcv1 suite unchanged against rpcv2, which only exercises the v1 surface, so `getEventsV2` has no end-to-end coverage today.

## Depends on
#883, for the environment the test runs in.

urvisavla:
### What
Fixes #1021

getEventsV2 integration tests: paging in both directions, filters, parity with getEvents, paging while the tip moves. Plus a tampered-cursor test on the JSON-RPC boundary.

Also fixes two harness setup flakes: `waitForRPC` waits for one consensus ledger after catch-up, because Core has no Soroban transaction queue until then and rejected the limits upgrade with txNOT_SUPPORTED; and the upgrade key is set only once the Core container has closed the ledger that holds it.

### Known limitations

JSON topic filters are rejected; one subtest pins that.


cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
```diff
@@ -0,0 +1,364 @@
...
+				Format:    format,
+			})
+			require.NoError(t, err)
+			requireFixtureEvents(t, fx, v2.Events)
+
+			require.Len(t, v2.Events, len(v1.Events))
+			for i := range v1.Events {
+				assert.Equal(t, protocol.EventInfoV2(v1.Events[i]), v2.Events[i], "event %d", i)
+			}
+			assert.Equal(t, v1.LatestLedger, v2.LatestLedger, "latestLedger")
```
tamirms: `latestLedger` here is the live tip of a Core that closes a ledger about every second, and the two readings come from two separate round trips. Any close that lands between the getEvents response and the getEventsV2 response fails this line, so the assertion carries a small per-run flake that a green CI run does not rule out.

The event comparison above is safe, since a closed range over already-closed ledgers cannot change, and `oldestLedger` is safe too because the retention window is a day and nothing prunes during a test. So it is only this line.

Would you be up for making the tip fields non-racy? Reading v2 second means `v2.LatestLedger >= v1.LatestLedger` always holds, so that plus a floor of `fx.last()` pins what the test cares about. Taking both readings inside a `require.Eventually` that retries until the pair agrees also works if exact equality is the thing worth checking.

cmd/stellar-rpc/internal/rpcv2/jsonrpc_test.go
```diff
@@ -132,6 +133,61 @@ func TestJSONRPCHandler_GetEventsV2ReportsTypedErrorData(t *testing.T) {
...
+			params, err := json.Marshal(map[string]string{"cursor": cursor})
+			require.NoError(t, err)
+			out := rpcv2test.PostRPC(t, url, protocol.GetEventsV2MethodName, string(params))
+			require.NotNil(t, out.Error)
+			assert.EqualValues(t, jrpc2.InvalidParams, out.Error.Code, "message: %s", out.Error.Message)
+			var data struct {
+				Reason string `json:"reason"`
+			}
+			require.NoError(t, json.Unmarshal(out.Error.Data, &data))
+			assert.Contains(t, []string{protocol.ErrorReasonCursorMalformed, protocol.ErrorReasonInvalidParams}, data.Reason,
```
tamirms: All eleven cases here come back with `cursor_malformed` today. Every one of them routes through `DecodeEventCursor`, so the error is `ErrCursorMalformed` or `ErrCursorUnknownVersion`, and `responseError` maps both to `CursorMalformedErrorData`. I replaced this line with `assert.Equal(t, protocol.ErrorReasonCursorMalformed, data.Reason, ...)` locally and the test still passes.

Since the comment says the point is to pin the wire shape, would you want the exact assertion? The disjunction stays green if a change starts reporting these as plain `invalid_params`, which is the regression the test exists to catch. The `invalid_params` alternative was needed while the empty-cursor case was in the table, and that case is gone now.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
```diff
@@ -0,0 +1,364 @@
...
+	resp := test.InvokeHostFunc(f.contractRaw, "increment")
+	f.ledgers = append(f.ledgers, resp.Ledger)
+	f.txHashes = append(f.txHashes, resp.TransactionHash)
+}
+func (f *eventsFixture) last() uint32 { return f.ledgers[len(f.ledgers)-1] }
+
+// Each call waits for inclusion, so the increments land in distinct ledgers.
+func deployAndIncrement(t *testing.T, test *infrastructure.Test, n int) *eventsFixture {
+	rpc := test.GetRPCLient()
+	test.UploadEventsContract()
```
tamirms: This block is the same twenty lines as `cmd/stellar-rpc/internal/integrationtest/transaction_test.go:312-333`, and the harness already exposes the shape for two other contracts: `Test.CreateHelloWorldContract()` and `Test.CreateAutorestoreContract()` both upload, create, and hand back the contract ID. A `CreateEventsContract()` beside them mirrors `CreateAutorestoreContract` almost line for line:

```go
func (i *Test) CreateEventsContract() (protocol.GetTransactionResponse, [32]byte, xdr.Hash) {
	_, contractHash := i.uploadContract(GetEventsContract())
	salt := xdr.Uint256(testSalt)
	account := i.MasterAccount().GetAccountID()
	op := createCreateContractV2Operation(account, salt, contractHash)
	contractID := GetContractID(i.t, account, salt, StandaloneNetworkPassphrase)
	return i.PreflightAndSendMasterOperation(op), contractID, contractHash
}
```

`deployAndIncrement` then starts with `_, rawID, _ := test.CreateEventsContract()`, which also drops the `txnbuild` import here and removes the salt round trip through `MustCreateContractV2().ContractIdPreimage.MustFromAddress().Salt`, since that salt is just `testSalt`. The copy in `transaction_test.go` could move over to it whenever someone is next in that file.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
```diff
@@ -604,13 +604,38 @@ func (i *Test) waitForRPC() {
...
+				i.waitForOneMoreLedger(result.LatestLedger, deadline)
+			}
 			return
 		}
 		require.False(i.t, time.Now().After(deadline), "RPC never got healthy: %+v", err)
 		time.Sleep(time.Second)
 	}
 }
 
+// waitForOneMoreLedger returns once the daemon has committed a ledger past
```
tamirms: The loop in `waitForRPC` checks `exited` on every pass, so a daemon that dies is reported with its own error instead of a timeout. This second loop does not, so a daemon that exits during the extra wait leaves the test polling a dead endpoint until the health deadline runs out, and the failure reads "RPC never committed a ledger past N: <nil>".

One option is to fold the extra ledger into the existing loop rather than add a second one: remember the ledger seen at the first caught-up poll and keep going until `result.LatestLedger` passes it. That keeps the exit watch, the deadline handling and the per-poll `getHealth` log, and it removes the second function. The poll interval would go from 200ms to the outer loop's 1s, which costs at most one extra second on a 1s ledger cadence.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
```diff
@@ -0,0 +1,364 @@
...
+	for round := range rounds {
+		got, last := drain(t, rpc, req)
+		seen = append(seen, got...)
+		require.Equal(t, protocol.ScanStatusWaitingForLedgers, last.ScanStatus, "round %d", round)
+		require.NotEmpty(t, last.Cursor, "round %d", round)
+
+		idle, err := rpc.GetEventsV2(t.Context(), protocol.GetEventsV2Request{Cursor: last.Cursor})
+		require.NoError(t, err)
+		assert.Empty(t, idle.Events, "round %d: idle tip cursor served events", round)
+		assert.NotEmpty(t, idle.Cursor, "round %d: idle tip cursor lost the cursor", round)
```
tamirms: `idle.Cursor` becomes the next round's cursor on line 357, so when this assertion fails the test keeps going and sends a request with neither a cursor nor a `minLedger`, which the handler rejects with "minLedger is required for ascending order". The failure then surfaces inside `drain` on the following round instead of here. `require` on both of these stops at the cause, and lines 346 and 347 in the same loop already use it.

tamirms (commented):
The test set covers what the unit tests cannot reach, and I like that the paged cases assert against the unpaged read rather than against a hand-written expectation. That is the shape that catches a real pager bug.

A few things I checked before reading: `go vet` is clean on every changed package, `go mod tidy` produces no diff (so the `prometheus/client_model` flip to `// indirect` is right, nothing imports it), and the branch still merges cleanly on the base now that #978 has landed there.

I also worked through the txNOT_SUPPORTED diagnosis and it holds up. `daemon_rpcv2.go` points `ingestion.core_url` at the daemon's own captive core, so rpcv2 submits there rather than to the Core container the way rpcv1 does, and because `caughtUpWithCore` compares against the container's tip read after the health reading, one more ingested ledger does imply the captive core externalized a ledger through consensus. Good find.

Five comments below. The first is the only one I would not leave as it is, since it carries a real per-run flake that a green run does not rule out. The rest are smaller.

tamirms (approved):

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
```diff
@@ -0,0 +1,346 @@
...
+		_, dup := ids[e.ID]
+		require.False(t, dup, "duplicate event id %s", e.ID)
+		ids[e.ID] = struct{}{}
+		assert.Contains(t, fx.ledgers, uint32(e.Ledger))
+		assert.Contains(t, fx.txHashes, e.TransactionHash)
+		assert.Equal(t, protocol.EventTypeContract, e.EventType)
+		assert.NotEmpty(t, e.ContractID)
+		if e.ContractID == fx.contractID {
+			fromContract++
+		}
```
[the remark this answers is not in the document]
urvisavla: Core builds the fee event as `ContractEventType::CONTRACT` (EventManager.cpp). CI passes this on real Core.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
```diff
@@ -0,0 +1,346 @@
...
+	})
+
+	t.Run("filters are OR-ed", func(t *testing.T) {
+		req := rng
+		req.Filters = []protocol.EventFilterV2{
+			{ContractID: fx.contractID},
+			{EventType: protocol.EventTypeContract},
+		}
+		assert.Equal(t, eventIDs(unfiltered.Events), eventIDs(call(t, req).Events))
+	})
```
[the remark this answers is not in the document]
urvisavla: Same: all nine events are contract type, so the union is the unfiltered page.

urvisavla:
All five addressed in 90070583.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
urvisavla: Done in 90070583: `v2.LatestLedger >= v1.LatestLedger`, both floored at `fx.last()`.

cmd/stellar-rpc/internal/rpcv2/jsonrpc_test.go
urvisavla: Done in 90070583: pinned to `cursor_malformed`.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
urvisavla: Done in 90070583: `Test.CreateEventsContract`, used here. The copy in `transaction_test.go` is left for the next visit.

cmd/stellar-rpc/internal/integrationtest/infrastructure/test.go
urvisavla: Done in 90070583: folded into the `waitForRPC` loop.

cmd/stellar-rpc/internal/rpcv2/integrationtest/get_events_v2_test.go
urvisavla: Done in 90070583: `require` on both.

